PR #3, opened 2015-08-18
Write network configuration to have wicked understand ovs (bsc#930986)

rsalevsky:
This is based on https://en.opensuse.org/Portal:Wicked/OpenvSwitch and
comment 9 in the bug.

https://bugzilla.suse.com/show_bug.cgi?id=930986

Based on https://github.com/crowbar/barclamp-neutron/pull/245

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -133,6 +133,8 @@ if neutron[:neutron][:networking_plugin] == "ml2" and
 
   # Create the bridges Neutron needs.
   # Usurp config as needed.
+  has_ovs_bridge = false
+
   networks = [["nova_fixed", "fixed"], ["nova_floating", "public"]]
   neutron[:neutron][:additional_external_networks].each do |net|
     networks << [net, net]
@@ -140,6 +142,9 @@ if neutron[:neutron][:networking_plugin] == "ml2" and
   networks.each do |net|
     bound_if = (node[:crowbar_wall][:network][:nets][net[0]].last rescue nil)
     next unless bound_if
+
+    has_ovs_bridge = true
+
     name = "br-#{net[1]}"
     execute "Neutron: create #{name}" do
       command "ovs-vsctl add-br #{name}; ip link set #{name} up"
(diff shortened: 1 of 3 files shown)


toabctl:
@rsalevsky Can you fix the style issues, please?

rsalevsky:
@toabctl Done, gating fails due to the patch is already in the package.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -157,53 +162,139 @@
...
+      # on SLE12, we create proper network configuration for OVS bridges, see
+      # https://en.opensuse.org/Portal:Wicked/OpenvSwitch
+
+      # We need to create a static ifcfg file so that wicked will know that the
+      # ovs bridge usurps the interface address. This also means that we need
+      # to override the variables we pass to the template for the ifcfg of the
+      # interface.
+      bound_if_resource = resources(template: "/etc/sysconfig/network/ifcfg-#{bound_if}")
+      bound_if_variables = bound_if_resource.variables
+
```
rhafer: As others already pointed in https://github.com/crowbar/barclamp-neutron/pull/245 I also believe a lot of this code should actually move to the network barclamp.

When applying the neutron proposal (```apply_role_pre_chef_call()```) we'd "just" need to adjust the settings of the network proposal to what the neutron requires. E.g. be setting ```add_bridge``` (or some new attribute) on the ```fixed``` and ```floating``` networks. Of course we'd also need to teach the network barclamp to create ovs bridges.

We already do something like that today for the "use_vlan" attribute of  the ```nova_fixed``` network.

I'll take a look into this and try to come up with something.

aspiers:
> Done, gating fails due to the patch is already in the package.

@rsalevsky ICBW but it looks to me like gating fails not because the patch is already in the package ([all patches except those numbered `5*` are automatically stripped out](https://github.com/SUSE-Cloud/automation/blob/master/scripts/crowbar-testbuild.py#L96), and [this one is numbered `Patch9`](https://build.suse.de/package/view_file/Devel:Cloud:6/crowbar-openstack/crowbar-openstack.spec?expand=1)), but it's actually failing because for some reason `/usr/lib/build/spec_add_patch` is trying to patch using `-p0` instead of `-p1`.

aspiers:
@dirkmueller @rsalevsky I found the bug in `spec_add_patch` and [fixed it](https://github.com/openSUSE/obs-build/pull/198).

Now you guys owe me a favour ;-)  Which is to help explain why this PR is causing my mkcloud runs to fail on the neutron proposal ...  I'm not sure but it looks like it's probably caused by `wicked` restarting in the middle of the chef-client run, due to this line in `template "/etc/wicked/local.conf"`:

    notifies :restart, "service[wicked]"

aspiers:
Here's the `chef-client` failure on [`controller1`](http://pastebin.nue.suse.com/15494/src).  On `controller2` [there is no failure after the `wicked restart`](http://pastebin.nue.suse.com/15495/src), but that is the last `chef-client` run which `controller2` ever does from Crowbar.

aspiers:
Here's Crowbar's [`production.log`](http://pastebin.nue.suse.com/15496/src).  Also `/var/log/messages` on both controllers shows a lot of turmoil due to the network interfaces being pulled away from underneath the feet of `corosync`, `neutron` etc., e.g. http://pastebin.nue.suse.com/15497/src

If you want to look closer, the supportconfigs are under `/root/manual.aspiers/.artefacts` on `mkcloud.cloud.suse.de`:

```
-rw------- 1 root root 1524084 Sep  4 02:23 nts_crowbar_150904_0220.tbz
-rw------- 1 root root 1188278 Sep  4 02:23 nts_d52-54-01-77-77-01_150904_0220.tbz
-rw------- 1 root root 1073925 Sep  4 02:23 nts_d52-54-02-77-77-02_150904_0220.tbz
-rw------- 1 root root  778333 Sep  4 02:23 nts_d52-54-03-77-77-03_150904_0220.tbz
-rw------- 1 root root  869587 Sep  4 02:23 nts_d52-54-04-77-77-04_150904_0220.tbz
-rw-r--r-- 1 root root  252860 Sep  4 02:23 console_150904_0129.log
```

In summary, it looks like telling Chef to do a `wicked restart` is a really bad idea.

aspiers:
So [the VIP monitoring bug](https://github.com/crowbar/crowbar-ha/pull/15) which I finally nailed partially contributed to the brokenness caused by the restart of `wicked`.  But thinking about it more, it's simply completely unsafe to restart `wicked` on a cluster node without at very least putting the node in maintenance mode.  Even then I'm not sure that's enough to eliminate the risk of it being fenced (e.g. if the STONITH device is SBD over iSCSI) or other bad things happening.  Maybe it's possible to only do a partial restart which doesn't affect the interface the cluster depends on?  Let's talk about this next week.

rhafer:
> Maybe it's possible to only do a partial restart which doesn't affect the interface the cluster depends on? Let's talk about this next week.

Not sure about that. The restart seems to be needed because we need wicked-nanny daemon to run for the ovs setup to work correctly. And according to https://www.suse.com/documentation/sles-12/book_sle_admin/data/sec_basicnet_manconf.html#sec_basicnet_manconf_using_nanny that requires a wicked restart. But as mentioned I am currently reworking this and try to move all of the low-level pieces (ovs-vsctl calls, ifcfg file creation, ...) to the network barclamp so this code will change a bit.

Also I am somehow sure (not completely yet) that we don't need the wicked restart. When applying the network role to the node for the first time, the network barclamp will make sure that the ovs-bridge and its slave interfaces have the right configuration (by doing the appropriate ```ip``` and ```ovs-vsctl``` calls) wicked is only required to bring up the network correctly from the ifcfg files during a reboot. And for that it would be enough to just update ```/etc/wicked/local.conf``` without restarting wicked. If, for whatever reason, this does not work another possible solution could be to unconditionally enable the wicked nanny on all nodes (either via a chef-recipe, or the autoyast profile). AFAICS this shouldn't do any harm(™)

aspiers:
OK.  I know nothing about wicked so I'll leave the decision to you.  But if you keep the restart, like I said, at very least you need to put the node in maintenance mode if it's in a cluster.  And of course you have to test this carefully with HA enabled :) I get the impression noone did that yet.

vuntz:
Superseded by https://github.com/crowbar/crowbar-core/pull/33

PR #5, opened 2015-08-18
Add iptable rules to deny VNC access on Compute hosts (bnc #938200)

dguitarbite:
VNC ports are accessible from the management network. This patch
implements iptable rules which will deny access to these ports to all
but the novnc proxy service (nova-controller).

### DIFF
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -285,16 +285,22 @@ end unless node.platform == "suse"
 
 env_filter = " AND nova_config_environment:#{node[:nova][:config][:environment]}"
 nova_controller = search(:node, "roles:nova-multi-controller#{env_filter}")
+nova_controller_ips = []
+
+if !nova_controller.nil? && nova_controller.length > 0
+  nova_controller.each do |nova_controller_node|
+    nova_controller_ips.push Chef::Recipe::Barclamp::Inventory.get_network_by_type(nova_controller_node, "admin").address
+  end
+end
 
 # Note: since we do not allow shared storage with a cluster, we know that the
 # first controller is the right one to use (ie, the only one)
 if !nova_controller.nil? and nova_controller.length > 0 and nova_controller[0].name != node.name
-  nova_controller_ip =  Chef::Recipe::Barclamp::Inventory.get_network_by_type(nova_controller[0], "admin").address
   mount node[:nova][:instances_path] do
     action node[:nova]["use_shared_instance_storage"] ? [:mount, :enable] : [:umount, :disable]
     fstype "nfs"
     options "rw,auto"
-    device nova_controller_ip + ":" +  node[:nova][:instances_path]
+    device nova_controller_ips[0] + ":" + node[:nova][:instances_path]
   end
 
 end
@@ -417,4 +423,37 @@ end
 # restarted (in case of a config change and if we're also a controller)
 execute "trigger-nova-own-az-config" do
   command "true"
+end
+
+# Set iptables rules for blocking VNC Access for all but the nova-controller node.
+bash "nova_compute_vncblock_reject_all" do
+  code <<-EOH
+    iptables -D INPUT -j VNCBLOCK -p tcp --match multiport --dports 5900:15900
+    iptables -I INPUT -j VNCBLOCK -p tcp --match multiport --dports 5900:15900
+    iptables -F VNCBLOCK
+    iptables -I VNCBLOCK -p tcp \
+        -m connbytes --connbytes 0:1024 \
+        --connbytes-dir both --connbytes-mode bytes \
+        -m state --state ESTABLISHED \
+        -m u32 --u32 "0>>22&0x3C@ 12>>26&0x3C@ 0=0x52464220" \
+        -m string --algo kmp --string "RFB 003." --to 130 \
+        -j REJECT --reject-with tcp-reset
+    EOH
+end
+
+# Filter out packets which use VNC protocol as per RFB 003 using the u32 module
+# for all possible hosts with role nova_multi_controller.
+nova_controller_ips.each do |nova_controller_ip|
+  bash "nova_compute_vncblock_allow_few" do
+    code <<-EOH
+      iptables -I VNCBLOCK -p tcp \
+        -m connbytes --connbytes 0:1024 \
+        --connbytes-dir both --connbytes-mode bytes \
+        -m state --state ESTABLISHED \
+        -m u32 --u32 "0>>22&0x3C@ 12>>26&0x3C@ 0=0x52464220" \
+        -m string --algo kmp --string "RFB 003." --to 130 \
+        -s #{nova_controller_ip} \
+        -j ACCEPT
+    EOH
+  end
 end


chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -418,3 +418,12 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
 execute "trigger-nova-own-az-config" do
   command "true"
 end
+
+# Set iptables rules for blocking VNC Access for all but the nova-controller node.
+bash "vnc_access" do
+  code <<-EOH
+    touch /tmp/hello_pranav
```
mjura: We don't need this

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -418,3 +418,12 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
...
+    iptables -I INPUT -p tcp --match multiport --dports 5900:15900 -j DROP
+    iptables -I INPUT -s #{nova_controller_ip} -p tcp --match multiport --dports 5900:15900 -j ACCEPT
```
mjura: This two iptables rules can be replaced with one rule

mjura:
We should consider to use for this task /etc/host.deny TCP wrapper instead iptables which is managed by OpenStack on compute nodes.

chef/cookbooks/nova/recipes/compute.rb
dguitarbite: Removing this. It sneaked into this patch.

chef/cookbooks/nova/recipes/compute.rb
dguitarbite: I am taking a look into this. If I find a good way to do this in one line, ill update it but as far as I could search for a single line rule, people have mentioned and preferred using multiple lines for such tasks.

dguitarbite:
@mjura I have tried to use host.deny/allow. There are two issues here:

1. host.deny/allow does not allow us to select port range.
2. It has some similar limitations: The VNC ports are opened by qemu service that gives two more challenges:
a. Blocking qemu may break other services (live migration etc.) esp. since we cannot block specific ports.
b. Qemu by default in the Cloud product is not compiled with tcp flag, hence it bypasses hosts.allow/deny rules completely.

chef/cookbooks/nova/recipes/compute.rb
ovisan: 1. hello_pranav ? please change the file name to something professional
2. delete the file after execution

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -418,3 +418,12 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
 execute "trigger-nova-own-az-config" do
   command "true"
 end
+
+# Set iptables rules for blocking VNC Access for all but the nova-controller node.
+bash "vnc_access" do
```
tboerger: This is a pretty bad resource name. This name have to be unique within chef.

tboerger:
Gating fails, bad naming, dummy files... -1

chef/cookbooks/nova/recipes/compute.rb
mjura: We would like to run system commands using Chef calls with escaping.

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -418,3 +418,12 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
...
+  code <<-EOH
+    touch /tmp/hello_pranav
+    iptables -I INPUT -p tcp --match multiport --dports 5900:15900 -j DROP
```
MaximilianMeister: i would decide either to use the full name of the options or the abbreviation, mixed like this it looks a bit weird

rsalevsky:
Commit message is wrong we use `bsc#XXXXXX` without a space. Maybe add the complete URL to the commit description.

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -418,3 +418,23 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
 execute "trigger-nova-own-az-config" do
   command "true"
 end
+
+admin_net = node[:network][:networks][:admin]
+admin_subnet = admin_net[:subnet] + "/" + admin_net[:netmask]
+
+# Set iptables rules for blocking VNC Access for all but the nova-controller node.
+bash "vncblock" do
```
tboerger: Still bad resource naming, if some other cookbook gives the same name to another resource the resource will just be reused. So avoid that!

dguitarbite: Do you suggest a name here? I do not know how exactly to name this resource.

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -418,3 +418,23 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
...
+  code <<-EOH
+    iptables -N VNCBLOCK
+    # Ceph uses range 6800-7300~ which potentially conflicts with the VNC ports range.
+    # Setting the upper limit to 6799 to avoid blocking Ceph traffic for the time being.
+    # FIXME (psalunke): Update it to 15900 or more after changing the default Ceph ports.
+    # Also consider open ports on controller node in case of dual Controller/Compute node
+    # multiport --dports 8773:8779
+    iptables -D INPUT -j VNCBLOCK
+    iptables -I INPUT -j VNCBLOCK
+    iptables -F VNCBLOCK
```
mjura: Let's understand this properly:
- first you create a chain, 
- after this you delete chain with target, 
- then you create this same chain with a target
- and at the end you are flushing all rules to add your rule
For me it is insane

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -418,3 +418,23 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
...
+
+# Set iptables rules for blocking VNC Access for all but the nova-controller node.
+bash "vncblock" do
+  code <<-EOH
+    iptables -N VNCBLOCK
+    # Ceph uses range 6800-7300~ which potentially conflicts with the VNC ports range.
+    # Setting the upper limit to 6799 to avoid blocking Ceph traffic for the time being.
+    # FIXME (psalunke): Update it to 15900 or more after changing the default Ceph ports.
+    # Also consider open ports on controller node in case of dual Controller/Compute node
+    # multiport --dports 8773:8779
```
mjura: On port 5984 couchdb is also listening

chef/cookbooks/nova/recipes/compute.rb
mjura: I will repeat it once again, for system commands we would like to use Chef calls

tboerger: What about `nova_compute_vncblock`?

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -418,3 +418,23 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
 execute "trigger-nova-own-az-config" do
   command "true"
 end
+
+admin_net = node[:network][:networks][:admin]
+admin_subnet = admin_net[:subnet] + "/" + admin_net[:netmask]
+
+# Set iptables rules for blocking VNC Access for all but the nova-controller node.
```
mjura: In comment you mention about `nova-controller` but you are allowing for whole admin network, please make it strict

mjura:
If you want to delete and recreate iptable chains every each run of chef-client, then I would prefer to check on which ports nova services are listening and use only these specific ports. This was proposed by @dirkmueller during discussion about this card.

dguitarbite:
WIP: Please do not merge this.

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -285,16 +285,22 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
 
 env_filter = " AND nova_config_environment:#{node[:nova][:config][:environment]}"
 nova_controller = search(:node, "roles:nova-multi-controller#{env_filter}")
+nova_controller_ips = Array.new
```
tboerger: Just `nova_controller_ips = []`

tboerger:
I can't believe that we need such complicated rules...

dguitarbite:
You can merge this code now. @tboerger I agree, its too complex and we are doing some really not-kool-stuff in there :(.

toabctl:
Hm. No PR testing here? Why?

dguitarbite:
Just meant to say that its no longer WIP.

tboerger:
IMHO it looks too complicated. Somebody with enough iptabpes knowledge needs to give some hints, please!

ovisan:
we are doing? who is we?

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -418,3 +424,36 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
...
+bash "nova_compute_vncblock_reject_all" do
+  code <<-EOH
+    iptables -D INPUT -j VNCBLOCK -p tcp --match multiport --dports 5900:15900
+    iptables -I INPUT -j VNCBLOCK -p tcp --match multiport --dports 5900:15900
+    iptables -F VNCBLOCK
+    iptables -I VNCBLOCK -p tcp \
+        -m connbytes --connbytes 0:1024 \
+        --connbytes-dir both --connbytes-mode bytes \
+        -m state --state ESTABLISHED \
+        -m u32 --u32 "0>>22&0x3C@ 12>>26&0x3C@ 0=0x52464220" \
```
ovisan: Can you please explain what this cryptic line does?

PR #6, opened 2015-08-19
If tempest is deployed, load docker image at docker compute node

jsuchome:
so that tempest could run tests with docker based image.

The image is packaged in tempest barclamp. We can't do this from
tempest recipe because tempest might be deployed on another node
which does not have access to the docker comput node.

Tex version is here https://github.com/crowbar/barclamp-nova/pull/481

This should work together with https://github.com/crowbar/crowbar-openstack/pull/18

### DIFF
--- a/chef/cookbooks/nova/recipes/docker.rb
+++ b/chef/cookbooks/nova/recipes/docker.rb
@@ -19,3 +19,23 @@
 #
 
 node.set[:nova][:libvirt_type] = "docker"
+
+# if tempest is deployed, load packaged docker image at docker compute node
+tempest_node = search(:node, "roles:tempest").first
+unless tempest_node.nil?
+  docker_image = tempest_node[:tempest][:tempest_test_docker_image] || ""
+
+  bash "load docker image" do
+    code <<-EOH
+      TEMP=$(mktemp -d)
+      IMG_FILE=$(basename #{docker_image})
+
+      wget --no-verbose #{docker_image} --directory-prefix=$TEMP 2>&1 || exit $?
+      docker load --input=$TEMP/$IMG_FILE
+      mkdir -p /var/lib/crowbar
+      touch /var/lib/crowbar/docker_for_tempest_loaded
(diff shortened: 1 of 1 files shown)


chef/cookbooks/nova/recipes/docker.rb
```diff
@@ -19,3 +19,21 @@
...
+tempest_node = search(:node, "roles:tempest").first
+unless tempest_node.nil?
+  docker_image = tempest_node[:tempest][:tempest_test_docker_image] || ""
+
+  bash "load docker image" do
+    code <<-EOH
+      TEMP=$(mktemp -d)
+      IMG_FILE=$(basename #{docker_image})
+
+      wget --no-verbose #{docker_image} --directory-prefix=$TEMP 2>&1 || exit $?
```
vuntz: Aren't we doing the wget on every run?

mjura:
I would prefer to use curl instead of wget but looks OK. After discussion and getting some more explanation +1 from me

chef/cookbooks/nova/recipes/docker.rb
jsuchome: True. We could do some `docker images` run for test, but it might be too slow as well. So I propose to create just some tag, create some file that docker image was already pulled. Do we have a place on nodes where it is common to write such info?

chef/cookbooks/nova/recipes/docker.rb
```diff
@@ -19,3 +19,22 @@
...
+      docker load --input=$TEMP/$IMG_FILE
+      touch /var/lib/docker_for_tempest_loaded
```
jsuchome: So, this is probably ugly, but I hoped good enough... but if there is better path to put it, I'd gladly use it

vuntz: I'd put it in /var/lib/crowbar/ since we might want other flags like this.

aplanas:
LGTM +1

PR #8, opened 2015-08-21
nova: Correctly configure haproxy for metadata when ssl is on

vuntz:


tboerger:
:+1:

aspiers:
+1 and gate failed for other reasons.

PR #9, opened 2015-08-21
Replace crowbar.yml with registry classes

tboerger:


tboerger:
I won't do it in near future, closing it for now.

PR #11, opened 2015-08-25
Enables deactivate image test

sayalilunkad:


aspiers:
Please can you [update the commit message](http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/) to explain why this change is needed?  Thanks!

dirkmueller:
+1, assuming it works :-)

aspiers:
Thanks a lot for the update!  It now nicely explains what the change is, but still doesn't explain *why* it's needed.  It's important for the commit message to capture this as mentioned at the bottom of https://etherpad.nue.suse.com/p/cloud-pull-requests and also in this best practices document: https://wiki.openstack.org/wiki/GitCommitMessages#Information_in_commit_messages

(P.S. I guess I probably missed a Trello card somewhere which explains why the change is required, but even adding a link from the commit message to that would not be right since the Trello boards are internal and this is a F/OSS project.  That's covered by the "Do not assume the reviewer has access to external web services/site." point in the above wiki page.)

dirkmueller:
@aspiers I think the git commit message explains why the change is needed (at least to me..).

tboerger:
I'm fine with this commit message :+1:

aspiers:
@dirkmueller Thanks, I misread it the first time. +1

PR #18, opened 2015-08-31
Move PR#162 from barclamp-tempest to crowbar-openstack

aplanas:


jsuchome:
+1 (needs https://github.com/crowbar/crowbar-openstack/pull/6 for docker  load on nodes)

chef/data_bags/crowbar/migrate/tempest/011_add_docker_image.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade ta, td, a, d
```
rsalevsky: should `def upgrade(ta, td, a, d)`

rsalevsky:
Please the one hound issue with my comment and rebase against master to avoid the other violations.

chef/data_bags/crowbar/migrate/tempest/011_add_docker_image.rb
aplanas: This is what HoundCI says, but is quite inconsistent with the rest of the migrations.  I am not sure what is worse here: have a Hound complain or having a migration different from the others

rsalevsky: All new migrations should have the new format, we started already with this.

vuntz:
+1

rsalevsky:
+1

PR #19, opened 2015-09-04
[UNTESTED] ensure mongodb is installed/configured before clustering (bsc#944417)

aspiers:
Just like with all other HA OpenStack resources, we need to prevent the resources being configured in Pacemaker before we know that all nodes have the necessary packages installed and correctly configured.

This was one of the lessons learnt during analysis of:

  https://bugzilla.suse.com/show_bug.cgi?id=944417

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/mongodb.rb
+++ b/chef/cookbooks/ceilometer/recipes/mongodb.rb
@@ -46,6 +46,11 @@ service mongo_service do
 end
 
 if ha_enabled
+  # Wait for all nodes to reach this point so we know that all nodes
+  # will have all the required packages installed and configured before
+  # we create the pacemaker resources.
+  crowbar_pacemaker_sync_mark "sync-mongodb_before_ha"
+
   crowbar_pacemaker_sync_mark "wait-mongodb_service"
 
   pacemaker_primitive "mongodb" do


vuntz:
Actually, it's not perfect as, as discussed on Friday, this will not work if the cluster has 4 nodes (since the mongodb bit will only run on 3 nodes, and the sync mark will block forever). I don't have a good idea on how to fix this :/

aspiers:
@vuntz Simple - we should add Crowbar validation to prevent clusters with even numbers of nodes > 2.

aspiers:
@vuntz How about something like this? https://github.com/crowbar/crowbar-ha/pull/18

chef/cookbooks/ceilometer/recipes/mongodb.rb
```diff
@@ -37,6 +37,11 @@
   notifies :restart, "service[#{mongo_service}]", :immediately
 end
 
+# Wait for all nodes to reach this point so we know that all nodes
+# will have all the required packages installed and configured before
+# we create the pacemaker resources
+crowbar_pacemaker_sync_mark "sync-mongodb_before_ha"
+
```
vuntz: Can you move it inside the `if ha_enabled` block?

aspiers: Technically that's not necessary since the `sync_mark` primitives are a nop if there is no cluster, but yeah, I agree that's nicer - done.

aspiers:
@vuntz Thoughts on what we should do with this one?

vuntz:
@aspiers we should get it in.

aspiers:
@vuntz Agreed.

matelakat:
What's the story with this PR? It's quite old, and no activity on it for a long time? If it's no longer valid, please close it. Thanks.

rsalevsky:
MongoDB got removed in C8

PR #20, opened 2015-09-06
add sample input files for `crowbar batch build`

aspiers:
This should allow people to "kick the tyres" even quicker.  They originate from https://github.com/SUSE-Cloud/suse-cloud-vagrant/tree/master/vagrant/provisioning/admin and @dirkmueller suggested that some of the tweaks in the appliances / Vagrant boxes should be ported into the product.  So this is the first step towards that.

### DIFF
--- a/crowbar_framework/examples/HA-cloud.yaml
+++ b/crowbar_framework/examples/HA-cloud.yaml
@@ -1,0 +1,122 @@
+# Example input file for 'crowbar batch build' command which sets up a
+# cloud with a single 2-node HA cluster using DRBD and SBD for
+# STONITH, and a single KVM compute node.
+#
+# Assumes that the two controller nodes are aliased 'controller1' and
+# 'controller2', and the compute node is aliased 'compute1'.
+#
+# If you are short of physical or virtual hardware and only want to
+# test the controller cluster, you can skip the compute node via:
+#
+#   crowbar batch build -e cinder -e nova
+#
+---
+global_options:
+- action_for_existing_proposals: skip   # could also be e.g. 'recreate' or 'overwrite'
+proposals:
+- barclamp: provisioner
+  attributes:
+    shell_prompt: USER@ALIAS:CWD SUFFIX
+- barclamp: pacemaker
+  name: cluster1
+  action_if_exists: overwrite
+  attributes:
+    stonith:
+      mode: sbd
+      sbd:
+        nodes:
+          "@@controller1@@":
+            devices:
+            - /dev/sdc
+          "@@controller2@@":
+            devices:
+            - /dev/sdc
+    drbd:
+      enabled: true
+  deployment:
+    elements:
+      hawk-server:
+      - "@@controller1@@"
+      - "@@controller2@@"
+      pacemaker-cluster-member:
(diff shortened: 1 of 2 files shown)


toabctl:
+1

crowbar_framework/examples/HA-cloud-no-compute.yaml
```diff
@@ -0,0 +1,117 @@
...
+        nodes:
+          "@@controller1@@":
+            devices:
+            - /dev/sdc
+          "@@controller2@@":
+            devices:
+            - /dev/sdc
+    drbd:
+      enabled: true
+    setup_hb_gui: true
```
vuntz: Don't we want to drop that attribute (I guess it's gone in SLE12)?

crowbar_framework/examples/HA-cloud-no-compute.yaml
```diff
@@ -0,0 +1,117 @@
...
+- barclamp: database
+  # Proposal name defaults to 'default'.
+  # Default attributes are good enough, so we just need to assign
+  # nodes to roles:
+  attributes:
+    ha:
+      storage:
+        mode: drbd
+        drbd:
+          size: 1
```
vuntz: I'd make the size bigger by default.

crowbar_framework/examples/HA-cloud-no-compute.yaml
```diff
@@ -0,0 +1,117 @@
+# Example input file for 'crowbar batch build' command which sets up a
+# cloud with a single 2-node HA cluster using DRBD and SBD for
+# STONITH, and no compute node.
```
vuntz: Btw, what's the benefit of this one compared to the one with a compute node?

crowbar_framework/examples/HA-cloud-no-compute.yaml
aspiers: I'll make it 10GB.

crowbar_framework/examples/HA-cloud-no-compute.yaml
```diff
@@ -0,0 +1,117 @@
...
+    elements:
+      database-server:
+        - cluster:cluster1
+- barclamp: rabbitmq
+  attributes:
+    ha:
+      storage:
+        mode: drbd
+        drbd:
+          size: 1
```
aspiers: Same here.

crowbar_framework/examples/HA-cloud-no-compute.yaml
aspiers: Done.

crowbar_framework/examples/HA-cloud-no-compute.yaml
aspiers: It will work nicely if you only have 16GB RAM, or maybe even if you have 8GB.  Or if you only have 3 spare physical machines :-)  However I realised that the same can be achieved via `crowbar batch build -e` so I'm dropping this file and adding an appropriate comment to `HA-cloud.yaml`.

vuntz:
+1

toabctl:
+1

PR #22, opened 2015-09-09
[WIP] Adds sqlalchemy pool values for nova.

dguitarbite:
Add sqlalchemy pool values to avoid hardcoding DB pool values for nova.

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -2748,7 +2748,7 @@ connection=<%= @database_connection %>
 # (integer value)
 # Deprecated group/name - [DEFAULT]/sql_max_pool_size
 # Deprecated group/name - [DATABASE]/sql_max_pool_size
-#max_pool_size=<None>
+max_pool_size=<%= node[:nova]][:db][:max_pool_size] %>
 
 # Maximum db connection retries during startup. (setting -1
 # implies an infinite retry count) (integer value)
@@ -2766,7 +2766,7 @@ connection=<%= @database_connection %>
 # (integer value)
 # Deprecated group/name - [DEFAULT]/sql_max_overflow
 # Deprecated group/name - [DATABASE]/sqlalchemy_max_overflow
-#max_overflow=<None>
+max_overflow=<%= node[:nova][:db][:max_overflow] %>
 
 # Verbosity of SQL debugging information. 0=None,
 # 100=Everything (integer value)
@@ -2781,7 +2781,7 @@ connection=<%= @database_connection %>
 # If set, use this value for pool_timeout with sqlalchemy
 # (integer value)
 # Deprecated group/name - [DATABASE]/sqlalchemy_pool_timeout
-#pool_timeout=<None>
+pool_timeout=<%= node[:nova][:db][:pool_timeout] %>
 
 # Enable the experimental use of database reconnect on
 # connection lost (boolean value)
--- a/chef/data_bags/crowbar/bc-template-nova.json
+++ b/chef/data_bags/crowbar/bc-template-nova.json
@@ -30,7 +30,10 @@
       "db": {
         "password": "",
         "user": "nova",
-        "database": "nova"
+        "database": "nova",
+        "max_pool_size": 30,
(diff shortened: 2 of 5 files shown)


toabctl:
I think you need a migration. 
What do you think about adding this to ```helpers.rb``` ? So we could set the pool values for all components (not only nova) globally. Or do we need this more fine grained on a per-component basis?

dguitarbite:
@toabctl I saw the migration scripts for Cinder already present in there which should be reused IMO.

I am not sure if its a good idea to add this to helpers.rb to be reused, because these values should be independently customizable for each service. It may vary as per the architecture of deployment.

chef/data_bags/crowbar/bc-template-nova.json
```diff
@@ -31,6 +31,10 @@
         "password": "",
         "user": "nova",
         "database": "nova"
+        "max_pool_size": 30,
```
dirkmueller: syntax error

dirkmueller:
migration is missing

dguitarbite:
Migrations added. Thanks Dirk.

toabctl:
```033_add_configdrive.rb``` shouldn't be deleted

File: chef/data_bags/crowbar/migrate/nova/033_add_configdrive.rb [hunk 2]
```diff
@@ -1,13 +0,0 @@
...
-  end
-  return a, d
-end
-
-def downgrade(ta, td, a, d)
-  unless ta.has_key? "force_config_drive"
-    a.delete("force_config_drive")
-  end
-  return a, d
-end
```
aplanas: Why this migration is removed?

dguitarbite: Its a mistake, readding this file.

chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -2748,7 +2748,7 @@ connection=<%= @database_connection %>
 # (integer value)
 # Deprecated group/name - [DEFAULT]/sql_max_pool_size
 # Deprecated group/name - [DATABASE]/sql_max_pool_size
-#max_pool_size=<None>
+max_pool_size=<%= node[:nova][:max_pool_size] %>
```
dirkmueller: its [:nova][:db][:max_pool_size] on how you added it to the configuration..

chef/data_bags/crowbar/migrate/nova/034_add_sqlalchemy_pool_values.rb
```diff
@@ -0,0 +1,25 @@
+def upgrade(ta, td, a, d)
+  unless ta.has_key? "max_pool_size"
+    a["max_pool_size"] = ta["max_pool_size"]
```
dirkmueller: the migration isn't checking the right keys. the keys are under db in your json update. Where did you want to put it really?

dirkmueller:
-1 as it is, I suggest you test this before asking for reviews.. code review shouldn't find such glaring errors imho.

chef/data_bags/crowbar/migrate/nova/034_add_sqlalchemy_pool_values.rb
dguitarbite: I wanted to up it under DB section but this migration checking code is unclear for me. Ill test it and send the correct code for review.

chef/data_bags/crowbar/migrate/nova/034_add_sqlalchemy_pool_values.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+  unless ta["db"].has_key? "pool_timeout"
+    a["db"]["pool_timeout"] = ta["db"]["pool_timeout"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless ta["db"].has_key? "max_pool_size"
+    a["max_pool_size"].delete("max_pool_size")
```
dirkmueller: a['db']['...']

dguitarbite:
Ill reopen it once I have tested it properly.

PR #23, opened 2015-09-09
Let the network barclamp create OVS bridges

rhafer:
This depends on: https://github.com/crowbar/crowbar-core/pull/33

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -96,7 +96,7 @@ if node[:platform] == "suse" && node[:platform_version].to_f >= 12.0
   end
   bridges.each do |name|
     service "ovs-usurp-config-#{name}" do
-      # There's no need to stop anything here. I might even cut us off the 
+      # There's no need to stop anything here. I might even cut us off the
       # network.
       action [:disable]
       only_if { ::File.exists?("/etc/init.d/ovs-usurp-config-#{name}") }
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -49,45 +49,8 @@ bash "reload disable-rp_filter-sysctl" do
   subscribes :run, resources(cookbook_file: disable_rp_filter_file), :delayed
 end
 
-# openvswitch installation and configuration. It is only needed when using nsx or
-# when deploying on SLE11. Otherwise it's done as part of the network barclamp.
-# FIXME: This code can be cleaned up and moved to the vmware_suppport recipe
-# once SLE11 is no longer supported
-if neutron[:neutron][:networking_plugin] == "vmware" ||
-    (node[:platform] == "suse" && node[:platform_version].to_f < 12.0 &&
-      neutron[:neutron][:networking_plugin] == "ml2" &&
-      neutron[:neutron][:ml2_mechanism_drivers].include?("openvswitch"))
-
-  if node.platform == "ubuntu"
-    # If we expect to install the openvswitch module via DKMS, but the module
-    # does not exist, rmmod the openvswitch module before continuing.
-    if node[:network][:ovs_pkgs].any?{ |e|e == "openvswitch-datapath-dkms" } &&
-        !File.exists?("/lib/modules/#{%x{uname -r}.strip}/updates/dkms/openvswitch.ko") &&
-        File.directory?("/sys/module/openvswitch")
-      if IO.read("/sys/module/openvswitch/refcnt").strip != "0"
-        Chef::Log.error("Kernel openvswitch module already loaded and in use! Please reboot me!")
-      else
-        bash "Unload non-DKMS openvswitch module" do
-          code "rmmod openvswitch"
-        end
-      end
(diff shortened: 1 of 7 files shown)


rhafer:
BTW, the mkcloud ci won't get green on this until is merged https://github.com/crowbar/crowbar-core/pull/33

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -268,6 +268,65 @@ def validate_proposal_after_save proposal
...
+        # need to make sure that the network recipe does NOT create the
+        # VLAN interfaces (ethX.VLAN)
+        if node.crowbar["crowbar"]["network"]["nova_fixed"]["use_vlan"]
+          @logger.info("Forcing use_vlan to false for the nova_fixed network on node #{nodename}")
+          node.crowbar["crowbar"]["network"]["nova_fixed"]["use_vlan"] = false
+          node.save
+        end
+      end
+      node = NodeObject.find_node_by_name nodename
+      update_ovs_bridge_attributes(attributes, node)
```
vuntz: Just wondering: should we also change the attribute in the network proposal?

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -69,21 +74,14 @@
...
-  else
-    openvswitch_service = "openvswitch-switch"
-  end
-
-  service "openvswitch_service" do
-    service_name openvswitch_service
+  service node[:network][:ovs_service] do
     supports status: true, restart: true
     action [:start, :enable]
   end
```
vuntz: Is this still needed since it's done at https://github.com/crowbar/crowbar-core/pull/33/files#diff-43125a344f7a7312bf484096c5e371ecR256 ?

crowbar_framework/app/models/neutron_service.rb
rhafer: I decided for not doing that (but can be convinced otherwise) because that setting is pretty node specific. There are cases where we might need to allocate an IP from the "nova-floating" network on a specific node but where we don't need the bridge. (E.g. when using separate public and floating networks the tempest node needs to have an IP in floating).

vuntz: Ah, good reason. Can you document that in a comment (or in the commit message, although it's less easy to find)?

chef/cookbooks/neutron/recipes/common_agent.rb
rhafer: This is where it gets ugly. Technically this is supposed to ony run on SLE11 setup and setups using the "vmware" networking plugin. But I am still unsure how we should handle SLE11 I am somewhat reluctant to add large ```if sle12 or newer``` conditions all around the https://github.com/crowbar/crowbar-core/pull/33 PR as it makes the code even more ugly. OTOH cleanly implementing the SLE11 support in the network barclamp is probably not too hard (I think it's "just" a matter of creating the usurp scripts there). But it might just be a waste of time as we still don't know whether we'll need it. 

It's probably best to cleanout the remaining SLE11 bits here. And reimplement SLE11 support in barclamp-network when we really need it.

Would that be ok for you?

vuntz: Works for me.

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -268,6 +268,67 @@ def validate_proposal_after_save proposal
...
+        if node.crowbar["crowbar"]["network"][net]
+          unless node.crowbar["crowbar"]["network"][net]["add_ovs_bridge"]
+            @logger.info("Forcing add_ovs_bridge to true for the #{net} network on node #{node.name}")
+            node.crowbar["crowbar"]["network"][net]["add_ovs_bridge"] = true
+            needs_save = true
+          end
+        end
+      end
+    end
+    node.save if needs_save
```
vuntz: I guess we need a way to also set the `add_ovs_bridge` attribute to false when ovs is removed (although it's probably not working right now, but this won't hurt much). Maybe just a quick loop over all the networks on the node?

vuntz:
One minor comment, but other than that, +1

rhafer:
> I guess we need a way to also set the add_ovs_bridge attribute to false when ovs is removed (although it's probably not working right now, but this won't hurt much). Maybe just a quick loop over all the networks on the node?

@vuntz please see the last commit for that. Does that make sense. I must admit I am not entirely sure yet, if that is the correct thing to do in all cases.

vuntz:
@rhafer that looks correct to me; of course, it'll likely break one way or the other when moving from ovs to something else, but that's a good-enough first step, imho.

rhafer:
> that looks correct to me; of course, it'll likely break one way or the other when moving from ovs to something else, but that's a good-enough first step, imho.

Actually there was still a tiny issue in it ( ```node.crowbar["crowbar"]["network"].each``` vs ```node.crowbar["crowbar"]["network"].keys.each```) but that should be fixed now.

vuntz:
+1

rhafer:
Removed the [WIP] tag.

dirkmueller:
+1

PR #24, opened 2015-09-10
Set dhcp_client to empty value will skip test_subnet_details

jsuchome:
test that is not supported by docker instances.

Requires https://review.openstack.org/#/c/221874/


aplanas:
If you put that into the package as a patch we can merge this now

vuntz:
+1

aplanas:
+1

PR #25, opened 2015-09-11
Exclude unsupported volume test when testing with docker

jsuchome:


rsalevsky:
+1

vuntz:
Code change is fine, but the gating fails in tempest. Can you check if it's related to your change or something else?

bmwiedemann:
:+1:

PR #26, opened 2015-09-11
Cleanup keystone v3 attribute and enable multi-domain support

rhafer:
This is  https://github.com/crowbar/barclamp-nova_dashboard/pull/227 ported to the new repo layout

### DIFF
--- a/crowbar_framework/app/models/horizon.rb
+++ b/crowbar_framework/app/models/horizon.rb
@@ -75,6 +75,16 @@ class HorizonService < PacemakerServiceObject
   def validate_proposal_after_save proposal
     validate_one_for_role proposal, "horizon-server"
 
+    if proposal["attributes"][@bc_name]["multi_domain_support"]
+      ks_svc = KeystoneService.new @logger
+      keystone = Proposal.where(barclamp: ks_svc.bc_name,
+                                name: proposal["attributes"]["horizon"]["keystone_instance"]).first
+      # Using domains requires API Version 3 or newer
+      if keystone["attributes"][ks_svc.bc_name]["api"]["version"].to_f < 3.0
+        validation_error("Multi domain support requires enabling Keystone V3 API in the keystone proposal first.")
+      end
+    end
+
     super
   end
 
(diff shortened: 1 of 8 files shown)


vuntz:
@rhafer argh, sorry, I merged the renaming to horizon and so this needs to be rebased.

chef/data_bags/crowbar/migrate/nova_dashboard/034_add_multi_domain_support.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
```
vuntz: Path should be horizon, not nova_dashboard (for the two migrations).

rsalevsky:
Can we combine the migrations?

crowbar_framework/app/models/horizon.rb
```diff
@@ -75,6 +75,16 @@ def create_proposal
   def validate_proposal_after_save proposal
     validate_one_for_role proposal, "horizon-server"
 
+    if proposal["attributes"][@bc_name]["multi_domain_support"]
+      ks_svc = KeystoneService.new @logger
+      keystone = Proposal.where(barclamp: ks_svc.bc_name,
```
vuntz: Heh, any reason to not just put "keystone" instead of getting the KeystoneService and then fetching an attribute from there?

(not that I mind, just wondering)

vuntz:
+1 (but I'm fine if changes suggested are added)

dirkmueller:
can you add '''bsc#945052''' to the git commit summary of the 2nd commit?

rhafer:
> Can we combine the migrations?

We can, but I would really like to keep them separate. Is there a good reason to not have the separate?

rhafer:
> perhaps check here if keystone is nonempty (I'm not sure if we can guarantee that the keystone proposal exists and is not deactivated at this point in time..)

Hm, actually horizon requires keystone. It's not possible to create a horzion proposal when no active keystone proposal exists. It's also not possible to deactivate keystone when there is an active horizon proposal.

rhafer:
Feedback addressed.

dirkmueller:
+1

toabctl:
+1

toabctl:
Unrelated PR testing failure (hopefully fixed by https://github.com/SUSE-Cloud/automation/pull/563 )

PR #27, opened 2015-09-14
Support installation of Neutron ZVM Driver

dirkmueller:
Allow deploying with the ZVM L2 driver.


tboerger:
Beside the style offenses LGTM

chef/data_bags/crowbar/bc-template-neutron.schema
```diff
@@ -83,6 +83,11 @@
                       "tz_uuid": { "type" : "str", "required" : true },
                       "l3_gw_uuid": { "type" : "str", "required" : true }
                     }},
+                    "zvm": { "type": "map", "required": true, "mapping": {
+                      "zvm_xcat_server": { "type": "str", "required": true },
+                      "zvm_xcat_username": { "type": "str", "required": true },
+                      "zvm_xcat_password": { "type": "str", "required": true }
+                    }},
```
MaximilianMeister: will these 3 parameters then be taken out of nova? you have mentioned some day IIRC that nova could use these attributes from neutron, as neutron is required by nova anyway.

chef/cookbooks/neutron/templates/default/neutron_zvm_plugin.ini.erb
```diff
@@ -0,0 +1,52 @@
+[AGENT]
```
vuntz: I don't see any code using the template? Not there yet, or forgotten?

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -63,6 +63,8 @@
     ovs_agent_name: "openstack-neutron-openvswitch-agent",
     lb_agent_pkg: "openstack-neutron-linuxbridge-agent",
     lb_agent_name: "openstack-neutron-linuxbridge-agent",
+    zvm_agent_pkg: "openstack-neutron-zvm-agent",
+    zvm_agent_name: "openstack-neutron-zvm-agent",
```
rhafer: There is nothing that currently uses these attributes. But I guess that's why this is still tagged as [WIP] :)

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -159,6 +160,11 @@
   # on whether one of the external networks will share the physical interface
   # with "nova_fixed".
   external_networks = ["nova_floating"]
+
+  # XXX define the xcat management network when needed
+  if true
+    external_networks += ["xcatvsw1"]
```
bmwiedemann: test failed with
undefined method `to_hash' for nil:NilClass
in
13:      external_networks.each do |net|
14>>       networks[net] = node["network"]["networks"][net].to_hash

MaximilianMeister: not that it fixes the above error, but stylewise maybe use `external_networks.push("xcatvsw1")`

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -112,6 +112,7 @@
 # ML2 configuration: L2 agent and L3 agent
 if neutron[:neutron][:networking_plugin] == "ml2"
   ml2_mech_drivers = neutron[:neutron][:ml2_mechanism_drivers]
+  ml2_mech_drivers = ["zvm"] if node["kernel"]["machine"] =~ /s390x/
```
mmnelemane: Here still "ml2_mech_drivers" is overwritten

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -151,6 +156,13 @@
         interface_mappings += mapping
       end
     end
+  when ml2_mech_drivers.include?("zvm")
```
vuntz: Because of this code, I understand that the user can pick only one of ovs, linuxbridge and zvm. That means we need to change the UI of the barclamp accordingly, but I don't see code for that.

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 
+zvm_compute_node = search(:node, "roles:nova-compute-zvm") || []
```
vuntz: In case you update the PR, can you do something like: `use_zvm = not zvm_compute_nodes.empty?` and then use `use_zvm` (in both recipes).

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -13,8 +13,11 @@
 # limitations under the License.
 #
 
+zvm_compute_node = search(:node, "roles:nova-compute-zvm") || []
+
 pkgs = node[:neutron][:platform][:pkgs] + node[:neutron][:platform][:pkgs_fwaas]
 pkgs += node[:neutron][:platform][:pkgs_lbaas] if node[:neutron][:use_lbaas]
+pkgs << node[:neutron][:platform][:zvm_agent_pkg] if node[:neutron][:networking_plugin] == "ml2"
```
vuntz: Shouldn't we do that only if `ml2_mech_drivers.include?("zvm")` (like in the other recipes)?

vuntz:
Actually, I'm confused as to when the code is running: some tests are checking for zvm compute nodes, some tests are testing for the mechanism to be specified in the proposal. Which one is the right one?

vuntz:
@dirkmueller can we quickly mumble after your refinement on this? I'm still confused and it will be faster to chat :-)

chef/cookbooks/neutron/recipes/common_agent.rb
dirkmueller: done

chef/cookbooks/neutron/recipes/common_agent.rb
dirkmueller: @vuntz you can pick whatever you want, it will only be used for hypervisors that are not the ZVM type. the ZVM one uses vlan networking (hardcoded)

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -169,6 +179,9 @@
   #TODO(vuntz): temporarily disable the hyperv mechanism since we're lacking networking-hyperv from stackforge
   #ml2_mechanism_drivers = node[:neutron][:ml2_mechanism_drivers].dup.push("hyperv")
   ml2_mechanism_drivers = node[:neutron][:ml2_mechanism_drivers].dup
+  if zvm_compute_node.first
```
toabctl: ```if use_zvm``` ...

chef/cookbooks/neutron/recipes/common_agent.rb
vuntz: Just to document some offline discussion: my first concern is invalid because this is added to the array a few lines earlier (`ml2_mech_drivers.push("zvm")`), but my concern about "only one of ovs, linuxbridge and zvm" still stands, because now this means that as soon as we have a zvm compute node, we stop configuring ovs and lb (since it's a `case`, not a `if`).

toabctl:
+1 (assuming you tested it and it works)

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -13,8 +13,14 @@
 # limitations under the License.
 #
 
+zvm_compute_node = search(:node, "roles:nova-compute-zvm") || []
+use_zvm = !zvm_compute_node.empty?
```
vuntz: In case the PR gets updated again (no need to update it just for that):

I know it doesn't change anything to the code path here, but for better readability, I would move the `node[:neutron][:networking_plugin] == "ml2"` test to be part of `use_zvm`.

chef/cookbooks/neutron/recipes/common_agent.rb
dirkmueller: which is correct on the nova-zvm node. I've updated the condition that should fix this concern.

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -13,8 +13,14 @@
 # limitations under the License.
 #
 
+zvm_compute_node = search(:node, "roles:nova-compute-zvm") || []
+use_zvm = node[:neutron][:networking_plugin] == "ml2" && !zvm_compute_node.empty?
+
 pkgs = node[:neutron][:platform][:pkgs] + node[:neutron][:platform][:pkgs_fwaas]
 pkgs += node[:neutron][:platform][:pkgs_lbaas] if node[:neutron][:use_lbaas]
+if use_zvm
+  pkgs << node[:neutron][:platform][:zvm_agent_pkg] if node[:neutron][:networking_plugin] == "ml2"
```
vuntz: No need for the `if node[:neutron][:networking_plugin] == "ml2"` anymore here.

chef/cookbooks/neutron/recipes/common_agent.rb
vuntz: @dirkmueller thanks, now I finally understand :-)

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -112,6 +115,9 @@
 # ML2 configuration: L2 agent and L3 agent
 if neutron[:neutron][:networking_plugin] == "ml2"
   ml2_mech_drivers = neutron[:neutron][:ml2_mechanism_drivers]
+  if use_zvm && node.roles.include?("nova-compute-zvm")
```
vuntz: What about killing `use_zvm` which is only used here and only keeping `if node.roles.include?("nova-compute-zvm")`?

vuntz:
I'm happy to see this merged and I can create a followup PR for the nitpicks if preferred.

+1

aplanas:
LGTM +1

vuntz:
> I can create a followup PR for the nitpicks if preferred.

https://github.com/crowbar/crowbar-openstack/pull/234

rossella:
+1

PR #28, opened 2015-09-16
Add support for external database configuration

mjura:
Add support for external database configuration


vuntz:
Before really reviewing: why is this ceilometer-only? Don't we want this for everything?

mjura:
Let's close it for now

PR #30, opened 2015-09-17
Add warning for horizon timeout

vwallfahrer:
Horizon timeout is impacted by keystone token timeout since when the
token gets outdated, then the user is automatically unable to use
horizon. This patch includes a check for the timeout of horizon and
keystone barclamps.

### DIFF
--- a/crowbar_framework/app/models/horizon_service.rb
+++ b/crowbar_framework/app/models/horizon_service.rb
@@ -90,7 +90,8 @@ class HorizonService < PacemakerServiceObject
 
     # keystone_timeout is in seconds and horizon_timeout is in minutes
     if horizon_timeout * 60 > keystone_timeout
-      validation_error("Horizon timeout is bigger than keystone token expiration and these may cause problems.")
+      validation_error("Setting the Horizon timeout (#{horizon_timeout * 60} minutes) longer than the Keystone token expiration timeout" \ 
+        "(#{keystone_timeout} minutes) is not supported. Please lower the Horizon token timeout.")
     end
 
     super
--- a/crowbar_framework/app/models/horizon_service.rb
+++ b/crowbar_framework/app/models/horizon_service.rb
@@ -84,14 +84,19 @@ class HorizonService < PacemakerServiceObject
         validation_error("Multi domain support requires enabling Keystone V3 API in the keystone proposal first.")
       end
 
-    horizon_timeout = proposal["attributes"]["nova_dashboard"]["session_timeout"]
-    keystone_proposal = Proposal.where(barclamp: "keystone", name: "default").first
-    keystone_timeout = keystone_proposal["attributes"]["keystone"]["token_expiration"]
+      horizon_timeout = proposal["attributes"]["nova_dashboard"]["session_timeout"]
+      keystone_proposal = Proposal.where(barclamp: "keystone", name: "default").first
+      keystone_timeout = keystone_proposal["attributes"]["keystone"]["token_expiration"]
 
-    # keystone_timeout is in seconds and horizon_timeout is in minutes
-    if horizon_timeout * 60 > keystone_timeout
-      validation_error("Setting the Horizon timeout (#{horizon_timeout * 60} minutes) longer than the Keystone token expiration timeout" \ 
-        "(#{keystone_timeout} minutes) is not supported. Please lower the Horizon token timeout.")
+      timeout_warning = <<-EOF
+Setting the Horizon timeout (#{horizon_timeout} minutes) longer than the Keystone token expiration timeout (#{keystone_timeout * 60} minutes) is not supported.
+Please lower the Horizon token timeout.
+EOF
+
+      # keystone_timeout is in seconds and horizon_timeout is in minutes
+      if horizon_timeout * 60 > keystone_timeout
+        validation_error(timeout_warning)
+      end
     end
 
     super
--- a/crowbar_framework/app/models/horizon_service.rb
+++ b/crowbar_framework/app/models/horizon_service.rb
@@ -83,6 +83,14 @@ class HorizonService < PacemakerServiceObject
       if keystone["attributes"][ks_svc.bc_name]["api"]["version"].to_f < 3.0
         validation_error("Multi domain support requires enabling Keystone V3 API in the keystone proposal first.")
       end
+
+    horizon_timeout = proposal["attributes"]["nova_dashboard"]["session_timeout"]
+    keystone_proposal = Proposal.where(barclamp: "keystone", name: "default").first
+    keystone_timeout = keystone_proposal["attributes"]["keystone"]["token_expiration"]
+
+    # keystone_timeout is in seconds and horizon_timeout is in minutes
+    if horizon_timeout * 60 > keystone_timeout
+      validation_error("Horizon timeout is bigger than keystone token expiration and these may cause problems.")
     end
 
     super
--- a/crowbar_framework/app/models/keystone_service.rb
+++ b/crowbar_framework/app/models/keystone_service.rb
@@ -83,6 +83,15 @@ class KeystoneService < PacemakerServiceObject
       validation_error("API version 3 or newer is required when enabling domain specific drivers.")
     end
 
+    keystone_timeout = proposal["attributes"]["keystone"]["token_expiration"]
+    horizon_proposal = Proposal.where(barclamp: "nova_dashboard", name: "default").first
+    horizon_timeout = horizon_proposal["attributes"]["nova_dashboard"]["session_timeout"]
+
+    # keystone_timeout is in seconds and horizon_timeout is in minutes
+    if horizon_timeout * 60 > keystone_timeout
+      validation_error("Horizon timeout is bigger than keystone token expiration and these may cause problems.")
+    end
+
     super
   end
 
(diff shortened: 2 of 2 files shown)


crowbar_framework/app/models/horizon.rb
```diff
@@ -75,6 +75,15 @@ def create_proposal
   def validate_proposal_after_save proposal
     validate_one_for_role proposal, "horizon-server"
 
+    horizon_timeout =  proposal["attributes"]["nova_dashboard"]["session_timeout"]
+    keystone_proposal = Proposal.where(barclamp: "keystone", name: "default").first
+    keystone_timeout = keystone_proposal["attributes"]["keystone"]["token_expiration"]
+
+    # keystone_timeout is in seconds and horizon_timeout is in minutes
+    if horizon_timeout * 60 > keystone_timeout
+        validation_error("Horizon timeout is bigger than keystone token expiration and these may cause problems.")
```
vuntz: I'd suggest rephrasing to: "The session timeout cannot be larger than the Keystone token expiration timeout."

rsalevsky:
Hrm... code looks good so far but it's the same code twice. Maybe but it in an validation helper method and just call the method.

abel-navarro:
@rsalevsky you are right there is some code repeated (3 lines). Since the code is repeated in two different files and it's very specific for the validation, I'd say it would make it harder to read if you need to go to a third file for just this three lines. Would the code had been in the same file or had been more critical, I would totally go for a non-repetition approach.

abel-navarro:
@vuntz 's comment seems to have disappeared after the pull request update. This is his original comment:
> I'd suggest rephrasing to: "The session timeout cannot be larger than the Keystone token expiration timeout."

Btw, I agree with the rephrasing. Although I'd explicitly mention this is horizon's session timeout:
"The **horizon** session timeout cannot be larger than the Keystone token expiration timeout."

abel-navarro:
Just thought of another corner case to try: what happens if we have keystone barclamp installed but not the horizon one? Probably there should be a check when retrieving the proposal to see it's not empty.

abel-navarro:
I checked our default values for the expirations in cloud5:
* horizon: 24 hours (1440 minutes)
* keystone: 4 hours (14400 seconds)

So, the situation that triggers this problem is there by default. There won't be any warning in crowbar because it will only show upon a change. We will need to change the default values.

If we change the default values, should we also change any documentation for the end user?

abel-navarro:
I still haven't fully backported this to cloud5 and probably cannot finish this soon. FTR, this is what the code would look like for cloud5:
```ruby
    horizon_timeout = proposal["attributes"]["nova_dashboard"]["session_timeout"]
    keystone_proposal = ProposalObject.find_proposal("keystone", "default")
    unless keystone_proposal.nil?
      keystone_timeout = keystone_proposal["attributes"]["keystone"]["token_expiration"]

      # keystone_timeout is in seconds and horizon_timeout is in minutes
      if horizon_timeout * 60 > keystone_timeout
        validation_error("The horizon session timeout cannot be larger than the Keystone token expiration timeout.")
      end
    end
```

vuntz:
> If we change the default values, should we also change any documentation for the end user?

I don't think we document that, so go ahead.

abel-navarro:
Regarding the default timeouts, what about making both (horizon + keystone) to be 4 hours?

vuntz:
> Regarding the default timeouts, what about making both (horizon + keystone) to be 4 hours?

We need to make the settings consistent, at least. I'd possibly go for 24 hours, but no strong opinion.

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -83,6 +83,15 @@ def validate_proposal_after_save proposal
       validation_error("API version 3 or newer is required when enabling domain specific drivers.")
     end
 
+    keystone_timeout = proposal["attributes"]["keystone"]["token_expiration"]
+    horizon_proposal = Proposal.where(barclamp: "nova_dashboard", name: "default").first
+    horizon_timeout = horizon_proposal["attributes"]["nova_dashboard"]["session_timeout"]
```
vuntz: In this case, there's no guarantee that we have a horizon proposal, so we should check if `horizon_proposal` is `nil`.

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -83,6 +83,15 @@ def validate_proposal_after_save proposal
...
+
+    # keystone_timeout is in seconds and horizon_timeout is in minutes
+    if horizon_timeout * 60 > keystone_timeout
+      validation_error("Horizon timeout is bigger than keystone token expiration and these may cause problems.")
```
vuntz: s/these/this/

vuntz: Also, s/keystone/Keystone/

crowbar_framework/app/models/horizon.rb
```diff
@@ -75,6 +75,15 @@ def create_proposal
   def validate_proposal_after_save proposal
     validate_one_for_role proposal, "horizon-server"
 
+    horizon_timeout = proposal["attributes"]["nova_dashboard"]["session_timeout"]
+    keystone_proposal = Proposal.where(barclamp: "keystone", name: "default").first
+    keystone_timeout = keystone_proposal["attributes"]["keystone"]["token_expiration"]
+
+    # keystone_timeout is in seconds and horizon_timeout is in minutes
+    if horizon_timeout * 60 > keystone_timeout
+      validation_error("Horizon timeout is bigger than keystone token expiration and these may cause problems.")
```
vuntz: (same comments for the string)

crowbar_framework/app/models/keystone_service.rb
vuntz: And finally, s/timeout/session timeout/. Sorry for all the comments, should have made just one :-)

vuntz:
Also needs a rebase

crowbar_framework/app/models/keystone_service.rb
dirkmueller: you can edit comments ;)

crowbar_framework/app/models/horizon.rb
dirkmueller: I think this should be more specific. it doesn't cause "problems" per se, its just that a Horizon timeout bigger than the keystone token timeout is ignored. 

So I'd rephrase it similar to 

"Setting the Horizon timeout (%d minutes) longer than the Keystone token expiration timeout (%d minutes) is not supported. Please raise the Keystone token timeout in the keystone proposal."

abel-navarro: I'd remove the last part of the message ("please raise..."). The user may want to lower the horizon timeout instead.

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -83,6 +83,21 @@ def validate_proposal_after_save proposal
...
     end
 
+    keystone_timeout = proposal["attributes"]["keystone"]["token_expiration"]
+    horizon_proposal = Proposal.where(barclamp: "nova_dashboard", name: "default").first
+    unless horizon_proposal.nil?
+      horizon_timeout = horizon_proposal["attributes"]["nova_dashboard"]["session_timeout"]
+
+    timeout_warning = <<-EOF
+Setting the Horizon timeout (#{horizon_timeout} minutes) longer than the Keystone token expiration timeout (#{keystone_timeout * 60} minutes) is not supported.
+Please lower the Horizon token timeout.
```
vuntz: Here I would change the error and say "Please increase the Keystone token expiration timeout." (since the user is editing the keystone proposal)

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -83,6 +83,21 @@ def validate_proposal_after_save proposal
       validation_error("API version 3 or newer is required when enabling domain specific drivers.")
     end
 
+    keystone_timeout = proposal["attributes"]["keystone"]["token_expiration"]
+    horizon_proposal = Proposal.where(barclamp: "nova_dashboard", name: "default").first
+    unless horizon_proposal.nil?
+      horizon_timeout = horizon_proposal["attributes"]["nova_dashboard"]["session_timeout"]
+
+    timeout_warning = <<-EOF
```
vuntz: Indeed, missing two spaces here :-)

crowbar_framework/app/models/horizon_service.rb
```diff
@@ -83,6 +83,19 @@ def validate_proposal_after_save proposal
       if keystone["attributes"][ks_svc.bc_name]["api"]["version"].to_f < 3.0
         validation_error("Multi domain support requires enabling Keystone V3 API in the keystone proposal first.")
       end
+
+    horizon_timeout = proposal["attributes"]["nova_dashboard"]["session_timeout"]
+    keystone_proposal = Proposal.where(barclamp: "keystone", name: "default").first
+    keystone_timeout = keystone_proposal["attributes"]["keystone"]["token_expiration"]
+
+    timeout_warning = <<-EOF
+Setting the Horizon timeout (#{horizon_timeout} minutes) longer than the Keystone token expiration timeout (#{keystone_timeout * 60} minutes) is not supported.
```
vuntz: Hrm, shouldn't that be `keystone_timeout / 60`?

tboerger:
I would prefer the messages to be part of the i18n file as they are getting touched anyway.

abel-navarro:
since @vwallfahrer is sick, I'm continuing the PR in this new one: https://github.com/crowbar/crowbar-openstack/pull/51

tboerger:
Continued in #51

PR #31, opened 2015-09-18
manila: Add HA support for manila-api and manila-scheduler

toabctl:


chef/cookbooks/manila/recipes/common.rb
```diff
@@ -36,17 +36,23 @@
...
-  bind_host = "0.0.0.0"
+if node[:manila][:ha][:enabled]
+  node[:manila][:api][:bind_host] = my_ipaddress
+  bind_port = node[:manila][:ha][:ports][:api]
 else
-  bind_host = node[:manila][:api][:bind_host]
+  if node[:manila][:api][:bind_open_address]
+    bind_host = "0.0.0.0"
+  else
+    bind_host = node[:manila][:api][:bind_host]
```
vuntz: I'm a bit confused by your logic with `bind_host` here; isn't `node[:manila][:api][:bind_host]` not set anymore?

toabctl: Good catch. Fixed

chef/cookbooks/manila/recipes/sql.rb
```diff
@@ -31,4 +37,32 @@
...
+    node.save
+  end
+  action :nothing
+  subscribes :create, "execute[manila-manage db sync]", :immediately
+end
+
+crowbar_pacemaker_sync_mark "create-manila_database"
+
+# save data so it can be found by search
+node.save
```
vuntz: Why is this needed?

chef/cookbooks/manila/recipes/controller_ha.rb
```diff
@@ -0,0 +1,82 @@
...
+unless node[:manila][:ha][:enabled]
+  log "HA support for manila is disabled"
+  return
+end
+
+log "HA support for manila is enabled"
+
+haproxy_loadbalancer "manila-api" do
+  address node[:manila][:api][:bind_open_address] ?
+    "0.0.0.0" : node[:manila][:api][:bind_host]
```
vuntz: I don't believe this is correct. This should be the IP address of haproxy (see chef/cookbooks/glance/libraries/helpers.rb)

chef/cookbooks/manila/recipes/controller_ha.rb
```diff
@@ -0,0 +1,82 @@
...
+  port node[:manila][:api][:bind_port]
+  # FIXME(toabctl): implement SSL support
+  # use_ssl (node[:manila][:api][:protocol] == "https")
+  use_ssl false
+  servers CrowbarPacemakerHelper.haproxy_servers_for_service(node,
+                                                             "manila",
+                                                             "manila-server",
+                                                             "api")
+  action :nothing
+end.run_action(:create)
```
aspiers: Wouldn't it be better to do this after the two `sync_mark`s immediately below?  Are you basing this on another barclamp?  If so, which one?  Because that would either mean that the other barclamp is wrong, or that I'm wrong ;-)

chef/cookbooks/manila/recipes/controller_ha.rb
```diff
@@ -0,0 +1,82 @@
...
+
+pacemaker_clone "cl-#{group_name}" do
+  rsc group_name
+  action [:create, :start]
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+crowbar_pacemaker_order_only_existing "o-cl-#{group_name}" do
+  # NOTE(toabctl): Add cinder, nova, neutron and glance here? Because these
+  # are needed by the generic driver
```
vuntz: Can you find out if we're using the generic driver? If yes, then just change the array depending on the driver.

chef/cookbooks/manila/recipes/controller_ha.rb
vuntz: Sorry @aspiers, you're wrong :-)

This resource actually does nothing except setting attribute on the node, so it doesn't matter if it's protected or not.

toabctl: @aspiers it's based on the cinder barclamp code

chef/cookbooks/manila/recipes/controller_ha.rb
toabctl: We can use the generic driver but it's not recommended for production.

chef/cookbooks/manila/recipes/sql.rb
toabctl: copied from the cinder barclamp. But it's not needed. We only change ```node``` at line#58  and there is already a ```node.save```

chef/cookbooks/manila/recipes/controller_ha.rb
toabctl: Hm. Then it's also wrong in the cinder barclamp (see https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/cinder/recipes/controller_ha.rb#L24 ).

chef/cookbooks/manila/recipes/controller_ha.rb
aspiers: @vuntz And there's no risk of the attribute getting set, the sync_mark failing, and then the attribute taking effect on a later chef-client run even before the `sync_mark` gets passed?

vuntz: Nope, the effect is only active in the second phase of chef, so we'd have to pass the sync mark again.

chef/cookbooks/manila/recipes/controller_ha.rb
vuntz: That looks wrong, indeed.

chef/cookbooks/manila/recipes/controller_ha.rb
aspiers: Oh right, I didn't notice the `action :nothing` bit.  Even so it's conceptually uglier this way IMHO.  But whatever :)

chef/cookbooks/manila/recipes/controller_ha.rb
toabctl: Fixed in https://github.com/crowbar/crowbar-openstack/commit/a5126098ee3293bfdb6d867bff7f223e03c0d906

chef/cookbooks/manila/recipes/controller_ha.rb
toabctl: Fixed in https://github.com/crowbar/crowbar-openstack/commit/a5126098ee3293bfdb6d867bff7f223e03c0d906 . I'll prepare a PR for cinder.

chef/cookbooks/manila/recipes/sql.rb
toabctl: Fixed in https://github.com/crowbar/crowbar-openstack/commit/a5126098ee3293bfdb6d867bff7f223e03c0d906

chef/cookbooks/manila/attributes/default.rb
```diff
@@ -20,5 +20,15 @@
...
+if %w(redhat centos suse).include? node.platform
+  default[:manila][:ha][:api_ra] = "lsb:openstack-manila-api"
+  default[:manila][:ha][:scheduler_ra] = "lsb:openstack-manila-scheduler"
+else
+  default[:manila][:ha][:api_ra] = "lsb:manila-api"
+  default[:manila][:ha][:scheduler_ra] = "lsb:manila-scheduler"
+end
+default[:manila][:ha][:op][:monitor][:interval] = "10s"
+# Ports to bind to when haproxy is used for the real ports
+default[:manila][:ha][:ports][:api] = 5521
```
vuntz: Minor nitpick, but can you bump to 5525 (hopefully not used by any well-known service) so that we can have a few ports after 5520 in case cinder needs another service later on?

vuntz:
Minor nitpick, but +1.

chef/cookbooks/manila/attributes/default.rb
toabctl: Done in https://github.com/toabctl/crowbar-openstack/commit/975debef18417b3232e49d9fc9b75960a7176ec2

dirkmueller:
+1

vuntz:
+1

toabctl:
Unrelated PR testing failure:
```
10:41:36 [2015-09-21T08:41:34+00:00] INFO: Processing execute[modprobe xfs] action+ complain 73 'Committing the crowbar '\''default'\'' proposal for '\''neutron'\'' failed (1).'
10:41:36 + local ex=73
10:41:36 + shift
10:41:36 + printf 'Error: %s\n' 'Committing the crowbar '\''default'\'' proposal for '\''neutron'\'' failed (1).'
10:41:36 Error: Committing the crowbar 'default' proposal for 'neutron' failed (1).
10:41:36 + [[ 73 = - ]]
10:41:36 + exit 73
```

PR #36, opened 2015-09-18
cinder: Use VIP for cinder-api haproxy config

toabctl:


vuntz:
+1

dirkmueller:
LGTM codewise

toabctl:
@aspiers any comments on this one?

aspiers:
Looks right to me but why is this handled so inconsistently across the different barclamps?  And also why did the gate fail?

toabctl:
> Looks right to me but why is this handled so inconsistently across the different barclamps? And also why did the gate fail?

About the PR testing failure: It's ```mount: special device /dev/vdb does not exist``` which should be addressed by https://github.com/SUSE-Cloud/automation/pull/563

About the inconsistently: No idea.

PR #38, opened 2015-09-18
Set the right physical_network for the floating network

rhafer:
It needs to be on "floating" not "physnet1". It seems this didn't matter much as
long as we still had the "external_network_bridge" setting in the config. But
that needed to go for mulitple external network support.


vuntz:
Note that this was introduced in https://github.com/crowbar/barclamp-neutron/pull/254

dirkmueller:
@vuntz readded, the code was previously there already. its just that the change of bridge mappings introduced the issue

dirkmueller:
:+1:

rsalevsky:
+1

PR #41, opened 2015-09-23
Check for platform_version higher than SLE12

vwallfahrer:


vuntz:
That's not correct, we depend on the `else` after that for SP1.

PR #43, opened 2015-09-23
Rename horizon.rb to horizon_service.rb

rhafer:
To be consistent with all the other Service models. Seems it got wrong with the
nova_dashboard -> horizon rename.


vuntz:
+1

vwallfahrer:
+1

toabctl:
-1 . Seems to be a patch in the package for ```horizon.rb```. From the log: ```patching file crowbar_framework/app/models/horizon.rb```

vuntz:
@toabctl that's prtest.patch :-) I think the gating just doesn't deal properly with renaming in patches?

PR #46, opened 2015-09-26
glance: Fix default_store option

vuntz:
It now needs to be in the glance_store group. For some reason, the
deprecated location doesn't work.


aplanas:
This is only a sync with the current reference Kilo configuration for Ceph?

In either case +1

vuntz:
> This is only a sync with the current reference Kilo configuration for Ceph?

Yes, a minor sync (not full sync); and not just for ceph :-)

rsalevsky:
+1

dirkmueller:
+1

mjura:
+1

PR #47, opened 2015-09-28
Added new codeclimate config

tboerger:


MaximilianMeister:
:+1:

tboerger:
Got enough votes and config is exactly like the example from Codelimate, merging.

PR #49, opened 2015-09-28
Raise Nova defaults to avoid hangs on mass-deployment (bsc#944489)

dirkmueller:
It turns out Nova does not handle running out of RPC connections
gracefully, but rather misbehaving badly and failing VMs. Since
you're quickly running out of RPC threads when talking to neutron
while launching a massive number of VMs in parallel, we should
raise the upper limit to a saner value.

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -12,7 +12,7 @@
 #amqp_auto_delete=false
 
 # Size of RPC connection pool. (integer value)
-#rpc_conn_pool_size=30
+rpc_conn_pool_size=64
 
 # Qpid broker hostname. (string value)
 #qpid_hostname=localhost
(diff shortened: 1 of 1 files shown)


vuntz:
Would it make sense to have attributes for these? Or do you feel confident in the new values?

dirkmueller:
@vuntz I thought about it, I did not see a reason to make it configurable so far. are you concerned about the memory usage or that the values are not high enough? basically for rpc pool it means 64 connections per minute can be in flight, and for thread pool it means 256 per 4 minute period. That oughta be enough for everone. 

it means that you can launch  ~ 1 VM per second on a compute node.

vuntz:
@dirkmueller I'm mostly concerned about the fact that the values might not fit everyone's need (so not high enough for some cases). But maybe it's no big deal. I wouldn't think the memory impact is huge, so not worried about that.

My point is also that we allow setting the db-related options, so why not these ones,

Anyway, let's try it for now, and see how it goes. +1.

rsalevsky:
+1

PR #50, opened 2015-09-30
Use platform_family check instead of platform

jsuchome:
platform_family "suse" is for both "suse" and "opensuse" platforms


chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -140,14 +140,14 @@ def make_volume(node, backend_id, volume)
 node[:cinder][:volumes].each do |volume|
   if volume[:backend_driver] == "local"
     make_loopback_file(volume)
-    if (%w(suse).include? node.platform) && (node.platform_version.to_f >= 12.0)
+    if (%w(suse).include? node[:platform_family]) && (node[:platform_version].to_f >= 12.0)
```
jsuchome: This is probably not right. platform_family checks for ``suse`` and ``opensuse``, but that version check is relevant for SLES only, I assume. If I revert back to just platfom check, opensuse would be ignored. So does opensuse need an extra check here? Or should we think that the platform_version check will not break anything in the  openSUSE case?

chef/cookbooks/memcached/attributes/default.rb
```diff
@@ -3,7 +3,7 @@
 default[:memcached][:listen] = "0.0.0.0"
 default[:memcached][:daemonize] = true
 
-case node[:platform]
+case node[:platform_family]
 when "suse"
   default[:memcached][:user] = "memcached"
   default[:memcached][:daemonize] = false if node[:platform_version].to_f >= 12.0
```
jsuchome: Same issue with version vs. platform_family...

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -89,7 +89,7 @@
 
 # Cleanup the ovs-usurp init scripts that might still be existing from an old
 # install (before the network barclamp created the ovs-bridge configuration).
-if node[:platform] == "suse" && node[:platform_version].to_f >= 12.0
+if node[:platform_family] == "suse" && node[:platform_version].to_f >= 12.0
```
jsuchome: Same concern as with ceilometer

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -216,8 +216,8 @@
 execute "neutron-db-manage migrate" do
   user node[:neutron][:user]
   group node[:neutron][:group]
-  case node["platform"]
-  when "suse"
+  case node[:platform]
```
vuntz: I think it's safe to assume that debian works like ubuntu here, so changing to platform_family (and debian) would probably be better

chef/cookbooks/postgresql/attributes/default.rb
```diff
@@ -134,7 +134,7 @@
       "ruby#{node["languages"]["ruby"]["version"].to_f}-rubygem-pg"]
     default["postgresql"]["server"]["packages"] = ["postgresql91-server"]
     default["postgresql"]["contrib"]["packages"] = ["postgresql91-contrib"]
-  when node["platform_version"].to_f == 12.0
+  when node["platform_version"].to_f == 12.0 || node["platform"] == "opensuse"
```
jsuchome: Here I'm just guessing. Should we have separate branch for opensuse with detailed checks for versions?

jsuchome:
OK, ready for review

aplanas:
Are templates relevant here? If this is the case, there are some references of `:platform` in places like:
* chef/cookbooks/mysql/templates/redhat/my.cnf.erb
* chef/cookbooks/mysql/templates/default/grants.sql.erb
* chef/cookbooks/horizon/templates/redhat/local_settings.py.erb

This last case have references to SUSE.

Other places:
* chef/cookbooks/mysql/recipes/server.rb
  `if (platform?("ubuntu") && node.platform_version.to_f >= 10.04)`
* chef/cookbooks/cinder/recipes/volume.rb
  `if node[:platform] == "opensuse" || (node[:platform] == "suse" && node[:platform_version].to_f >= 12.0)`

There are more: `grep -R -E "platform[^_]" | less` in your branch

nkrinner:
https://github.com/jsuchome/crowbar-openstack/blob/aec6fa833404d7f07390f36d04a5436940fde07a/chef/cookbooks/heat/templates/default/heat.conf.erb#L420 Seems to also need a change afaics

jsuchome:
@aplanas thanks for those templates, I missed them.
- ubuntu check is correct I think, because it also checks for its version
- also that cinder/recipes/volume.rb tries to distinguish between suse versions, so I think it is correct

jsuchome:
Templates adapted as well

nkrinner:
I think https://github.com/jsuchome/crowbar-openstack/blob/9bd28bf36ac4a44c231a3bbac53c505b2b399d26/crowbar_framework/app/models/nova_service.rb#L106 and maybe even https://github.com/jsuchome/crowbar-openstack/blob/9bd28bf36ac4a44c231a3bbac53c505b2b399d26/crowbar_framework/app/models/nova_service.rb#L262 should be changed?

jsuchome:
@nkrinner good catch, though I think the 2nd one is not needed

chef/cookbooks/postgresql/attributes/default.rb
```diff
@@ -120,7 +120,7 @@
     default["postgresql"]["server"]["service_name"] = "postgresql"
   end
 
-when "suse"
+when "suse", "opensuse"
 
   case
   when node["platform_version"].to_f <= 11.2
```
nkrinner: I think we can remove that, as we don't support the old SLES anymore

chef/cookbooks/postgresql/attributes/default.rb
nkrinner: We'll have to adjust here when we move to SP1, but for now it should be fine.

nkrinner:
Looks good to me besides the comments i made inline.

chef/cookbooks/postgresql/attributes/default.rb
vuntz: No, SP1 is dealt with in the `else`...

vuntz: @jsuchome I think I'd do a separate branch here, yes.

jsuchome: @vuntz and is there specific openSUSE verstion we want to support?

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -365,7 +364,7 @@
 end
 
 ruby_block "synchronize PKI keys for non-founder" do
-  only_if { ha_enabled && !CrowbarPacemakerHelper.is_cluster_founder?(node) && (node[:keystone][:signing][:token_format] == "PKI" || node.platform == "suse") }
+  only_if { ha_enabled && !CrowbarPacemakerHelper.is_cluster_founder?(node) && (node[:keystone][:signing][:token_format] == "PKI" || node[:platform_family] == "suse") }
```
jsuchome: I do not really want to fix this (and above), I think it's better readable this way...

vuntz:
@jsuchome see https://github.com/jsuchome/crowbar-openstack/pull/1 for some suggestions

vuntz:
+1

aplanas:
+1. But lets see what the gating says

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -21,22 +21,19 @@
...
   # only default and xen flavor is supported by this helper right now
-  if node.platform == "suse" && node.platform_version.to_f >= 12.0
-    default_boot = "SLES12"
-    grub_env = %x[grub2-editenv list]
-    if grub_env.include?("saved_entry")
-      current_default = grub_env.strip.split("=")[1]
-    else
-      current_default = nil
-    end
+  if sles11
```
vuntz: Unfortunately, the variable is not available in the method (see failed gating), so we need to replace this with `if node[:platform] == "suse" && node[:platform_version].to_f < 12.0`

jsuchome:
Gate fales with

`` NameError
Cannot find a resource for sles11 on suse version 12.1``

What is wrong with cookbooks/nova/recipes/compute.rb:28?

dirkmueller:
@jsuchome needs rebase

vuntz:
@jsuchome if you have time, can you rebase? The gating was breaking just in cct, which means the PR was likely okay.

jsuchome:
@vuntz rebased, this time with horizon/recipes/server.rb conflicts, so hopefully it is correct

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -324,7 +325,7 @@
 resource.variables({apache_listen_ports: node.normal[:apache][:listen_ports_crowbar].values.map{ |p| p.values }.flatten.uniq.sort})
 
 template "#{node[:apache][:dir]}/sites-available/openstack-dashboard.conf" do
-  if node.platform == "suse"
+  if node[:platform_family] == "suse"
     path "#{node[:apache][:dir]}/vhosts.d/openstack-dashboard.conf"
   end
   source "openstack-dashboard.conf.erb"
```
vuntz: You lost a change below (not visible, line 346 `node.platform == "suse"`)

chef/cookbooks/cinder/recipes/scheduler.rb
```diff
@@ -19,6 +19,10 @@
 
 include_recipe "#{@cookbook_name}::common"
 
+if node[:platform] == "ubuntu"
```
vuntz: Wrong rebase, it seems: this block can be dropped.

jsuchome: Ah, it was probably some older rebase than the latest one...

PR #51, opened 2015-10-01
Add warning for horizon/keystone timeouts

abel-navarro:
Horizon timeout is impacted by keystone token timeout since when the
token gets outdated, then the user is automatically unable to use
horizon. This patch includes a check for the timeout of horizon and
keystone barclamps.

Increased default keystone token expiration timeout to 24 hours up
from 4 hours.

### DIFF
--- a/crowbar_framework/app/models/horizon_service.rb
+++ b/crowbar_framework/app/models/horizon_service.rb
@@ -86,6 +86,18 @@ class HorizonService < PacemakerServiceObject
       end
     end
 
+    horizon_timeout = proposal["attributes"]["horizon"]["session_timeout"]
+    keystone_proposal = Proposal.where(barclamp: "keystone", name: "default").first
+    unless keystone_proposal.nil?
+      keystone_timeout = keystone_proposal["attributes"]["keystone"]["token_expiration"]
+
+      # keystone_timeout is in seconds and horizon_timeout is in minutes
+      if horizon_timeout * 60 > keystone_timeout
+        validation_error("Setting the Horizon timeout (#{horizon_timeout} minutes) longer than the "\
+          "Keystone token expiration timeout (#{keystone_timeout / 60} minutes) is not supported.")
+      end
+    end
+
     super
   end
 
--- a/crowbar_framework/app/models/keystone_service.rb
+++ b/crowbar_framework/app/models/keystone_service.rb
@@ -84,6 +84,18 @@ class KeystoneService < PacemakerServiceObject
       validation_error("Keystone API version 3 needs to be enabled in order to use domain specific drivers.")
     end
 
+    keystone_timeout = proposal["attributes"]["keystone"]["token_expiration"]
+    horizon_proposal = Proposal.where(barclamp: "horizon", name: "default").first
+    unless horizon_proposal.nil?
+      horizon_timeout = horizon_proposal["attributes"]["horizon"]["session_timeout"]
+
+      # keystone_timeout is in seconds and horizon_timeout is in minutes
+      if horizon_timeout * 60 > keystone_timeout
+        validation_error("Setting the Horizon timeout (#{horizon_timeout} minutes) longer than the "\
+          "Keystone token expiration timeout (#{keystone_timeout / 60} minutes) is not supported.")
+      end
+    end
+
     super
   end
 
--- a/chef/data_bags/crowbar/template-horizon.json
+++ b/chef/data_bags/crowbar/template-horizon.json
@@ -11,7 +11,7 @@
       "site_theme": "",
       "site_branding_link": "",
       "help_url": "http://docs.openstack.org/",
-      "session_timeout": 1440,
+      "session_timeout": 240,
       "db": {
         "database": "horizon",
         "user": "horizon",


crowbar_framework/app/models/horizon_service.rb
```diff
@@ -85,6 +85,18 @@ def validate_proposal_after_save proposal
       end
     end
 
+    horizon_timeout = proposal["attributes"]["nova_dashboard"]["session_timeout"]
```
nkrinner: I think this has to be `proposal["attributes"]["horizon"]["session_timeout"]`
(In Cloud 5 the dashboard was named "nova_dashboard", for Cloud 6 we renamed it to "horizon")

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -83,6 +83,18 @@ def validate_proposal_after_save proposal
       validation_error("Keystone API version 3 needs to be enabled in order to use domain specific drivers.")
     end
 
+    keystone_timeout = proposal["attributes"]["keystone"]["token_expiration"]
+    horizon_proposal = Proposal.where(barclamp: "nova_dashboard", name: "default").first
```
nkrinner: Same here

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -83,6 +83,18 @@ def validate_proposal_after_save proposal
...
+    unless horizon_proposal.nil?
+      horizon_timeout = horizon_proposal["attributes"]["nova_dashboard"]["session_timeout"]
```
nkrinner: Same here

tboerger:
The error messages should be put to the i18n file.

crowbar_framework/app/models/horizon_service.rb
abel-navarro: Right, copy&paste problem. Will fix it.

abel-navarro:
@tboerger where's that i18n file? I cannot find it. Anyway, shouldn't all the error messages be in that file too?

tboerger:
https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/config/locales/keystone/en.yml yes, all of them should be there.

abel-navarro:
@tboerger thanks for the link. Shouldn't we tackle moving all error strings to the i18n file in another task? it would be weird just to have this string there.

vuntz:
@abel-navarro to not make things worse, I think it's good if at least new strings are directly in the yaml file. I agree we should also have a task for moving all the other strings there (want to create a card in the backlog for this?)

abel-navarro:
@vuntz ok, I'll put the string in that file and create a card to move the others.

jsuchome:
+1 (based on code, not tested)

I think that translation support could be done separately, as we currently do not translate any other validation error anyway.

vuntz:
So, given that we have a followup card for moving all errors to the l10n file, I'm changing my mind and giving a +1.

vuntz:
Gating is red, but it actually deployed fine, and things got broken during tempest. Looks unrelated to the change, but triggering a rebuild.

abel-navarro:
@nkrinner already made some progress with the i18n for this case. I couldn't verify it, but looked good to me. Anyway, I think it makes more sense to do them as a bulk change in the other trello card.

chef/data_bags/crowbar/template-keystone.json
```diff
@@ -12,7 +12,7 @@
       "rabbitmq_instance": "none",
       "domain_specific_drivers": false,
       "domain_config_dir": "/etc/keystone/domains",
-      "token_expiration": 14400,
+      "token_expiration": 86400,
```
vuntz: It was mentioned that it might be better to instead lower the horizon timeout, to avoid having tons of tokens tracked by keystone.

dirkmueller:
@abel-navarro I wonder what happened to tboergers comment about putting the warning in the i18n file?

dirkmueller:
raising the timeout is something that I wouldn't take light-heartedly. can we perhaps lower the horizon dashboard to sync the timeouts if that is important for you?

abel-navarro:
@dirkmueller we created a task/card to move all the error strings into the i18n file. It makes more sense to track it in a different file since all the other strings would still be hardcoded.

About the timeout, initially it was 4 hours and after some discussion it was decided to set to 24. I understand the reasoning that this will cause lots of tokens to be accumulated. I don't have a strong feeling about the final value. 4 hours sound reasonable to me so I can set it back there.

vuntz:
@abel-navarro please revert the timeout and instead change the horizon one (btw, should be done in a second commit)

abel-navarro:
Remember that backport to cloud5 is contained in these pull requests:
https://github.com/crowbar/barclamp-keystone/pull/291
https://github.com/crowbar/barclamp-nova_dashboard/pull/230

vuntz:
@abel-navarro can you actually remove the change in chef/data_bags/crowbar/template-keystone.json from the first commit? So we have a cleaner history?

abel-navarro:
@vuntz if you don't mind, I'll merge the two commits into one. The idea for the two commits was to reflect the reverted default values. If there's nothing to revert, one commit should be fine.

vuntz:
@abel-navarro I find it slightly clearer to have the change in the default in a separate commit, but I don't mind strongly.

aplanas:
recheck in jenkins

dirkmueller:
+1

vuntz:
Gating fails due to branding patch conflicting with this, but I'm confident it won't break stuff.

PR #52, opened 2015-10-01
nova: Only accept suse nodes for xen compute role

vuntz:
This was already enforced during validation, but we can add an explicit
constraint for it.


aplanas:
+1

jsuchome:
+1

PR #53, opened 2015-10-01
cinder, glance: Install qemu-img only when needed

vuntz:
qemu-img is used by some bits of both cinder and glance.


tboerger:
:+1:

toabctl:
Why not adding it as a Requires to the package?

vuntz:
@toabctl good point, will do that and check if other distros have the dep too.

vuntz:
@toabctl updated

aplanas:
+1

toabctl:
+1

PR #54, opened 2015-10-03
increase timeout for synchronization of HA resources (bsc#935462)

dirkmueller:
Heat HA resources can take over 1 minute to be set up, as
seen by mkcloud failures in ci.suse.de Jenkins.


aplanas:
+1

5 minutes are enough for mkcloud?

dirkmueller:
@aplanas seems to get further until it runs into the tempest issue (see automation patch)

aspiers:
I think 3 mins would have been more than enough for this but 5 mins is probably OK too.

PR #55, opened 2015-10-04
neutron: Fix order constraint for neutron-ha-tool

vuntz:
Right now, the order explicitly requires haproxy to be started before
neutron-server, which results in neutron-server being stopped and
started when migration haproxy.

Instead, do not require an order in the dependencies of neutron-ha-tool.
Also depends on postgresql and rabbitmq since without them, this won't
work.

Depends on crowbar/crowbar-ha#29

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -157,7 +157,13 @@ pacemaker_primitive ha_tool_primitive_name do
 end
 
 crowbar_pacemaker_order_only_existing "o-#{ha_tool_primitive_name}" do
-  ordering ["g-haproxy", "cl-neutron-server", agents_clone_name, ha_tool_primitive_name]
+  # While neutron-ha-tool technically doesn't directly depend on postgresql or
+  # rabbitmq, if these bits are not running, then neutron-server can run but
+  # can't do what it's being asked. Note that neutron-server does have a
+  # constraint on these services, but it's optional, not mandatory (because it
+  # doesn't need to be restarted when postgresql or rabbitmq are restarted).
+  # So explicitly depend on postgresql and rabbitmq (if they are in the cluster).
+  ordering "( postgresql rabbitmq g-haproxy cl-neutron-server #{agents_clone_name} ) #{ha_tool_primitive_name}"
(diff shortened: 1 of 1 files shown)


vuntz:
This depends on some non-existing PR to add support for String instead of Array for ordering.

vuntz:
Depends on https://github.com/crowbar/crowbar-ha/pull/29

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -157,7 +157,7 @@
 end
 
 crowbar_pacemaker_order_only_existing "o-#{ha_tool_primitive_name}" do
-  ordering ["g-haproxy", "cl-neutron-server", agents_clone_name, ha_tool_primitive_name]
+  ordering "( postgresql rabbitmq g-haproxy cl-neutron-server #{agents_clone_name} ) #{ha_tool_primitive_name}"
```
aspiers: When we originally discussed this we decided that we would only include direct dependencies of the current service, not transitive dependences.  E.g. see lines 95-96 of the [cloud-ha-ordering etherpad](https://etherpad.nue.suse.com/p/cloud-ha-ordering).  However at the time I don't think it occurred to us (or me at least) to allow for the possibility of (say) a direct dependency such as keystone living in a different cluster whilst postgresql/rabbitmq live in this same cluster as the neutron server.  So I guess it does make sense to add postgresql and rabbitmq here, but it is probably worth adding this explanation as a comment to clarify.

aspiers:
+1 once all three of the following conditions are satisfied:

- crowbar/crowbar-ha#29 is merged
- a comment is added as per my suggestion
- gate goes green

:)

chef/cookbooks/neutron/recipes/network_agents_ha.rb
vuntz: Added a comment.

rhafer:
+1 (But I guess in order to re-run the tests you'd need to rebase the patch as the test packages are pretty dated already and will very likely not work with current crowbar-core)

rhafer:
otoh running the standard tests does provide much value here anyway as the change reciped is not even executed during the test run.

dirkmueller:
+1

PR #57, opened 2015-10-06
increase timeout for synchronization of horizon HA resources (bsc#935462)

aspiers:
Horizon HA resources can take over 1 minute to be set up, as seen by [mkcloud failures in ci.suse.de Jenkins](https://trello.com/c/Z1L5phHx).

See also #54 for equivalent tweak to heat.

### DIFF
--- a/chef/cookbooks/horizon/recipes/ha.rb
+++ b/chef/cookbooks/horizon/recipes/ha.rb
@@ -37,7 +37,9 @@ end
 crowbar_pacemaker_sync_mark "sync-horizon_before_ha"
 
 # Avoid races when creating pacemaker resources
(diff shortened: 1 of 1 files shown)


dirkmueller:
+1, merging under the "obvious workaround"

aspiers:
Well apparently this timeout might have been caused by [bsc#947044: Install of openstack-dashboard takes way too long](https://bugzilla.suse.com/show_bug.cgi?id=947044) not [bsc#935462: missing synchronization means chef-client runs on cluster nodes drift apart](https://bugzilla.suse.com/show_bug.cgi?id=935462).

aspiers:
(in which case this PR is mostly useless)

vuntz:
Agreed with @aspiers, this won't fix it in the cases I was seeing...

PR #58, opened 2015-10-06
Update heat.conf template to liberty

vuntz:


aplanas:
+1 Tempest is green

abel-navarro:
Smoke tempest tests also ok for me (except three unrelated). +1

PR #59, opened 2015-10-07
manila: Allow using keystone v3 API

vuntz:
cc @rhafer


chef/data_bags/crowbar/migrate/manila/011_add_max_header_line.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["max_header_line"] = 16384
```
toabctl: I think we usually use the default value so ```ta["max_header_line"]```

rhafer: @vuntz Is this supposed to be backported to tex? Then we need the the ```unless a.has_key()``` thing here.

vuntz: fixed

rhafer:
+1

rsalevsky:
+1

toabctl:
+1

PR #60, opened 2015-10-07
Sync ceilometer

aplanas:


toabctl:
Is there a way to show the diff of the .conf.erb file in github? I just see ```1,079 additions, 870 deletions not shown```

aplanas:
@toabctl No idea, the same is happening in the heat one: https://github.com/crowbar/crowbar-openstack/pull/58

vuntz:
@toabctl @aplanas just look at the patch on your machine, I'd say :/

chef/cookbooks/ceilometer/recipes/server.rb
```diff
@@ -311,8 +311,11 @@
 end
 
 # Cronjob to repair the database and free space for mongodb.  This
-# only makes sense when the time_to_live > 0
-if node[:ceilometer][:use_mongodb] && node[:ceilometer][:database][:time_to_live] > 0
+# only makes sense when the metering_time_to_live or
+# event_time_to_leave > 0
+time_to_live = node[:ceilometer][:database][:metering_time_to_live] > 0 \
```
jsuchome: Wouldn't it be better to have the variable name indicate that it's a boolean? (like time_to_live_set or so)

chef/data_bags/crowbar/migrate/ceilometer/021_rename_time_to_live_attribute.rb
```diff
@@ -0,0 +1,15 @@
...
+    a["database"]["metering_time_to_live"] = a["database"]["time_to_live"]
+    a["database"].delete("time_to_live")
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  if a["database"]["metering_time_to_live"]
+    a["database"]["time_to_live"] = a["database"]["metering_time_to_live"]
+    a["database"].delete("meterin_time_to_live")
```
jsuchome: typo (``meterin_time_to_live``)

chef/data_bags/crowbar/migrate/ceilometer/022_add_event_time_to_live_attribute.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["database"]["event_time_to_live"] = ta["database"]["even_time_to_live"]
```
jsuchome: typo: ``even_time_to_live``

jsuchome:
Do we need separate migrations for those 2 time to live variables?

aplanas:
@jsuchome One is a rename and another one is new, I feel better separating in two migrations

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -94,7 +100,8 @@
       node_hostname: node["hostname"],
       hypervisor_inspector: hypervisor_inspector,
       libvirt_type: libvirt_type,
-      time_to_live: time_to_live,
+      metering_time_to_live: metering_time_to_live,
+      event_time_to_live: event_time_to_live,
```
jsuchome: it looks like in the template you're still using just ``@time_to_live`` value, is that right?

aplanas: no, is changed in the same commit: https://github.com/aplanas/crowbar-openstack/commit/9bda334707ea36ecdf3a7b6237e3924bba08511b#diff-8ba6ab0f73ac9ad5fa0f2fe85bb234deR503

jsuchome: Ok, sorry, it's confusing without that diff

vuntz:
Can't comment on the right line of the conf file, but this is wrong:
`host_ip = <%= node[:nova][:vcenter][:port] rescue "443" %>`
should be host_port

vuntz:
Shouldn't we set `[api]/workers`?

vuntz:
`time_to_live` is used in the webui, so code needs to be adapted there.

vuntz:
Should we set `shuffle_time_before_polling_task` to something non-0?

vuntz:
I believe pipeline.yaml.erb should also be sync'ed.

vuntz:
There's also a `alarm_history_time_to_live` setting, that we could add with the other ttl settings...

vuntz:
Wondering out loud: do we want to set `refresh_pipeline_cfg`?

vuntz:
Maybe also `batch_polled_samples`.

aplanas:
@vuntz  `alarm_history_time_to_live` is deprecated for removal and is going to be ignored

aplanas:
* Fix typos
* Use `time_to_live_set`
* Fix `host_port` reference
* Set `[api]/workers`
* Export `*_time_to_live` to the WebUI
* Sync `pipeline.yaml.erb`
* Set `shuffle_time_before_polling_task` to 15 seconds (rndint(0, 15))

aplanas:
> Wondering out loud: do we want to `set refresh_pipeline_cfg`?

Maybe we can wait a bit:
https://bugs.launchpad.net/ceilometer/+bug/1477321

vuntz:
> @vuntz alarm_history_time_to_live is deprecated for removal and is going to be ignored

Isn't it deprecated just because of aodh? I guess the question is are we switching to it? :-)

vuntz:
I checked and am not super convinced about the need for changing `batch_polled_samples`. So only question left is the one about `alarm_history_time_to_live` / aodh.

Other than that, LGTM.

aplanas:
+1 Tempest green

vuntz:
+1

PR #61, opened 2015-10-07
Update glance config templates to liberty

vuntz:


aplanas:
LGTM: +1

aplanas:
Still +1. Tempest is green (one ceilometer + glace timeout the first run, the second run produre 116 OK tests in smoke test)

abel-navarro:
Also for me smoke tempest tests run ok (except three not related). +1

PR #62, opened 2015-10-07
Fixup migration for tex backport (again)

rhafer:
Check for the correct attribute. "multi_domain_support" never was an attribute
of the keystone proposal.


vuntz:
Ouch, +1

rsalevsky:
+1

PR #64, opened 2015-10-07
Restart cinder-volume on setting up the losetup

dirkmueller:
When the local backend is being used, we need to restart cinder-volume
in order to reinitialize the backend and access the newly created vg.

### DIFF
--- a/chef/cookbooks/cinder/recipes/volume.rb
+++ b/chef/cookbooks/cinder/recipes/volume.rb
@@ -67,6 +67,10 @@ def setup_loopback_device(volume)
     cmd = ["losetup", "-f", fname]
     attach_loopback_device = Mixlib::ShellOut.new(cmd)
     attach_loopback_device.run_command
+    # restart cinder-volume service so that it reloads the vgs state on start
+    cmd = ["service", "openstack-cinder-volume", "restart"]
+    restart_cinder = Mixlib::ShellOut.new(cmd)
+    restart_cinder.run_command
   end
 end
 


chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -67,6 +67,10 @@ def setup_loopback_device(volume)
     cmd = ["losetup", "-f", fname]
     attach_loopback_device = Mixlib::ShellOut.new(cmd)
     attach_loopback_device.run_command
+    # restart cinder-volume service so that it reloads the vgs state on start
+    cmd = ["service", "openstack-cinder-volume", "restart"]
+    restart_cinder = Mixlib::ShellOut.new(cmd)
+    restart_cinder.run_command
```
vuntz: That's an horrible way to do this; can't we send somehow a notification to the right service?

vuntz:
It's unclear to me if it's fixing the local backend because it's always broken (sounds weird, I'd expect cinder-volume to be started after this code anyway), or after reboot (https://bugzilla.suse.com/show_bug.cgi?id=947499).

chef/cookbooks/cinder/recipes/volume.rb
dirkmueller: we could, I was just trying to get it to work minimally. the whole code is broken imho anyway, we should install a systemd service for that.

dirkmueller:
@vuntz its just fixing the gate right now (somebody changed the cloud6 trigger from all_noreboot to all_batch, which triggers a reboot and thereby exposes that reboot was always broken since teh SLE12 switch)

vuntz:
@dirkmueller I propose https://github.com/crowbar/crowbar-openstack/pull/66 instead

chef/cookbooks/cinder/recipes/volume.rb
vuntz: @dirkmueller happy to close this in favor of https://github.com/crowbar/crowbar-openstack/pull/66 ?

dirkmueller:
@vuntz agreed, much better.

PR #65, opened 2015-10-08
Sync swift

aplanas:


chef/cookbooks/swift/templates/default/dispersion.conf.erb
```diff
@@ -17,6 +17,16 @@ auth_key = <%= node["swift"]["dispersion"]["service_password"] %>
 # auth_version = 2.0
 auth_version = 2.0
 #
+# NOTE: If you want to use keystone (auth version 3.0), then its configuration
```
vuntz: Don't we want to use keystone v3 when it's enabled in the keystone barclamp?

chef/cookbooks/swift/templates/default/proxy-server.conf.erb
```diff
@@ -328,39 +332,83 @@ admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
...
+#
+# delay_auth_decision defaults to False, but leaving it as false will
+# prevent other auth systems, staticweb, tempurl, formpost, and ACLs from
+# working. This value must be explicitly set to True.
+# delay_auth_decision = False
+delay_auth_decision = <%= @keystone_delay_auth_decision ? "True" : "False" %>
+#
 # cache = swift.cache
 # include_service_catalog = False
-signing_dir = /var/cache/swift/keystone-signing
```
vuntz: Are you sure we should remove this?

chef/cookbooks/swift/templates/default/proxy-server.conf.erb
```diff
@@ -310,16 +320,10 @@ user_test_tester3 = testing3
 # and have it in your main pipeline so instead of having tempauth in
 # there you can change it to: authtoken keystoneauth
 #
-
 [filter:authtoken]
```
vuntz: I don't see anything about the region in this section. That sounds wrong (probably missing upstream too)

vuntz:
I'm not 100% sure about the changes that relates to logging, as swift has always been special there -- we'll need to see the result to double-check this.

chef/cookbooks/swift/templates/default/proxy-server.conf.erb
aplanas: Oh I see, is in https://github.com/openstack/keystonemiddleware and not in Swift. I was confusing searching in the Swift source code for this options.

I will revert

chef/cookbooks/swift/templates/default/proxy-server.conf.erb
aplanas: `region_name` is defined also here: https://github.com/openstack/keystonemiddleware/blob/ec4ec10ee2df5c8ec66eae64e0c85b0d9106f7e7/keystonemiddleware/auth_token/__init__.py#L281

What is the expected value?

vuntz: `@keystone_settings['region_name']` (iirc)

chef/cookbooks/swift/templates/default/dispersion.conf.erb
jsuchome: That makes sense

aplanas:
* Enable keystone auth v3 in dispersion.conf
* Add `region_name` in prosy-server.conf

chef/cookbooks/swift/templates/default/dispersion.conf.erb
```diff
@@ -9,13 +9,37 @@
...
+# auth_url = http://localhost:5000/v3/
+# auth_user = user
+# auth_key = password
+# auth_version = 3.0
+# project_name = project
+# project_domain_name = project_domain
+# user_domain_name = user_domain
+<% if node[:keystone][:api][:version]).to_f <= 3.0 - %>
+
+auth_url = <%= @keystone_settings['internal_auth_url'] %>
```
jsuchome: So, this is same for either API version, does it need to be in separate branch? (also auth_key)

toabctl:
This is now 3 years old. Closing...

PR #66, opened 2015-10-08
cinder: Create a cinder-looplvm utility / service

vuntz:
This will be used to mount the loopback volumes used with LVM on boot.


vuntz:
Removed wip tag, should work now.

toabctl:
Is there a reason why we don't add the initscript and .service file to the package?

toabctl:
Ah. It depends on ```<%= @loop_lvm_paths %>```

vuntz:
@toabctl and also because the loopback file stuff is really just something specific to one option in crowbar, not something really useful out of crowbar.

toabctl:
Don't know how we do that usually in crowbar, but can you rename chef/cookbooks/cinder/files/suse/cinder-looplvm to cinder-looplvm.init  and use a source keyword to address the file? that way it's clear that it's an init script.

chef/cookbooks/cinder/files/suse/cinder-looplvm
```diff
@@ -0,0 +1,63 @@
+#! /bin/sh
+#
+# Copyright (c) 2013 SUSE, Inc.
+#
+# /etc/init.d/cinder-looplvm
+#
+### BEGIN INIT INFO
+# Provides:          cinder-looplvm
```
toabctl: Doesn't it provide on a SUSE system ```openstack-cinder-looplvm``` ? So you should use here the ```cinder_looplvm_service``` variable?

chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -140,32 +121,70 @@ def make_volume(node, backend_id, volume)
...
+    cookbook_file "/etc/init.d/#{cinder_looplvm_service}" do
       owner "root"
       group "root"
-      mode 0755
-      variables(loop_lvm_paths: loop_lvm_paths.map{ |x| Shellwords.shellescape(x) }.join(" "))
+      mode "0755"
+    end
+
+    link "/usr/sbin/rc#{cinder_looplvm_service}" do
+      action :create
```
toabctl: Can't we just link to ```service``` here, too? As done in https://github.com/crowbar/crowbar-openstack/pull/66/files#diff-e44ebd5ea17511e644658be6d5ac87bfR178 ?

chef/cookbooks/cinder/files/suse/cinder-looplvm
vuntz: Fixed.

chef/cookbooks/cinder/recipes/volume.rb
vuntz: No, that's only for systemd services that we do that.

vuntz:
All comments should be addressed.

chef/cookbooks/cinder/files/suse/cinder-looplvm.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Loopback mount for LVM VGs used by cinder
+Before=openstack-cinder-volume.service
```
toabctl: Sorry for missing that in the previous review but I think this also depends on the OS. So you need to use ```<%= @service_name %>``` similiar to the sysv init script.

vuntz: I was thinking about it, but since it's in the suse directory, I was too lazy... But I shouldn't be lazy, will fix.

vuntz: done

chef/cookbooks/cinder/templates/default/cinder-looplvm.erb
```diff
@@ -0,0 +1,66 @@
+#! /bin/sh
+
```
toabctl: *If* you have to change it again, adding a comment that this is generated by chef would be nice. but just a nit...

toabctl:
+1

dirkmueller:
+1

mjura:
Looks OK +1

PR #67, opened 2015-10-09
Require SLE12 SP1 or newer for OpenStack roles on SUSE

dirkmueller:


crowbar_framework/app/models/nova_service.rb
```diff
@@ -33,6 +33,7 @@ def role_constraints
           "unique" => false,
           "count" => 1,
           "exclude_platform" => {
+            "suse" => "< 12.1",
             "windows" => "/.*/"
           },
           "cluster" => true
```
vuntz: Can you update it for docker also (the role below)?

crowbar_framework/app/models/nova_service.rb
```diff
@@ -77,6 +81,7 @@ def role_constraints
           "unique" => false,
           "count" => 1,
           "exclude_platform" => {
+            "suse" => "< 12.1",
             "windows" => "/.*/"
           }
         },
```
vuntz: Missing exclusion for xen (below)

crowbar_framework/app/models/nova_service.rb
dirkmueller: I wasn't sure yet what the docker plan is.. thats why I left it out.

crowbar_framework/app/models/nova_service.rb
dirkmueller: added

vuntz:
+1

toabctl:
+1

PR #68, opened 2015-10-10
Rename nova roles to not have "multi-" in their names

vuntz:
This is really useless, not consistent with what we do for everything
else and confusing for users. So let's fix it.

### DIFF
--- a/chef/cookbooks/cinder/recipes/common.rb
+++ b/chef/cookbooks/cinder/recipes/common.rb
@@ -42,7 +42,7 @@ else
 end
 Chef::Log.info("Glance server at #{glance_server_host}")
 
-nova_apis = search(:node, "roles:nova-multi-controller") || []
+nova_apis = search(:node, "roles:nova-controller") || []
 if nova_apis.length > 0
   nova_api = nova_apis[0]
   nova_api_insecure = nova_api[:nova][:ssl][:enabled] && nova_api[:nova][:ssl][:insecure]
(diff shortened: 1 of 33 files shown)


vuntz:
Requires https://github.com/SUSE-Cloud/automation/pull/599 for the CI.

chef/cookbooks/cinder/recipes/common.rb
```diff
@@ -42,7 +42,7 @@
 end
 Chef::Log.info("Glance server at #{glance_server_host}")
 
-nova_apis = search(:node, "roles:nova-multi-controller") || []
+nova_apis = search(:node, "roles:nova-controller") || []
```
toabctl: What about renaming it to ```nova-server``` ? That's inconsistent between the different roles, too. So imo we should stick with ```-server``` or ```-controller``` .

vuntz: IIRC, we use `controller` when it's more than on services (in cinder too, for instance). And we use `server` when it's only one service.

Hrm, actually, glance is not following that scheme. Oh well. I don't have a strong feeling on this. Let's check if there are other opinions.

aspiers:
Strongly in favour of the idea; didn't review in detail but I guess it's mostly just search and replace, so if you can get the gate green then +1 from me.

rsalevsky:
+1

chef/cookbooks/cinder/recipes/common.rb
toabctl: Oh. manila has multiple services and uses ```-server```. So it's definitly inconsistent...

tboerger:
LGTM

vuntz:
Tested with https://github.com/SUSE-Cloud/automation/pull/599 and https://github.com/SUSE-Cloud/cct/pull/39, and it works fine.

vuntz:
(which is, btw, a call for reviews for these PR ;-))

PR #69, opened 2015-10-14
Enable multipath for image xfer also on cinder-volume

dirkmueller:
Without the settings on Nova and Cinder-volume in sync, there are
interesting side effects in several drivers since they pass through
inconsistent information between Cinder and Nova, that causes undesired
artefacts.

### DIFF
--- a/chef/cookbooks/nova/attributes/default.rb
+++ b/chef/cookbooks/nova/attributes/default.rb
@@ -36,6 +36,8 @@ default[:nova][:use_shared_instance_storage] = false
 # Hypervisor Settings
 #
 default[:nova][:libvirt_type] = "kvm"
+
+# Keep in sync with cinder cookbook
 if node[:platform_family] == "suse"
   default[:nova][:libvirt_use_multipath] = true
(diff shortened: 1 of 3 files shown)


vuntz:
Can you add a comment to the nova cookbook about keeping this in sync?

vuntz:
+1

chef/cookbooks/nova/attributes/default.rb
```diff
@@ -36,6 +36,8 @@
 # Hypervisor Settings
 #
 default[:nova][:libvirt_type] = "kvm"
+
+# Keep in sync with cinder cookbook
 if node[:platform_family] == "suse"
   default[:nova][:libvirt_use_multipath] = true
```
tboerger: No else branch here?

dirkmueller: the else branch is on line 43 ?

vuntz:
Anyone else for a +1?

tboerger:
:+1:

PR #71, opened 2015-10-18
Liberty related fixes (Non-Neutron)

dirkmueller:


chef/cookbooks/glance/recipes/api.rb
```diff
@@ -98,6 +98,14 @@
 glance_stores = node.default[:glance][:glance_stores]
 glance_stores += ["glance.store.vmware_datastore.Store"] unless node[:glance][:vsphere][:host].empty?
 
+# glance_stores unconditionally enables cinder and swift stores, so we need
+# to install the necessary clients
+case node[:platform_family]
+when "suse"
+  package "python-cinderclient"
+  package "python-swiftclient"
```
vuntz: Shouldn't they be deps of the package instead?

dirkmueller: well, we explicitely enable those stores, by default only those are enabled that are available. Thats why I thought a requires in the package wouldn't make a lot of sense.

vuntz: Hrm, okay. Any reason to make this specific to the `suse` family? :-)

dirkmueller: didn't want to figure out the package names for non-suse.. :) but we can remove the check

vuntz:
+1

tboerger:
LGTM

toabctl:
+1

PR #73, opened 2015-10-20
Use ceilometermiddleware for object store integration

dirkmueller:
The previous entry point was deprecated in Kilo and
got removed in Liberty

### DIFF
--- a/chef/cookbooks/swift/recipes/proxy.rb
+++ b/chef/cookbooks/swift/recipes/proxy.rb
@@ -90,6 +90,7 @@ proxy_config[:protocol] = swift_protocol
 proxy_config[:ssl_enabled] = node[:swift][:ssl][:enabled]
 proxy_config[:ssl_certfile] = node[:swift][:ssl][:certfile]
 proxy_config[:ssl_keyfile] = node[:swift][:ssl][:keyfile]
+proxy_config[:rabbit_settings] = fetch_rabbitmq_settings
 proxy_config[:max_containers_per_extraction] = node[:swift][:middlewares][:bulk][:max_containers_per_extraction]
 proxy_config[:max_failed_extractions] = node[:swift][:middlewares][:bulk][:max_failed_extractions]
 proxy_config[:max_deletes_per_request] = node[:swift][:middlewares][:bulk][:max_deletes_per_request]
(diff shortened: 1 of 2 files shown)


chef/cookbooks/swift/recipes/proxy.rb
```diff
@@ -117,8 +117,10 @@
 
 if %w(rhel suse).include?(node[:platform_family])
   package "openstack-swift-proxy"
+  package "python-ceilometermiddleware"
 else
   package "swift-proxy"
+  package "python-ceilometermiddleware"
```
vuntz: I think this belongs to the swift recipe in the ceilometer cookbook, not here.

dirkmueller: That doesn't work, swift doesn't start anymore if that package isn't installed, and swift is generally installed before ceilometer. and since it is optional we can't add it as a package dependency.

vuntz: @dirkmueller but the ceilometer middleware is not enabled if the ceilometer-swift role is not used (so the section in the config file shouldn't matter). It's weird that it doesn't start anymore, sounds like a bug in swift or elsewhere.

dirkmueller: @vunt hmm. looking further you're right, initial start of swift succeeds (without the ceilometer middleware), and restart which adds it also succeeds, but then it crashes as python-ceilometer does not contain the ceilometer middleware anymore. 

I still think it belongs into the cookbook that deploys swift proxy though. I find it awkward that the ceilometer cookbook would need to be able to figure out on which node swift is being deployed in order to deploy the client bindings for swift on that node. it seems more natural to me that when we reference it from the swift proxy server configuration file, that the swift proxy server cookbook is installing the dependencies for what it references in the config file.

chef/cookbooks/swift/templates/default/proxy-server.conf.erb
```diff
@@ -9,6 +9,31 @@ bind_port = <%= @bind_port %>
...
 
+# The RabbitMQ broker address where a single node is used.
+# (string value)
+#rabbit_host=localhost
+rabbit_host=<%= @rabbit_settings[:address] %>
+
+# The RabbitMQ broker port where a single node is used.
+# (integer value)
+#rabbit_port=5672
+rabbit_port=5672
```
toabctl: Should be ```<%= @rabbit_settings[:port] %>```

toabctl: beside that, +1

toabctl:
+1

aplanas:
+1 Partially tested in C16

PR #74, opened 2015-10-21
We do not need to load docker image explicitely,

jsuchome:
nova-docker can do it itself

### DIFF
--- a/chef/cookbooks/nova/recipes/docker.rb
+++ b/chef/cookbooks/nova/recipes/docker.rb
@@ -19,23 +19,3 @@
 #
 
 node.set[:nova][:libvirt_type] = "docker"
-
-# if tempest is deployed, load packaged docker image at docker compute node
-tempest_node = search(:node, "roles:tempest").first
-unless tempest_node.nil?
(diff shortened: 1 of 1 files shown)


jsuchome:
Well.. unless there's a requirement that compute nodes do not have the access to the network...

vuntz:
@jsuchome reading the code, they would download the image from glance and import it from there, or am I missing something? (so network access wouldn't matter)

jsuchome:
@vuntz you mean in _pull_missing_image? I can see there's some docker API used, not glance, but I may be missing it...

vuntz:
@jsuchome https://git.openstack.org/cgit/openstack/nova-docker/tree/novadocker/virt/docker/driver.py#n406

`images.fetch` is downloading from glance, as far as I understand.

jsuchome:
Ah, it comes from nova.virt. Than it makes sense, OK.

vuntz:
Anyway, assuming this works, +1 :-)

dirkmueller:
LGTM

PR #76, opened 2015-10-21
Fix Tempest endpoint for Liberty.

aplanas:


aplanas:
We need this to run tempest in Liberty

dirkmueller:
I added the patch to the package, is the change then still needed?

aplanas:
No, with 0001-Fix-identity-new-endpoint_type-options-for-old-users.patch this workaround is not needed

PR #78, opened 2015-10-26
Add "role recipes" with the goal of simplifying the roles

vuntz:
We want the roles to only reference one recipe, so that we can put some
intelligence in that recipe:

 - in a first step, we will stop changing the run list of a node
   depending on its state, and instead have the role recipe "decide"
   whether something should be done or not, depending on the state. This
   will solve the issue that the search results for nodes are not
   reliable when a node is in some state that is not "readying",
   "applying" or "ready".

 - in a later step, we might want to have a new attribute that will tell
   us which role to apply instead of applying everything.

### DIFF
--- a/chef/cookbooks/ceilometer/metadata.rb
+++ b/chef/cookbooks/ceilometer/metadata.rb
@@ -11,3 +11,5 @@ depends "database"
 depends "crowbar-openstack"
 depends "crowbar-pacemaker"
 depends "utils"
+
+recommends "hyperv"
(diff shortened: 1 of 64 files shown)


tboerger:
LGTM

vuntz:
Removed wip tag.

tboerger:
:+1:

aspiers:
Same comment about code duplication as in https://github.com/crowbar/crowbar-core/pull/99#issuecomment-154484553

mjura:
Looks ok +1

Btw. I wanted to ask about Ceph, but it seems that all ceph roles including only one recipe already

vuntz:
Closing; this is partly done, will create a new PR for the remaining.

PR #79, opened 2015-10-28
Fix nova endpoints for liberty

rhafer:
Make the v2.1 the new default "nova" endpoint. Provide v2.0 via the
"nova_legacy" service. (As e.g. done in devstack)

### DIFF
--- a/chef/cookbooks/nova/recipes/api.rb
+++ b/chef/cookbooks/nova/recipes/api.rb
@@ -103,15 +103,15 @@ keystone_register "register ec2 service" do
   action :add_service
 end
 
-keystone_register "register computev21 service" do
+keystone_register "register nova_legacy service" do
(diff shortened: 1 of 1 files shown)


rhafer:
This should fix some of the recent novaclient functional test failures we've been seeing on the ci system recently (not all of them though).

I am still a bit unclear about what to do with the nova21 endpoints in case of an upgrade from tex. Just delete them?

vuntz:
+1

aplanas:
+1 Tempest green

rhafer:
@vuntz do you thing it's ok to just leave the nova21 endpoints in place after an upgrade?

vuntz:
> @vuntz do you thing it's ok to just leave the nova21 endpoints in place after an upgrade?

I don't think they will hurt, so yes.

PR #80, opened 2015-10-29
Set flag to install and start openvswitch (bsc#950798)

rhafer:
All compute nodes, even if we don't ourself create bridges on them need ovs
running when the ovs mech driver is selected.

### DIFF
--- a/crowbar_framework/app/models/neutron_service.rb
+++ b/crowbar_framework/app/models/neutron_service.rb
@@ -291,9 +291,17 @@ class NeutronService < PacemakerServiceObject
   def update_ovs_bridge_attributes(attributes, node)
     needs_save = false
     ovs_bridge_networks = []
+    ml2_mechanism_drivers = []
     if attributes["networking_plugin"] == "ml2"
       ml2_type_drivers = attributes["ml2_type_drivers"]
-      if attributes["ml2_mechanism_drivers"].include?("openvswitch")
+      ml2_mechanism_drivers = attributes["ml2_mechanism_drivers"]
+      if ml2_mechanism_drivers.include?("openvswitch")
+        # This node needs the ovs packages installed and the service started
+        node.crowbar["network"] ||= {}
+        unless node.crowbar["network"]["needs_openvswitch"]
+          node.crowbar["network"]["needs_openvswitch"] = true
+          needs_save = true
+        end
         # We need to create ovs bridges for floating and (when vlan type driver
         # is enabled) nova_fixed.  Adjust the network attribute accordingly.
         # We only do that on the node attributes and not the proposal itself as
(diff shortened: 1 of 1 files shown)


rhafer:
This needs to be merged together with https://github.com/crowbar/crowbar-core/pull/105
For this reason the ci will not be able to succeed testing this PR.

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -276,6 +276,14 @@ def update_ovs_bridge_attributes(attributes, node)
     if attributes["networking_plugin"] == "ml2"
       ml2_type_drivers = attributes["ml2_type_drivers"]
       if attributes["ml2_mechanism_drivers"].include?("openvswitch")
+        # This node needs the ovs packages installed and the service started
+        unless node.crowbar["network"]
+          node.crowbar["network"] = {}
+        end
```
vuntz: `node.crowbar["network"] ||= {}` instead of the 3 lines?

vuntz:
Do we need to do that for compute nodes also?

rhafer:
> Do we need to do that for compute nodes also?

It's already there. NovaService actually calls into  NeutronService for that already. See: https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/models/nova_service.rb#L218

crowbar_framework/app/models/neutron_service.rb
rhafer: Good, question :). I'll change that.

vuntz:
+1

rhafer:
> This needs to be merged together with crowbar/crowbar-core#105

BTW, I was wrong with that. This one mergable without crowbar/crowbar-core#105. In fact it's better to merge this one before. That way the CI should be able to succeed for the other PR.

rhafer:
I needed to rework the condition to disable "needs_openvswitch" a bit. It might have wrongly disabled the switch on compute nodes in some cases.

jsuchome:
@rhafer so it's in do-not-merge state now?

rhafer:
@jsuchome No. It's fixed already. Just requires review.

jsuchome:
+1 (for the code; not tested)

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -308,6 +316,12 @@ def update_ovs_bridge_attributes(attributes, node)
         end
       end
     end
+    if ovs_bridge_networks.empty? && !ml2_mechanism_drivers.include?("openvswitch")
```
rhafer: @vuntz still ok with this? This is the part I updated. I didn't have that ```&& !ml2_mechanism_drivers.include?("openvswitch")``` before, which would have caused the ```needs_openvswitch``` flag to be unset again on compute nodes :/.

vuntz: Yes, +1

aspiers:
Would be nice to have the full bug URL in the commit body ;-)

PR #81, opened 2015-10-30
cinder: Make tgtd listen on admin IP address only (bsc#828214)

vuntz:
http://bugzilla.suse.com/show_bug.cgi?id=828214

### DIFF
--- a/chef/cookbooks/cinder/templates/default/tgt.sysconfig.erb
+++ b/chef/cookbooks/cinder/templates/default/tgt.sysconfig.erb
@@ -1,0 +1,10 @@
+## Path:        Network/File systems/iSCSI
+## Description: tgt iSCSI server options
+## Type:        string
+## Default:     ""
+## ServiceRestart: tgtd
+#
+# Options for tgtd
+# set listen addr with --iscsi portal=192.0.2.1:3260
+#
+TGTD_OPTS="--iscsi portal=<%= @admin_ip %>"
--- a/chef/cookbooks/cinder/recipes/volume.rb
+++ b/chef/cookbooks/cinder/recipes/volume.rb
@@ -275,6 +275,19 @@ if node[:platform] == "suse" && node[:platform_version].to_f < 12.0
   end
 end
 
+if node[:platform_family] == "suse"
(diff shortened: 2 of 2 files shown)


chef/cookbooks/cinder/templates/default/tgt.sysconfig.erb
```diff
@@ -0,0 +1,10 @@
+## Path:        Network/File systems/iSCSI
+## Description: tgt iSCSI server options
+## Type:        string
+## Default:     ""
+## ServiceRestart: tgtd
+#
+# Options for tgtd
+# set listen addr with --iscsi portal=192.0.2.1:3260
+#
+TGTD_OPTS="--iscsi portal=<%= @admin_ip %>"
```
tboerger: What about the port? In the example above it's written like `--iscsi portal=192.0.2.1:3260`.

tboerger:
One comment but LGTM

chef/cookbooks/cinder/templates/default/tgt.sysconfig.erb
vuntz: I think it's optional, and the CI seems to agree. I can add it if desired, though.

tboerger: If you can confirm that it works fine I have no problem with it.

rsalevsky:
+1

mjura:
Ehhh, admin network is not right for exporting and using storage shares! In most cases users defined `admin` on really slow network interfaces like 100Mb/s or 1Gb/s. Why don't we want to use tgt in `storage` network? For me this is proper place for this.

vuntz:
@mjura this is really unrelated to this PR. At the moment, `admin` is already what is being used by default as the nodes are not on the `storage` network...

chef/cookbooks/cinder/templates/default/tgt.sysconfig.erb
vuntz: I checked, it is optional.

vuntz:
Anyone else up for a second +1?

aplanas:
+1

toabctl:
+1

PR #82, opened 2015-10-30
Update template configs with Liberty

mjura:


mjura:
Creating separate PR
https://github.com/crowbar/crowbar-openstack/pull/95
https://github.com/crowbar/crowbar-openstack/pull/94

PR #83, opened 2015-10-31
neutron: Fix multiple external networks with DVR

vuntz:
The non-network nodes also need to have access to the external networks
when using DVR.


vuntz:
I'm not 100% sure about this, but this feels logical. cc @rossella

tboerger:
Codewise LGTM, but I don't know if the logic is correct.

rossella:
@vuntz confirmed, with DVR also compute nodes need access to the external network to be able to serve floating ips

rhafer:
+1 (@vuntz feel free to merge as it requires admin powers)

PR #89, opened 2015-11-05
Use patched tgtd.service to work around bnc#950946

jsuchome:
(added a small sleep between starting tgtd by bmwiedemann

This is an alternative way of "fixing" the problem when tgtd fails to start and breaks cinder deployment.
Better option is probably https://build.suse.de/request/show/77987
Even better would be having the fixed package in SLE11SP1 GM.

### DIFF
--- a/chef/cookbooks/cinder/recipes/volume.rb
+++ b/chef/cookbooks/cinder/recipes/volume.rb
@@ -275,6 +275,15 @@ if node[:platform] == "suse" && node[:platform_version].to_f < 12.0
   end
 end
 
+# use patched tgtd.service to work around bnc#950946
+template "/usr/lib/systemd/system/tgtd.service" do
+  source "tgtd.service.erb"
+  owner "root"
+  group "root"
+  mode "0644"
+  only_if { node[:platform] == "suse" && node[:platform_version].to_f >= 12.0 }
+end
+
 service "tgt" do
   supports status: true, restart: true, reload: true
   action [:enable, :start]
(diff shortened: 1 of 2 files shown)


aplanas:
+1

We need to remember revert this when the new version of the package arrives

jsuchome:
@aplanas Or we can not place it in git at all and only add as a patch to package? https://build.suse.de/package/show/home:jsuchome:branches:Devel:Cloud:6:Staging/crowbar-openstack

aplanas:
@jsuchome I like more this idea of putting the patch `cinder-tgt-service-hotfix.patch` in the package

jsuchome:
Now we have https://build.suse.de/request/show/77987 let's see if it's the best option

chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -275,6 +275,15 @@ def make_volume(node, backend_id, volume)
   end
 end
 
+# use patched tgtd.service to work around bnc#950946
+template "/usr/lib/systemd/system/tgtd.service" do
```
vuntz: You would want it in /etc/systemd/system/, instead of changing the file owned by the package :-)

jsuchome: Yes, I've tried it, but probably has done something wrong, because I think still this one was used when starting the service.

Anyway, we went with placing new tgt package on Cloud media for now

vuntz: > Yes, I've tried it, but probably has done something wrong, because I think still this one was used when starting the service.

You need to reload systemd :-)

rhafer:
The patched tgt seems to have solve the issue. So I guess this one is no longer needed?

jsuchome:
Right

PR #77, opened 2015-10-22
Fix ceilometer middleware configuration

dirkmueller:
The rabbitmq settings in Liberty are expected as an url instead.

PR #92, opened 2015-11-09
Move rabbmitmq conf for ceilometer to the correct place

rhafer:
The rabbmitmq settings for ceilometer need to appear in the filter section for
ceilometer not in [DEFAULT].


rhafer:
This caused gate failures like: https://ci.suse.de/job/openstack-mkcloud/11057/consoleFull

dirkmueller:
@rhafer This is already fixed in https://github.com/crowbar/crowbar-openstack/pull/77 -- waiting for review..

PR #94, opened 2015-11-11
Update keystone.conf config template with Liberty

mjura:
Updating keystone.conf template config with OpenStack Liberty

### DIFF
--- a/chef/cookbooks/keystone/attributes/default.rb
+++ b/chef/cookbooks/keystone/attributes/default.rb
@@ -26,7 +26,7 @@ default[:keystone][:group] = "keystone"
 
 default[:keystone][:debug] = false
 default[:keystone][:frontend] = "apache"
-default[:keystone][:verbose] = false
+default[:keystone][:verbose] = true
 default[:keystone][:domain_specific_drivers] = false
 default[:keystone][:domain_config_dir] = "/etc/keystone/domains"
 
@@ -44,8 +44,8 @@ default[:keystone][:api][:api_host] = "0.0.0.0"
 default[:keystone][:api][:version] = "2.0"
 default[:keystone][:api][:region] = "RegionOne"
 
-default[:keystone][:identity][:driver] = "keystone.identity.backends.sql.Identity"
-default[:keystone][:assignment][:driver] = "keystone.assignment.backends.sql.Assignment"
+default[:keystone][:identity][:driver] = "sql"
+default[:keystone][:assignment][:driver] = "sql"
 
 default[:keystone][:sql][:idle_timeout] = 30
 
@@ -67,7 +67,7 @@ default[:keystone][:ldap][:password] = ""
 default[:keystone][:ldap][:suffix] = "cn=example,cn=com"
 default[:keystone][:ldap][:use_dumb_member] = false
 default[:keystone][:ldap][:allow_subtree_delete] = false
-default[:keystone][:ldap][:dumb_member] = "cn=dumb,dc=example,dc=com"
+default[:keystone][:ldap][:dumb_member] = "cn=dumb,dc=nonexistent"
 default[:keystone][:ldap][:page_size] = 0
 default[:keystone][:ldap][:alias_dereferencing] = "default"
 default[:keystone][:ldap][:query_scope] = "one"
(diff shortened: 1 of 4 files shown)


vuntz:
Can you update template-keystone.json with default values that have changed? See http://docs.openstack.org/draft/config-reference/content/keystone-conf-changes-liberty.html

vuntz:
I'm a bit mixed on keeping the drivers for revoke and trust (and maybe others) commented out while we had them set explicitly before. If this changes upstream without us noticing, then this is bad. On the other hand, that's what we usually do. So maybe it's okay.

vuntz:
Re:
```
<% if @signing_token_format == "PKI" %>
provider = pki
<%end %>
```
I'd prefer to have an explicit `else` for `uuid` since we're likely going to add support for something else (fernet?) soon.

vuntz:
Wondering if we should set `notification_format` to `cadf`. Any opinion?

mjura:
Let me answer on your questions:
- all options in `keystone.conf` template are with defaults 
- I think we should stay with upstream as close as we can, that's way we can keep default drivers
- I've updated `provider` option and added `uuid` but this makes no sense for me
- With `notification_format` we can stay also with `basic` option. I didn't hear about any billing system integration, so it makes no sense to change it
http://docs.openstack.org/developer/keystone/event_notifications.html

mjura:
PR was rebased, gating was green

vuntz:
> - all options in keystone.conf template are with defaults
> - I think we should stay with upstream as close as we can, that's way we can keep default drivers

Will re-check. I guess you missed my point about the default values in the template: we should move from keystone.foo.bar.sql.foo to just sql.

> - I've updated provider option and added uuid but this makes no sense for me
> - With notification_format we can stay also with basic option. I didn't hear about any billing system integration, so it makes no sense to change it http://docs.openstack.org/developer/keystone/event_notifications.html

This is not just for billing, but also for auditing. And there it matters quite a bit.

mjura:
@vuntz : I've change for `[resource]` and `[role]` sections drivers from `<None>` to `sql`. Rest drivers are sql or setup by crowbar options.

`notification_format` still want to leave as basic. But I'm fine with everything.

toabctl:
@mjura: I think in the attributes.rb file, ```default[:keystone][:identity][:driver]``` should be changed to ```sql```

toabctl:
@mjura same for ```default[:keystone][:assignment][:driver]``` I think. And here also the ```<% if``` in keystone.conf.erb must be adjusted.

toabctl:
@mjura https://github.com/mjura/crowbar-openstack/blob/keystone_liberty_conf/chef/cookbooks/keystone/templates/default/keystone.conf.erb#L926 must be adjusted due to https://github.com/mjura/crowbar-openstack/commit/1ed90d5dd51b6f7436636ba51576925ad23dc267

mjura:
@toabctl : Thanks, I've changed default attributes
```
default[:keystone][:identity][:driver] 
default[:keystone][:assignment][:driver]
``` 
to `sql`

for `driver` in [assignment] section we could go with 
`
driver = <%= node[:keystone][:assignment][:driver] rescue "sql" %>
`

mjura:
I've set up also
```
<% if node[:keystone][:assignment][:driver] == 'hybrid' %>
```

mjura:
Rebased

toabctl:
+1

vuntz:
Minor nitpick:
```
<% if @signing_token_format == "PKI" %>
provider = pki
<% else %>
provider = uuid
<%end %>
```

You should end the `<%` tags with `-%>`

vuntz:
> I guess you missed my point about the default values in the template: we should move from keystone.foo.bar.sql.foo to just sql.

This is still valid, just do `git grep "keystone.*backends"` to find all the locations where it needs to be changed. I'm happy to have it done in a followup PR if it's tracked in the trello card, though.

vuntz:
> Edit: the important ones I immediately see are user_attribute_ignore

This is still valid (this was about the default attributes of the barclamp).

vuntz:
> `notification_format` still want to leave as basic

Can you explain why (so I get convinced ;-))?

mjura:
Things which were done:
   - chef statements was fixed in couple places
   - default attributes for cookbook and barclamp were synchronize

mjura:
>> notification_format still want to leave as basic
>Can you explain why (so I get convinced ;-))?

I think that the best would be that we don't use this notifications anywhere, so there is not need to modify default values. Or I am wrong and missed something.

mjura:
> This is still valid, just do git grep "keystone.*backends" to find all the locations where it needs to be > changed. I'm happy to have it done in a followup PR if it's tracked in the trello card, though.

Follow up card was created

mjura:
> Edit: note that this will also require a migration for the values that are in the proposal.

I added migration script which is changing defaults

mjura:
@vuntz : Thank you for code-review, I think that I addressed all you comments.

I rebased this PR, let's wait for gating

dirkmueller:
+1

vuntz:
> >> notification_format still want to leave as basic
> 
>    > Can you explain why (so I get convinced ;-))?
> 
> I think that the best would be that we don't use this notifications anywhere, so there is not need to modify default values. Or I am wrong and missed something.

People can use the notifications for auditing purpose, and the additional details there are useful in this context. Now there might be downsides to changing the option, but unless the downsides are really bad, I think it's useful to change the default.

PR #97, opened 2015-11-11
Update neutron.conf config template with Liberty

mjura:


vuntz:
@mjura just to clarify: do you intend to update the other neutron templates in this PR, or in another PR?

mjura:
Ready for code-review

toabctl:
@mjura needs a rebase

mjura:
Rebased

vuntz:
For the record, the following templates have not been updated (could be because there's no change, though):
- linuxbridge_conf.ini.erb
- ml2_conf_cisco.ini.erb
- ml2_conf.ini.erb
- nsx.ini.erb
- ovs_neutron_plugin.ini.erb

vuntz:
In metadata_agent.ini.erb: can you either comment out or remove the default line `auth_url = http://localhost:5000/v2.0`? Same for the other settings which have a default line that is not commented in the upstream sample! right now some are commented out (`# auth_region = RegionOne` and some are removed `admin_tenant_name = %SERVICE_TENANT_NAME%`)

vuntz:
In metadata_agent.ini.erb: you lost this from the sample:

```
-[AGENT]
-# Log agent heartbeats from this Metadata agent
-# log_agent_heartbeats = False
```

vuntz:
Nearly sure that we need a template for /etc/neutron/neutron_lbaas.conf

vuntz:
And we likely also need a template for /etc/neutron/services_lbaas.conf

vuntz:
There's still a diff between `lbaas_agent.ini.erb` and the file from neutron-lbaas.

vuntz:
neutron.conf: Nearly sure there should be some region_name setting in [keystone_authtoken]

vuntz:
`driver = neutron.services.metering.drivers.iptables.iptables_driver.IptablesMeteringDriver` got lost in metering_agent.ini.erb -- I assume it's because it feels it's the default value, but it's actually totally unclear from the template what the default value is, so better to keep ot.

vuntz:
neutron.conf.erb: same comment as for nova for `rpc_backend=rabbit`: I'd keep it explicit.

vuntz:
neutron.conf.erb: I guess you missed the comment before `auth_strategy` when you removed the explicit value:

```
# Note: explicitely set this because of RHEL (where noauth is used)
auth_strategy = keystone
```

vuntz:
The nova related options in neutron.conf should be moved to [nova]

vuntz:
[DEFAULT] dnsmasq_base_log_dir might be valuable to set when debugging is enabled.

vuntz:
[DEFAULT] dns_domain probably needs to be set to `node[:neutron][:dhcp_domain]`

vuntz:
For the record, it seems we might want an attribute to enable tunnel_csum in openvswitch_agent.ini as it can improve performance: http://openvswitch.org/pipermail/dev/2015-August/059335.html

vuntz:
Wondering if we should have an attribute for [ml2] external_network_type (new setting in ml2_conf.ini)

vuntz:
Question: does `allow_automatic_dhcp_failover = True` have any impact wrt neutron-ha-tool? (I think no, but not 100% sure)

vuntz:
comment_iptables_rules should probably be set when debugging is enabled.

vuntz:
Can we / should we set prevent_arp_spoofing in ovs_neutron_plugin.ini? (Asking if we can because our ovs version might not allow that)

vuntz:
Have you checked if we want to deal with OVSvApp? (it seems related to vsphere)

vuntz:
There's also some NSXv bits to check (see https://blueprints.launchpad.net/neutron/+spec/vmware-nsx-v)

vuntz:
Last note for now: of course some of the bits I mentioned are possibly not appropriate for this PR and should be moved to followup cards, but I'll let you decide which ones :-)

mjura:
Updated templates:
- linuxbridge_conf.ini.erb
- ml2_conf.ini.erb
- ovs_neutron_plugin.ini.erb

Templates for 
- nsx.ini.erb
- ml2_conf_cisco.ini.erb
will be done in another trello cards.

mjura:
> Have you checked if we want to deal with OVSvApp? (it seems related to vsphere)

Yeap, I added already card for doing this

mjura:
> There's also some NSXv bits to check (see https://blueprints.launchpad.net/neutron/+spec/vmware-nsx-v)

NSX will be done in another card too.

mjura:
> Can we / should we set prevent_arp_spoofing in ovs_neutron_plugin.ini? (Asking if we can because our ovs version might not allow that)

I think this is related with something else https://blueprints.launchpad.net/neutron/+spec/arp-spoof-patch-ebtables

mjura:
> comment_iptables_rules should probably be set when debugging is enabled.

It has default value `true` always should we respect this only when `debug = true`

vuntz:
> > Can we / should we set prevent_arp_spoofing in ovs_neutron_plugin.ini? (Asking if we can because our ovs version might not allow that)
> 
> I think this is related with something else https://blueprints.launchpad.net/neutron/+spec/arp-spoof-patch-ebtables

I think your bit is the generic version, I'm talking about https://review.openstack.org/#/c/171003/

vuntz:
> > comment_iptables_rules should probably be set when debugging is enabled.
> 
> It has default value true always should we respect this only when debug = true

My bad, I thought the default was false. Let's keep it that way then!

rhafer:
> Wondering if we should have an attribute for [ml2] external_network_type (new setting in ml2_conf.ini)

We might want to set it to "flat" by default. Cause that is what we create default external network with. Currently I don't think we need to parametrize it.

mjura:
@rhafer: I've added `external_network_type = flat` in ml2_conf.ini

mjura:
Things which left still

TODO:
- LBASS
  * add template for /etc/neutron/neutron_lbaas.conf
  * add template for /etc/neutron/services_lbaas.conf
- diff between lbaas_agent.ini.erb and the file from neutron-lbaas
- nova related options in neutron.conf should be moved to [nova]
- it seems we might want an attribute to enable tunnel_csum in openvswitch_agent.ini as it can   improve performance: http://openvswitch.org/pipermail/dev/2015-August/059335.html
- does allow_automatic_dhcp_failover = True have any impact wrt neutron-ha-tool? (I think no, but not 100% sure)
should we set prevent_arp_spoofing in ovs_neutron_plugin.ini? (Asking if we can because our ovs version might not allow that)

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
```diff
@@ -65,11 +75,13 @@ enable_metadata_network = <%= @enable_metadata_network %>
 # Location to store DHCP server config files
 # dhcp_confs = $state_path/dhcp
 
-# Domain to use for building the hostnames
+# Domain to use for building the hostnames. This option will be deprecated in
+# a future release. It is being replaced by dns_domain in neutron.conf
 # dhcp_domain = openstacklocal
 dhcp_domain = <%= @dhcp_domain %>
```
rhafer: Looking at the comment for this, do we really still need this setting?  We already set ```dns_domain``` in neutron.conf.

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
```diff
@@ -79,22 +91,32 @@ dnsmasq_config_file = /etc/neutron/dnsmasq-neutron.conf
...
+# Disable this if you hit the issue in
+# https://bugs.launchpad.net/neutron/+bug/1052535 or if
+# you are sure that your version of iproute suffers from the problem.
+# This should not be a problem any more.  Refer to bug:
+# https://bugs.launchpad.net/neutron/+bug/1418079
+# This option is deprecated and will be removed in the M release
+# dhcp_delete_namespaces = True
 <% if node[:platform_family] == "suse" -%>
 dhcp_delete_namespaces = True
 <% end -%>
```
rhafer: Hm, the default for ```dhcp_delete_namespaces``` is already ```True``` so if we need to set something here (I don't know) we need do it for ```node[:platform_family] != "suse"```

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
```diff
@@ -74,16 +96,22 @@ periodic_fuzzy_delay = <%= @periodic_fuzzy_delay %>
...
+# you are sure that your version of iproute suffers from the problem.
 # If True, namespaces will be deleted when a router is destroyed.
-# router_delete_namespaces = False
+# This should not be a problem any more.  Refer to bug:
+# https://bugs.launchpad.net/neutron/+bug/1418079
+# This option is deprecated and will be removed in the M release
+# router_delete_namespaces = True
 <% if node[:platform_family] == "suse" -%>
 router_delete_namespaces = True
 <% end -%>
```
rhafer: Here the same applies as for ```dhcp_delete_namespaces```

mjura: Removed

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
mjura: Removed

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
mjura: If everybody are fine I would like to keep it in case of some regression

mjura:
Follow up cards were created for:

- enable auth_plugin in [nova] section in neutron.conf
- enable attribute to enable tunnel_csum in openvswitch_agent.ini
- allow_automatic_dhcp_failover = True impact on neutron-ha-tool
- enable prevent_arp_spoofing in ovs_neutron_plugin.ini

mjura:
All comments were addressed and PR was rebased

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
```diff
@@ -39,8 +13,8 @@ physical_interface_mappings = <%= @interface_mappings %>
 [vxlan]
 # (BoolOpt) enable VXLAN on the agent
 # VXLAN support can be enabled when agent is managed by ml2 plugin using
-# linuxbridge mechanism driver. Useless if set while using linuxbridge plugin.
-# enable_vxlan = False
+# linuxbridge mechanism driver.
+# enable_vxlan = True
 <% if @ml2_type_drivers.include?("vxlan") -%>
 enable_vxlan = True
```
rhafer: It seems the default for ```enable_vxlan``` changed to ```True``` it seems. So we need to change the condition to something like this:

```
 <% unless @ml2_type_drivers.include?("vxlan") -%>
 enable_vxlan = False
```

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
dguitarbite: This option is clearly not required if ``dns_domain`` is already being set.

mjura: We are going to keep it, until this will be deprecated and removed completely in case of some regression.

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
mjura: Thank you Ralf, that's good point

mjura:
Rebased

rhafer:
In general this looks good to me now (I didn't do much testing though). I am still a bit unsure about the config for the split out neutron-services (lbaas, fwaas), but I just asked @mjura to create a followup card for those.

So +1 from my side (when the cards have been created).

dguitarbite:
+1 from my side. I did not test the code. Based on reading through it.

mjura:
Ok, we have tested together with @rhafer lbaas packages and configuration. Works fine with change  https://build.opensuse.org/package/rdiff/Cloud:OpenStack:Liberty:Staging/openstack-neutron-lbaas?linkrev=base&rev=12

mjura:
It was rebased

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -147,7 +147,8 @@
       use_namespaces: true,
       allow_overlapping_ips: neutron[:neutron][:allow_overlapping_ips],
       dvr_enabled: neutron[:neutron][:use_dvr],
-      network_nodes_count: network_nodes_count
+      network_nodes_count: network_nodes_count,
+      dns_domain: neutron[:neutron][:dhcp_domain]
```
rossella: s/dhcp_domain/dns_domain ?

mjura: I would prefer to not rename this attribute, while is still used by nova to setup `dhcp_domain` option and it is still used in chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb to setup also `dhcp_domain` option

mjura:
It was only rebased

chef/cookbooks/neutron/recipes/common_config.rb
rossella: I see, then why do we have a variable for 'dns_domain' in chef/cookbooks/neutron/templates/default/neutron.conf.erb
 line 137? It's confusing...we should use dhcp_domain in the neutron.conf.erb too, don't you think?

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -119,6 +132,10 @@ auth_strategy = keystone
 # tell dnsmasq to use infinite lease times.
 # dhcp_lease_duration = 86400
 
+# Domain to use for building the hostnames
+# dns_domain = openstacklocal
+dns_domain = <%= @dns_domain %>
```
rossella: based on my previous comment, if you want to keep both dhcp_domain and dns_domain, it makes sense to use just one variable...so here it should be

dns_domain = <%= @dhcp_domain %>

what do you think?

PR #100, opened 2015-11-12
Fix migration of nova-multi-* to nova-* in HA case

vuntz:
We also need to migrate the elements_expanded attribute


tboerger:
:+1:

rsalevsky:
+1

mjura:
+1

PR #101, opened 2015-11-12
neutron: Always automatically enable l2pop whenever possible

vuntz:
The only reason we had an attribute for it is that it was not possible
to automatically set this up during a maintenance upgrade as it was
possibly creating some downtime. Since we're moving to a new version,
there will be downtime anyway, so let's use this as an opportunity to
enable l2pop.

There should be no big downside to just always enable l2pop when it's
possible to use it.


vuntz:
cc @rossella

rossella:
@vuntz I can't see any big downside, go for it

tboerger:
:+1:

mjura:
Looks OK
+1

PR #104, opened 2015-11-13
tempest: update tempest.conf.erb to Liberty

toabctl:
This is a copy of the Liberty tempest.conf.sample
with our current modification.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -132,7 +132,6 @@ end
 
 machine_id_file = node[:tempest][:tempest_path] + "/machine.id"
 docker_image_id_file = node[:tempest][:tempest_path] + "/docker_machine.id"
-heat_machine_id_file = node[:tempest][:tempest_path] + "/heat_machine.id"
 
 glance_node = search(:node, "roles:glance-server").first
 insecure = "--insecure"
@@ -217,31 +216,6 @@ EOH
     "OS_PROJECT_DOMAIN_NAME" => keystone_settings["api_version"] != "2.0" ? "Default" : ""
   })
   not_if { File.exists?(machine_id_file) }
-end
-
-bash "upload tempest heat-cfntools image" do
-    code <<-EOF
-OS_USERNAME=${OS_USERNAME:-admin}
-OS_TENANT_NAME=${OS_TENANT_NAME:-admin}
-OS_PASSWORD=${OS_PASSWORD:-admin}
-
-id=$(glance #{insecure} image-show ${IMAGE_NAME} | awk '/id/ { print $4}')
-[ -n "$id" ] && echo $id > #{heat_machine_id_file}
-
-true
-EOF
-  environment ({
-    "IMAGE_NAME" => node[:tempest][:heat_test_image_name],
-    "OS_USERNAME" => tempest_adm_user,
-    "OS_PASSWORD" => tempest_adm_pass,
-    "OS_TENANT_NAME" => tempest_comp_tenant,
-    "OS_AUTH_URL" => keystone_settings["internal_auth_url"],
-    "OS_IDENTITY_API_VERSION" => keystone_settings["api_version"],
-    "OS_USER_DOMAIN_NAME" => keystone_settings["api_version"] != "2.0" ? "Default" : "",
-    "OS_PROJECT_DOMAIN_NAME" => keystone_settings["api_version"] != "2.0" ? "Default" : ""
-  })
-
-  not_if { node[:tempest][:heat_test_image_name].nil? or File.exists?(heat_machine_id_file) }
 end
 
 flavor_ref = "6"
@@ -474,7 +448,6 @@ template "/etc/tempest/tempest.conf" do
     object_versioning: swift_allow_versions,
     # orchestration settings
     heat_flavor_ref: heat_flavor_ref,
-    heat_machine_id_file: heat_machine_id_file,
     # scenario settings
     cirros_version: cirros_version,
     image_regex: image_regex,
--- a/chef/cookbooks/tempest/templates/default/tempest.conf.erb
+++ b/chef/cookbooks/tempest/templates/default/tempest.conf.erb
@@ -3,30 +3,18 @@
 <% img_id = `cat #{@machine_id_file}`.strip %>
 
 #
-# From tempest.config
-#
-
-# Whether to disable inter-process locks (boolean value)
-#disable_process_locking = false
-
-# Directory to use for lock files. (string value)
-lock_path = /var/run/tempest
-
-#
-# From tempest.config
+# From oslo.log
 #
 
 # Print debugging output (set logging level to DEBUG instead of
-# default WARNING level). (boolean value)
+# default INFO level). (boolean value)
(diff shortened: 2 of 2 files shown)


toabctl:
FYI: I haven't tested this yet.

toabctl:
Initial smoke tests passed (but only 126 tests: ```Ran 126 tests in 952.512s```. not sure if that's correct)

toabctl:
Result of a full tempest run has ```Ran 1372 tests in 7244.439s```

mjura:
Could you check `api_v3` option once again, I see that it was 
```
# If false, skip all nova v3 tests. (boolean value)
api_v3 = <%= @nova_api_v3 ? 'true' : 'false' %>
```

mjura:
Another option which I have found is `disable_ssl_certificate_validation`
```
# Set to True if using self-signed SSL certificates. (boolean value)
disable_ssl_certificate_validation = <%= @ssl_insecure ? 'true' : 'false' %>
```

mjura:
Please check also `domain_name = Default`

mjura:
And `deactivate_image = true`

mjura:
To check `build_timeout = 2000`

toabctl:
> And deactivate_image = true

@mjura That's already set

toabctl:
> To check build_timeout = 2000

@mjura already set

toabctl:
> And deactivate_image = true

@mjura already set

toabctl:
> Could you check api_v3 option once again

@mjura: done in https://github.com/crowbar/crowbar-openstack/commit/fe10900614c03a0e919d361edfc7d04ca592c3a6

toabctl:
> Another option which I have found is disable_ssl_certificate_validation

@mjura: done in https://github.com/crowbar/crowbar-openstack/commit/fe10900614c03a0e919d361edfc7d04ca592c3a6

toabctl:
> Please check also domain_name = Default

@mjura: done in https://github.com/crowbar/crowbar-openstack/commit/fe10900614c03a0e919d361edfc7d04ca592c3a6

toabctl:
@mjura any other comments or a +1?

mjura:
If gating will be green, then +1

vuntz:
In case there's a need for a new update to the template, can you change `<%=[^ ]` to `<%= ` (in many places, we're missing a space for readability). But only do it in case you need to update the template.

vuntz:
`api_v3 = <%= @nova_api_v3 ? 'true' : 'false' %>` and `nova_api_v3: nova[:nova][:enable_v3_api],`

Not your fault, but the attribute is gone for nova. I'm not sure which tempest tests are still using this -- if this is like nova, this actually controls v2.1 testing and we want to always enable this. Otherwise, we should always disable it.

vuntz:
Not strictly related to the config template, but to the move to liberty: iirc, there's some magic in tempest now to do the cleanup. Which means we could remove the tempest_cleanup.sh script / template too.

vuntz:
Shouldn't we set [telemetry-feature-enabled] events to true?

vuntz:
What about [volume-feature-enabled] bootable?

vuntz:
> api_v3 = <%= @nova_api_v3 ? 'true' : 'false' %> and nova_api_v3: nova[:nova][:enable_v3_api],
> 
> Not your fault, but the attribute is gone for nova. I'm not sure which tempest tests are still using this -- if this is like nova, this actually controls v2.1 testing and we want to always enable this. Otherwise, we should always disable it.

I just realized the option is in [identity-feature-enabled]; so clearly we're not setting it correctly :-)

vuntz:
`tenant_name = <%= @keystone_settings['default_tenant'] %>` used to be `tenant_name = <%= @comp_tenant %>`. Is the change intended?

vuntz:
`password = <%= @comp_pass %>` got lost in [identity]

vuntz:
Not sure if it's fine to have [compute] ssh_user not be changed to cirros. Can you confirm?

vuntz:
Wondering out loud (not related to this PR, though): we don't change `#volume_device_name = vdb`, but maybe it's needed for xen?

vuntz:
```
# Does Nova preserve preexisting ports from Neutron when deleting an
# instance? This should be set to True if testing Kilo+ Nova. (boolean
# value)
#preserve_ports = false
```
=> setting should be changed :-)

toabctl:
@vuntz addressed all your comments (except removing the tempest_cleanup script) in https://github.com/crowbar/crowbar-openstack/commit/399cae8e0e92c7d969c455a22f83136f9d2337dd

vuntz:
+1

Want me to create a followup card about the remaining comments (mostly the cleanup script and volume_device_name for xen)?

toabctl:
> Want me to create a followup card about the remaining comments (mostly the cleanup script and volume_device_name for xen)?

Done: https://trello.com/c/LfO82kea/807-cleanup-tempest-barclamp

toabctl:
@mjura still +1? Or any other comments?

vuntz:
@toabctl in http://docs.openstack.org/kilo/config-reference/content/nova-conf-changes-kilo.html, I see that the ec2_path changed from /services/Cloud to /, so we might have to change ec2_url accordingly.

toabctl:
> @toabctl in http://docs.openstack.org/kilo/config-reference/content/nova-conf-changes-kilo.html, I see that the ec2_path changed from /services/Cloud to /, so we might have to change ec2_url accordingly.

We still have /services/Cloud in our nova.conf . Should we postpone it until the nova.conf.erb is updated?

toabctl:
Retested the latest config. ```Ran 1372 tests in 6858.053s ; FAILED (failures=10)```. Failures are S3 related and also telemetry notifications.

toabctl:
The failing test is: ```# python -mtesttools.run -- tempest.api.telemetry.test_telemetry_notification_api.TelemetryNotificationAPITestJSON.test_check_nova_notification```. Gets a timeout (waiting for a notification). maybe something we have in the new nova.conf.erb ?

vuntz:
> We still have /services/Cloud in our nova.conf . Should we postpone it until the nova.conf.erb is updated?

We have a commented out value, though. So the default is used, and the new default is `/` :-)

vuntz:
@toabctl nearly sure the telemetry failure you're seeing exists before this PR.

toabctl:
> We have a commented out value, though. So the default is used, and the new default is / :-)

Fixed in https://github.com/crowbar/crowbar-openstack/commit/a7a63826a5324ea9073bbe613172e41df92918ee

vuntz:
+1

dirkmueller:
@toabctl can you remove the heat_machine_id_file varible now that it is no longer used?

toabctl:
@dirkmueller done. @vuntz still +1 ?

vuntz:
@toabctl this is a never-ending story, because now you can remove the remaining reference to `heat_test_image_name`

toabctl:
@vuntz but can we stop here, merge it and continue with the cleanup in another PR?

vuntz:
I'm fine with that, sure. My +1 is still valid in that case.

dirkmueller:
+1

PR #105, opened 2015-11-13
Help Docker instances get proper MTU value

abel-navarro:
We already send the right MTU value to instances via an option in DHCP,
but this mechanism may not work in all Docker instances because some of
them won't even run a DHCP client. Having the right MTU setting on
instance creation is needed. There's an option (network_device_mtu)
that can be configured in nova.conf to achieve that.

The 1500-byte MTU is too big when a tunneling underlay is used. The
overhead caused by the tunneling is usually 20-byte for IPv4 header,
8-byte for either GRE or VxLAN and 14-byte for Ethernet, totalling
42 bytes. In this situation the right MTU should be 1458 bytes. In
this patch we use 1450 in case two additional VLAN headers are added
in the overlay. This should also be revisited when the underlay or
overlay are using IPv6.

This is only needed when the compute node is running Docker instances
and the networking requires tunneling.

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -1487,6 +1487,9 @@ security_group_api=neutron
 # DEPRECATED: THIS VALUE SHOULD BE SET WHEN CREATING THE
 # NETWORK. MTU setting for network interface. (integer value)
 #network_device_mtu=<None>
+<% if @libvirt_type.eql?('docker') && @neutron_has_tunnel -%>
+network_device_mtu=1400
+<% end -%>
 
 
 #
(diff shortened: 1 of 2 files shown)


abel-navarro:
This has manually tested in a local environment.

chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -1486,8 +1486,11 @@ security_group_api=neutron
 
 # DEPRECATED: THIS VALUE SHOULD BE SET WHEN CREATING THE
 # NETWORK. MTU setting for network interface. (integer value)
+<% if @libvirt_type.eql?('docker') && node[:network][:needs_openvswitch] -%>
+network_device_mtu=1450
+<% else -%>
 #network_device_mtu=<None>
-
+<% end -%>
```
vuntz: Just always leave the commented value, and don't remove an empty line. So in short:
```
# NETWORK. MTU setting for network interface. (integer value)
#network_device_mtu=<None>
network_device_mtu=1450
```

Generally, this should look more like https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/neutron/recipes/network_agents.rb#L112:

- we should use 1400 and not 1450
- this is not ovs-specific (it's also required with linuxbridges when using vxlan)

So I would update the condition accordingly.

vuntz:
+1

toabctl:
+1

PR #106, opened 2015-11-13
bin: Add 3600s timeout to all crowbar subcommands

vuntz:
Some didn't have that, and I was seeing manila failing from time to
time.


toabctl:
+1

rsalevsky:
I think this change makes no sense right now. @tboerger Is working on the new crowbar client and this will replace this scripts.

vuntz:
@rsalevsky except that this helps avoid issues in the CI now. And this doesn't conflict with the work on replacing the scripts.

rsalevsky:
@vuntz Ah ok, yeah I have nothing against it at the changes are fine.

:+1:

tboerger:
:+1:

PR #107, opened 2015-11-14
horizon: increase timeout in apache config

toabctl:
On a slow system, the first request can take a long time due
to online compression. Apache is then running into a:

Timeout when reading response headers from daemon process 'horizon'

error.


dirkmueller:
Checks seems to pass, so +1

vuntz:
Let's merge it for now to unbreak things, but I really think we need a way to trigger that first request so that users don't experience it...

PR #108, opened 2015-11-15
Allow cinder to be set as a default backend store

dirkmueller:


crowbar_framework/app/models/glance_service.rb
```diff
@@ -77,6 +77,16 @@ def create_proposal
   def validate_proposal_after_save proposal
     validate_one_for_role proposal, "glance-server"
 
+    cinder_proposal = Proposal.where(barclamp: "cinder", name: "default").first
+    if cinder_proposal.nil?
+
+      if proposal["attributes"]["glance"]["default_store"] == "cinder"
```
vuntz: Maybe rewrite to:
```ruby
if proposal["attributes"]["glance"]["default_store"] == "cinder"
  cinder_proposal = Proposal.where(barclamp: "cinder", name: "default").first
  if cinder_proposal.nil?
```
(to avoid an unneeded query)

crowbar_framework/config/locales/glance/en.yml
```diff
@@ -57,3 +57,5 @@ en:
         use_cachemanagement: 'Turn On Cache Management'
         logging_header: 'Logging'
         verbose: 'Verbose Logging'
+      validation:
+        default_store_no_cinder: 'Can not set default store to Cinder cinder being active.'
```
vuntz: s/Can not/Cannot/ and hrm, parse error for the second part of the sentence ;-)

crowbar_framework/app/models/glance_service.rb
```diff
@@ -77,6 +77,15 @@ def create_proposal
   def validate_proposal_after_save proposal
     validate_one_for_role proposal, "glance-server"
 
+    if proposal["attributes"]["glance"]["default_store"] == "cinder"
+      cinder_proposal = Proposal.where(barclamp: "cinder", name: "default").first
+      if cinder_proposal.nil?
```
vuntz: The two lines should be:
`if Proposal.where(barclamp: "cinder").empty?`
as we don't know for sure that the cinder proposal will be named `default`.

toabctl:
+1

rsalevsky:
+1

PR #95, opened 2015-11-11
Update cinder.conf config template with Liberty

mjura:
Updating cinder.conf template config with Liberty

PR #110, opened 2015-11-15
Allow enablement of glance image backend optimizations (bsc#945043)

dirkmueller:
By storing images in cinder backend, an efficient COW initialisation of volumes can be used, which saves both time and disk usage.

### DIFF
--- a/chef/cookbooks/cinder/templates/default/cinder.conf.erb
+++ b/chef/cookbooks/cinder/templates/default/cinder.conf.erb
@@ -512,6 +512,7 @@ glance_api_servers=<%= @glance_server_protocol %>://<%= @glance_server_host %>:<
 
 # Version of the glance API to use (integer value)
 #glance_api_version=1
+glance_api_version=2
 
 # Number retries when downloading an image from glance
 # (integer value)
--- a/chef/cookbooks/cinder/templates/default/cinder.conf.erb
+++ b/chef/cookbooks/cinder/templates/default/cinder.conf.erb
@@ -741,6 +741,10 @@ nova_api_insecure=<%= @nova_api_insecure %>
 # the direct_url.  Currently supported schemes: [file]. (list
 # value)
 #allowed_direct_url_schemes=
+<% if @show_storage_location -%>
+### Enable file and rbd if available
+allowed_direct_url_schemes=cinder
+<% end -%>
 
 
 #
@@ -1015,6 +1019,11 @@ use_syslog=<%= node[:cinder][:use_syslog] ? "True" : "False" %>
 
 <% if @use_multi_backend -%>
 [<%= backend_id %>]
+<% end -%>
+
+<% if @show_storage_location -%>
+image_upload_use_cinder_backend = True
+image_upload_use_internal_tenant = False
 <% end -%>
 
 #
--- a/chef/data_bags/crowbar/template-glance.json
+++ b/chef/data_bags/crowbar/template-glance.json
@@ -32,6 +32,7 @@
       },
       "enable_caching": false,
       "use_cachemanagement": false,
+      "show_storage_location": false,
       "image_cache_datadir": "/var/lib/glance/image-cache",
       "image_cache_stall_timeout": 86400,
       "image_cache_max_size": 10737418240,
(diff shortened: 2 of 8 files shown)


chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -740,7 +740,10 @@ nova_api_insecure=<%= @nova_api_insecure %>
 # the direct_url.  Currently supported schemes: [file]. (list
 # value)
 #allowed_direct_url_schemes=
-
```
vuntz: I don't think it matters as we haven't updated the cinder template yet, but let's keep that empty line.

chef/data_bags/crowbar/template-glance.json
```diff
@@ -72,7 +73,7 @@
     "glance": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 23,
+      "schema-revision": 24,
```
vuntz: Hrm. I think I did a mistake and forgot to bump to 30 earlier on. At least 22 and 23 should be moved to the 30s. Feel free to fix that in your PR or I'll do a followup PR to fix the migrations to avoid issues in tex.

crowbar_framework/config/locales/glance/en.yml
```diff
@@ -23,6 +23,7 @@ en:
         rabbitmq_instance: 'RabbitMQ'
         keystone_instance: 'Keystone Instance'
         store_header: 'Image Storage'
+        show_storage_location: 'Show Backend Store Location'
```
vuntz: The string doesn't help people understand what this is about. Unfortunately, I can only think of a long sentence to explain that :/

chef/data_bags/crowbar/template-glance.json
vuntz: Oh, actually, we did bump to 20. Ignore that comment.

mjura:
Depends on https://github.com/crowbar/crowbar-openstack/pull/95

crowbar_framework/config/locales/glance/en.yml
dirkmueller: We can use "Pass through Backend Store Location"

dirkmueller:
@mjura why does it depend on #95?

vuntz:
+1 on this, maybe just changing the string.

crowbar_framework/config/locales/glance/en.yml
vuntz: It might be slightly better, yes. Still not really clear, but I'm afraid it's not going to be clear with a short description anyway.

chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -740,6 +740,10 @@ nova_api_insecure=<%= @nova_api_insecure %>
 # the direct_url.  Currently supported schemes: [file]. (list
 # value)
 #allowed_direct_url_schemes=
+<% if @show_storage_location -%>
+### Enable file and rbd if available
+allowed_direct_url_schemes=cinder
```
vuntz: Hrm, actually, shouldn't that be rbd?

dirkmueller: I think enabling the direct url scheme for cinder should be the way to move forward. we can also enable rbd, but it has the constraint that we need to setup a rbd client on the cinder node first, which is not what this patch is doing. Hence the fixme.

dirkmueller: There is btw significant work going forward to make the cinder backend in glance a fully featured one, which means we can get rid of all the plethora of custom glance storage backends and just put that problem to cinder.

chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -511,7 +511,7 @@ my_ip=<%= node[:cinder][:my_ip] %>
 glance_api_servers=<%= @glance_server_protocol %>://<%= @glance_server_host %>:<%= @glance_server_port %>
 
 # Version of the glance API to use (integer value)
-#glance_api_version=1
+glance_api_version=2
```
vuntz: Can you keep the default value/comment like we usually do?

vuntz:
+1 (with a nitpick)

vuntz:
+1

toabctl:
+1

PR #111, opened 2015-11-16
horizon: Install Manila UI plugin when needed

toabctl:
If there is a manila-server role, install the UI plugin package
on the dashboard node to be able to handle shares via the UI.

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -59,6 +59,26 @@ else
   end
 end
 
+# install horizon manila plugin if needed
+case node[:platform_family]
+when "suse"
+  manila_ui_pkgname = "openstack-horizon-plugin-manila-ui"
+when "rhel"
+  manila_ui_pkgname = "openstack-manila-ui"
+else
+  manila_ui_pkgname = nil
+end
+
+unless manila_ui_pkgname.nil?
+  manila_servers = search(:node, "roles:manila-server") || []
+  unless manila_servers.empty?
+    package manila_ui_pkgname do
+      action :install
+      notifies :reload, resources(service: "apache2")
+    end
+  end
+end
+
 if node[:platform_family] == "suse"
   # Get rid of unwanted vhost config files:
   ["#{node[:apache][:dir]}/vhosts.d/default-redirect.conf",


chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -59,6 +59,31 @@
...
+  manila_ui_pkgname = "openstack-horizon-plugin-manila-ui"
+when "rhel"
+  manila_ui_pkgname = "openstack-manila-ui"
+else
+  manila_ui_pkgname = nil
+end
+
+unless manila_ui_pkgname.nil?
+  manila_servers = search(:node, "roles:manila-server") || []
+  if manila_servers.length > 0
```
tboerger: I think it makes more sense to just change the action depending on `manila_servers.empty?`

tboerger: Beside that you can do it more chef-like with `not_if { manila_ui_pkgname.nil? }` instead of wrapping it within an if block.

toabctl: > Beside that you can do it more chef-like with not_if { manila_ui_pkgname.nil? } instead of wrapping it within an if block.

Does that work? Because I have then 
```
package nil do
not_if { manila_ui_pkgname.nil? }
end
```

toabctl: > I think it makes more sense to just change the action depending on manila_servers.empty?

Done in https://github.com/toabctl/crowbar-openstack/commit/444dcda3f295f912bbe0b588269d0d03b8a30364

tboerger: > Does that work? Because I have then

I'm not totally sure if that works :)

tboerger: I thought more about something like this

```ruby
unless manila_ui_pkgname.nil?
  manila_servers = search(:node, "roles:manila-server") || []

  package manila_ui_pkgname do
    action if manila_servers.empty?
      :remove
    else
      :install
    end
  end
end
```

BTW, you don't need to quote the package names :)

toabctl: Ah. Sure. Done in https://github.com/toabctl/crowbar-openstack/commit/14a39090e118658460401e45f2bb553c1b3dc7d0 . Thanks!

mjura:
+1

rhafer:
+1

tboerger:
:+1:

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -59,6 +59,28 @@
...
 
+# install horizon manila plugin if needed
+case node[:platform_family]
+when "suse"
+  manila_ui_pkgname = "openstack-horizon-plugin-manila-ui"
+when "rhel"
+  manila_ui_pkgname = "openstack-manila-ui"
+else
+  manila_ui_pkgname = nil
+end
```
dirkmueller: imho this should be in attributes/

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -59,6 +59,28 @@
...
+
+unless manila_ui_pkgname.nil?
+  manila_servers = search(:node, "roles:manila-server") || []
+  package manila_ui_pkgname do
+    if manila_servers.empty?
+      action :remove
```
dirkmueller: I'm not sure that this works.. I think I tried that once and chef was screwing up the evaluation in interesting ways...

IMHO we should just always :install it. in any case the if condition would need to check whether the endpoint is in the keystone catalog, not if the manila-server role is deployed

vuntz: Isn't horizon hiding things if there's no endpoint for manila anyway? If yes, then we could just always install the package...

chef/cookbooks/horizon/recipes/server.rb
toabctl: Hm. I don't understand why that would improve anything. The ```node[:platform_family]``` check needs to be done at all runs. No?

chef/cookbooks/horizon/recipes/server.rb
toabctl: > I'm not sure that this works.. I think I tried that once and chef was screwing up the evaluation in interesting ways...

Fixed in https://github.com/crowbar/crowbar-openstack/commit/0f648e52df95d581fee231f261482a1672aeae6b

toabctl: > Isn't horizon hiding things if there's no endpoint for manila anyway? If yes, then we could just always install the package...

Are we removing endpoints if we remove a role? But removing packages seems to be broken anyway (see bsc#955639). I already updated the code to do ```:nothing```

toabctl:
@rhafer @tboerger @mjura @dirkmueller other comments or plus ones?

chef/cookbooks/horizon/recipes/server.rb
vuntz: I think the background is that we usually put distro-specific package names as default attributes.

Btw, there's no debian/ubuntu package for this?

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -59,6 +59,31 @@
...
+when "rhel"
+  manila_ui_pkgname = "openstack-manila-ui"
+else
+  manila_ui_pkgname = nil
+end
+
+unless manila_ui_pkgname.nil?
+  manila_servers = search(:node, "roles:manila-server") || []
+  if manila_servers.empty?
+    # FIXME (toabctl): use :remove here (but that's not working. bsc#955639)
```
vuntz: I would actually kill that comment: if the user manually installs the package for some reason, we don't want to uninstall it.

dirkmueller: Agreed, lets just install it when there is a manila-server role active.

chef/cookbooks/horizon/recipes/server.rb
dirkmueller: Its just more chef-y.

chef/cookbooks/horizon/recipes/server.rb
toabctl: @vuntz done in https://github.com/crowbar/crowbar-openstack/commit/cf8a1965f2aead56984760a6e5b00ed486d1c3c5

dirkmueller:
+1

chef/cookbooks/horizon/recipes/server.rb
toabctl: > Btw, there's no debian/ubuntu package for this?

@vuntz no.

PR #112, opened 2015-11-16
neutron: Fix metadata agent when neutron is using SSL

vuntz:
The metadata agent (neutron-metadata-agent) is receiving requests from
metadata proxies in namespaces (neutron-ns-metadata-proxy) through a
UNIX socket. The requests there are always plain HTTP and since it's not
going through the network, it's fine.

However, if [DEFAULT] use_ssl in neutron.conf is set to true, then
neutron-metadata-agent inherits this and expects HTTPS on this UNIX
socket too. So always override this setting.


rhafer:
+1

rsalevsky:
+1

mjura:
+1

toabctl:
@mjura is this already in https://github.com/crowbar/crowbar-openstack/pull/97 ? So we don't lose this change when we update the template?

mjura:
Depends on https://github.com/crowbar/crowbar-openstack/pull/97

PR #115, opened 2015-11-18
Use internalURL for communication with OpenStack

dirkmueller:
In case the public network is down, debugging is rather painful.
I think if you're on the node that has the .openrc then the
internalURL should always work, so use that instead.


vuntz:
Code is fine, but I'm kind of mixed as I'm thinking some people might just copy the file out of their deployment. But I could be wrong.

So +0.5.

mjura:
This works for me, nothing against merging it +1

PR #116, opened 2015-11-18
Switch to OpenStack Identity API version 3 by default

dirkmueller:
New deployments should be using v3 as it provides more
features onwards, and starting with Liberty v3-only should
be possible (however this change still enables both versions).


vuntz:
I think we need to try it for a couple of weeks to decide if it's safe, so I'm fine merging this, and we can revert if we see any issue. So +1

rhafer:
I agree. +1

aplanas:
Maybe is too late, but +1

rhafer:
CI is failing because there's a test using the keystone command (trying the v2 API against the v3 endpoint)

sayalilunkad:
+1

PR #117, opened 2015-11-18
nova: don't install python-suds for vmware

toabctl:
python-nova already depends on python-oslo.vmware which depends
on python-suds-jirko which is the successor of python-suds.


vuntz:
+1

vmoravec:
+1 Tested on qa2 hardware

mjura:
Right now I'm testing vmware integration, I will let you know, but looks GOOD :)

PR #120, opened 2015-11-19
Fix for rename nova roles migration script

mjura:
This change is fixing migration script which was merged in https://github.com/crowbar/crowbar-openstack/pull/100

It was conflicting with `chef/data_bags/crowbar/migrate/nova/035_add_vmware_port_attribute.rb`


vuntz:
+1

rsalevsky:
+1

PR #121, opened 2015-11-20
Adding hyperv driver to the front end

ovisan:
This is needed for the hyperv agent on windows to run correctly and
allocate ip's
This PR is a dependency of https://github.com/crowbar/crowbar-hyperv/pull/10

### DIFF
--- a/crowbar_framework/app/models/neutron_service.rb
+++ b/crowbar_framework/app/models/neutron_service.rb
@@ -37,7 +37,7 @@ class NeutronService < PacemakerServiceObject
   end
 
   def self.networking_ml2_mechanism_drivers_valid
-    ["linuxbridge", "openvswitch", "cisco_nexus"]
+    ["linuxbridge", "openvswitch", "cisco_nexus", "hyperv"]
   end
 
   class << self


MaximilianMeister:
if it is enough to add it here :+1: 

so you checked the config file after chef client and the UI?

ovisan:
After this change, i have selected the `hyperv` as well as `openvswitch` and applied the barclamp neutron again. Then checked the `config file /etc/neutron/plugins/ml2/ml2_conf.ini` and hyperv is there
`mechanism_drivers = openvswitch,hyperv,l2population`
Also launched an instance for hyperv and the instance got an IP, so i would say it's tested.

tboerger:
I have no idea if that is enough. That should be verified by @vuntz or somebody else who knows enough about that part.

mjura:
As I know hyperv should works also with `openvswitch` and `linuxbridge` as well. I'm sorry but it is wrong solution for this problem.

vuntz:
This is not needed, we always put hyperv in `/etc/neutron/plugins/ml2/ml2_conf.ini`.

vuntz:
Ah well, it's actually temporarily disabled, see https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/neutron/recipes/server.rb#L160 So the right fix is to uncomment things there, imho.

PR #122, opened 2015-11-20
Avoid HoundCI bitching about has_key?

dirkmueller:
According to our now master of style, Hash#has_key? is deprecated in
favor of Hash#key?. (https://github.com/bbatsov/ruby-style-guide#hash-key)


tboerger:
Thanks! :+1:

rsalevsky:
:+1:

PR #125, opened 2015-11-23
Add DVR option in neutron proposal UI

mjura:


vuntz:
We need some javascript magic to forcefully disable the setting in case where DVR is not supported (see the validation in the model for the cases where we don't support it).

Alternatively: we validate that these cases are supported now (I think it's mostly vlan, which should work now but someone needs to check that)

mjura:
I think that we are talking about this 
https://specs.openstack.org/openstack/neutron-specs/specs/kilo/neutron-ovs-dvr-vlan.html

mjura:
Ok, here I have found
http://docs.openstack.org/networking-guide/scenario_dvr_ovs.html
```
In the Juno (initial) release, DVR supports VXLAN and GRE project networks. In the Kilo release, support was added for DVR with VLAN project networks. All releases support flat and VLAN external networks.
```
So it means that only `vmware` plugin is not supported. Am I right @vuntz ?

dirkmueller:
I believe linuxbridge is also not supported with DVR (yet)

vuntz:
> I believe linuxbridge is also not supported with DVR (yet)

"DVR now supports VLANs in addition to VXLAN/GRE" (kilo release notes)

vuntz:
> So it means that only vmware plugin is not supported. Am I right @vuntz ?

I would rather say: only ml2 with linuxbridge or ovs is supported. If we add support for some other plugin, then I have no idea if this will work.

But adding support for dvr with vlan is possibly a bigger task; you might want to just do the javascript magic for this PR...

mjura:
I have to discuss latest changes with @tboerger

MaximilianMeister:
can't say much about what is supported, but I tested the javascript magic and it works :+1:

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
```diff
@@ -236,6 +236,7 @@ function networking_plugin_check() {
     $('#ml2_type_drivers_container').show();
     $('#ml2_type_drivers_default_provider_network_container').show();
     $('#ml2_type_drivers_default_tenant_network_container').show();
+    $('#dvr_container').show();
```
vuntz: Actually, I would only show it from `ml2_type_drivers_check` if gre or vxlan is in the list of drivers.

mjura: Ok, it was moved to `ml2_type_drivers_check`

mjura:
Rebased

crowbar_framework/config/locales/neutron/en.yml
```diff
@@ -31,6 +31,7 @@ en:
         rabbitmq_instance: 'RabbitMQ Instance'
         keystone_instance: 'Keystone Instance'
         dhcp_domain: 'DHCP Domain'
+        use_dvr: 'Enable Distributed Virtual Router'
```
dirkmueller: I think "Use Distributed Virtual Router (setup/layout)" would be more descriptive

mjura: Thank you for a tip, it was adopted

dirkmueller:
that was not a verbatim-to-copy proposal.. you should have chosen between setup or layout, not actually copy (setup/layout) :-)

mjura:
He he, it was already in quotes ready to use :) no worries, I was checking you.

Fixed and rebased

dirkmueller:
+1

mjura:
I did final tests. So DVR works fine with openvswitch ML2 driver and VLAN/GRE/VXLAN, but doesn't work with Llinux Bridge and VMware NSX.

crowbar_framework/app/assets/javascripts/barclamps/neutron/application.js
```diff
@@ -291,6 +293,8 @@ var ovsSelection = null
 
 function ml2_mechanism_drivers_check() {
   var values = $('#ml2_mechanism_drivers').val() || [];
+  $('#dvr_container').show();
```
dirkmueller: this causes flicker I guess. better move it to line 335

mjura: Ok, let it be. I haven't seen any side effects of this. It was moved to 335

mjura:
I've split up changes to different commits, should cover all usage scenarios.

dirkmueller:
:+1:

aplanas:
+1

PR #126, opened 2015-11-24
Nova db settings

toabctl:


vuntz:
+1

chef/data_bags/crowbar/migrate/nova/037_add_sqlalchemy_pool_values.rb
```diff
@@ -0,0 +1,31 @@
+def upgrade(ta, td, a, d)
+  unless a["db"].key? "max_pool_size"
+    a["db"]["max_pool_size"] = ta["db"]["max_pool_size"]
```
dirkmueller: I wonder if we need this code when its nil in the json...

dirkmueller:
+0.1. I'm a bit unsure about the migration (see inline comment) but the rest looks good.

toabctl:
@vuntz still +1 ?

chef/data_bags/crowbar/migrate/nova/037_add_sqlalchemy_pool_values.rb
toabctl: It worked, but I removed the whole upgrade code because.

vuntz:
yes, still +1

dirkmueller:
+1

chef/data_bags/crowbar/migrate/nova/037_add_sqlalchemy_pool_values.rb
```diff
@@ -0,0 +1,21 @@
+def upgrade(ta, td, a, d)
+  # NOTE(toabctl): we do nothing here because all the added keys
```
tboerger: I think we don't need this kind of comment...

aplanas: I find this useful!

PR #128, opened 2015-11-24
Fix interface mappings when floating is not on a VLAN

rhafer:
Partial fix for: https://bugzilla.suse.com/show_bug.cgi?id=955811

Partial because:
* It's still broken for openvswitch setups (the bug was about linuxbridge), I intend to fix that in a separate pull request


chef/cookbooks/neutron/libraries/helpers.rb
```diff
@@ -0,0 +1,13 @@
+module NeutronHelper
+  def self.get_floating_physnet(node)
+    # When nova_fixed and nova_floating are on the same conduit and nova_floating
+    # should not use a separate VLAN, we need them to use the same "physnet" setting 
+    # in neutron
+    if (node[:network][:networks][:nova_fixed][:conduit] ==
+        node[:network][:networks][:nova_floating][:conduit]) &&
```
vuntz: Why not resolve the conduits?

rhafer: Do you know of a good way how to do that reliably?  Remember the conduits mappings might be different for the controller and the network node.

And by the time chef-client executes this on the controller I think it might not even be possible to expand them reliably for the network node. Because the conduit mapping depends on the actual roles deployed on a node (I admit I haven't really checked that yet).

vuntz: I can't think properly right now, so I'll just ask :-) Isn't this bit of the config specific to each node anyway? Or is it essential that this is in sync on all nodes?

rhafer:
(Trying to move this discussion from the hidden outdated diff to make it more visible)

> I can't think properly right now, so I'll just ask :-) Isn't this bit of the config specific to each node anyway? Or is it essential that this is in sync on all nodes?

@vuntz Yes and no. The ```interface_mappings``` (or ```bridge_mappings``` or ovs) setting could in theory be different from node to node (e.g. compute nodes can use something different than network node). But for creating the different neutron network we need to know the "physnet" they're on. And that differs when nova_fixed and nova_floating endup on the same "physnet" (i.e. are using the same physical interface). So neutron (the server) somehow needs to know if that is the case.
This also means that it is possible to create a network config in crowbar that neutron can not even handle. (E.g. if the compute nodes use the same physical interface for both networks but the network node has them separated)

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -151,7 +151,7 @@
 #                so the order is important
 tenant_network_types = [[node[:neutron][:ml2_type_drivers_default_tenant_network]] + node[:neutron][:ml2_type_drivers]].flatten.uniq
 
-external_networks = ["floating"]
+external_networks = [ Neutron.get_floating_physnet(node) ]
```
bhdn: Shouldn't it be NeutronHelper here?

rhafer: Right. But I am currently reworking the whole patch as this one still had a few other issues. Not quite ready with that yet, though.

rhafer:
I completely reworked this. It now tries resolve the conduits to the network interface (on the network node). Additionally I added code to check if the external networks endup on separate interfaces.

The code is still not working for openvswitch thought. Working on that next.

rhafer:
Note: I'd prefer to fix the openvswitch side of the bug in another pull request. This make backporting this patch to Tex a bit easier. And while the issue was a regression (broke with an update) for linuxbridge it was previously never possible to create a working setup for openvswitch (+vlan) with fixed and floating (use_vlan: false) on the same physical interface. 

Still we should address the ovs part in tex as well. But the ovs fix will depend on some of the recent changes we did in the rails part for which the backport (https://github.com/crowbar/barclamp-neutron/pull/260) still needs some work.

rhafer:
rebased (to get the package for ci rebuilt with the latest fixes)

vuntz:
+1

rhafer:
Rebased.

vuntz:
Still +1

aplanas:
I feel the code hard to understand (implicit `return` and explicit ones in the same method, using symbols and strings for the same references), but this points to my lack of reading skills in Ruby, also I think that this code is important, so:

+1

PR #129, opened 2015-11-25
Specify hypervisor_type=kvm for the tempest image

dirkmueller:
This will break it for Xen though, not sure how
to resolve that :/


vuntz:
What about uploading it a second time with hypervisor_type=xen?

dirkmueller:
you can add only one image id to the tempest.conf :)

toabctl:
> you can add only one image id to the tempest.conf :)

But you can have multiple tempest.conf files. and select the .conf you want to use.

vuntz:
I guess I'm a bit confused: are we concerned about xen-only setups (in which case we can detect this and set the right hypervisor type when uploading the image; this PR would need reworking) or mixed kvm-xen setups?

dirkmueller:
@vuntz well.. both I guess.

PR #131, opened 2015-11-27
Make comments more verbose for VMware insecure option

mjura:


dirkmueller:
would be good to do the same for glance

mjura:
@dirkmueller : I would like to keep as it is, because option for glance seems to have different description

vuntz:
@mjura can you clarify why it is different for glance? I want to give you a +1 but am missing this piece of information :-)

mjura:
@vuntz: Because for Glance this option means to enable HTTP or HTTPS connection to vCenter

/usr/lib/python2.7/site-packages/glance_store/_drivers/vmware_datastore.py
```
    def _get_http_conn_class(self):
        if self.api_insecure:
            return http_client.HTTPConnection
        return http_client.HTTPSConnection
```

mjura:
@dirkmueller, @vuntz : Updated also comment for Glance, it seams that upstream explanation is not so accurate.

crowbar_framework/config/locales/glance/en.yml
```diff
@@ -42,7 +42,7 @@ en:
           password: 'vCenter Password'
           datastore: 'Datastore Name'
           datacenter_path: 'Datacenter Path'
-          api_insecure: 'Insecure SSL requests to vCenter'
+          api_insecure: 'Insecure HTTP connections to vCenter'
```
dirkmueller: HTTP is by definition insecure. How about 

```
Use HTTP for connections to vCenter
```

Since we do not have the option for HTTP access in cinder and nova, I think we should just remove this option for glance alltogether. Sorry for bringing this up so late, I should have researched this in the previous PR already.

mjura: No worries, I wasn't sure too. Upstream documentation description wasn't accurate for this option. I will remove this option.

mjura: @dirkmueller: My proposal would be to keep this option in `RAW` mode, it can be used some day. https://github.com/crowbar/crowbar-openstack/pull/134/

dirkmueller:
+1

jsuchome:
+1

crowbar_framework/config/locales/cinder/en.yml
```diff
@@ -117,7 +117,7 @@ en:
               password: 'vCenter Password'
               volume_folder: 'Folder for Volumes'
               ca_file: 'CA file for verifying the vCenter certificate'
-              insecure: 'Do not verify vCenter certificate'
+              insecure: 'vCenter SSL Certificate is insecure (for instance, self-signed)'
```
tboerger: This is the form label, IMHO far too long. This should be an alert below the field.

PR #134, opened 2015-11-30
Dropping VMware API insecure option from Glance

mjura:
Because of discussion in https://github.com/crowbar/crowbar-openstack/pull/131#discussion-diff-46124162, we decided to drop it

### DIFF
--- a/chef/data_bags/crowbar/template-glance.json
+++ b/chef/data_bags/crowbar/template-glance.json
@@ -59,8 +59,7 @@
         "password": "",
         "datastore": "",
         "datacenter_path": "ha-datacenter",
-        "store_image_dir": "/openstack_glance",
-        "api_insecure": false
+        "store_image_dir": "/openstack_glance"
       },
       "sql_idle_timeout": 3600,
       "keystone_instance": "none",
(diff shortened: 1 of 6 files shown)


dirkmueller:
IMHO even the back implementation could be removed.. or how would you use vcenter integration without ssl on nova or cinder?

mjura:
@dirkmueller : Glance integration can still be used independently :) anyway, I will drop this attribute completely

mjura:
Ok, `api_secure` id dropped

mjura:
It was rebased

mjura:
Rebased

chef/data_bags/crowbar/template-glance.json
```diff
@@ -73,7 +72,7 @@
     "glance": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 24,
+      "schema-revision": 23,
```
dirkmueller: thats a downgrade

mjura: Ok, I fixed it

dirkmueller:
:+1:

PR #136, opened 2015-12-02
Updates mtu configuration to use neutron config options.

dguitarbite:
Removes custom dnsmasq-neutron.conf.erb and sets mtu options via.
neutron ml2_conf.ini.erb.

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents.rb
@@ -101,16 +101,9 @@ template "/etc/neutron/metering_agent.ini" do
   )
 end
 
-ml2_drivers = node[:neutron][:ml2_type_drivers]
-template "/etc/neutron/dnsmasq-neutron.conf" do
-  source "dnsmasq-neutron.conf.erb"
-  owner "root"
-  group "root"
-  mode "0644"
-  variables(
-    # nil means to leave the default MTU
-    force_mtu: ml2_drivers.include?("gre") || ml2_drivers.include?("vxlan") ? 1400 : nil
-  )
+# Delete pre-existing configuration file.
+file "/etc/neutron/dnsmasq-neutron.conf" do
+  action :delete
 end
 
 dns_list = node[:dns][:forwarders].join(",")
@@ -174,7 +167,6 @@ service node[:neutron][:platform][:dhcp_agent_name] do
   action [:enable, :start]
   subscribes :restart, resources("template[/etc/neutron/neutron.conf]")
   subscribes :restart, resources("template[/etc/neutron/dhcp_agent.ini]")
-  subscribes :restart, resources("template[/etc/neutron/dnsmasq-neutron.conf]")
   provider Chef::Provider::CrowbarPacemakerService if ha_enabled
 end
 
--- a/chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
@@ -28,6 +28,30 @@ mechanism_drivers = <%= @ml2_mechanism_drivers.join(",") %>
 # to be loaded from the neutron.ml2.extension_drivers namespace.
 # extension_drivers =
 # Example: extension_drivers = anewextensiondriver
+#
+# =========== items for MTU selection and advertisement =============
+# (IntOpt) Path MTU.  The maximum permissible size of an unfragmented
+# packet travelling from and to addresses where encapsulated Neutron
+# traffic is sent.  Drivers calculate maximum viable MTU for
+# validating tenant requests based on this value (typically,
+# path_mtu - max encap header size).  If <=0, the path MTU is
+# indeterminate and no calculation takes place.
+# path_mtu = 0
+path_mtu = <%= @path_mtu %>
+
+# (IntOpt) Segment MTU.  The maximum permissible size of an
+# unfragmented packet travelling a L2 network segment.  If <=0,
+# the segment MTU is indeterminate and no calculation takes place.
+# segment_mtu = 0
+
+# (ListOpt) Physical network MTUs.  List of mappings of physical
+# network to MTU value.  The format of the mapping is
+# <physnet>:<mtu val>.  This mapping allows specifying a
+# physical network MTU value that differs from the default
+# segment_mtu value.
+# physical_network_mtus =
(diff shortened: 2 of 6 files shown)


chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -101,18 +101,6 @@
...
-  source "dnsmasq-neutron.conf.erb"
-  owner "root"
-  group "root"
-  mode "0644"
-  variables(
-    # nil means to leave the default MTU
-    force_mtu: ml2_drivers.include?("gre") || ml2_drivers.include?("vxlan") ? 1400 : nil
-  )
-end
-
```
vuntz: Can you add code to delete the old file in case it's there? Would be something like:
```
file "/etc/..." do
  action :delete
end
```

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -161,6 +161,13 @@ allow_overlapping_ips = False
 
 # Maximum number of routes per router
 # max_routes = 30
+#
+# =========== items for MTU selection and advertisement =============
+# Advertise MTU.  If True, effort is made to advertise MTU
+# settings to VMs via network methods (ie. DHCP and RA MTU options)
+# when the network's preferred MTU is known.
+advertise_mtu = True
```
vuntz: Should this be false when we don't set mtu? Or it doesn't matter?

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
```diff
@@ -28,6 +28,29 @@ mechanism_drivers = <%= @ml2_mechanism_drivers.join(",") %>
...
 # Example: extension_drivers = anewextensiondriver
+#
+# =========== items for MTU selection and advertisement =============
+# (IntOpt) Path MTU.  The maximum permissible size of an unfragmented
+# packet travelling from and to addresses where encapsulated Neutron
+# traffic is sent.  Drivers calculate maximum viable MTU for
+# validating tenant requests based on this value (typically,
+# path_mtu - max encap header size).  If <=0, the path MTU is
+# indeterminate and no calculation takes place.
+path_mtu = <%= @path_mtu %>
```
mjura: Please keep default value commented out, when you change something, look on another options modified by chef

chef/cookbooks/neutron/templates/default/neutron.conf.erb
mjura: Please keep default value

mjura:
Config will be provided in PR https://github.com/crowbar/crowbar-openstack/pull/97, please depend on it

chef/cookbooks/neutron/templates/default/neutron.conf.erb
dguitarbite: It we dont set MTU it defaults to ``0``. This is handled internally by Neutron. Keeping this option True just makes Neutron check for MTU values and should not be a problem. ``True`` with option ``0`` is equivalent to ``False`` with commented values.

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
dguitarbite: As mentioned above and also as mentioned in the description, there is no need for keeping this value commented out as setting it to zero is equivalent.
``If <=0, the path MTU is indeterminate and no calculation takes place.``

dguitarbite:
@mjura How do I depend on your PR?

chef/cookbooks/neutron/templates/default/neutron.conf.erb
mjura: As I said please keep default value commented out in conf template, this is what we agreed in the past. This is for better understanding configuration and tracking issues by engineers.

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
mjura: As I said please keep default value commented out in conf template, this is what we agreed in the past. This is for better understanding configuration and tracking issues by engineers.
```
# path_mtu = 0
```

chef/cookbooks/neutron/templates/default/neutron.conf.erb
mjura:
```
# advertise_mtu = False
```

mjura:
> @mjura How do I depend on your PR?

Let's say that it is conflicting, I would like to have PR #97 merged first.

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
rhafer: What @mjura means here is, that we ususally do it like this:

```
# indeterminate and no calculation takes place.
# path_mtu = 0
path_mtu = <%= @path_mtu %>
```

dguitarbite: Oh, Ill update this. Thanks for clarifying it out for me.

dguitarbite:
@mjura Thanks, I will wait for your PR.

chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -101,16 +101,9 @@
...
-  owner "root"
-  group "root"
-  mode "0644"
-  variables(
-    # nil means to leave the default MTU
-    force_mtu: ml2_drivers.include?("gre") || ml2_drivers.include?("vxlan") ? 1400 : nil
-  )
+# Delete pre-existing configuration file.
+file "/etc/neutron/dnsmasq-neutron.conf" do
+      action :delete
```
rossella: this file needs to be deleted only if we are upgrading from an older version to the current one. Not sure how we usually handle this case.
I think you should check if the file exists before deleting it .

dirkmueller: that should be okay I think, chef doesn't do anything if the file isn't there.

vuntz:
@dguitarbite can you rebase now that the neutron config rebase was merged?

dguitarbite:
Yes, I'll rebase the patch.

rossella:
I rebased the patch and had to create a new pull request since I don't have write rights on Pranav's fork...the new PR is #185 , closing this one

PR #139, opened 2015-12-03
Use pacemaker transactions (glance, heat, keystone)

abel-navarro:


chef/cookbooks/glance/recipes/ha.rb
```diff
@@ -54,31 +55,39 @@
...
-  ordering ["postgresql", "rabbitmq", "cl-keystone", "cl-#{group_name}"]
+crowbar_pacemaker_order_only_existing "o-#{clone_name}" do
+  ordering ["postgresql", "rabbitmq", "cl-keystone", clone_name]
   score "Optional"
   action :create
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
+pacemaker_transaction "#{service_name} clone" do
+  cib_objects
```
vuntz: Should be: `cib_objects cib_obects`

You might want to rename your cib_objects to transaction_objects to avoid confusion and have `cib_objects transaction_objects`.

chef/cookbooks/glance/recipes/ha.rb
```diff
@@ -54,31 +55,39 @@
...
+end
```
vuntz: @aspiers stupid question, but... where do we start the clone with the new way?

chef/cookbooks/glance/recipes/ha.rb
aspiers: Yep, that's exactly what I said during the team workshop ;-)

chef/cookbooks/glance/recipes/ha.rb
aspiers: It gets started when the transaction gets committed, because `target-role` defaults to `Started`.

vuntz: Oh, of course. Might make sense to somehow document that (but not sure where); I got too used to our resources defaulting `target-role` to `Stopped`.

aspiers:
-1 until the above issue is fixed.

@dirkmueller This is a good example of why we need to be able to tell the gate to test with a different scenario (in this case with HA enabled), and specifying a magic line in the PR description would be the easiest way to do that.

chef/cookbooks/glance/recipes/ha.rb
```diff
@@ -54,31 +55,39 @@
...
+cib_objects << "pacemaker_clone[#{clone_name}]"
 
-crowbar_pacemaker_order_only_existing "o-cl-#{group_name}" do
-  ordering ["postgresql", "rabbitmq", "cl-keystone", "cl-#{group_name}"]
+crowbar_pacemaker_order_only_existing "o-#{clone_name}" do
+  ordering ["postgresql", "rabbitmq", "cl-keystone", clone_name]
   score "Optional"
   action :create
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
```
vuntz: The `pacemaker_transaction` should be run before this, as this references a resource that won't exist.

abel-navarro:
Related PR for Horizon/Apache to use pacemaker transactions: https://github.com/crowbar/crowbar-ha/pull/50

aspiers:
+1 iff this has been successfully tested with `hacloud=1`.

jsuchome:
Codewise this looks good +1

chef/cookbooks/glance/recipes/ha.rb
```diff
@@ -54,28 +55,36 @@
...
   rsc group_name
-  action [:create, :start]
+  action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
+transaction_objects << "pacemaker_clone[#{clone_name}]"
+
+pacemaker_transaction "#{service_name} clone" do
+  cib_objects transaction_objects
+end
```
vuntz: I would change it to:
```
pacemaker_transaction "glance server" do
  cib_objects transaction_objects
  # note that this will also automatically start the resources
  action :commit_new
end
```

chef/cookbooks/heat/recipes/ha.rb
```diff
@@ -48,34 +48,43 @@
...
   rsc group_name
-  action [:create, :start]
+  action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
+transaction_objects << "pacemaker_clone[#{clone_name}]"
+
+pacemaker_transaction "#{service_name} clone" do
+  cib_objects transaction_objects
+end
```
vuntz: Same here.

chef/cookbooks/keystone/recipes/ha.rb
```diff
@@ -55,18 +57,25 @@
...
+  pacemaker_clone clone_name do
     rsc service_name
-    action [:create, :start]
+    action :update
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
+  transaction_objects << "pacemaker_clone[#{clone_name}]"
+
+  pacemaker_transaction "#{service_name} clone" do
+    cib_objects transaction_objects
```
vuntz: And again :-)

chef/cookbooks/glance/recipes/ha.rb
abel-navarro: This was not in the original work by @aspiers (https://github.com/crowbar/crowbar-openstack/commit/2d29fa187881371dbe390d598a6de3499f38dca4), that why I didn't add it.

This part is a bit obscure to me, but if there's agreement it's the right way to do it I'll update it straight away.

aspiers: `commit_new` is the default action so this is not strictly necessary, but it's maybe nice to have for clarity.

@vuntz is right that we only want this happening on the founder.  Well, that has the downside that non-founders proceed quicker and increase the risk of sync drift ... and technically it shouldn't be necessary if we use sync_marks correctly because they would prevent races and non-founders running this would not attempt to create already existing objects.  But given crowbar/crowbar-ha#49 I think we should go with @vuntz's suggestion.

abel-navarro: Updated the PR to include the suggested changes.

chef/cookbooks/glance/recipes/ha.rb
```diff
@@ -54,28 +55,39 @@
...
+clone_name = "cl-#{group_name}"
+pacemaker_clone clone_name do
   rsc group_name
-  action [:create, :start]
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects << "pacemaker_clone[#{clone_name}]"
+
+pacemaker_transaction "#{service_name} clone" do
```
jsuchome: `service_name` is not defined in this file. I assume it should be `service_name = glance`

chef/cookbooks/heat/recipes/ha.rb
```diff
@@ -48,34 +48,46 @@
...
+clone_name = "cl-#{group_name}"
+pacemaker_clone clone_name do
   rsc group_name
-  action [:create, :start]
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects << "pacemaker_clone[#{clone_name}]"
+
+pacemaker_transaction "#{service_name} clone" do
```
jsuchome: `service-name = heat` is missing

abel-navarro:
After a couple of runs, this passed the tempest smoke tests with `hacloud=1`. It was failing previously on some cinder tests and later on neutron tests. Last run was clear.

chef/cookbooks/glance/recipes/ha.rb
```diff
@@ -54,28 +55,39 @@
...
 end
+transaction_objects << "pacemaker_group[#{group_name}]"
 
-pacemaker_clone "cl-#{group_name}" do
+clone_name = "cl-#{group_name}"
+pacemaker_clone clone_name do
   rsc group_name
-  action [:create, :start]
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
aspiers: Rather than constantly specifying this `only_if` test, I wonder if it would be cleaner to simply wrap all of them in an `if` block.  What do you think?

aspiers:
+1 based on comment about successful test run.  Technically we could merge this now, but maybe we should wait until after Xmas considering there's a (very low) risk this could break something.

chef/cookbooks/glance/recipes/ha.rb
```diff
@@ -54,28 +55,39 @@
...
+end
+transaction_objects << "pacemaker_clone[#{clone_name}]"
+
+pacemaker_transaction "#{clone_name} clone" do
```
vuntz: I would change the name of this resource to "glance server" (as it's not just about the clone, but about everything).

chef/cookbooks/heat/recipes/ha.rb
```diff
@@ -48,34 +48,46 @@
...
+clone_name = "cl-#{group_name}"
+pacemaker_clone clone_name do
   rsc group_name
-  action [:create, :start]
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects << "pacemaker_clone[#{clone_name}]"
+
+pacemaker_transaction "#{clone_name} clone" do
```
vuntz: Same here ("heat server')

chef/cookbooks/keystone/recipes/ha.rb
```diff
@@ -55,18 +57,28 @@
...
+  clone_name = "cl-#{service_name}"
+  pacemaker_clone clone_name do
     rsc service_name
-    action [:create, :start]
+    action :update
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+  transaction_objects << "pacemaker_clone[#{clone_name}]"
+
+  pacemaker_transaction "#{clone_name} clone" do
```
vuntz: Same here ("keystone server")

vuntz:
Minor nitpicks. I'm +1 even without them, as this can be changed in a trivial follow-up PR.

aspiers:
+1 for merging now.

PR #140, opened 2015-12-07
heat, horizon, neutron: Symlink the suse templates as opensuse

vuntz:
We want the same set of templates for both SUSE and openSUSE.


dirkmueller:
:+1:

vuntz:
Bah, the symlink doesn't work...

vuntz:
Creating symlinks for the files instead...

rsalevsky:
+1

PR #142, opened 2015-12-08
rabbitmq: Extend epmd.socket unit file to listen on 0.0.0.0 for Leap

vuntz:
Right now, on Leap, epmd.socket makes systemd only listen on 127.0.0.1
which makes rabbitmq-server fail to start. So we extend the socket unit
file to listen on 0.0.0.0.

This is the recommended setup as per
https://bugzilla.suse.com/show_bug.cgi?id=869112

Note that for some reason, it fails when we only listen on the admin IP
address.

See also discussion at
https://bugzilla.redhat.com/show_bug.cgi?id=1104843

### DIFF
--- a/chef/cookbooks/rabbitmq/files/opensuse/epmd.socket-listen.conf
+++ b/chef/cookbooks/rabbitmq/files/opensuse/epmd.socket-listen.conf
@@ -1,0 +1,3 @@
+[Socket]
+ListenStream=
+ListenStream=0.0.0.0:4369
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -60,6 +60,31 @@ bash "enabling rabbit management" do
   notifies :restart, "service[rabbitmq-server]"
 end
 
+if node[:platform] == "opensuse"
+  # On Leap, epmd.socket is enabled by default but only listens on 127.0.0.1,
+  # while we need it to listen on 0.0.0.0.
+
+  directory "/etc/systemd/system/epmd.socket.d" do
+    owner "root"
+    group "root"
+    mode 0755
+    action :create
+  end
+
+  cookbook_file "/etc/systemd/system/epmd.socket.d/listen.conf" do
+    source "epmd.socket-listen.conf"
+    owner "root"
+    group "root"
+    mode 0644
+  end
+
+  bash "reload systemd for epmd.socket extension" do
+    code "systemctl daemon-reload"
+    action :nothing
+    subscribes :run, "cookbook_file[/etc/systemd/system/epmd.socket.d/listen.conf]", :immediate
+  end
+end
+
 service "rabbitmq-server" do
   supports restart: true, start: true, stop: true, status: true
   action [:enable, :start]


mjura:
This will be fixed in rabbitmq-server package, there is problem in rabbitmq-server.service with order to start rabbitmq-server earlier than epmd.socket.

rabbitmq-server.service
```
 [Unit]
 Description=RabbitMQ broker
-After=epmd.socket syslog.target network.target
-Requires=epmd.socket
+After=syslog.target network.target
 
 [Service]
 Type=simple
```

vuntz:
> This will be fixed in rabbitmq-server package, there is problem in rabbitmq-server.service with order to start rabbitmq-server earlier than epmd.socket.

No, it's not enough: I'm already using the fixed rabbitmq-server. The problem is really the Leap default configuration (which is different from what we have in SLES), where epmd.socket is enabled by default.

mjura:
> No, it's not enough: I'm already using the fixed rabbitmq-server. The problem is really the Leap default configuration (which is different from what we have in SLES), where epmd.socket is enabled by default.

Just wondering, if we should disable also epmd.socket service than changing this to listen on all interfaces. More reasonable and secure solution would be use it in this same way like on SLE12

vuntz:
It listens on all interfaces in our SLES12 setup too :-)

mjura:
> It listens on all interfaces in our SLES12 setup too :-)

Yes, epmd listen on all interfaces but it is started by rabbitmq-server. 
/usr/lib/systemd/system/epmd.socket
```
...
[Socket]
ListenStream=127.0.0.1:4369
Accept=false
...
```

Services epmd.service and epmd.socket are disabled. So on opensuse, we can disable only epmd.socket service and let rabbitmq-server to start epmd and we will have mirror solution than on SLES.

vuntz:
> So on opensuse, we can disable only epmd.socket service and let rabbitmq-server to start epmd and we will have mirror solution than on SLES.

Between disabling (and stopping) epmd.socket and patching it, I don't see a big difference: in both cases, we diverge from SLES. What I implemented is however what is recommended and documetned by the openSUSE packager (see the bug, and see the README in the package).

I can change it of course, but I don't see what's the difference then :-)

mjura:
> I can change it of course, but I don't see what's the difference then :-)

Let it be :) 
+1

nkrinner:
+1

dirkmueller:
I still think this should be fixed in packaging, -1

vuntz:
@dirkmueller how? epmd.socket is enabled by default on Leap (see discussion in the bug), that's really the issue (rabbitmq depending on epmd.socket is another issue, but unrelated -- I don't use a package with that bug)

mjura:
> I still think this should be fixed in packaging, -1

Yes, I'm thinking this same. Btw. @dirkmueller do you remember which changes are needed to fix this except changing `rabbitmq-server.service` file

```
 [Unit]
 Description=RabbitMQ broker
-After=epmd.socket syslog.target network.target
-Requires=epmd.socket
+After=syslog.target network.target

 [Service]
 Type=simple
```

I tried to find it in Changes history and I see only this difference

vuntz:
That change to rabbitmq-server.service is needed anyway. But it won't solve the fact that epmd.socket is enabled and started by default on openSUSE, and not on SLES. I don't see a packaging way around this except changing the systemd presets, but again, it got changed on purpose on openSUSE.

mjura:
For me stiil, it would easier to just disable `epmd.socket` instead of changing configuration, but this change works too

+1

vuntz:
Let's close for now, no interest in pushing this further.

PR #143, opened 2015-12-09
swift: Fix crash when applying proposal with a deactivated ceph proposal

vuntz:
It is documented that the proposal status is not something that should
be used to detect if a proposal is deactivated or not. So we need to
check if the role exists for this.


rsalevsky:
+1

aplanas:
+1

PR #144, opened 2015-12-09
Arch fixes

dirkmueller:


crowbar_framework/app/models/nova_service.rb
```diff
@@ -107,12 +107,13 @@ def proposal_dependencies(role)
     answer
   end
 
-  def node_platform_supports_xen(node)
-    node[:platform_family] == "suse"
+  def node_supports_xen(node)
+    return false if node[:platform_family] != "suse"
+    node[:block_device].include?("vda") && node["kernel"]["machine"] =~ /x86_64/
```
vuntz: Isn't the vda test reversed?

Also, why do we need a pattern for the architecture?

dirkmueller: yes, intentionally reversed, since I switched from "unless" to "if" in the calling condition

vuntz: > yes, intentionally reversed, since I switched from "unless" to "if" in the calling condition

I still think it's reversed :-)

Before, we were not assigning a node to xen if it had vda disk; now we're requiring the node to have a vda disk.

toabctl: @dirkmueller it's reversed, as @vuntz mentioned ...

toabctl:
+1

aplanas:
+1

vuntz:
+1

mjura:
+1

PR #145, opened 2015-12-09
nova: Install open-iscsi before starting its daemon

vuntz:
We explicitly start the iscsid daemon to help nova-compute, but we
didn't make sure it was installed, and it's not a dependency of nova.

It's generally okay because it's a dependency of libvirt, but on docker
nodes, the package is actually not installed by default (at least on
Leap).


dirkmueller:
:+1:

PR #146, opened 2015-12-09
swift: Fix ring builder parser to be compatible with both Juno & Liberty

vuntz:


vuntz:
For the record: I can't fix the remaining hound issues as these are long lines that are there for a reason (existing output of a command).

vuntz:
For some reason, hound is not reporting here...

aplanas:
No fully tested, but +1

jsuchome:
+1 (not tested, but the code looks correct)

PR #147, opened 2015-12-09
nova: Only install openvswitch-kmp-xen when using xen, ovs and SLES

vuntz:
So far, we were installing the package as soon as xen was used, even
though ovs might not have been used.

Also, this is not needed on Leap, where we use the openvswitch module
from the upstream kernel.

### DIFF
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -182,10 +182,19 @@ case node[:nova][:libvirt_type]
           end
 
         when "xen"
-          %w{kernel-xen xen xen-tools openvswitch-kmp-xen}.each do |pkg|
+          %w{kernel-xen xen xen-tools}.each do |pkg|
             package pkg do
               action :install
             end
+          end
+
+          if node[:platform] == "suse"
+            # on SLES (but not Leap!), we need a xen-specific package for ovs
+            install_xen_module = false
+            node[:crowbar][:network].keys.each do |name|
+              install_xen_module ||= node[:crowbar][:network][name]["add_ovs_bridge"]
+            end
+            package "openvswitch-kmp-xen" if install_xen_module
           end
 
           service "xend" do


toabctl:
+1

aplanas:
+1

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -182,12 +182,21 @@ def set_boot_kernel_and_trigger_reboot(flavor="default")
...
+          %w{kernel-xen xen xen-tools}.each do |pkg|
             package pkg do
               action :install
             end
           end
 
+          if node[:platform] == "suse"
+            # on SLES (but not Leap!), we need a xen-specific package for ovs
+            install_xen_module = false
+            node[:crowbar][:network].keys.each do |name|
```
rhafer: I think you could just do check for ```node[:network][:needs_openvswitch]``` here. That attribute is being set when saving/applying the nova proposal as part of call to neutron_service.enable_neutron_networks().

It would be even more appropriate to do it somewhere here:
 https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/network/recipes/default.rb#L27

Actually, maybe the bug is that we hardcode it to openvswitch-kmp-default there. Maybe just using openvswitch-kmp works? Both packages provide that symbol.

vuntz: > Actually, maybe the bug is that we hardcode it to openvswitch-kmp-default there. Maybe just using openvswitch-kmp works? Both packages provide that symbol.

Ok, marking as WIP until we investigate this.

vuntz: @rhafer trying this in https://github.com/crowbar/crowbar-core/pull/190

vuntz:
Obsoleted by https://github.com/crowbar/crowbar-openstack/pull/429

PR #148, opened 2015-12-09
Do not claim disk that doesn't belong to right component.

jsuchome:
Based on https://github.com/crowbar/barclamp-cinder/pull/265 and https://github.com/crowbar/barclamp-swift/pull/298

Requires https://github.com/crowbar/crowbar-core/pull/161

### DIFF
--- a/chef/cookbooks/swift/recipes/disks.rb
+++ b/chef/cookbooks/swift/recipes/disks.rb
@@ -30,12 +30,14 @@ def get_uuid(disk)
   nil
 end
 
-unclaimed_disks = BarclampLibrary::Barclamp::Inventory::Disk.unclaimed(node)
-to_use_disks = []
-unclaimed_disks.each do |k|
-  to_use_disks << k
-end
-claimed_disks = BarclampLibrary::Barclamp::Inventory::Disk.claimed(node, "Swift")
+claim_string = "Swift"
+
+# take only disks with Swift role; when there is none, take disks without role
+swift_disks, disks_without_role =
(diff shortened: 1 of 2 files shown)


chef/cookbooks/swift/recipes/disks.rb
```diff
@@ -30,12 +30,14 @@ def get_uuid(disk)
...
-unclaimed_disks.each do |k|
-  to_use_disks << k
-end
-claimed_disks = BarclampLibrary::Barclamp::Inventory::Disk.claimed(node, "Swift")
+claim_string = "Swift"
+
+# take only disks with Swift role; when there is none, take disks without role
+swift_disks, disks_without_role =
+  BarclampLibrary::Barclamp::Inventory::Disk.prepared_for(node, claim_string)
+to_use_disks = swift_disks.empty? ? disks_without_role : swift_disks
```
jsuchome: This means that once a new disk is added to the node which already has swift-storage role and chef-client is executed, that disk will be claimed by swift. (Unless user manages to assign different role to it, but I'm not sure if there's time for that).

That's consistent with our current behavior, when user doesn't have to care about disk roles. However it might be bit confusing for user that would like to have more devices with different role on storage node. That said, maybe it should be discouraged... ?

jsuchome:
Gating cannot pass before https://github.com/crowbar/crowbar-core/pull/161 is merged

bmwiedemann:
for some reason tests were triggered again and failed in swift. https://ci.suse.de/view/Cloud/view/Builds/job/openstack-mkcloud/14642/

jsuchome:
Too old, apparently no one needs it

PR #149, opened 2015-12-09
Fix nova configuration for ZVM

dirkmueller:


mjura:
I think that we had patch for this driver which was moving this options to [zvm] section. In my opinion this specific options should stay in [zvm].

dirkmueller:
@mjura the change is not only moving it to a different section (but I dropped the patch already since it wasn't workign anymore).

vuntz:
+1 (not sure if you want to wait for the nova.conf rebase or not, though)

aplanas:
+1

PR #150, opened 2015-12-11
Some work for compatibility with chef 11

vuntz:
This is mostly about changing "node[:foo] =" to "node.set[:foo] ="

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/server.rb
+++ b/chef/cookbooks/ceilometer/recipes/server.rb
@@ -161,7 +161,7 @@ end
 # could possibly not be reached in case of errors).
 ruby_block "mark node for ceilometer db_sync" do
   block do
-    node[:ceilometer][:db_synced] = true
+    node.set[:ceilometer][:db_synced] = true
     node.save
   end
   action :nothing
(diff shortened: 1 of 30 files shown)


rsalevsky:
Looks good so far but the hound needs some attention.

tboerger:
Some offenses can be fixed, otherwise :+1:

vuntz:
I added a commit to help with hound, not sure if we really want to fix the "line too long" ones (would require adding some temporary variable just for this). @rsalevsky @tboerger what do you think?

mjura:
@vuntz : I had this same discussion last week with @rsalevsky about "Line too long". I'm voting for check fir lines with 150 characters.

@tboerger : What would you say about this?

aplanas:
Is this change of `default` and `.set` valid for chef 10? Or is expected problems once is merged?

vuntz:
> Is this change of default and .set valid for chef 10? Or is expected problems once is merged?

Good that you're asking :-) It is compatible (things that were not default `default` are actually a bug, `.set` is just implicit in Chef 10).

Gating seems to agree.

aplanas:
+1

rsalevsky:
+1

PR #151, opened 2015-12-11
cookbooks: Simplify monitor recipes

vuntz:
We return directly if the node is not nagios-client.


jsuchome:
+1

mjura:
It is not good that we don't support nagios barclamp, but we have to maintain nagios-client recipes.

Looks ok, +1

Please rebase

PR #152, opened 2015-12-11
rabbitmq: Fix nodename used for rabbitmq

vuntz:
It was missing the rabbit@ part, which is wrong as this could mean a
conflict with other erlang apps.

Note that this change breaks existing deployment of rabbitmq, as
rabbitmq can't get restarted (due to the nodename change). So it's only
okay to get in if we know that the upgrade process will stop rabbitmq.

(+ a cleanup commit)

### DIFF
--- a/chef/cookbooks/rabbitmq/attributes/default.rb
+++ b/chef/cookbooks/rabbitmq/attributes/default.rb
@@ -24,7 +24,7 @@
 default[:rabbitmq][:user] = "nova"
 default[:rabbitmq][:vhost] = "/nova"
 
-default[:rabbitmq][:nodename]  = node[:hostname]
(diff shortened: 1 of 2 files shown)


mjura:
Looks OK, but we have to remember about upgrade part. Did you create already card for this?

vuntz:
> Looks OK, but we have to remember about upgrade part. Did you create already card for this?

If we merge it now, I think it's fine (we can tell people to manually stop / restart it if they're already running this branch, and if they come from tex, then the upgrade process will stop rabbitmq anyway).

aplanas:
+1

rsalevsky:
+1

dirkmueller:
+1

PR #158, opened 2015-12-16
Update all usages of ::File.exists? to ::File.exist?

dirkmueller:
The exists? one is deprecated.

### DIFF
--- a/chef/cookbooks/cinder/recipes/common.rb
+++ b/chef/cookbooks/cinder/recipes/common.rb
@@ -79,14 +79,14 @@ if node[:cinder][:api][:protocol] == "https"
     package "openssl"
     ruby_block "generate_certs for cinder" do
       block do
-        unless ::File.exists? node[:cinder][:ssl][:certfile] and ::File.exists? node[:cinder][:ssl][:keyfile]
+        unless ::File.exist? node[:cinder][:ssl][:certfile] and ::File.exist? node[:cinder][:ssl][:keyfile]
           require "fileutils"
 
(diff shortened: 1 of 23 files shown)


aplanas:
Some of the Hound complains maybe needs to be addressed, even if they are not related with the change.

But in either case: +1

jsuchome:
+1 (I agree with @aplanas , Style/AndOr error might be fixed, I would not care that much about line length)

vuntz:
+1; if you want to add a commit to fix more issues, then that's fine, but definitely no need for that to merge the PR (as it's a never-ending story; that's what I wanted to try with my swift PR yesterday).

dirkmueller:
why are we actually having houndci enabled with those super annoying warnings (e.g. the line lenght)?

chef/cookbooks/cinder/recipes/common.rb
```diff
@@ -79,7 +79,7 @@
     package "openssl"
     ruby_block "generate_certs for cinder" do
       block do
-        unless ::File.exists? node[:cinder][:ssl][:certfile] and ::File.exists? node[:cinder][:ssl][:keyfile]
+        unless ::File.exist? node[:cinder][:ssl][:certfile] && ::File.exist? node[:cinder][:ssl][:keyfile]
```
vuntz: You need to do `File.exist?(foo) &&` instead of `File.exist? foo &&` (that's the annoying thing that changes when moving to `&&`)

mjura:
+1 for last commit message

aplanas:
Still +1

PR #138, opened 2015-12-03
Revert "increase timeout for synchronization of HA resources (bsc#935462)"

abel-navarro:
This reverts commit f04bd35ef94822f8e229e6cfb5219a16f1d92cb9.

When we merge this PR https://github.com/crowbar/crowbar-ha/pull/45,
the sync_mark timeout will increase to 180 second by default.

PR #159, opened 2015-12-16
increase nova_database sync timeout to 240 (bsc#935462)

aspiers:
The default sync timeout is now 180 seconds, so we definitely don't want to decrease it here; if anything it should be longer since nova is towards the end of a full run list, so the nodes are more likely to have drifted by this point.

https://bugzilla.suse.com/show_bug.cgi?id=935462

See also #138 which does something similar for `nova_ha_resources`.

### DIFF
--- a/chef/cookbooks/nova/recipes/database.rb
+++ b/chef/cookbooks/nova/recipes/database.rb
@@ -27,7 +27,7 @@ db_settings = fetch_database_settings
 
 crowbar_pacemaker_sync_mark "wait-nova_database" do
   # the db sync is very slow for nova
-  timeout 120
+  timeout 240
 end
 
 # Creates empty nova database


aspiers:
@vuntz @dirkmueller The gate won't tell us anything useful for this since it won't have HA enabled.  OK to just merge?  This should fix the failure seen in https://ci.suse.de/job/openstack-mkcloud/12930/ which is the failed gate for https://github.com/crowbar/crowbar-ha/pull/46.

dirkmueller:
+1

vuntz:
> if anything it should be longer since nova is towards the end of a full run list

If nova is towards the end, what about horizon, heat, ceilometer? :-)

aspiers:
That's why I submitted #160 too.

aspiers:
I guess you could argue ceilometer should have a higher timeout than nova.  Oh well.

PR #160, opened 2015-12-16
Revert increase timeout for sync of ceilometer HA resources (bsc#935462)

aspiers:
crowbar/crowbar-ha@d2c1a01c7 from crowbar/crowbar-ha#45 increased the default timeout from 60 to 180, so it no longer makes sense to set this to 120 which is now lower than the default.

This reverts commit 7b7cd2b571b2e7b0416bf22a9e6fdd5b04129fe4.

See also #138 and #159.


rsalevsky:
+1

aplanas:
+1

aspiers:
As with #159 it doesn't make any sense to wait for the gate on this one.

PR #162, opened 2015-12-17
postgresql, rabbitmq, cinder, nova, ceilometer: Use new pacemaker_transaction LWRP to create pacemaker resources

vuntz:
This depends on https://github.com/crowbar/crowbar-ha/pull/46 (EDIT: already merged)

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/mongodb.rb
+++ b/chef/cookbooks/ceilometer/recipes/mongodb.rb
@@ -49,13 +49,22 @@ if ha_enabled
   pacemaker_primitive "mongodb" do
     agent node[:ceilometer][:ha][:mongodb][:agent]
     op node[:ceilometer][:ha][:mongodb][:op]
-    action :create
+    action :update
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
 
   pacemaker_clone "cl-mongodb" do
     rsc "mongodb"
-    action [:create, :start]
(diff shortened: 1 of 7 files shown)


vuntz:
Removing wip tag: worked well with manual testing.

mjura:
Hmmm, in https://github.com/crowbar/crowbar-ha/pull/46#issue-120224127 it is comment
```
  action :update
  # this will never create anything, only converge an existing primitive to the right parameters
```
but in this change we overwrite all :create actions to :update

Could you clarify this, please.

vuntz:
@mjura: that's the whole point, the creation itself is done by the new transaction resource :-) We want to create the resources in one transaction, to avoid racesz.

aspiers:
Gate failed to apply keystone.

jsuchome:
I've retriggered gating run and voila - it passes!

+1 for the code

aspiers:
The gate does not test with `hacloud=1` so it doesn't tell us much at all.  This must be tested with `hacloud=1` before merging.

abel-navarro:
Manually deployed this with `hacloud=1` and does it successfully. Now running smoke tempest tests.

abel-navarro:
smoke tempest tests passed on an `hacloud=1` environment.

vuntz:
@aspiers said he was fine merging this before leaving, so considering this as a second +1 and merging.

PR #163, opened 2015-12-17
Create Ceph client.cinder key, when it is not yet ready (bsc#959567)

mjura:


mjura:
Depends on https://github.com/crowbar/crowbar-ceph/pull/30

aplanas:
Too many trailing spaces : )

mjura:
It was only rebased

mjura:
Ok, it was tested

aplanas:
+1 for me

chef/cookbooks/nova/recipes/ceph.rb
```diff
@@ -136,7 +136,9 @@
 
         if !admin_keyring.empty? && File.exists?(admin_keyring)
           # Now add our secret and its value
-          cmd = ["ceph", "-k", admin_keyring, "-c", ceph_conf, "auth", "get-key", "client.#{rbd_user}"]
+          cmd = ["ceph", "-k", admin_keyring, "-c", ceph_conf,
+                 "auth", "get-or-create-key", "client.#{rbd_user}"]
+
```
rsalevsky: Line breaks should be like this:
```ruby
cmd = [
  "ceph",
  "-k", admin_keyring,
  "-c", ceph_conf, "auth", 
  "get-or-create-key",
  "client.#{rbd_user}"
]
```

mjura:
@rsalevsky: I've adopted your comment, thank you

vuntz:
I don't think this is valid; the bug (from bugzilla) was really just because of the missing patch in the other PR, imho, and would not be triggered here.

It definitely feels wrong to create the key here.

mjura:
> I don't think this is valid; the bug (from bugzilla) was really just because of the missing patch in the other PR, imho, and would not be triggered here.

@vuntz: What other PR? For me this bug is still valid
```
Error executing action `create` on resource 'ruby_block[save nova key as libvirt secret]'

Mixlib::ShellOut::ShellCommandFailed
------------------------------------
Expected process to exit with [0], but received '2'
---- Begin output of ["ceph", "-k", "/etc/ceph/ceph.client.admin.keyring", "-c", "/etc/ceph/ceph.conf", "auth", "get-key", "client.cinder"] ----
STDOUT: 
STDERR: Error ENOENT: don't have client.cinder
---- End output of ["ceph", "-k", "/etc/ceph/ceph.client.admin.keyring", "-c", "/etc/ceph/ceph.conf", "auth", "get-key", "client.cinder"] ----
Ran ["ceph", "-k", "/etc/ceph/ceph.client.admin.keyring", "-c", "/etc/ceph/ceph.conf", "auth", "get-key", "client.cinder"] returned 2

Cookbook Trace:
---------------
/var/chef/cache/cookbooks/nova/recipes/ceph.rb:142:in `block (3 levels) in from_file'

Resource Declaration:
---------------------
# In /var/chef/cache/cookbooks/nova/recipes/ceph.rb

 94:   ruby_block "save nova key as libvirt secret" do
 95:     block do
 96:       # Check if libvirt is installed and started
 97:       if system("virsh hostname &> /dev/null")
 98: 
 99:         # First remove conflicting secrets due to same usage name
100:         virsh_secret = Mixlib::ShellOut.new("virsh secret-list")
101:         secret_list = virsh_secret.run_command.stdout
102:         virsh_secret.error!
103: 
104:         secret_lines = secret_list.strip.split("\n")
105:         if secret_lines.length < 2 || !secret_lines[0].lstrip.start_with?("UUID") || !secret_lines[1].start_with?("----")
106:           raise "cannot fetch list of libvirt secret"
107:         end
108:         secret_lines.shift(2)
109: 
110:         secret_lines.each do |secret_line|
111:           secret_uuid = secret_line.split(" ")[0]
112:           cmd = ["virsh", "secret-dumpxml", secret_uuid]
113:           virsh_secret_dumpxml = Mixlib::ShellOut.new(cmd)
114:           secret_xml = virsh_secret_dumpxml.run_command.stdout
115:           # some secrets might not be ceph-related, skip these
116:           next if secret_xml.index("<usage type='ceph'>").nil?
117: 
118:           # lazy xml parsing
119:           re_match = %r[<usage type='ceph'>.*<name>(.*)</name>]m.match(secret_xml)
120:           next if re_match.nil?
121:           secret_usage = re_match[1]
122:           undefine = false
123: 
124:           if secret_uuid == rbd_uuid
125:             undefine = true if secret_usage != "client.#{rbd_user} secret"
126:           else
127:             undefine = true if secret_usage == "client.#{rbd_user} secret"
128:           end
129: 
130:           if undefine
131:             cmd = ["virsh", "secret-undefine", secret_uuid]
132:             virsh_secret_undefine = Mixlib::ShellOut.new(cmd)
133:             virsh_secret_undefine.run_command
134:           end
135:         end

Compiled Resource:
------------------
# Declared in /var/chef/cache/cookbooks/nova/recipes/ceph.rb:94:in `block in from_file'

ruby_block("save nova key as libvirt secret") do
  action "create"
  retries 0
  retry_delay 2
  block_name "save nova key as libvirt secret"
  cookbook_name "nova"
  recipe_name "ceph"
  block #<Proc:0x000000052020b0@/var/chef/cache/cookbooks/nova/recipes/ceph.rb:95>
end

[2016-01-12T13:52:41+00:00] ERROR: Running exception handlers
[2016-01-12T13:52:41+00:00] FATAL: Saving node information to /var/chef/cache/failed-run-data.json
[2016-01-12T13:52:41+00:00] ERROR: Exception handlers complete
[2016-01-12T13:52:41+00:00] FATAL: Stacktrace dumped to /var/chef/cache/chef-stacktrace.out
[2016-01-12T13:52:41+00:00] FATAL: Mixlib::ShellOut::ShellCommandFailed: ruby_block[save nova key as libvirt secret] (nova::ceph line 94) had an error: Mixlib::ShellOut::ShellCommandFailed: Expected process to exit with [0], but received '2'
---- Begin output of ["ceph", "-k", "/etc/ceph/ceph.client.admin.keyring", "-c", "/etc/ceph/ceph.conf", "auth", "get-key", "client.cinder"] ----
STDOUT: 
STDERR: Error ENOENT: don't have client.cinder
---- End output of ["ceph", "-k", "/etc/ceph/ceph.client.admin.keyring", "-c", "/etc/ceph/ceph.conf", "auth", "get-key", "client.cinder"] ----
Ran ["ceph", "-k", "/etc/ceph/ceph.client.admin.keyring", "-c", "/etc/ceph/ceph.conf", "auth", "get-key", "client.cinder"] returned 2
```

mjura:
To reproduce this bug you have to deploy RADOS cinder backend on `cinder-controller` and `cinder-volume` on different nodes after nova proposal will be deployed.

With this patch cinder proposal is applying fine

mjura:
@vuntz: I'm still using this patch for my purposes and always works fine.

vuntz:
> @vuntz: What other PR?

https://github.com/crowbar/crowbar-ceph/pull/30

vuntz:
> To reproduce this bug you have to deploy RADOS cinder backend on cinder-controller and cinder-volume on different nodes after nova proposal will be deployed.

Yes, I understand. But how does chef-client on the nova nodes that are not cinder related run before the chef-client from the cinder proposal when you click apply? And isn't the key created there?

mjura:
> Yes, I understand. But how does chef-client on the nova nodes that are not cinder related run before the chef-client from the cinder proposal when you click apply? And isn't the key created there?

Yes, this scenario is described in bsc#959567! In this case `cinder.client` will be created with keyring, but without capabilities. During chef-client run on cinder related node missing capabilities will be updated. I've tested this case many times. Trust me or please test it by yourself, in bsc#959567 I've described how to reproduce this issue, it is quite easy.

jsuchome:
I think @mjura's explanation is convincing: +1  Or do you still see it as problematic, @vuntz ?

dirkmueller:
to be honest I still don't get it. in my assumption cinder roles are applied before nova roles, and cinder controller should create the key, so nova will have it later on in the described scenario. 

are you saying that the key is created by the cinder-volume role?

mjura:
> to be honest I still don't get it. in my assumption cinder roles are applied before nova roles, and cinder controller should create the key, so nova will have it later on in the described scenario.

> are you saying that the key is created by the cinder-volume role?

We discussed with @dirkmueller on IRC about this.  In @dirkmueller opinion real fix would be to rework the crowbar orchestration and it is similar topic like the HA orchestration markers. Unfortunately it will be develop for Cloud 7. 

I'm for merging this change because Ceph as distributed system can deal with such problems. Asynchronized solutions are better than synchronizing markers and doing this as one big thread.

mjura:
Can we merge it?

vuntz:
@dirkmueller are you +1 after the discussion you had? I'm still trying to wrap my head around this. I understand this fixes the problem, but it feels wrong. I wonder if it'd be an option to simply move the code there from compile phase to convergence phase (in a ruby_block).

mjura:
> I wonder if it'd be an option to simply move the code there from compile phase to convergence phase (in a ruby_block).

@vuntz: Do you have idea, how we could do this?

Btw. `ceph get-or-create-key` will create keyring only once and after when it exists, it will be fetched by other compute nodes. One of the argument against this change was that this can cause problem of mass `create-key` requests.  But this will be created and propagated to other monitors in ceph cluster.

Since ceph is distributed filesystem and allows for such operations I think that this should be fine.

vuntz:
I see we're already in a ruby_block, so I'm confused now.

So the issue is that chef has not run on cinder-volume yet for this ceph backend? That also means that what we setup in nova before that is not useful yet. Would it be an option to wait for cinder to be configured?

vuntz:
> Since ceph is distributed filesystem and allows for such operations I think that this should be fine.

Sure, I think it's fine, but as I said, it "feels" wrong. So I'm thinking out loud about other options to see if there's a working one that feels "less wrong". (If no, we then just merge this patch)

vuntz:
Ok, let's not lose too much time and merge it.

PR #164, opened 2015-12-17
Allow the deployment of Trove in insecure configurations

aplanas:


dirkmueller:
:+1:

mjura:
Looks OK +1

chef/cookbooks/trove/recipes/default.rb
```diff
@@ -91,6 +91,7 @@ class ::Chef::Recipe
 node.set["openstack"]["mq"]["database"]["rabbit"]["vhost"] = rabbitmq[:trove][:vhost]
 node.set["openstack"]["secret"][rabbitmq[:trove][:user]]["user"] = rabbitmq[:trove][:password]
 
+node.set["openstack"]["insecure"] = keystone_settings["insecure"]
```
vuntz: Don't you need to set node["openstack"]["database"]["insecure"] too? Because the default attribute that you add is evaluated before this line, so it's set to the old value of `node["openstack"]["insecure"]` (which might not be set, btw).

vuntz: Oh, it might work with convergence: failing the first time, but then the second time the attribute is set.

aplanas: Ahh this can explain the problem with bnc#955863 too. I will update this.

vuntz:
+1

aplanas:
* Set `node["openstack"]["database"]["insecure"]` to avoid a similar problem that the one in bnc#955863

dirkmueller:
+1

aplanas:
codeclimate looks blocked somehow, but ci and gating is green

vuntz:
Still +1

PR #166, opened 2015-12-18
trove: Register trove endpoint directly for the right region (bsc#955…

vuntz:
…863)

We were relying on convergence to get the endpoint registered for the
right region, which had the side-effect of also having an endpoint
registered for RegionOne.

https://bugzilla.suse.com/show_bug.cgi?id=955863


jsuchome:
+1

aplanas:
+1

PR #167, opened 2015-12-21
Allow to remove unused images after creating them

mjura:
This is fixing develcloud6 (4/linuxbridge/-t -s/all/ ) gating

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -2023,6 +2023,7 @@ firewall_driver=nova.virt.firewall.NoopFirewallDriver
 # Unused unresized base images younger than this will not be
 # removed (integer value)
 #remove_unused_original_minimum_age_seconds=86400
+remove_unused_original_minimum_age_seconds=0
 
 
 #


mmnelemane:
+1

rhafer:
Could you explain in a little more detail why this helps? Doesn't setting this to 0 miss the whole purpose of the image cache?

While it might fix the ci failures, I am not really convinced this is the right fix.

mjura:
test_volume_boot_pattern[compute,id-557cd2c2-4eb8-4dce-98be-f86765ff311b,image,smoke,volume]FAIL
> Could you explain in a little more detail why this helps? Doesn't setting this to 0 miss the whole purpose of the image cache?

This is because tempest test

```
Slowest 1 tests took 388.43 secs:
tempest.scenario.test_volume_boot_pattern.TestVolumeBootPatternV2
    test_volume_boot_pattern[compute,id-557cd2c2-4eb8-4dce-98be-f86765ff311b,image,smoke,volume]  388.43

======================================================================
FAIL: tempest.scenario.test_volume_boot_pattern.TestVolumeBootPatternV2.test_volume_boot_pattern[compute,id-557cd2c2-4eb8-4dce-98be-f86765ff311b,image,smoke,volume]
----------------------------------------------------------------------
```

This test is failing, because tempest creates Ceph boot volume and early after this is trying to remove it and it is not possible, because default time for removing unused base images is set up for 86400 seconds and test is timeout.

mjura:
@rhafer : As I can see this will be limited right now to `image_cache_manager_interval=2400`

rhafer:
> As I can see this will be limited right now to image_cache_manager_interval=2400

That makes this change even more weird IMO. In that case it seems you were just lucky that the cache_manager ran inbetween. Or am I missunderstanding something here?

mjura:
@rhafer : This option is weird at all, at it was added by VMware guy in https://review.openstack.org/#/c/59994 probably because VMware vCenter is caching this images in really awkward way.

dirkmueller:
thanks, was about to suggest to not merge this..

mjura:
This change is not valid any more

PR #168, opened 2015-12-28
add manilaclient.conf for tests

mmnelemane:
Functional tests use manilaclient.conf at /etc/manilaclient for reading credentials. In order to run functional tests without actually modifying the test code, the deployment needs to have the conf file installed. The installation of the directory is handled by https://build.opensuse.org/request/show/351091 and https://build.opensuse.org/request/show/351092. This commit copies manilaclient.conf file to the required path and updates the credentials for admin user and service user.


mmnelemane:
Populating the manilaclient.conf is not required unless running functional tests. Hence this change is now moved to cct.

PR #169, opened 2015-12-30
Enables manila tests

sayalilunkad:
Enables manila tests in tempest if manila service is running.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -248,6 +248,8 @@ raise("Cannot fetch EC2 credentials ") if ec2_access.empty? || ec2_secret.empty?
 use_ceilometer = $?.success?
 `#{keystonev2} endpoint-get --service database &> /dev/null`
 use_trove = $?.success?
+`#{keystonev2} endpoint-get --service share &> /dev/null`
+use_manila = $?.success?
 
 # FIXME: should avoid search with no environment in query
 neutrons = search(:node, "roles:neutron-server") || []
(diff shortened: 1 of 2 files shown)


chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -248,6 +248,8 @@
 use_ceilometer = $?.success?
 %x{#{keystonev2} endpoint-get --service database &> /dev/null}
 use_trove = $?.success?
+%x{#{keystonev2} endpoint-get --service share &> /dev/null}
```
rsalevsky:
```ruby
`#{keystonev2} endpoint-get --service share &> /dev/null}`
```

Please do it for the others as well in a separate commit.

dirkmueller: should be unrelated cleanup that can be in a different PR..

dirkmueller:
+1, assuming that it passes testing (which I guess it won't since we need to do more to enable manila)

chef/cookbooks/tempest/recipes/config.rb
vuntz: There was a PR about this recently, so maybe this should be changed for the new line here too?

vuntz:
So gating passed, but no manila tests were run in tempest.

Are some of them smoketests? Or do we need a full tempest run?

chef/cookbooks/tempest/recipes/config.rb
sayalilunkad: Done

sayalilunkad:
@vuntz tempest plugin for manila has not been enabled yet in the package so the tests will run.

vuntz:
@sayalilunkad ah, okay. I'm a bit reluctant to merge this without seeing a run with the manila tests. Or did you have a run on your cloud? If yes, were the results okay?

(code-wise, +1, btw)

sayalilunkad:
@vuntz Nope @toabctl said they won't pass :-) That's the reason we haven't enabled the plugin in the package yet.

vuntz:
Ah, okay. If the strategy is to merge this, and then enable the tempest in the packages later on once they pass, then I'm fine merging.

mmnelemane:
+1

aplanas:
+1

PR #170, opened 2015-12-31
limit OpenStack services by node role

aspiers:
**WIP - the helpers do not work due to Chef internals.  Need to be replaced with definitions or wrapper LWRPs.**

Now that `pacemaker_remote` can extend the cluster beyond the core corosync nodes, we have to ensure that controller services only run on the core nodes, and compute services only run on the compute nodes.  We do this by introducing new `crowbar-openstack::controller` and `crowbar-openstack::compute` roles, which new rule-based `location` constraints can then reference.

**N.B. These new constraints require https://github.com/crowbar/crowbar-ha/pull/54 to work.**

This PR is currently based on #139, but will be rebased when that is merged.

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/pacemaker_helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/pacemaker_helpers.rb
@@ -1,0 +1,46 @@
+#
+# Copyright 2015, SUSE
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# These helper methods provides a convenient way to create Pacemaker
+# location constraints which limit resources to only run on OpenStack
+# controller nodes or compute nodes.
+#
+# Example usage:
+#
+#   service_name = "keystone"
+#   location_name = "l-#{service_name}-controller"
+#   pacemaker_location location_name do
+#     definition controller_only_location(location_name, service_name)
+#     action :update
+#     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+#   end
+#   transaction_objects << "pacemaker_location[#{location_name}]"
+
+class Chef
+  class Resource
+    class PacemakerLocation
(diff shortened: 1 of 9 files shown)


chef/cookbooks/crowbar-openstack/libraries/pacemaker_helpers.rb
```diff
@@ -0,0 +1,46 @@
...
+
+class Chef
+  class Resource
+    class PacemakerLocation
+      def controller_only_location(location, service)
+        "location #{location} #{service} resource-discovery=exclusive " +
+          "rule 0: OpenStack-role eq controller"
+      end
+
+      def compute_only_location(location, service)
```
abel-navarro: I know it won't be a valid configuration but, do we prevent users to set a node as both controller and compute? (i.e. a one node cloud). If we don't prevent that, then when we reach this code it may cause problems.

chef/cookbooks/crowbar-openstack/recipes/compute.rb
```diff
@@ -0,0 +1,28 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+ha_enabled = !node[:pacemaker].nil?
+
+execute "set OpenStack-role to compute" do
+  command "crm node attribute #{node[:nodename]} set OpenStack-role compute"
```
vuntz: @aspiers hrm, shouldn't this be run on a non-remote node?

vuntz:
The bit which is totally unclear to me here is how the new roles are being added to the nodes. I guess I'm waiting for some more commits :-)

chef/cookbooks/crowbar-openstack/recipes/compute.rb
aspiers: Damn, you're right :-(

aspiers:
The roles will be assigned by Crowbar in the normal way.  But that's part of the UI story, so I didn't touch it.  Like I said this morning, currently I'm just manually editing proposals to assign them, although I didn't even test these roles yet.  That's why it's marked WIP :-)

chef/cookbooks/crowbar-openstack/libraries/pacemaker_helpers.rb
vuntz: We're in the context of a pacemaker cluster, and in this case, the question is "can we have a node both as remote and normal member of the cluster", and the answer is no.

You can still have a compute node on a cluster member, but not with hypervisor HA.

vuntz:
Obsoleted by https://github.com/crowbar/crowbar-openstack/pull/175 and https://github.com/crowbar/crowbar-openstack/pull/179

PR #171, opened 2016-01-04
Fixes hound issues for %x

sayalilunkad:
Replaces %x with backticks as per updated
ruby style.

### DIFF
--- a/chef/cookbooks/swift/recipes/proxy.rb
+++ b/chef/cookbooks/swift/recipes/proxy.rb
@@ -250,7 +250,7 @@ if node[:swift][:ssl][:enabled]
             end
 
             # Generate private key
-            %x(openssl genrsa -out #{node[:swift][:ssl][:keyfile]} 4096)
+            `openssl genrsa -out #{node[:swift][:ssl][:keyfile]} 4096`
             if $?.exitstatus != 0
               message = "SSL private key generation failed"
               Chef::Log.fatal(message)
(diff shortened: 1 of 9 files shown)


chef/cookbooks/swift/recipes/proxy.rb
```diff
@@ -271,7 +271,7 @@
             end
 
             # Generate self-signed certificate with above CSR
-            %x(openssl x509 -req -days 3650 -in #{ssl_csr_file} -signkey #{node[:swift][:ssl][:keyfile]} -out #{node[:swift][:ssl][:certfile]})
+            `(openssl x509 -req -days 3650 -in #{ssl_csr_file} -signkey #{node[:swift][:ssl][:keyfile]} -out #{node[:swift][:ssl][:certfile]})`
```
aplanas: Not Ruby expert, but I think that you need to replace the full `%x()`, including the parents (the same for the rest of the lines)

aplanas:
Not sure if we are using this kind of refactoring PR to also address the long line complains.

+1 when gating is green

sayalilunkad:
@aplanas there is currently a discussion on the cloud-devel ML about if we want to increase the line width to silence that error. Maybe after there is some conclusion on that we can act on it.

rsalevsky:
+1

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -266,7 +266,7 @@
 
 # FIXME: the command above should be good enough, but radosgw is broken with
 # tempest; also should avoid search with no environment in query
-#%x{#{keystone} endpoint-get --service object-store &> /dev/null}
+#%x{#{keystone} endpoint-get --service object-store &> /dev/null}`
```
rsalevsky: # `#{keystone} endpoint-get --service object-store &> /dev/null}`

PR #172, opened 2016-01-04
Run SSL certificate checks more thoroughly (bsc#960006)

dirkmueller:
We need to verify the certificate is non-empty, and we need
to check for it before we reconfigure services to use it.

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -350,8 +350,8 @@ if node[:horizon][:apache][:ssl]
       not_if { File.size?(node[:horizon][:apache][:ssl_crt_file]) }
     end
   else
-    unless ::File.exist? node[:horizon][:apache][:ssl_crt_file]
-      message = "Certificate \"#{node[:horizon][:apache][:ssl_crt_file]}\" is not present."
+    unless ::File.size? node[:horizon][:apache][:ssl_crt_file]
+      message = "The file \"#{node[:horizon][:apache][:ssl_crt_file]}\" does not exist or is empty."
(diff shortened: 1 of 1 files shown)


chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -339,24 +350,18 @@
...
+if node[:horizon][:apache][:ssl] && node[:horizon][:apache][:generate_certs]
+  package "apache2-utils"
+
+  bash "Generate Apache certificate" do
+    code <<-EOH
+      (umask 377 ; /usr/bin/gensslcert -C openstack-dashboard )
+EOH
+    not_if { File.size?(node[:horizon][:apache][:ssl_crt_file]) }
+  end
+    message = "Certificate \"#{node[:horizon][:apache][:ssl_crt_file]}\" does not exist or is empty."
```
aplanas: Do we need some `else` `unless ::File.exist? ...` as in the old code? Otherwise I think that we will raise the exception every time

vuntz:
I don't understand how that helps: the check is done in the compile phase anyway, so before we reconfigure services (since that's done in the second phase of chef). Or am I missing something?

vuntz:
(the change from `exist?` to `size?` might be enough, though)

dirkmueller:
@vuntz yes, I thought its clearer this way. Want me to change it back?

vuntz:
I don't mind moving the code, but I guess I'd do that in a separate commit to clarify things :-)

dirkmueller:
@vuntz done

vuntz:
+1

vuntz:
+1

aplanas:
+1

PR #173, opened 2016-01-05
neutron: Revert one piece from a98aa50da5471ea58657d4abf6cbf186fd1096a5

vuntz:
There's no need to drop DVR support for linuxbridge in the cookbook:
what matters is that crowbar makes it impossible; having the cookbook
support this will be helpful when upstream adds support for DVR with
linuxbridge.


aplanas:
+1

jsuchome:
+1

dirkmueller:
agreed, I raised that concern already to mjura back then but should have been more persistent.

PR #175, opened 2016-01-05
Set node attribute for OpenStack-role pacemaker attribute for controller

vuntz:
Depends on https://github.com/crowbar/crowbar-ha/pull/62

### DIFF
--- a/crowbar_framework/app/models/neutron_service.rb
+++ b/crowbar_framework/app/models/neutron_service.rb
@@ -381,8 +381,10 @@ class NeutronService < PacemakerServiceObject
 
     server_elements, server_nodes, server_ha_enabled = role_expand_elements(role, "neutron-server")
     reset_sync_marks_on_clusters_founders(server_elements)
+    Openstack::HA.set_controller_role(server_nodes) if server_ha_enabled
     network_elements, network_nodes, network_ha_enabled = role_expand_elements(role, "neutron-network")
     reset_sync_marks_on_clusters_founders(network_elements)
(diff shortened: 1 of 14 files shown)


crowbar_framework/app/models/neutron_service.rb
```diff
@@ -380,7 +380,9 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
     end
 
     server_elements, server_nodes, server_ha_enabled = role_expand_elements(role, "neutron-server")
+    OpenStack::HA.set_controller_role(server_nodes) if server_ha_enabled
     network_elements, network_nodes, network_ha_enabled = role_expand_elements(role, "neutron-network")
+    OpenStack::HA.set_controller_role(network_nodes) if network_ha_enabled
```
jsuchome: What about calling `set_controller_role` only once here, with the union of both sets as argument?

jsuchome:
Not tested, but the code looks correct. 
+1 (with one nitpick)

crowbar_framework/app/models/neutron_service.rb
vuntz: The `if` conditions are different, so we can't do that.

jsuchome: Ouch, silly me :-)

vuntz:
Had to change OpenStack to Openstack to make rails happy :/

aplanas:
+1

jsuchome:
Still +1

PR #177, opened 2016-01-05
Use new pacemaker_transaction LWRP to create resources (manila, swift, neutron)

vuntz:


chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -28,93 +28,111 @@
...
+if use_l3_agent
+  l3_agent_primitive = "neutron-l3-agent"
+  pacemaker_primitive l3_agent_primitive do
+    agent node[:neutron][:ha][:network][:l3_ra]
+    op node[:neutron][:ha][:network][:op]
+    action :update
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+  group_members << l3_agent_primitive
+  transaction_objects << "pacemaker_primitive[#{l3_agent_primitive}]"
```
jsuchome: Should this really be updated if ``use_l3_agent`` is true but node is not a cluster founder?

vuntz: Yes, we only do the update on the founder, but the transaction about creating things is actually unrelated to the update. (although, technically, we do the transaction on the founder only too... so you're kind of right, but that makes the code a bit less readable)

chef/cookbooks/manila/recipes/controller_ha.rb
```diff
@@ -47,37 +47,52 @@
...
-pacemaker_primitive "manila-api" do
+transaction_objects = []
+
+api_primitive = "manila-api"
+pacemaker_primitive api_primitive do
   agent node[:manila][:ha][:api_ra]
   op node[:manila][:ha][:op]
-  action :create
+  action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
toabctl: Just for my understanding: are theses ```only_if ...founder``` still needed if we check this now in the ```pacemaker_transaction``` ?

chef/cookbooks/neutron/recipes/network_agents_ha.rb
jsuchome: So if we do so many things on the founder only, wouldn't it be better to just exit once we are not at founder? Or is there still something we want to do?

jsuchome:
I think those spacing hound errors could be fixed...

chef/cookbooks/manila/recipes/controller_ha.rb
vuntz: Yes: the transaction is only used for creating the pacemaker resources, but the LWRP resource here is used to update an existing pacemaker resource, and that's actually the update case which is racy and where the founder check is really important.

vuntz:
> I think those spacing hound errors could be fixed...

Done

chef/cookbooks/neutron/recipes/network_agents_ha.rb
vuntz: There's the sync mark call at the end :-) We could arguably drop the sync mark and just return like you suggest, but that would actually recreate the drift issue that causes some failures.

Once we fix crowbar orchestration, I think we can do what you suggest with no major risk.

jsuchome:
+1

vuntz:
Since @aplanas gave a +1 on https://github.com/crowbar/crowbar-openstack/pull/180 which contains this PR, I'll consider we have a second +1 here :-)

PR #182, opened 2016-01-07
nova: Use search_env_filtered helper for search

vuntz:


ovisan:
:+1:

aplanas:
+1, maybe is better squash both commits

PR #184, opened 2016-01-08
Add class Crowbar::Upgrade to unset `db_synced` flag.

aplanas:


jsuchome:
Codewise this looks good, I assume you have tested it.

+1

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -0,0 +1,51 @@
...
+    # pre-upgrade actions needed for the nodes
+    def self.pre_upgrade
+    end
+
+    # post-upgrade actions needed for the nodes
+    def self.post_upgrade
+      unset_db_synced
+    end
+
+    # set to false the flag `db_synced` for every OpenStack component
```
jsuchome: Maybe the comment why it is needed would be helpful here...

aplanas:
* Expand `unset_db_synced` method comment

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -0,0 +1,53 @@
...
+    def self.pre_upgrade
+    end
+
+    # post-upgrade actions needed for the nodes
+    def self.post_upgrade
+      unset_db_synced
+    end
+
+    # set to false the flag `db_synced` for every OpenStack component,
+    # so the next time that Chef runst, will synchronize and update
```
jsuchome: Typo :-(

aplanas:
* Fix typo : (

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -0,0 +1,53 @@
...
+    # set to false the flag `db_synced` for every OpenStack component,
+    # so the next time that Chef runs, will synchronize and update the
+    # OpenStack service database
+    def self.unset_db_synced
+      roles = {
+        "ceilometer-server" => :ceilometer,
+        "cinder-controller" => :cinder,
+        "glance-server" => :glance,
+        "heat-server" => :heat,
+        "keystone-server" => :keysone,
```
vuntz: s/keysone/keystone/

aplanas:
* `s/keysone/keystone`

vuntz:
One thing I'm wondering is if this should be Crowbar::Upgrade while living in crowbar-openstack. I'm kind of thinking the perfect solution would be do have an upgrade framework where we can plug classes like yours. But I guess that it's good enough for now.

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -0,0 +1,53 @@
...
+        "cinder-controller" => :cinder,
+        "glance-server" => :glance,
+        "heat-server" => :heat,
+        "keystone-server" => :keystone,
+        "manila-server" => :manila,
+        "neutron-server" => :neutron,
+        "nova-controller" => :nova
+      }
+      roles.each do |role, component|
+        nodes = NodeObject.find("roles:{#role}")
```
vuntz: Actually, I read this to fast the first time. Why not do something like:

```ruby
NodeObject.all.each do |node|
  save_it = false
  component.each do |component|
    next unless node[component][db_synced]
    node[component][:db_synced] = false
  end
  node.save if save_it
end
```

Two reasons for this: I don't think we benefit from optimizing the code by only looking at the nodes we would need to look at (on the contrary, we do more searches); and more importantly, at this point in time in the upgrade process, the nodes don't have the roles assigned anymore.

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -0,0 +1,53 @@
...
+        nodes.each do |node|
+          if node.key?(component)
+            node.set[component][:db_synced] = false
```
vuntz: I'm not sure we need `.set` as `node` is not a chef node, but a crowbar node.

aplanas:
* Do not search by role, but iterate over all the `components`

vuntz:
Assuming the code is working, +1

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    # set to false the flag `db_synced` for every OpenStack component,
+    # so the next time that Chef runs, will synchronize and update the
+    # OpenStack service database
+    def self.unset_db_synced
+      # we can't search by roles (like 'keystone-server') because at
+      # this point the nodes maybe don't have roles assigned anymore
+      components = [:ceilometer, :cinder, :glance, :heat, :keystone,
+                    :manila, :neutron, :nova]
```
rsalevsky:
```ruby 
components = [
  :ceilometer, :cinder, :glance, :heat, :keystone, :manila, :neutron, :nova
]
```

rsalevsky:
Beside my small nitpick +1

aplanas:
* Style change

toabctl:
Broke the CI with
```
Checking for file conflicts: [..........error]
Detected 1 file conflict:

File /opt/dell/crowbar_framework/lib/crowbar/upgrade.rb
  from install of
     crowbar-core-3.0+git.1452787488.32b7279-196.5.noarch (Cloud)
  conflicts with file from install of
     crowbar-openstack-3.0+git.1452787647.0307b35-148.1.noarch (Cloud)

File conflicts happen when two packages attempt to install files with the same name but different contents. If you continue, conflicting files will be replaced losing the previous content.
Continue? [yes/no] (no): no
```

The file is already in crowbar-core

aplanas:
This is sad, I lost the race here. I moved the namespace to `Openstack` here https://github.com/crowbar/crowbar-openstack/pull/198

crowbar_framework/lib/crowbar/upgrade.rb
```diff
@@ -0,0 +1,49 @@
...
+      # we can't search by roles (like 'keystone-server') because at
+      # this point the nodes maybe don't have roles assigned anymore
+      components = [
+        :ceilometer, :cinder, :glance, :heat,
+        :keystone, :manila, :neutron, :nova
+      ]
+      NodeObject.all.each do |node|
+        save_it = false
+        components.each do |component|
+          next unless node[component][db_synced]
```
dirkmueller: :db_synced ?

aplanas: Ugg: https://github.com/crowbar/crowbar-openstack/pull/205

PR #185, opened 2016-01-11
Updates mtu configuration to use neutron config options.

rossella:
Removes custom dnsmasq-neutron.conf.erb and sets mtu options via.
neutron ml2_conf.ini.erb.

### DIFF
--- a/chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
@@ -82,7 +82,6 @@ dhcp_domain = <%= @dhcp_domain %>
 
 # Override the default dnsmasq settings with this file
 # dnsmasq_config_file =
-dnsmasq_config_file = /etc/neutron/dnsmasq-neutron.conf
 
 # Comma-separated list of DNS servers which will be used by dnsmasq
 # as forwarders.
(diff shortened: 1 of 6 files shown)


rossella:
WARN I haven't tested this

dirkmueller:
+1

aplanas:
+1

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
```diff
@@ -80,10 +80,6 @@ enable_metadata_network = <%= @enable_metadata_network %>
 # dhcp_domain = openstacklocal
 dhcp_domain = <%= @dhcp_domain %>
 
-# Override the default dnsmasq settings with this file
-# dnsmasq_config_file =
-dnsmasq_config_file = /etc/neutron/dnsmasq-neutron.conf
```
vuntz: I think we should keep the comment and the default (commented out) value, to not diverge from the default configuration file. So just remove that third line.

mmnelemane:
+1

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
rossella: agreed, thanks

PR #186, opened 2016-01-11
nova: Change grub config after xen kernel is installed

vuntz:
The code to change the grub config is parsing the list of available
kernels, but since it was running in the compilation phase, it had to
reboot twice to actually work as intended. So run this in the
convergence phase with a ruby_block.


aplanas:
+1

toabctl:
+1

vuntz:
No code change, just rephrased the commit to be clearer.

rsalevsky:
+1

aplanas:
+1

toabctl:
Still +1

PR #187, opened 2016-01-11
glance: Avoid greenthreads hanging endlessly on "slowloris" attack

toabctl:
See https://bugs.launchpad.net/nova/+bug/1361360 for the full
story.

### DIFF
--- a/chef/cookbooks/glance/templates/default/glance-registry.conf.erb
+++ b/chef/cookbooks/glance/templates/default/glance-registry.conf.erb
@@ -171,6 +171,7 @@ workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
 # have to set this option to False when you create a wsgi server.
 # (boolean value)
 #http_keepalive = true
+http_keepalive = false
 
 # Timeout for client connections' socket operations. If an incoming
 # connection is idle for this number of seconds it will be closed. A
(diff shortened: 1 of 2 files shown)


chef/cookbooks/glance/templates/default/glance-registry.conf.erb
```diff
@@ -171,6 +171,7 @@ workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
 # have to set this option to False when you create a wsgi server.
 # (boolean value)
 #http_keepalive = true
+http_keepalive = false
```
aplanas: Is this needed? As I can see in the link the change is only in glance-api.conf

toabctl: It's also there for the registry service. See https://review.openstack.org/#/c/130839/ (linked in the bug). Or am I missing something?

aplanas: My bad, I checked the comments but not the patch.

aplanas:
+1

vuntz:
+1.

Do you have a list of other components where we still need the change?

toabctl:
> Do you have a list of other components where we still need the change?

I think @dirkmueller did already the other services in bb7420313010995c0ba62bad70e9ecb4d99df04a . Manila is also already fixed in 2259327999453121a395edeae1fe23f76590c40d . So I think glance is the only missing service.

dirkmueller:
+1

PR #189, opened 2016-01-12
tempest: Add a file to flag when a deployment is Xen-only

vuntz:
This will be useful to disable some tests that are failing for Xen, and
that are caused by known unfixed upstream bugs.

The check is not really thorough as we only check that there's no KVM
node, but for all practical matters, it's good enough.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -316,10 +316,17 @@ if backend_names.length > 1
 end
 
 kvm_compute_nodes = search(:node, "roles:nova-compute-kvm") || []
+xen_compute_nodes = search(:node, "roles:nova-compute-xen") || []
 docker_compute_nodes = search(:node, "roles:nova-compute-docker") || []
 
 use_resize = kvm_compute_nodes.length > 1
 use_livemigration = nova[:nova][:use_migration] && kvm_compute_nodes.length > 1
+
+# create a flag to disable some test for xen (lp#1443898)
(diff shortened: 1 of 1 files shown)


chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -316,11 +316,17 @@
...
+xen_compute_nodes = search(:node, "roles:nova-compute-xen") || []
 docker_compute_nodes = search(:node, "roles:nova-compute-docker") || []
 
 use_resize = kvm_compute_nodes.length > 1
 use_livemigration = nova[:nova][:use_migration] && kvm_compute_nodes.length > 1
 
+if !xen_compute_nodes.empty? && kvm_compute_nodes.empty?
+  # create a flag to disable some test (lp#1443898)
+  file "#{node[:tempest][:tempest_path]}/flag-xen_only"
+end
```
vuntz: Ah, I forgot: do we want to remove that flag file in a `else`?

aplanas: IMHO yes, if we reuse the same deployment first with xen, and after that w/out xen, we will disable the test. But honestly is a corner case.

aplanas:
+1

rsalevsky:
+1

chef/cookbooks/tempest/recipes/config.rb
toabctl: yes. please add a ```else```

vuntz: Done.

aplanas:
+1

rsalevsky:
Still +1

PR #192, opened 2016-01-12
Remove duplicated options in [DEFAULT] section

mjura:
We would like to remove duplicated options from [DEFAULT] section

Options:
```
# Host on which to listen for incoming requests (string value)
#serialproxy_host = 0.0.0.0

# Port on which to listen for incoming requests (integer value)
# Minimum value: 1
# Maximum value: 65535
#serialproxy_port = 6083
```
were moved to [serial_console] section, it was fixed in upstream https://review.openstack.org/#/c/244177/

Options:
```
# Host on which to listen for incoming requests (string value)
#html5proxy_host = 0.0.0.0

# Port on which to listen for incoming requests (integer value)
# Minimum value: 1
# Maximum value: 65535
#html5proxy_port = 6082
```
were moved to [spice] section, it will be fixed in upstream
https://review.openstack.org/#/c/264271/

All options they have already their own namespaces and in [DEFAULT] section, they are not restricted

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -83,14 +83,6 @@ novncproxy_host = <%= @bind_host %>
 novncproxy_port = <%= @bind_port_novncproxy %>
 
 # Host on which to listen for incoming requests (string value)
-#serialproxy_host = 0.0.0.0
-
-# Port on which to listen for incoming requests (integer value)
-# Minimum value: 1
-# Maximum value: 65535
-#serialproxy_port = 6083
-
-# Host on which to listen for incoming requests (string value)
 #html5proxy_host = 0.0.0.0
 
 # Port on which to listen for incoming requests (integer value)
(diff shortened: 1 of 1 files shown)


aplanas:
+1 but maybe put this in two commits (or even two pr)

mjura:
@aplanas: I've spllitted this change in two

aplanas:
cool: still +1

rsalevsky:
+1

vuntz:
Doesn't that create a diff with the default config file? If no, +1.

cc @toabctl @dirkmueller

dirkmueller:
This creates a diff with the default config file. I don't think it makes sense to merge this unless it gets fixed upstream (which is unlikely as it is part of bigger rework in master..).

aplanas:
I propose to include lp#264271 and lp#244177 in the package as a patch, and merge this, because eventually they will land in the real package and change the default conf.

mjura:
> I propose to include lp#264271 and lp#244177 in the package as a patch, and merge this, because eventually they will land in the real package and change the default conf.

Object, lp#264271 and lp#244177 are part of bigger change in master. It will need a lot of effort to backport it.

Because this change will cause a diff between nova.conf.sample I think it would be the best to close this PR.

PR #193, opened 2016-01-12
nova: Fix call to non existing method (caused by f064ae87)

vuntz:
cc @dirkmueller


rsalevsky:
After the hound message is fixed +1

vuntz:
Hound is happy now (and I fixed the error message which was broken...)

aplanas:
+1

bmwiedemann:
+1 - mkcloud had passed with kvm but failed to boot the VM with xen (maybe unrelated)

vuntz:
Deployment work in gating; since the code is really just in the validation of the proposal, that should be good enough even though tempest failed (which is due to failures related to xen).

toabctl:
+1

PR #194, opened 2016-01-13
[cinder] Add rbd custom capabilities for Cinder RADOS backend, (bsc#961108)

mjura:
Depends on https://github.com/crowbar/crowbar-ceph/pull/32

And this is how it looks like

custom_caps


mjura:
Depends on https://github.com/crowbar/crowbar-ceph/pull/32

dirkmueller:
I am not sure we want this to be exposed to UI, since this is a temporary workaround for ephemeral-storage-on-rbd not implemented. (consider documentation and QA overheard for a user-visible option).

mjura:
This change can be treated separately with https://github.com/crowbar/crowbar-ceph/pull/32

mjura:
This is not needed any more, all changes were done in https://github.com/crowbar/crowbar-ceph/pull/32

PR #191, opened 2016-01-12
Tempest ui

dirkmueller:

PR #196, opened 2016-01-14
tempest: Hide menu item that leads to tempest dashboard

vuntz:
The tempest dashboard is non-functional right now, so at least hide it.

This can be reverted once https://github.com/crowbar/crowbar-openstack/pull/191 is completed.


dirkmueller:
Fine by me

aplanas:
+1

PR #199, opened 2016-01-14
neutron: Do not set external_network_type in ml2_conf.ini

vuntz:
This was set to flat, but this results in a config that is invalid for
tempest tests related to external networks
(tempest.api.network.admin.test_external_network_extension.ExternalNetworksTestJSON):
it's simply not possible to create an external network because there's
no way to allocate a segment for the flat type.

Note that this was introduced following https://github.com/crowbar/crowbar-openstack/pull/97#issuecomment-160137507

cc @rhafer


dirkmueller:
+1. I was looking at the same thing just now

rhafer:
Hm, I guess one could also argue that tempest could request a type that works for it. But anyway. I am fine with this change.

+1

PR #201, opened 2016-01-15
Function to check if additional repos are present after the upgrade of admin node

jsuchome:
Requires https://github.com/crowbar/crowbar-core/pull/188 and https://github.com/crowbar/crowbar-ceph/pull/34

Function like this needs to be called from upgrade wizard

### DIFF
--- a/crowbar_framework/lib/openstack/upgrade.rb
+++ b/crowbar_framework/lib/openstack/upgrade.rb
@@ -45,5 +45,38 @@ module Openstack
         node.save if save_it
       end
     end
+
+    def self.enable_repos_for_feature(feature)
+      Crowbar::Repository.check_all_repos.each do |repo|
+        # enable disabled repos now
+        if (repo.config["features"] || []).include? feature
+          Rails.logger.info("enabling repository #{repo.id}")
+          @provisioner_service.enable_repository(repo.platform, repo.arch, repo.id)
+        end
+      end
+    end
+
+    # Check if the additional repositories (HAE, SES) are present
+    # after the upgrade of the admin node
+    def self.check_additional_repositories_presence(thelogger)
+      @provisioner_service = ProvisionerService.new thelogger
+      # check if HA repos are needed
+      pacemaker_proposal = Proposal.where(barclamp: "pacemaker").first
+      unless pacemaker_proposal.nil?
+        unless Crowbar::Repository.provided? "ha"
+          raise I18n.t("barclamp.pacemaker.validation.hae_repo")
+        end
+        enable_repos_for_feature "ha"
+      end
+
+      # check if Storage repos are neededa
+      ceph_proposal = Proposal.where(barclamp: "ceph").first
+      unless ceph_proposal.nil?
+        unless Crowbar::Repository.provided? "ceph"
+          raise I18n.t("barclamp.pacemaker.validation.ses_repos")
+        end
+        enable_repos_for_feature "ceph"
+      end
+    end
   end
 end


aplanas:
lgtm so far, from where is going to be called?

jsuchome:
Not sure. I expect from upgrade wizard (which does not exist yet), in a loop, until it returns true.

aplanas:
Are the i18n string also there? Actually I do not find them

jsuchome:
The one for pacemaker should be there - I haven't checked, but the code is copied from crowbar-ha.

crowbar_framework/lib/openstack/upgrade.rb
```diff
@@ -45,5 +45,26 @@ def self.unset_db_synced
...
     end
+
+    # Check if the additional repositories (HAE, SES) are present
+    # after the upgrade of the admin node
+    def self.check_additional_repositories_presence
+
+      # check if HA repos are needed
+      pacemaker_proposal = Proposal.where(barclamp: "pacemaker").first
+      unless (pacemaker_proposal.nil? || Crowbar::Repository.provided_and_enabled? ("ha"))
+        validation_error I18n.t("barclamp.pacemaker.validation.hae_repo")
```
vuntz: Is `validation_error` available from here?

jsuchome: It is not.

crowbar_framework/lib/openstack/upgrade.rb
```diff
@@ -45,5 +45,21 @@ def self.unset_db_synced
         node.save if save_it
       end
     end
+
+    # Check if the additional repositories (HAE, SES) are present
+    # after the upgrade of the admin node
+    def self.check_additional_repositories_presence
+      # check if HA repos are needed
+      pacemaker_proposal = Proposal.where(barclamp: "pacemaker").first
+      unless pacemaker_proposal.nil? || (Crowbar::Repository.provided_and_enabled? "ha")
```
jsuchome: Looks like we will need to split the check for "provided" and "enabled"

vuntz:
+1. We could possibly move the checks to crowbar-ha and crowbar-ceph later on -- but we'll figure this out when we know how this will be called.

aplanas:
+1

jsuchome:
I've added dependency on https://github.com/crowbar/crowbar-ceph/pull/34

crowbar_framework/lib/openstack/upgrade.rb
```diff
@@ -45,5 +45,36 @@ def self.unset_db_synced
...
+        if (repo.config["features"] || []).include? "ha"
+          Rails.logger.info("enabling repository #{repo.id}")
+          @provisioner_service.enable_repository(repo.platform, repo.arch, repo.id)
+        end
+      end
+
+      # check if Storage repos are neededa
+      ceph_proposal = Proposal.where(barclamp: "ceph").first
+      unless ceph_proposal.nil?
+        unless Crowbar::Repository.provided_and_enabled? "ceph"
```
vuntz: Hrm, should be `provided?` I think.

jsuchome:
(rebased)

crowbar_framework/lib/openstack/upgrade.rb
```diff
@@ -45,5 +45,36 @@ def self.unset_db_synced
         node.save if save_it
       end
     end
+
+    # Check if the additional repositories (HAE, SES) are present
+    # after the upgrade of the admin node
+    def self.check_additional_repositories_presence
+      @provisioner_service = ProvisionerService.new Rails.logger
```
rsalevsky: The logger should be handed over by the function which calls the lib.

jsuchome:
@vuntz could you please merge this, it's out of date again...

crowbar_framework/lib/openstack/upgrade.rb
```diff
@@ -45,5 +45,36 @@ def self.unset_db_synced
...
+    def self.check_additional_repositories_presence(thelogger)
+      @provisioner_service = ProvisionerService.new thelogger
+      # check if HA repos are needed
+      pacemaker_proposal = Proposal.where(barclamp: "pacemaker").first
+      unless pacemaker_proposal.nil?
+        unless Crowbar::Repository.provided? "ha"
+          raise I18n.t("barclamp.pacemaker.validation.hae_repo")
+        end
+        # enable disabled repos now
+        if (repo.config["features"] || []).include? "ha"
```
vuntz: Sorry, before merging, one last question: where is `repo` defined? Or don't you want to iterate over all repos?

jsuchome: Ouch, indeed. I must have lost it somewhere...

jsuchome:
Hopefully fixed last part, but it's untested now...

vuntz:
+1

nkrinner:
+1

nkrinner:
Ran a rudimentary but successful test. Seems to work fine.

PR #202, opened 2016-01-15
Configure additional swift services

vuntz:
There's one tempest test at least depending on the container sync.

### DIFF
--- a/chef/cookbooks/swift/recipes/storage.rb
+++ b/chef/cookbooks/swift/recipes/storage.rb
@@ -84,15 +84,15 @@ template "/etc/swift/container-sync-realms.conf" do
 end
 
 svcs = %w{swift-object swift-object-auditor swift-object-expirer swift-object-replicator swift-object-updater}
-svcs = svcs + %w{swift-container swift-container-auditor swift-container-replicator swift-container-sync swift-container-updater}
-svcs = svcs + %w{swift-account swift-account-reaper swift-account-auditor swift-account-replicator}
+svcs += %w{swift-container swift-container-auditor swift-container-replicator swift-container-sync swift-container-updater}
+svcs += %w{swift-account swift-account-reaper swift-account-auditor swift-account-replicator}
 
 ## make sure to fetch ring files from the ring compute node
 if (!compute_nodes.nil? and compute_nodes.length > 0 )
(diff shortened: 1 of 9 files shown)


vuntz:
Gating is failing due to a packaging bug in swift; I fixed that but it will take some time to reach gating...

chef/cookbooks/swift/recipes/storage.rb
```diff
@@ -44,7 +44,11 @@
 
 storage_ip = Swift::Evaluator.get_ip_by_type(node,:storage_ip_expr)
 
-%w{account-server object-server container-server}.each do |service|
+env_filter = " AND swift_config_environment:#{node[:swift][:config][:environment]}"
+compute_nodes = search(:node, "roles:swift-ring-compute#{env_filter}")
```
aplanas: Style / readability question, is not more clear this?

```ruby
env_filter = "swift_config_environment:#{node[:swift][:config][:environment]}"
compute_nodes = search(:node, ["roles:swift-ring-compute", env_filter].join(" AND "))
proxy_nodes = search(:node, ["roles:swift-proxy", env_filter].join(" AND "))
```

vuntz: You are correct; I actually think we have a helper that we can use for this and that makes things look simpler.

chef/data_bags/crowbar/template-swift.schema
```diff
@@ -118,6 +118,10 @@
               "generate_certs": { "type" : "bool", "required" : true },
               "insecure": { "type" : "bool", "required" : true }
             }},
+            "container_sync": { "type": "map", "required": true, "mapping": {
+              "key": { "type" : "str", "required" : true },
+              "key2": { "type" : "str", "required" : true }
```
aplanas: do we need migrations for `key` and `key2`?

vuntz: Gah, I forgot to git add that. Fixed!

chef/cookbooks/swift/recipes/storage.rb
vuntz: fixed

aplanas:
LGTM +1

aplanas:
+1

jsuchome:
Not tested, but looks good to me. +1

PR #98, opened 2015-11-11
Update nova.conf.erb config template with Liberty

mjura:

PR #203, opened 2016-01-15
nova: Set paths to ssl certificate in nova.conf

vuntz:


kbaikov:
+1

aplanas:
+1

dirkmueller:
*sigh*, regression from the nova.conf rework in #98 

+1

PR #204, opened 2016-01-15
swift: Fix ceilometer integration

vuntz:
Driver must be messaging, not rabbit.

### DIFF
--- a/chef/cookbooks/swift/templates/default/proxy-server.conf.erb
+++ b/chef/cookbooks/swift/templates/default/proxy-server.conf.erb
@@ -738,7 +738,7 @@ super_admin_key = <%= @admin_key %>
 paste.filter_factory = ceilometermiddleware.swift:filter_factory
 control_exchange = swift
 url = <%= @rabbit_settings[:url] %>
-driver = rabbit
+driver = messaging
 topic = notifications
 set log_level = WARN
 


chef/cookbooks/swift/templates/default/proxy-server.conf.erb
```diff
@@ -738,7 +738,7 @@ super_admin_key = <%= @admin_key %>
 paste.filter_factory = ceilometermiddleware.swift:filter_factory
 control_exchange = swift
 url = <%= @rabbit_settings[:url] %>
-driver = rabbit
+driver = messaging
```
aplanas: Maybe I would comment this line and do not setup the driver. In Liberty is named 'messaging' and in master 'messagingv2'. The code shows that if this parameter is not set, the correct one is chosen.

aplanas: For reference:

* https://github.com/openstack/ceilometermiddleware/blob/stable/liberty/ceilometermiddleware/swift.py#L114
* https://github.com/openstack/ceilometermiddleware/blob/master/ceilometermiddleware/swift.py#L114

vuntz: We do set the driver in the other config files, though (at least for glance and nova), so maybe we'd switch to v2 earlier than post-Liberty... I don't have any strong feeling either way.

What do others think?

aplanas:
In either case: +1

chef/cookbooks/swift/templates/default/proxy-server.conf.erb
sayalilunkad: We can keep this as messaging for now and later change to v2, which I guess will be needed in the other services as well..

aplanas: @vuntz my proposal is drop the line and do not setup the driver, because the middleware takes care of this

But if is set in different configurations, lets merge this and postpone the decision for later

dirkmueller:
+1

PR #205, opened 2016-01-18
Fix :db_synced is symbol

aplanas:


tboerger:
:+1:

jsuchome:
+1

vuntz:
As the code is not called yet, no reason to wait for gating.

PR #206, opened 2016-01-18
manila: Fix proposal dependencies (bnc#949253)

toabctl:
Manila's generic driver needs Nova, Glance, Cinder and
Neutron. Add all of them as proposal dependencies.


toabctl:
> Not related with this PR, but the explicit dep with Neutron and Nova suggest that that Manilla needs to be, in the barclamp view, after Nova (now is after Cinder)

Done in https://github.com/crowbar/crowbar-openstack/commit/531592292cfde9dd102d6933e332f2d06c3141a6

aplanas:
+1

rsalevsky:
+1

vuntz:
Gating failed because manila got queue, and it got queued because it didn't know which nova, cinder, glance and neutron proposal to look at. You probably need to add attributes to reference these proposals (like we do for keystone).

toabctl:
@vuntz are you talking about something like https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/models/keystone_service.rb#L57 ?

vuntz:
@toabctl oh, sorry, I was not clear. Something like https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/models/manila_service.rb#L88-L93

toabctl:
@vuntz done in https://github.com/toabctl/crowbar-openstack/commit/7f974495a5d43f0611062afdd437cd9fd2db36c0

vuntz:
@toabctl I believe we also need to add the new attributes to the schema and template, + a migration.

toabctl:
@vuntz .ah. ok. adding dependencies is more complicated than I thought...

toabctl:
@vuntz maybe now!? untested...

vuntz:
Yep, looks good, let's see what gating says!

vuntz:
+1

aplanas:
Still: +1

dirkmueller:
+1

PR #207, opened 2016-01-18
swift: Include logging recipe

toabctl:
Swift wants to restart the rsyslog service which is defined in
the logging recipe. This fixes:

FATAL: Chef::Exceptions::ResourceNotFound: resource
template[/etc/rsyslog.d/11-swift.conf] is configured to notify resource
service[rsyslog] with action restart, but service[rsyslog] cannot be
found in the resource collection. template[/etc/rsyslog.d/11-swift.conf]
is defined in /var/chef/cache/cookbooks/swift/recipes/default.rb:39:in
`from_file'

### DIFF
--- a/chef/cookbooks/swift/recipes/default.rb
+++ b/chef/cookbooks/swift/recipes/default.rb
@@ -34,12 +34,14 @@ template "/etc/swift/swift.conf" do
  })
 end
 
-rsyslog_version = `rsyslogd -v | head -1 | sed -e "s/^rsyslogd \\(.*\\), .*$/\\1/"`
-# log swift components into separate log files
-template "/etc/rsyslog.d/11-swift.conf" do
-  source "11-swift.conf.erb"
-  mode "0644"
-  variables(rsyslog_version: rsyslog_version)
-  notifies :restart, "service[rsyslog]"
-  only_if { node[:platform_family] == "suse" } # other distros might not have /var/log/swift
+if node.roles.include?("logging-client")
+  rsyslog_version = `rsyslogd -v | head -1 | sed -e "s/^rsyslogd \\(.*\\), .*$/\\1/"`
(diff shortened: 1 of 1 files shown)


jsuchome:
+1

vuntz:
I'm actually not sure this is fully correct. My guess is that this is happening because the node somehow didn't get the logging-client role.

Now whether it's right to actually start rsyslog from the swift recipe (because this is what your change would do iwhen the logging-client role is missing) is something debatable. I would probably instead just remove the notification when the node doesn't have the logging-client role.

toabctl:
> I'm actually not sure this is fully correct. My guess is that this is happening because the node somehow didn't get the logging-client role.

I need to check the reason.

> Now whether it's right to actually start rsyslog from the swift recipe (because this is what your change would do iwhen the logging-client role is missing) is something debatable. I would probably instead just remove the notification when the node doesn't have the logging-client role.

Fixed in https://github.com/crowbar/crowbar-openstack/commit/6e742f09ab2241c111fe1e2375be71f8112c079e

dirkmueller:
hmm, but how did the node end up not having the logging-client role?

vuntz:
+1; we also need to understand why the role was missing, of course.

toabctl:
Hm. I can't reproduce it anymore. Had that failure yesterday 3-4x...

toabctl:
@dirkmueller @vuntz : should we merge that anyway? or wait for another failure and debug it?

vuntz:
It's still valid, imho. The user could remove the node from logging-client.

vuntz:
We should actually also check with logging-server, but I want this now -- so I'm going to merge and submit a followup PR.

toabctl:
For the record - the followup PR is https://github.com/crowbar/crowbar-openstack/pull/212

PR #208, opened 2016-01-19
swift: Create container-sync-realms.conf on proxy nodes too

vuntz:
The file is needed not just on container/storage nodes, but also on the
proxy node.

### DIFF
--- a/chef/cookbooks/swift/recipes/common_config.rb
+++ b/chef/cookbooks/swift/recipes/common_config.rb
@@ -1,0 +1,36 @@
+#
+# Copyright 2016, SUSE
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


aplanas:
+1

is planned to make the number `realms` configurable?

jsuchome:
+1 (Do we have a bug report? Or some failure indicating that it's missing on proxy  nodes?)

vuntz:
> (Do we have a bug report? Or some failure indicating that it's missing on proxy nodes?)

https://ci.suse.de/job/openstack-mkcloud/14923/artifact/.artifacts/tempest.log

vuntz:
> is planned to make the number realms configurable?

It's what I mentioned in https://github.com/crowbar/crowbar-openstack/commit/0749ce0f7e532b929f394eb64b0eeff68208ba1d

I'm not exactly sure how people will want to use that, so for now it's only one realm with a default config that is not super useful.

PR #209, opened 2016-01-19
manila: Use keystone v2 for cinder,neutron and nova

toabctl:
For the generic driver, Manila uses internally cinder, neutron and nova
to setup a service instance.
While i.e. creating a cinder volume, the cinderclient uses the auth_url
configured in manila.conf . This url must be set to v2. Otherwise
authentication fails. See lp#1535700 .


dirkmueller:
Interesting, so this is more prominent than I thought. we have the same issue in novaclient

dirkmueller:
+1 assuming that it works

aplanas:
+1

vuntz:
Fine with me; I would possibly have used a variable in the recipe to avoid duplicating the code, but no big deal :-)

toabctl:
@dirkmueller it works. tested manually. ready for merging..

PR #210, opened 2016-01-19
nova: Don't specify ssl certificates in nova.conf for compute-only nodes

vuntz:
Nova now tries to use the certificates also when creating a client for
glance: https://review.openstack.org/#/c/72974/

Since we really only have this setting to control whether the controller
uses SSL, always put empty paths for ssl certificates on compute-only
nodes.

### DIFF
--- a/chef/cookbooks/nova/recipes/config.rb
+++ b/chef/cookbooks/nova/recipes/config.rb
@@ -171,6 +171,33 @@ else
   oat_server = node
 end
 
+# only put certificates in nova.conf for controllers; on compute nodes, we
+# don't need them and specifying them results in the certificates being queried
+# when creating clients for glance
+if api[:nova][:ssl][:enabled] && node["roles"].include?("nova-controller")
+  api_ssl_certfile = api[:nova][:ssl][:certfile]
+  api_ssl_keyfile = api[:nova][:ssl][:keyfile]
+  api_ssl_cafile = api[:nova][:ssl][:ca_certs]
+else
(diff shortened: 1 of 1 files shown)


chef/cookbooks/nova/recipes/config.rb
```diff
@@ -171,8 +171,36 @@
...
 
+# only put certificates in nova.conf for controllers; on compute nodes, we
+# don't need them and specifying them results in the certificates being queried
+# when creating clients for glance
+if api[:nova][:ssl][:enabled] && node["roles"].include?("nova-controller")
+  api_ssl_certfile = api[:nova][:ssl][:certfile]
+  api_ssl_keyfile = api[:nova][:ssl][:keyfile]
+  api_ssl_cafile = api[:nova][:ssl][:ca_certs]
+else
+  api_ssl_certfile = ""
```
toabctl: Not sure if that makes a difference, but currently we set this always in the template (and also to an empty string). Shouldn't we check in the template if the variable is empty or nil and don't set it if so?

vuntz: I don't think it makes a difference as an empty string is the same as the default, afaict. But I can still do that if you want?

toabctl: No. if you don't have to rework it, fine for me. +1

vuntz:
Since this is blocking M9, I'm going ahead and merging.

PR #213, opened 2016-01-21
tempest: Temporary disable manila tests

toabctl:
Currently we have the tests disabled in our package but we want
to enable it without breaking the gate. So disable the tests for
now, enable the tests in the package then as see gating results
when reverting this patch.


aplanas:
+1

vuntz:
+1

PR #214, opened 2016-01-21
Restores postgresql database

sayalilunkad:
This adds a recipe to restore the postgresql database and
adds condition to do this when node is an upgraded one.

### DIFF
--- a/chef/cookbooks/postgresql/recipes/db_restore.rb
+++ b/chef/cookbooks/postgresql/recipes/db_restore.rb
@@ -1,0 +1,43 @@
+#
+# Cookbook Name:: postgresql
+# # Recipe:: db_restore
+#
+# Copyright 2013-2016, SUSE LINUX GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+service_name = "postgresql"
+dump_location = node[:crowbar][:upgrade][:db_dump_location]
+
+if node.roles.include?("database-server")
+  if node[:database][:ha][:enabled]
+    # Checks that the service is available, if it's running on this node.
+    service_available = "crm resource show #{service_name} | grep -q \" #{node.hostname} *$\" \
+                        | grep -q 'running'"
+    execute "restore database from #{dump_location}" do
+      command "/usr/lib/postgresql94/bin/psql -d postgres -f #{dump_location}"
+      user "postgres"
+      only_if service_available
+    end
+  else
+    execute "restore database from #{dump_location}" do
+      command "/usr/lib/postgresql94/bin/psql -d postgres -f #{dump_location}"
+      user "postgres"
+    end
+  end
+end
+
+file dump_location do
+  action :delete
+end
--- a/chef/cookbooks/postgresql/recipes/server.rb
+++ b/chef/cookbooks/postgresql/recipes/server.rb
@@ -128,6 +128,10 @@ else
   log "HA support for postgresql is disabled"
 end
 
+db_dump = node[:crowbar].fetch(:upgrade, {})[:db_dump_location]
+if !db_dump.nil? && File.exist?(db_dump)
+  include_recipe "postgresql::db_restore"
+end
 # NOTE: Consider two facts before modifying "assign-postgres-password":
 # (1) Passing the "ALTER ROLE ..." through the psql command only works
 #     if passwordless authorization was configured for local connections.


sayalilunkad:
Not tested yet

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,40 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+dump_location = node[:crowbar][:upgrade][:db_dump_location]
+# Only run the psql commands if the service is running on this node, so that
+# we don't depend on the node running the service to be as fast as this one
+service_name = "postgresql"
+only_if_command = "crm resource show #{service_name} | grep -q \" #{node.hostname} *$\""
```
jsuchome: So this happens only for HA setup?

sayalilunkad: Ah no it should not. Need to update that.

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,40 @@
+#
+# Cookbook Name:: postgresql
+# # Recipe:: db_restore
+#
+# Copyright 2013-2016, SUSE LINUX Products GmbH
```
rsalevsky: `Copyright 2013-2016, SUSE LINUX GmbH`

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  execute "start_db_server" do
+    command "/usr/lib/postgresql94/bin/postgres -D #{node['postgresql']['dir']}"
+  end
+
+  execute "restore_data" do
+    command "/usr/lib/postgresql94/bin/postgres -d postgres -g #{dump_location}"
+  end
+  only_if only_if_command1 || only_if_command2
```
jsuchome: In HA case, the service exists on all nodes in the cluster. So `only_if_command2` would not be executed...

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,41 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+dump_location = node[:crowbar][:upgrade][:db_dump_location]
+# Only run the psql commands if the service is running on this node, so that
+# we don't depend on the node running the service to be as fast as this one
+service_name = "postgresql"
+only_if_command1 = File.exist?("/etc/init.d/#{service_name}")
```
MaximilianMeister: I think we can give some meaningful names like `service_available` and `hostname_in_resource` or sth similar here. I know it has been like that before in `server.rb` but I would change it there as well

chef/cookbooks/postgresql/recipes/db_restore.rb
sayalilunkad: Is there a good way to check if the node is a db node in case of HA?

jsuchome: I think there should be something marked in `node` data, I think `node[:database][:ha][:enabled]`

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,41 @@
...
+only_if_command2 = "crm resource show #{service_name} | grep -q \" #{node.hostname} *$\""
+
+user "postgres" do
+  execute "new_db_cluster" do
+    command "/usr/lib/postgresql94/bin/initdb -D #{node['postgresql']['dir']}"
+  end
+
+  execute "start_db_server" do
```
vuntz: Why do you need to start it? Shouldn't we let the recipe start it?

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,41 @@
...
+
+dump_location = node[:crowbar][:upgrade][:db_dump_location]
+# Only run the psql commands if the service is running on this node, so that
+# we don't depend on the node running the service to be as fast as this one
+service_name = "postgresql"
+only_if_command1 = File.exist?("/etc/init.d/#{service_name}")
+only_if_command2 = "crm resource show #{service_name} | grep -q \" #{node.hostname} *$\""
+
+user "postgres" do
+  execute "new_db_cluster" do
```
vuntz: I think the initdb is already done automatically earlier in the recipe.

chef/cookbooks/postgresql/recipes/server.rb
```diff
@@ -128,35 +128,40 @@
...
-  code <<-EOH
-echo "ALTER ROLE postgres ENCRYPTED PASSWORD '#{node['postgresql']['password']['postgres']}';" | psql -p #{node['postgresql']['config']['port']}
-  EOH
-  # shouldn't try to update the password on the current 'passive' nodes
-  only_if only_if_command if ha_enabled
-  action :run
-end
+dump_location = node[:crowbar][:upgrade][:db_dump_location]
+if File.exist?(dump_location)
+  include_recipe "db_restore"
```
vuntz: This needs to happen only once, so you need to remove the dump, or had a flag to say that it was restored.

chef/cookbooks/postgresql/recipes/server.rb
```diff
@@ -128,35 +128,40 @@
...
+else
```
vuntz: Also, I wouldn't do a else, I would just restore first and then keep the other code.

chef/cookbooks/postgresql/recipes/db_restore.rb
jsuchome: Maybe because we want initial start which allows importing the data? 
@sayalilunkad please describe the actions in comments as part of the code

sayalilunkad: If it is starting the db server before this recipe is run then we don't need the first two step.

chef/cookbooks/postgresql/recipes/server.rb
sayalilunkad: This was discussed with @rhafer ans we thought since we don't yet need a flag which indicated the node had been restored for anything else we can use the presence of the backup file to show if the node was upgraded or a new one.

chef/cookbooks/postgresql/recipes/server.rb
sayalilunkad: But in the case of restoring a node we don't need to execute the next two bash blocks because that should already be done with the restore..

vuntz: It actually doesn't matter if it's executed, except that we make sure that things are configured the way they should be if we execute it.

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,43 @@
...
+dump_location = node[:crowbar][:upgrade][:db_dump_location]
+
+if node[:database][:ha][:enabled]
+  # Checks that the service is running.
+  service_active = "crm resource show #{service_name} 2> /dev/null | grep -q \"is running on\""
+  until ::Kernel.system(service_active)
+    Chef::Log.debug("#{service_name} still not started")
+    sleep(2)
+  end
+  execute "restore_data" do
```
jsuchome: This is same for both cases, right?

sayalilunkad: Mostly yes. Needs to be tested on HA setup.

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,47 @@
...
+
+dump_location = node[:crowbar][:upgrade][:db_dump_location]
+
+if node[:database][:ha][:enabled]
+  # Checks that the service is available, if it's running on this node.
+  service_active = "crm resource show #{service_name} | grep -q \" #{node.hostname} *$\""
+  until ::Kernel.system(service_active)
+    Chef::Log.debug("#{service_name} still not started")
+    sleep(2)
+  end
```
jsuchome: I'm not sure which nodes get this recipe, but if it is run for all nodes in databse cluster, this will end in endless loop if current node is not the one running postgresql. It's actually running on one node only

jsuchome: I think you need to check 
1. If this node is expected to have postgresql running (that might be done by that crm resource check). If you are in the cluster, but on the other node, you probably need to just exit.
2. Now you need to wait for service to be started (if you really expect it to be started by some other place)

jsuchome:
I think this looks good.

jsuchome:
+1 (not tested by me)

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,47 @@
...
+    supports status: true, restart: true, start: true
+    action :start
+  end
+  execute "restore_data" do
+    command "/usr/lib/postgresql94/bin/psql -d postgres -f #{dump_location}"
+    user "postgres"
+  end
+end
+backup_file_exists = File.exist?(dump_location)
+execute "remove_db_dump" do
```
jsuchome: nitpick: could you add a comment explainig why the extra check for file existence is here?

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,47 @@
...
+#
+
+service_name = "postgresql"
+dump_location = node[:crowbar][:upgrade][:db_dump_location]
+if node[:database][:ha][:enabled]
+  # Checks that the service is available, if it's running on this node.
+  service_available = "crm resource show #{service_name} | grep -q \" #{node.hostname} *$\" \
+                      | grep -q 'running'"
+  execute "restore_data" do
+    command "/usr/lib/postgresql94/bin/psql -d postgres -f #{dump_location}"
```
jsuchome: Do we have to use full path? This may hurt us if we'd like to use this code for next upgrade, with possibly different postgresql version.... (I know that's not likely)

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,47 @@
...
+    user "postgres"
+    only_if service_available
+  end
+elsif node[:database][:enabled]
```
vuntz: Where is this attribute coming from? (apologies if it came following a review, I didn't follow the whole discussion :/)

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,47 @@
...
+  # Checks if service is active and will start service if not active.
+  service "postgresql" do
+    supports status: true, restart: true, start: true
+    action :start
+  end
```
vuntz: I'm still confused as to why we start the service here: it should already be started in the server.rb recipe from which we're called.

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,47 @@
...
+    command "/usr/lib/postgresql94/bin/psql -d postgres -f #{dump_location}"
+    user "postgres"
+  end
+end
+backup_file_exists = File.exist?(dump_location)
+execute "remove_db_dump" do
+  command "rm #{dump_location}"
+  user "root"
+  only_if backup_file_exists
+end
```
vuntz: I think we could rewrite the last bit as:
```
file dump_location
  action :delete
end
```

jsuchome: Yep; if it doesn't fail when file does not exist

vuntz: It doesn't; it just makes sure the file is deleted and is happy if it's not present.

chef/cookbooks/postgresql/recipes/db_restore.rb
sayalilunkad: What this part is supposed to do is to check if the postgresql service is running and only in the case it is not it should start it. But please correct me if I am wrong as I am fairly inexperienced with chef.

vuntz: So what you're doing is telling chef "start postgresql unless it's already started", which is slightly different but would meet your requirements.

That however should not be needed because we already do that earlier (not in your newly-added code, but in the old code from server.rb). So you can safely assume it is running.

chef/cookbooks/postgresql/recipes/server.rb
```diff
@@ -128,6 +128,9 @@
   log "HA support for postgresql is disabled"
 end
 
+if File.exist?(node[:crowbar][:upgrade][:db_dump_location])
```
vuntz: Gating fails because (I think) `node[:crowbar][:upgrade]` is `nil`.

@jsuchome I guess that's correct because this is only set when the node is in crowbar_upgrade state, right?

A fix would be something like:
```ruby
db_dump = node[:crowbar].fetch(:upgrade, {})[:db_dump_location]
if !db_dump.nil? && File.exist?(db_dump)
   ...
end
```

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+service_name = "postgresql"
+dump_location = node[:crowbar][:upgrade][:db_dump_location]
+if node.roles.include?("database-server")
```
vuntz: Can you add an empty line before the `if` line?

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,41 @@
...
+      only_if service_available
+    end
+  else
+    execute "restore_data" do
+      command "/usr/lib/postgresql94/bin/psql -d postgres -f #{dump_location}"
+      user "postgres"
+    end
+  end
+end
+file dump_location do
```
vuntz: Can you add an empty line before that line?

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -0,0 +1,41 @@
...
+    # Checks that the service is available, if it's running on this node.
+    service_available = "crm resource show #{service_name} | grep -q \" #{node.hostname} *$\" \
+                         | grep -q 'running'"
+    execute "restore_db" do
+      command "/usr/lib/postgresql94/bin/psql -d postgres -f #{dump_location}"
+      user "postgres"
+      only_if service_available
+    end
+  else
+    execute "restore_data" do
```
vuntz: I'd probably use the same name for the two `execute` resources. `execute "restore database from #{dump_location}" do`?

chef/cookbooks/postgresql/recipes/server.rb
jsuchome: Hm, I don't think so, it should be always there: https://github.com/crowbar/crowbar-core/blob/master/chef/data_bags/crowbar/template-crowbar.json#L23

vuntz: @jsuchome but non-admin nodes are (except for the upgrade case) not part of the crowbar proposal, so won't have these attributes.

jsuchome: Ah, you're right. Your code change makes sense, of course.

vuntz:
Thanks for the changes, hopefully gating will be green now!

jsuchome:
It's green! +1 from me again

vuntz:
+1

PR #215, opened 2016-01-21
openstack: Add flag to configure haproxy & apache for controller HA only

vuntz:
**Note:** this goes hand in hand with https://github.com/crowbar/crowbar-ha/pull/71

Because the pacemaker resources for haproxy and apache are created in
crowbar-pacemaker (from crowbar-ha), we, in theory, shouldn't assume
there that these two services will be used for OpenStack and so we can't
always just add the location constraint to tell pacemaker to only run
these on controller nodes.

So we add a flag for that.


aspiers:
+1

jsuchome:
+1

PR #216, opened 2016-01-21
neutron: Add attribute to disable creation of fixed/floating networks

vuntz:
This is a common request we receive; ideally, we would simply move the
definitions of the networks to the barclamp, and add a way to disable
them there, but in the meantime, this should be good enough.

http://bugzilla.suse.com/show_bug.cgi?id=963007


rhafer:
+1 (you might wanna fix the hound complaints though)

vuntz:
> (you might wanna fix the hound complaints though)

done :-)

aplanas:
+1

rhafer:
Still +1

PR #221, opened 2016-01-22
Fix issue with "libvirtError: Secret not found" for multi rbd backend,  (bsc#963029)

mjura:
This change is fixing issue with "libvirtError: Secret not found: rbd no secret matches uuid"

For more details please go to https://bugzilla.suse.com/show_bug.cgi?id=963029

### DIFF
--- a/chef/cookbooks/nova/recipes/ceph.rb
+++ b/chef/cookbooks/nova/recipes/ceph.rb
@@ -122,9 +122,9 @@ cinder_controller[:cinder][:volumes].each_with_index do |volume, volid|
           undefine = false
 
           if secret_uuid == rbd_uuid
-            undefine = true if secret_usage != "client.#{rbd_user} secret"
+            undefine = true if secret_usage != "crowbar-#{rbd_uuid} secret"
           else
-            undefine = true if secret_usage == "client.#{rbd_user} secret"
+            undefine = true if secret_usage == "crowbar-#{rbd_uuid} secret"
           end
 
           if undefine
@@ -160,7 +160,12 @@ cinder_controller[:cinder][:volumes].each_with_index do |volume, volid|
 
         if secret != client_key
           secret_file_path = "/etc/ceph/ceph-secret-#{rbd_uuid}.xml"
-          secret_file_content = "<secret ephemeral='no' private='no'> <uuid>#{rbd_uuid}</uuid><usage type='ceph'> <name>client.#{rbd_user} secret</name> </usage> </secret>"
+          secret_file_content = "<secret ephemeral='no' private='no'>" \
+                                " <uuid>#{rbd_uuid}</uuid>" \
+                                " <usage type='ceph'>" \
+                                " <name>crowbar-#{rbd_uuid} secret</name>" \
+                                " </usage> " \
+                                "</secret>"
           File.write(secret_file_path, secret_file_content)
 
           cmd = ["virsh", "secret-define", "--file", secret_file_path]


crowbar_framework/app/models/cinder_service.rb
```diff
@@ -222,11 +222,20 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
     # Generate secrets uuid for libvirt rbd backend
     dirty = false
+    ceph = {}
     proposal = Proposal.where(barclamp: "cinder", name: role.inst).first
     role.default_attributes[:cinder][:volumes].each_with_index do |volume, volid|
       next unless volume[:backend_driver] == "rbd"
       if volume[:rbd][:secret_uuid].empty?
-        secret_uuid = `uuidgen`.strip
+
+        if ceph[volume[:rbd][:config_file]] && ceph[volume[:rbd][:config_file]][volume[:rbd][:user]]
```
dirkmueller: why do you involve the config_file in the lookup at all? the bugreport says that the <name> has to be unique, and the name is created this way:
``` ruby
<name>client.#{rbd_user} secret</name> 
```
which means you should group by rbd_user only.

mjura: `ceph[volume[:rbd][:config_file]][volume[:rbd][:user]]` - this is because cinder can be used with many ceph clusters and `volume[:rbd][:config_file]` is the way how we differentiate this. But yes, I agree that single compute node can be used  with different Ceph clusters which have this same user name. I will change this. Thank you

crowbar_framework/app/models/cinder_service.rb
```diff
@@ -222,11 +222,19 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
       if volume[:rbd][:secret_uuid].empty?
-        secret_uuid = `uuidgen`.strip
+
+        if ceph_clients[volume[:rbd][:user]]
+          secret_uuid = ceph_clients[volume[:rbd][:user]]
+        else
+          secret_uuid = `uuidgen`.strip
+          ceph_clients[volume[:rbd][:user]] = secret_uuid
+        end
+
```
dirkmueller: I prefer 

``` ruby

  unless ceph_clients[volume[:rbd][:user]]
    ceph_clients[volume[:rbd][:user]] = `uuidgen`.strip
  end
  secret_uuid = ceph_clients[volume[:rbd][:user]]
```

which makes this slightly more readable imho

dirkmueller:
+1

mjura:
@dirkmueller : I've included your suggestion, thank you

vuntz:
So the issue is that different rbd backends but with the same user are using different UUIDs? Would be nice to put that in the commit message.

However, I still feel there's an issue: you can have the same user but different ceph.conf files in different backends; don't we want different UUIDs in that case?

crowbar_framework/app/models/cinder_service.rb
```diff
@@ -222,11 +222,17 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
     dirty = false
+    ceph_clients = {}
     proposal = Proposal.where(barclamp: "cinder", name: role.inst).first
     role.default_attributes[:cinder][:volumes].each_with_index do |volume, volid|
       next unless volume[:backend_driver] == "rbd"
       if volume[:rbd][:secret_uuid].empty?
-        secret_uuid = `uuidgen`.strip
+
+        unless ceph_clients[volume[:rbd][:user]]
+          ceph_clients[volume[:rbd][:user]] = `uuidgen`.strip
```
aplanas: `%x(uuidgen).strip`?

dirkmueller:
@vuntz that is currently not possible. I agree that this is desireable to get fixed, but that might be a topic for another PR..

crowbar_framework/app/models/cinder_service.rb
mjura: I would prefer to leave as it is, this same system call we have in `swift_service.rb` and `tempest_service.rb`

vuntz:
> @vuntz that is currently not possible.

We have an attribute to specify the path to ceph.conf for each backend (+ the attribute telling us if we're using the crowbar-deployed ceph). What are we missing to make it possible?

mjura:
> We have an attribute to specify the path to ceph.conf for each backend (+ the attribute telling us if we're using the crowbar-deployed ceph). What are we missing to make it possible?

@dirkmueller, @vuntz : No worries, it was my first approach for this PR. I will take of it. I have to only add validation check to `cinder_service.rb`

mjura:
Guys I've updated new version of this, right solution is also including issue with multi Ceph instances.

Please comment on this.

crowbar_framework/app/models/cinder_service.rb
```diff
@@ -126,7 +128,14 @@ def validate_proposal_after_save proposal
       if backend_driver == "rbd"
         rbd_crowbar ||= volume["rbd"]["use_crowbar"]
         rbd_ceph_conf ||= !volume["rbd"]["use_crowbar"] && (volume["rbd"]["config_file"].strip == "/etc/ceph/ceph.conf")
+
+        if rbd_ceph_users[volume["rbd"]["user"]]
+          rbd_virsh_users_check = true if rbd_ceph_users[volume["rbd"]["user"]] != volume["rbd"]["config_file"]
```
vuntz: You can write `rbd_virsh_users_check ||= rbd_ceph_users[volume["rbd"]["user"]] != volume["rbd"]["config_file"]`

And thanks, I now understand what @dirkmueller meant :-)

crowbar_framework/app/models/cinder_service.rb
```diff
@@ -222,11 +235,19 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
     dirty = false
+    rbd_inst = {}
     proposal = Proposal.where(barclamp: "cinder", name: role.inst).first
     role.default_attributes[:cinder][:volumes].each_with_index do |volume, volid|
       next unless volume[:backend_driver] == "rbd"
       if volume[:rbd][:secret_uuid].empty?
-        secret_uuid = `uuidgen`.strip
+
+        if !rbd_inst[volume[:rbd][:config_file]] || !rbd_inst[volume[:rbd][:config_file]][volume[:rbd][:user]]
+          rbd_inst[volume[:rbd][:config_file]] = {} unless rbd_inst[volume[:rbd][:config_file]]
```
vuntz: `rbd_inst[volume[:rbd][:config_file]] ||= {}`

crowbar_framework/config/locales/cinder/en.yml
```diff
@@ -144,3 +144,4 @@ en:
         raw_device_backend: 'There cannot be multiple raw devices backends when one raw device backend is configured to use all disks.'
         missing_unclaimed_disk: 'Nodes %{nodes_without_suitable_drives} for cinder volume role are missing at least one unclaimed disk, required when using raw devices.'
         rados_backends: 'RADOS backends not deployed with Crowbar cannot use /etc/ceph/ceph.conf as configuration files when also using the RADOS backend deployed with Crowbar.'
+        rbd_virsh_users: 'Because of libvirt limitation that you can define one secret UUID per each user name, please change user names for seperate instances of RADOS backends.'
```
vuntz: I suggest: `Due to a limitation in the virtualization stack, each RADOS backend must be using a different user.`

The details (about the UUID) can be added as a comment to your new validation check.

(username would be better than user, but then you need to change the string for the barclamp UI to use username too ;-))

crowbar_framework/app/models/cinder_service.rb
mjura: It was changed

crowbar_framework/config/locales/cinder/en.yml
mjura: Thank you, I've changed also username in UI.

mjura:
I've adopted once again comments from @vuntz

vuntz:
+1

crowbar_framework/app/models/cinder_service.rb
```diff
@@ -126,7 +128,17 @@ def validate_proposal_after_save proposal
       if backend_driver == "rbd"
         rbd_crowbar ||= volume["rbd"]["use_crowbar"]
         rbd_ceph_conf ||= !volume["rbd"]["use_crowbar"] && (volume["rbd"]["config_file"].strip == "/etc/ceph/ceph.conf")
+
+        # Because of the limitation in libvirt that you can define one secret UUID per each
+        # username, we have to check and prevent this scenario when we define multiply RADOS
+        # instances.
```
dirkmueller: I don't think this is true.. the issue is caused by the way you define the secret. you can perfectly well submit multiple secrets for the same username, only the "name" attribute needs to be unique. I think it would make more sense if you'd define the name attribute to be crowbar-&lt;uuid&gt; instead, which avoids the whole mess

dirkmueller:
@vuntz we're missing fixes in the way the virt secrets are getting defined..

crowbar_framework/config/locales/cinder/en.yml
```diff
@@ -144,3 +144,4 @@ en:
         raw_device_backend: 'There cannot be multiple raw devices backends when one raw device backend is configured to use all disks.'
         missing_unclaimed_disk: 'Nodes %{nodes_without_suitable_drives} for cinder volume role are missing at least one unclaimed disk, required when using raw devices.'
         rados_backends: 'RADOS backends not deployed with Crowbar cannot use /etc/ceph/ceph.conf as configuration files when also using the RADOS backend deployed with Crowbar.'
+        rbd_virsh_users: 'Due to a limitation in the virtualization stack, each RADOS backend must be using a different username.'
```
dirkmueller: strong -1 when we give completely inaccurate advise to customers. this looks like we're fools.

crowbar_framework/app/models/cinder_service.rb
mjura: Ok, I'm sure if the username has to be this same as Ceph client user, but I will check it. How about <username>-<md5sum(UUID)> this pattern will be shorter and at least we can still see username from crowbar UI

mjura:
@dirkmueller : Thanks for a comments, I will rework this solution once again.

dirkmueller:
+1

dirkmueller:
@vuntz still ok?

vuntz:
-1: you need to update that code:

```ruby
          if secret_uuid == rbd_uuid
            undefine = true if secret_usage != "client.#{rbd_user} secret"
          else
            undefine = true if secret_usage == "client.#{rbd_user} secret"
          end
```

vuntz:
+1

chef/cookbooks/nova/recipes/ceph.rb
```diff
@@ -122,9 +122,9 @@
           undefine = false
 
           if secret_uuid == rbd_uuid
-            undefine = true if secret_usage != "client.#{rbd_user} secret"
+            undefine = true if secret_usage != "crowbar-#{rbd_uuid} secret"
           else
-            undefine = true if secret_usage == "client.#{rbd_user} secret"
+            undefine = true if secret_usage == "crowbar-#{rbd_uuid} secret"
```
mjura: I would prefer to keep as it is, looks better and it is easier to read

PR #222, opened 2016-01-25
tempest: Use correct endpoint type for manila

toabctl:
There is no "internal_URL". It's "internalURL".

This fixes:

ConfigFileValueError: Value for option endpoint_type is not valid


aplanas:
+1

mjura:
+1

toabctl:
Please merge because it's not tested by the gate and an obvious typo

PR #223, opened 2016-01-25
tempest: Update [share] options in tempest.conf.erb

toabctl:
The current list of available options is incomplete. Add the
missing options and set some sane defaults.


dirkmueller:
+1 (assuming gating passes)

vuntz:
+1

PR #224, opened 2016-01-25
Refactor optional repo check

rsalevsky:
This change makes it easier to handle all the diffrent cases in the UI.


dirkmueller:
@rsalevsky typo optional in commit message

jsuchome:
Good in general, but the functions can return false in 2 different cases. How will user know the difference?

rsalevsky:
@jsuchome Fixed, the `if ... else ... end` was anyway not needed.

jsuchome:
+1

vuntz:
+1

PR #225, opened 2016-01-25
Removes public endpoint from keystone conf (bsc#954957)

sayalilunkad:
Removing the public endpoint fixes the issue of mismatched endpoints.


dirkmueller:
I'd appreciate if you'd update the config file template as well. I'm btw don't undersand the explanation on why this is a good change.

sayalilunkad:
@dirkmueller  maybe this will explain better as to why we need this change https://bugs.launchpad.net/mos/+bug/1508489

dirkmueller:
@sayalilunkad thanks for the link, please add the explanation to the git commit message

mjura:
+1

dirkmueller:
+1

rhafer:
I must admit I am not fully getting it. But AFAIU this change is ok. 

+ 0.99

PR #227, opened 2016-01-26
neutron: Disable l2pop mechanism unless DVR is enabled

vuntz:
We see quite some failures in the HA gating that seem to be related to
the use of l2pop. Until this is fixed, let's just not use l2pop for now,
unless we need it for DVR.

cc @rossella  @abel-navarro


rossella:
+1 (maybe line length needs to be fixed) thanks @vuntz !

abel-navarro:
I didn't know it is only for DVR. If that's the case, then +1.

vuntz:
> I didn't know it is only for DVR.

It's required for DVR; it's still useful outside of DVR, but optional.

dirkmueller:
+1

PR #228, opened 2016-01-26
tempest: Set password in tempest.conf

toabctl:
Use the correct variable name in the template.


vuntz:
+1

mjura:
It is correct fix, I've checked it in `chef/cookbooks/tempest/recipes/config.rb`
+1

toabctl:
Testing failure unrelated. Tempest passed.

PR #229, opened 2016-01-26
cinder: Adjust Fujitsu Eternus config

toabctl:
For Liberty, the driver (which is now out-of-tree) changed the
attribute names of the configuration file.


vuntz:
+1

dirkmueller:
+1

PR #232, opened 2016-01-27
trove: Add insecure options for openstack services in trove.conf

vuntz:
This depends on upstream trove patch that we need to backport:
https://review.openstack.org/#/c/272686/


aplanas:
+1

jsuchome:
Looks good already. +1

PR #236, opened 2016-01-28
Start Agents before Neutron-Server

dirkmueller:
There is known issue with Neutron right now that when we allocate
a new network, it tries to bind the port for the Neutron router
right away. However, if at that point the agent isn't running yet,
the port will go into failed state and not be retried and the
neutron network is broken then.

So lets try to start the agents before neutron-server so that
we can bind the ports on the host.

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -189,14 +189,12 @@ if use_l3_agent
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
 
+  ha_tool_ordering_deps = "postgresql rabbitmq g-haproxy #{agents_clone_name} cl-neutron-server"
+
   crowbar_pacemaker_order_only_existing "o-#{ha_tool_primitive_name}" do
-    # While neutron-ha-tool technically doesn't directly depend on postgresql or
-    # rabbitmq, if these bits are not running, then neutron-server can run but
-    # can't do what it's being asked. Note that neutron-server does have a
-    # constraint on these services, but it's optional, not mandatory (because it
-    # doesn't need to be restarted when postgresql or rabbitmq are restarted).
-    # So explicitly depend on postgresql and rabbitmq (if they are in the cluster).
-    ordering "( postgresql rabbitmq g-haproxy cl-neutron-server #{agents_clone_name} ) #{ha_tool_primitive_name}"
+    # While neutron-ha-tool technically doesn't directly depend on neutron-server,
+    # if these bits are not running, then it can't do what it's being asked.
+    ordering "( #{ha_tool_ordering_deps} ) #{ha_tool_primitive_name}"
     score "Mandatory"
     action :create
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
--- a/chef/cookbooks/neutron/recipes/server_ha.rb
+++ b/chef/cookbooks/neutron/recipes/server_ha.rb
@@ -60,8 +60,13 @@ pacemaker_transaction "neutron server" do
 end
 
 crowbar_pacemaker_order_only_existing "o-#{clone_name}" do
-  ordering ["postgresql", "rabbitmq", "cl-keystone", clone_name]
-  score "Optional"
+  # While technically the agents depend on the neutron-server being up,
+  # the neutron-server also is trying to communicate with the agents and when those
+  # are down, the port is set to vif_binding failed state and left DOWN
+  # Since neutron never retries, the broken ports are more disturbing the
+  # functionality
+  ordering "( postgresql rabbitmq g-haproxy cl-keystone cl-g-neutron-agents ) cl-neutron-server"
+  score "Mandatory"
   action :create
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
--- a/chef/data_bags/crowbar/template-neutron.json
+++ b/chef/data_bags/crowbar/template-neutron.json
@@ -122,12 +122,12 @@
       },
       "elements": {},
       "element_order": [
-         ["neutron-server" ],
-         ["neutron-network" ]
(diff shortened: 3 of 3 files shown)


chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -189,14 +189,22 @@
...
+    # While technically the agents depend on the neutron-server being up,
+    # the neutron-server also is trying to communicate with the agents and when those
+    # are down, the port is set to vif_binding failed state and left DOWN
+    # Since neutron never retries, the broken ports are more disturbing the
+    # functionality
+    ordering "( postgresql rabbitmq g-haproxy #{agents_clone_name} ) cl-neutron-server"
+    score "Mandatory"
+    action :create
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
```
vuntz: Hrm, that bit would need to be in the server_ha recipe, imho: the last element (cl-neutron-server here) is a hint of where this lives, since that constraint should live where that last element lives.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -189,14 +189,22 @@
...
-    # While neutron-ha-tool technically doesn't directly depend on postgresql or
-    # rabbitmq, if these bits are not running, then neutron-server can run but
-    # can't do what it's being asked. Note that neutron-server does have a
-    # constraint on these services, but it's optional, not mandatory (because it
-    # doesn't need to be restarted when postgresql or rabbitmq are restarted).
-    # So explicitly depend on postgresql and rabbitmq (if they are in the cluster).
-    ordering "( postgresql rabbitmq g-haproxy cl-neutron-server #{agents_clone_name} ) #{ha_tool_primitive_name}"
+    # While neutron-ha-tool technically doesn't directly depend on neutron-server,
+    # if these bits are not running, then it can't do what it's being asked.
+    ordering "cl-neutron-server #{ha_tool_primitive_name}"
```
vuntz: That's wrong: you might want to remove agents_clone_name because of your other change, but in case neutron-server lives on another cluster, we still need all the other elements listed here.

vuntz:
Thanks for addressing the comments.

I'm not very comfortable with the change, but mostly because I'm not sure if it's really correct :-) Let's see what gating thinks (might be worth starting some HA gating also)

dirkmueller:
@vuntz well, its a workaround for a neutron bug.. in theory the starting order shouldn't matter. it seems 3 out of the 5 qa scenarios are affected by this problem.

aplanas:
gating fails in `wait_for 1000 1 "ping -q -c 1 -w 1 $vmip >/dev/null" "testvm booted and ping returned"`, but tempest is OK, so I guess that is OK for this commit?

aplanas:
Oh, there is an older job[1] for this PR that fails in the same point. So maybe is not so OK

[1] https://ci.suse.de/view/Cloud/view/Trigger/job/openstack-mkcloud/15755/consoleText

abel-navarro:
@aplanas this was investigated in https://trello.com/c/30w2SCA0. I try to freeze that environment and look at it again.

vuntz:
@abel-navarro but I thought @dirkmueller's fix was intended to fix that issue. Or maybe I'm confusing things with another issue?

aplanas:
Note: if this PR is related with the 'binding for port %s failed..' complain, this is also happening in C5[1], so maybe needs to be backported?

[1] https://ci.suse.de/view/Cloud/view/Trigger/job/openstack-mkcloud/15803/

toabctl:
retriggered build with HA

toabctl:
Hm. Not what exaclty the error is, but tempest passed with HA.

abel-navarro:
@toabctl as we were mentioning in #279 this is due to the [ping returned](https://trello.com/c/30w2SCA0) bug. Hopefully that other PR fixed that problem.

vuntz:
I'm pretty confident that we indeed need at least the other PR for the HA case.

TBH, I'm not really sold on this change here because there could still be scenarios where this doesn't work fine (deploying the roles on two different nodes, and deployment on the network role failed, for instance).

I feel that @rossella's idea of creating the network only once the agent have been registered might be needed too (maybe on top of this PR, not instead of this PR).

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -122,12 +122,12 @@
...
-         ["neutron-server" ],
-         ["neutron-network" ]
+         ["neutron-network" ],
+         ["neutron-server" ]
       ],
       "element_run_list_order": {
-        "neutron-server": 94,
-        "neutron-network": 95
+        "neutron-network": 94,
+        "neutron-server": 95
```
vuntz: We need a migration for this, I guess.

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -60,8 +60,13 @@
...
 crowbar_pacemaker_order_only_existing "o-#{clone_name}" do
-  ordering ["postgresql", "rabbitmq", "cl-keystone", clone_name]
-  score "Optional"
+  # While technically the agents depend on the neutron-server being up,
+  # the neutron-server also is trying to communicate with the agents and when those
+  # are down, the port is set to vif_binding failed state and left DOWN
+  # Since neutron never retries, the broken ports are more disturbing the
+  # functionality
+  ordering "( postgresql rabbitmq g-haproxy cl-keystone cl-g-neutron-agents ) cl-neutron-server"
+  score "Mandatory"
```
vuntz: My understanding is that we're only trying to fix the initialization issue, so I don't think we need the changes to the pacemaker constraints. That wouldn't help us for the non-HA case when nodes are rebooting, for instance.

For this, we should rely on the "retry binding" upstream patch.

dirkmueller:
dropping this for now

PR #237, opened 2016-01-28
Update VNCBLOCK to close rejected connections.

dguitarbite:
Illegal connections to VNC server on compute nodes which were blocked
left an open process on compute node leaving possibilities for
potential DDOS.

### DIFF
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -390,7 +390,7 @@ bash "nova_compute_vncblock_reject_all" do
        -m u32 --u32 "0>>22&0x3C@ 12>>26&0x3C@ 0=0x52464220" \
        -m string --algo kmp --string "RFB 003." --to 130 \
        -j VNCBLOCK
-    iptables -I VNCBLOCK -p tcp -j REJECT --reject-with tcp-reset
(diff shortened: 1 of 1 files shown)


aplanas:
is there any bugreport to reference here?

vuntz:
+1

dguitarbite:
I got the fix before the big report :)

aplanas:
/DDOS/DDoS/

+1

dirkmueller:
s,DDos,DoS,

dguitarbite:
DDOS (Das Denial of Service) looked so kool! DOS sounds like windows DOS ;)

aplanas:
Is DoS, not DOS, but anyway: +1

dirkmueller:
@dguitarbite please update teh commit message. 

does this really fix the problem? according to documentation it does send back an ICMP unreachable to the *source*, documentation doesn't say anything about destination.

dguitarbite:
@dirkmueller It does solve the problem. That is ironic but removing "tcp" flags in there closes the connection after a couple of seconds.

dguitarbite:
I think we should get this PR merged. There is no reason for it to be open.

dirkmueller:
+1

PR #238, opened 2016-01-28
Upgrade: Fix db_synced reset

rhafer:
The node[barclamp] attribute might not be there. E.g. if a specific components is not
deployed.


jsuchome:
+1

vuntz:
+1

PR #241, opened 2016-01-29
ceilometer: Fix broken migration

vuntz:
There's only one element that is central-agent/polling, and we were
assuming that find would return this element and only this one, but find
returns an array.


jsuchome:
+1

aplanas:
+1

PR #242, opened 2016-01-30
swift: Fix chef crash on storage nodes when proxy is HA (bsc#964277)

vuntz:
Due to how we assign roles on nodes in crowbar (through the
crowbar-$node role), it can happen that a node is assigned to a role,
but the node doesn't have the role yet (this happens only when
chef-client has run on the node). This is exactly what we hit here.

The first batch of the application of a swift proposal is for storage
nodes. So there's no swift-proxy node yet (unless a storage node is also
a proxy node). However, we search for a proxy node, which unfortunately
doesn't work; but we were using get_instance() for the search, and this
helper is trying to be nice by ensuring we always get a valid node by
falling back to the current node if it doesn't find anything. This gets
nasty when we use HA for swift-proxy because we then try to get the
virtual hostname of the haproxy service, but that just doesn't work
because the node we're using for this is not part of cluster. And so we
crash.

So simply don't use get_instance(), and use a search helper that can
tell us that there's no result yet. Which is fine: we just don't create
a config file yet, and because swift-storage is running in a second
batch later on, the config file will be created in that second batch.

https://bugzilla.suse.com/show_bug.cgi?id=964277

### DIFF
--- a/chef/cookbooks/swift/recipes/default.rb
+++ b/chef/cookbooks/swift/recipes/default.rb
@@ -34,26 +34,30 @@ template "/etc/swift/swift.conf" do
  })
 end
 
-ha_enabled = node[:swift][:ha][:enabled]
-ssl_enabled = node[:swift][:ssl][:enabled]
-swift_protocol = ssl_enabled ? "https" : "http"
-proxy_node = get_instance("roles:swift-proxy")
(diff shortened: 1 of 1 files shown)


aplanas:
+1

dirkmueller:
+1

PR #245, opened 2016-02-01
Cleanup cinder liberty conf rebase

dirkmueller:
This was added to the wrong place, moving it to the right place.

### DIFF
--- a/chef/cookbooks/cinder/templates/default/cinder.conf.erb
+++ b/chef/cookbooks/cinder/templates/default/cinder.conf.erb
@@ -595,11 +595,6 @@ strict_ssh_host_key_policy = <%= @strict_ssh_host_key_policy ? 'true' : 'false' 
 [<%= backend_id %>]
 <% end -%>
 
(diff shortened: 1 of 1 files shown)


chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -811,6 +806,9 @@ iscsi_write_cache = off
 # content. The cinder backend and locations support must be enabled in the
 # image service, and glance_api_version must be set to 2. (boolean value)
 #image_upload_use_cinder_backend = false
+<% if @show_storage_location -%>
+image_upload_use_cinder_backend = false
```
aplanas: the original logic was to put `true` here

vuntz: s/false/true/ ?

mjura:
+1

aplanas:
+1

toabctl:
+1

PR #246, opened 2016-02-01
manila: Fix default proposal for generic backend

vuntz:
The default proposal has a generic backend with no password nor key;
however, the UI makes you believe that a password is set (it actually
displays "undefined").

To have this work, we need to change the validation to not check that
there's one attribute between password and key, but also to check that
there's one non-empty attribute.

cc @toabctl

### DIFF
--- a/chef/cookbooks/manila/templates/default/manila.conf.erb
+++ b/chef/cookbooks/manila/templates/default/manila.conf.erb
@@ -2061,10 +2061,10 @@ share_driver=manila.share.drivers.generic.GenericShareDriver
 share_backend_name=<%= share['backend_name'] %>
 driver_handles_share_servers=False
 service_instance_user=<%= share['generic']['service_instance_user'] %>
-<% if share['generic'].key?('service_instance_password') -%>
+<% if share['generic'].key?('service_instance_password') && !share['generic']['service_instance_password'].empty? -%>
 service_instance_password=<%= share['generic']['service_instance_password'] %>
 <% end -%>
-<% if share['generic'].key?('path_to_private_key') -%>
+<% if share['generic'].key?('path_to_private_key') && !share['generic']['path_to_private_key'].empty? -%>
 path_to_private_key=<%= share['generic']['path_to_private_key'] %>
 <% end -%>
 service_instance_name_or_id=<%= share['generic']['service_instance_name_or_id'] %>
--- a/chef/cookbooks/manila/templates/default/manila.conf.erb
+++ b/chef/cookbooks/manila/templates/default/manila.conf.erb
@@ -2064,7 +2064,7 @@ service_instance_user=<%= share['generic']['service_instance_user'] %>
 <% unless share['generic']['service_instance_password'].empty? -%>
 service_instance_password=<%= share['generic']['service_instance_password'] %>
 <% end -%>
-<% if share['generic'].key?('path_to_private_key') && !share['generic']['path_to_private_key'].empty? -%>
+<% unless share['generic']['path_to_private_key'].empty? -%>
 path_to_private_key=<%= share['generic']['path_to_private_key'] %>
 <% end -%>
 service_instance_name_or_id=<%= share['generic']['service_instance_name_or_id'] %>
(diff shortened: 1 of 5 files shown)


chef/data_bags/crowbar/template-manila.json
```diff
@@ -64,9 +64,12 @@
           "backend_name": "default",
           "generic": {
             "service_instance_user": "root",
+            "service_instance_password": "",
+            "path_to_private_key": "",
```
toabctl: Why are you adding theses keys? They are not required. And if they are available, the keys are set in the template. See i.e. https://github.com/vuntz/crowbar-openstack/blob/manila-default-password/chef/cookbooks/manila/templates/default/manila.conf.erb#L2064

vuntz: > Why are you adding theses keys?

That's what I tried to explain in the commit message :-) Can you tell me what is not clear in it so I can improve it?



> And if they are available, the keys are set in the template.

I assume setting these to an empty string actually doesn't do anything compared to the default. Or am I mis-assuming?

toabctl: > I assume setting these to an empty string actually doesn't do anything compared to the default. Or am I mis-assuming?

```""``` (empty string) is different to ```None``` which is the default if you don't set it. But maybe the code only does a bool compare and then it doesn't matter. But I would prefer to not set it (so it's value is ```None```) if we don't need it.

Comment (toabctl) [inline chef/data_bags/crowbar/template-manila.json]: I'm not sure howto handle such situation in the UI. Should we remove the password field from the UI?

Comment (vuntz) [inline chef/data_bags/crowbar/template-manila.json]: Would it work for you to not set the options in the config file if the values are empty?

Comment (toabctl) [inline chef/data_bags/crowbar/template-manila.json]: Fine for me.

Comment (vuntz) [inline chef/data_bags/crowbar/template-manila.json]: @toabctl done

Comment (vuntz) [inline chef/data_bags/crowbar/template-manila.json]: Btw, this is actually required in case the user deletes and recreates the generic backend since the defaults when creating the backend will have (empty) attributes.

Comment (toabctl) [inline chef/data_bags/crowbar/template-manila.json]: @vuntz one more thing - is there now any reason to have the fields as ```required: false``` or should we mark all as ```required: true``` now?

Comment (vuntz) [inline chef/data_bags/crowbar/template-manila.json]: Hrm, so I'd say that if they are displayed in the webui, they should be required. See new commit in this PR for this.

File: chef/cookbooks/manila/templates/default/manila.conf.erb [hunk 2]
```diff
@@ -2061,10 +2061,10 @@ share_driver=manila.share.drivers.generic.GenericShareDriver
 share_backend_name=<%= share['backend_name'] %>
 driver_handles_share_servers=False
 service_instance_user=<%= share['generic']['service_instance_user'] %>
-<% if share['generic'].key?('service_instance_password') -%>
+<% if share['generic'].key?('service_instance_password') && !share['generic']['service_instance_password'].empty? -%>
```
toabctl: There's no need to check for ```.key?``` because now the key is mandatory...

vuntz: Fixed.

chef/data_bags/crowbar/template-manila.schema
```diff
@@ -85,7 +85,7 @@
                   "generic": {
                     "type": "map", "mapping": {
                       "service_instance_user": { "type": "str", "required": true },
-                      "service_instance_password": { "type": "str", "required": false },
+                      "service_instance_password": { "type": "str", "required": true },
```
toabctl: Another thing - don't we need a migration now to always add the key if it isn't already there? (sorry for the fragmented review style)

dirkmueller: yeah, we need one since it was in cloud5 :/

chef/data_bags/crowbar/template-manila.schema
```diff
@@ -85,7 +85,7 @@
...
                       "path_to_private_key": { "type": "str", "required": false },
```
dirkmueller: want to make this one required as well while working on it?

vuntz: I only changed the other one because it's displayed in the UI so required. @toabctl are you fine making that one required too?

vuntz:
@dirkmueller updated for your comments.

toabctl:
+1

rsalevsky:
+1

PR #247, opened 2016-02-01
Revert "tempest: Temporary disable manila tests"

toabctl:
This reverts commit 2d27de339d69f5a6ea938d38632563e0e3994d76.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -252,9 +252,7 @@ use_ceilometer = $?.success?
 `#{keystonev2} endpoint-get --service database &> /dev/null`
 use_trove = $?.success?
 `#{keystonev2} endpoint-get --service share &> /dev/null`
-# FIXME(toabctl): temporary disable manila tempest tests
-# use_manila = $?.success?
-use_manila = false
+use_manila = $?.success?
 
 # FIXME: should avoid search with no environment in query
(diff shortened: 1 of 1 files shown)


toabctl:
At least locally smoke tests passed for me. Let's see what gating says...

toabctl:
It's broken because of a missing share type. Fix is https://github.com/SUSE-Cloud/automation/pull/819

toabctl:
Ok. So the smoke tests pass with linuxbridge+vlan but fail with openvswitch. That's something I recognized and debugged with @rhafer . So our openvswitch+vlan setup needs to be fixed.

vuntz:
So after some discussion on IRC, I understand that this only works with linuxbridge. If that's correct, then we can't merge yet because that would break gating (which is using ovs)

toabctl:
> So after some discussion on IRC, I understand that this only works with linuxbridge. If that's correct, then we can't merge yet because that would break gating (which is using ovs)

Correct. Tests passed because I changed the job to use linuxbridge+vlan.

toabctl:
Currently blocked by https://github.com/SUSE-Cloud/automation/pull/824

toabctl:
@vuntz now it passed also with openvswitch. Still only smoke tests but if that's ok, this is ready to be merged

vuntz:
+1

mjura:
+1

aplanas:
+1

PR #248, opened 2016-02-01
Database restore fixes

rhafer:


chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -20,21 +20,19 @@
 service_name = "postgresql"
 dump_location = node[:crowbar][:upgrade][:db_dump_location]
 
-if node.roles.include?("database-server")
```
sayalilunkad: We  need this check for non-HA otherwise the else clause will be applied on all the nodes in the non-HA setup.

rhafer: Hm, this recipe is included by the ```postgresql::server``` recipe which will only ever run on nodes that have the "database-server" role. Am I missing something?

jsuchome:
Looks correct. +1

aplanas:
+1

rhafer:
Fixed the hound complaint about line continuation. Otherwise unchanged.

vuntz:
+1

chef/cookbooks/postgresql/recipes/db_restore.rb
sayalilunkad: Ok, then looks good. +1

PR #249, opened 2016-02-02
Keeps postgresql dump even after restore

sayalilunkad:
Does not remove database dump even after restore and creates file to indicate successful restore.

### DIFF
--- a/chef/cookbooks/postgresql/recipes/db_restore.rb
+++ b/chef/cookbooks/postgresql/recipes/db_restore.rb
@@ -36,6 +36,7 @@ else
   end
 end
 
-file dump_location do
-  action :delete
+# Creates file to indicate successful restore of database.
+file "#{dump_location}.restored-ok" do
+  action :create
 end
--- a/chef/cookbooks/postgresql/recipes/server.rb
+++ b/chef/cookbooks/postgresql/recipes/server.rb
@@ -129,7 +129,7 @@ else
 end
 
 db_dump = node["crowbar"].fetch("upgrade", {})["db_dump_location"]
-if !db_dump.nil? && File.exist?(db_dump)
+if !db_dump.nil? && File.exist?(db_dump) && !File.exist?("#{db_dump}.restored-ok")
   include_recipe "postgresql::db_restore"
 end
 # NOTE: Consider two facts before modifying "assign-postgres-password":


rhafer:
Hm, assuming that psql exits with something non-zero when the restore fails (does it?), I think we
don't have to change anything here. chef-client would fail with an exception in that case and never reach the "delete" step. 

Could you try if that works?

sayalilunkad:
@rhafer that's what I thought would happen previously but then we had the card for handling this. Did the card require to handle something else? And yes I will check that.

rhafer:
> that's what I thought would happen previously but then we had the card for handling this. Did the card require to handle something else?

No. Not really. I added the card because I had the dump removed during one of my tests even though the restore was not successful. But I think the issue was something else. It seems fixed with https://github.com/crowbar/crowbar-openstack/pull/248

> And yes I will check that.

Given the discussion in https://github.com/crowbar/crowbar-core/pull/265 though I guess we need to keep the dump around even after restore, though :(. Which means of course that we can't use the existence of the dump file as an indicator of whether we need to restore.

dirkmueller:
@sayalilunkad can you rebase?

sayalilunkad:
@rhafer I think the discussion in https://github.com/crowbar/crowbar-core/pull/265 is only for the data directory not for the dump. But in general I think it's a good idea to add a flag to indicate if we need restore.

sayalilunkad:
also does the I show if the dump was not successful? We might need this in that case.

dirkmueller:
@sayalilunkad could you rebase, please?

chef/cookbooks/postgresql/recipes/server.rb
```diff
@@ -129,7 +129,7 @@
 end
 
 db_dump = node["crowbar"].fetch("upgrade", {})["db_dump_location"]
-if !db_dump.nil? && File.exist?(db_dump)
+if !db_dump.nil? && !File.exist?("/var/lib/pgsql/restore_success")
```
rhafer: Please leave the ```File.exist?(db_dump) ``` check in place as well.

rhafer:
+1

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -36,6 +36,7 @@
   end
 end
 
-file dump_location do
-  action :delete
+# Creates file to indicate successful restore of database.
+file "/var/lib/pgsql/restore_success" do
```
vuntz: Can we use a filename that is based on `db_dump`? Something like "#{db_dump.restored-ok}".

rhafer:
Still +1

aplanas:
+1

PR #251, opened 2016-02-03
Check if ssl cert related files are non-empty

dirkmueller:
Even empty files are not valid certificate information.


vuntz:
+1

And yes, we should have a helper that deals with this instead of copy&past of this ssl-related code block :-)

toabctl:
+1

aplanas:
+1

PR #255, opened 2016-02-04
Fix nova role rename migration

rhafer:
Only add the new role to the "elements" if the old role was already listed
there before. Otherwise the value will be initialized to "nil" and cause
proposal validation failures.


rhafer:
This fix is required for the Upgrade to work.

jsuchome:
+1 (just fix the 'nove' typo in commit message)

rhafer:
> just fix the 'nove' typo in commit message

done

aplanas:
+1

dirkmueller:
+1

PR #258, opened 2016-02-04
Renames old postgres data dir

sayalilunkad:
Before creating new data directory checks if data directory
is present and moves it to another location if it is.

### DIFF
--- a/chef/cookbooks/postgresql/recipes/server_redhat.rb
+++ b/chef/cookbooks/postgresql/recipes/server_redhat.rb
@@ -76,6 +76,19 @@ template "#{node['postgresql']['sysconfig']}" do
   notifies :restart, "service[postgresql]", :delayed
 end
 
+# Renames any old data directory if present.
+backup = "#{node.postgresql.dir}.upgrade-backup"
+execute "Move old data directory" do
+  command "mv '#{node.postgresql.dir}' '#{backup}'"
+  only_if do
+    if node[:platform_family] == "suse" && File.exist?("#{node.postgresql.dir}/PG_VERSION")
+      File.foreach("#{node.postgresql.dir}/PG_VERSION").grep(/^9.1$/).any?
+    else
+      false
+    end
+  end
+end
+
 # We need initdb to populate /var/lib/pgsql/data before we generate the config
 # files (otherwise, later calls to initdb don't do anything and postgresql
 # doesn't want to start).


jsuchome:
As we have platform checks in this file, could you add a check for suse as well?

jsuchome:
Also, is there any way to make sure that this code is run only once?

sayalilunkad:
@jsuchome added a  condition to do that.

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,16 @@
   notifies :restart, "service[postgresql]", :delayed
 end
 
+# Renames any old data directory if present.
+old_dir = "var/lib/pgsql/data.old"
+execute "Move old data directory" do
+  command "mv '#{node.postgresql.dir}' '#{old_dir}'"
+  only_if do
+    node[:platform_family] == "suse" && ::File.directory?(node.postgresql.dir) \
+        && !::File.directory?(old_dir)
```
jsuchome: I know I'm being evil, but what if user decides that such backup is no longer needed and deletes it?

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,17 @@
...
 end
 
+# Renames any old data directory if present.
+old_dir = "var/lib/psql/data.old"
+execute "Move old data directory" do
+  command "mv '#{node.postgresql.dir}' '#{old_dir}'"
+  creates "/var/lib/psql/old_data_moved"
+  only_if do
+    node[:platform_family] == "suse" && ::File.directory?(node.postgresql.dir) \
+        && !::File.exist?("/var/lib/psql/old_data_moved")
```
rhafer: Hm, so now the we can ever remove "old_data_moved"? How about checking the "VERSION" file inside the data directory? If it has 9.1 in it move the dir, otherwise don't touch it.

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,14 @@
   notifies :restart, "service[postgresql]", :delayed
 end
 
+# Renames any old data directory if present.
+old_dir = "var/lib/pgsql/data.old"
+check_pg_version = File.readlines("#{node.postgresql.dir}/PG_VERSION").grep(/9.4/).size
```
rhafer: Can we reverse that check and look for 9.1? Only in case we decide to upgrade to some newer release at some point.

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,14 @@
   notifies :restart, "service[postgresql]", :delayed
 end
 
+# Renames any old data directory if present.
+old_dir = "var/lib/pgsql/data.old"
```
vuntz: Can you change this to `backup = "#{node.postgresql.dir}.upgrade-backup"`?

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,14 @@
...
+check_pg_version = File.readlines("#{node.postgresql.dir}/PG_VERSION").grep(/9.1/).size
```
vuntz: What about: `is_91 = File.readlines("#{node.postgresql.dir}/PG_VERSION").grep(/9.1/).size > 0`?

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,14 @@
   notifies :restart, "service[postgresql]", :delayed
 end
 
+# Renames any old data directory if present.
+backup = "#{node.postgresql.dir}.upgrade-backup"
+old_version = File.readlines("#{node.postgresql.dir}/PG_VERSION").grep(/9.1/).size > 0 ?
```
aplanas: I have some misunderstanding here:

* I feel confusing that this code for SUSE is in server_redhat
* I am not sure that is a good idea to have a fixed version in the code (9.1), this makes the code not reusable

sayalilunkad: There is also some code for SUSE in other parts in this file. Specifically to run the postgresql initdb. I guess we use this file for the rpm family. And you are right about the fixed version but in case we use it later for some other upgrade, this code can be updated to add more checks. I don't know if there is a better way to do this as we need to do this before we run the initdb call in postgres and it cased trouble when we did it before upgrading.

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,13 @@
   notifies :restart, "service[postgresql]", :delayed
 end
 
+# Renames any old data directory if present.
+backup = "#{node.postgresql.dir}.upgrade-backup"
+old_version = File.readlines("#{node.postgresql.dir}/PG_VERSION").grep(/9.1/).size > 0
```
rhafer: Hrm, this make the check be executed at the very beginning of the chef-client run (the so-called compile phase). At that point the data directory might not be mounted.

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,13 @@
...
+execute "Move old data directory" do
+  command "mv '#{node.postgresql.dir}' '#{backup}'"
+  only_if { node[:platform_family] == "suse" && old_version }
```
rhafer: Rework this to include the actual code from line 81 to have it executed in the "execute" of chef-client. Like this:

```
only_if do 
  if node[:platform_family] == "suse" && "check if file exisits here"
    File.readlines("#{node.postgresql.dir}/PG_VERSION").grep(/9.1/).size > 0
  end
end
```

chef/cookbooks/postgresql/recipes/server_redhat.rb
rhafer: This might help to get and idea on the different phase (compile vs. execute) of a chef-client run.
https://github.com/jhotta/chef-fundamentals-ja/blob/master/slides/anatomy-of-a-chef-run/01_slide.md

aplanas: Ahh good resource, thanks!!

jsuchome:
This seems to work well. +1

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,16 @@
   notifies :restart, "service[postgresql]", :delayed
 end
 
+# Renames any old data directory if present.
+backup = "#{node.postgresql.dir}.upgrade-backup"
+execute "Move old data directory" do
+  command "mv '#{node.postgresql.dir}' '#{backup}'"
+  only_if do
+    if node[:platform_family] == "suse" && File.exist?(node.postgresql.dir)
```
rhafer: Don't you need to check for "#{node.postgresql.dir}/PG_VERSION" here?

sayalilunkad: Well if the data directory is present then the PG_VERSION will be there so I don't think it would be necessary unless you think it is.

vuntz: I'd feel more comfortable changing the check the way @rhafer mentioned

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,16 @@
...
+      File.readlines("#{node.postgresql.dir}/PG_VERSION").grep(/9.1/).size > 0
```
vuntz: It might be just me, but having a `else` which returns false would be less ambiguous.

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,18 @@
   notifies :restart, "service[postgresql]", :delayed
 end
 
+# Renames any old data directory if present.
+backup = "#{node.postgresql.dir}.upgrade-backup"
+execute "Move old data directory" do
+  command "mv '#{node.postgresql.dir}' '#{backup}'"
+  only_if do
+    if node[:platform_family] == "suse" && File.exist?("#{node.postgresql.dir}/PG_VERSION")
+      File.readlines("#{node.postgresql.dir}/PG_VERSION").grep(/9.1/).size > 0
```
rsalevsky: Please use http://ruby-doc.org/core-2.2.0/File.html#method-c-size-3F

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,18 @@
...
+    else
+      return False
```
rsalevsky: `false`

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,18 @@
...
+    end
+  end
+end
```
rsalevsky: Linebreak missing

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,18 @@
   notifies :restart, "service[postgresql]", :delayed
 end
 
+# Renames any old data directory if present.
+backup = "#{node.postgresql.dir}.upgrade-backup"
+execute "Move old data directory" do
+  command "mv '#{node.postgresql.dir}' '#{backup}'"
```
rsalevsky: Can you do this with ruby?

sayalilunkad: Yes it has been done before. Eg: https://github.com/SUSE-Cloud/crowbar-openstack/blob/master/chef/cookbooks/mysql/recipes/server_ec2.rb#L27

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,19 @@
...
+  only_if do
+    if node[:platform_family] == "suse" && File.exist?("#{node.postgresql.dir}/PG_VERSION")
+      File.foreach("#{node.postgresql.dir}/PG_VERSION").grep(/9.1/).any?
```
rhafer: One last thing. Could you please put anchors to the pattern? ```/^9.1$/``` just to be on the safe side.

rhafer:
+1

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -76,6 +76,19 @@
...
 
+# Renames any old data directory if present.
+backup = "#{node.postgresql.dir}.upgrade-backup"
+execute "Move old data directory" do
+  command "mv '#{node.postgresql.dir}' '#{backup}'"
+  only_if do
+    if node[:platform_family] == "suse" && File.exist?("#{node.postgresql.dir}/PG_VERSION")
+      File.foreach("#{node.postgresql.dir}/PG_VERSION").grep(/^9.1$/).any?
+    else
+      return false
```
vuntz: You don't need `return`, here, just `false`.

chef/cookbooks/postgresql/recipes/server_redhat.rb
rsalevsky: No, I mean http://ruby-doc.org/stdlib-2.2.2/libdoc/fileutils/rdoc/FileUtils.html#method-c-mv.

vuntz: @rsalevsky I'm nearly sure we don't want to do that as we don't want to run this in the compile phase of chef, but in the convergence phase.

rsalevsky:
After removing the `return` +1

aplanas:
+1

PR #259, opened 2016-02-05
Additional fix for rename_nova_roles migration

rhafer:
When upgrading from tex the nodes do not have the nova element roles assigned.
Because they are in the "crowbar_upgarde" state.  Use the run_list_map
attribute as a search criteria for the nodes instead.


dirkmueller:
LGTM, but no idea if that works in all cases

jsuchome:
+1

PR #260, opened 2016-02-05
Fix auth_version value for tempest

dirkmueller:
This should be in sync with how all the other cookbooks define it.


rsalevsky:
+1

aplanas:
Tons of tempest fails, maybe is related??

PR #262, opened 2016-02-05
keystone: Fix middleware auth_version

toabctl:
Use "v3" instead of "v3.0" for auth_version from [keystone_authtoken].
Otherwise the middleware has a fallback to v2.0.


toabctl:
I should read all mails before creating PRs ...

PR #263, opened 2016-02-06
Swift report

toabctl:


toabctl:
@tboerger @rsalevsky @MaximilianMeister does that look ok? It shows for me at least the reports under /swift/dashboard (report generation is still broken. I guess /etc/swift/dispersion.conf is not configured correctly)

dirkmueller:
+1

aplanas:
+1

PR #264, opened 2016-02-07
nova: Sort fencing_topology used for compute HA

vuntz:
This just makes it more readable, and will be better for avoiding random
changes if we move to a LWRP later on.


toabctl:
+1

rsalevsky:
+1

PR #265, opened 2016-02-07
nova: Reset sync marks of clusters that are in compute roles

vuntz:
We want to reset sync marks of all clusters, not just the one that might
be in the controller role. Otherwise, we might still get a drift.


toabctl:
+1

dirkmueller:
+1

PR #266, opened 2016-02-08
nova: Remove wrong TODO about nova-evacuate order constraint

vuntz:
All services are indeed required because nova-evacuate will trigger nova
instances to be spawned. And that indeed requires a lot of services.

And we're actually missing swift-proxy...


dirkmueller:
+1

toabctl:
+1

PR #267, opened 2016-02-08
Put global settings again into DEFAULT section

dirkmueller:
As part of the big cinder.conf rebase, all settings after
the backends were in the section of the last backend..


bmwiedemann:
+1 tested working

vuntz:
Just checking: having two [DEFAULT] sections is fine, and they'll be merged?

bmwiedemann:
worked fine in our testing - e.g. ssl from the 2nd section was enabled and the backend from the first section still was used for volume create

toabctl:
+1

dirkmueller:
@vuntz apparently yes. we used to have one [DEFAULT] section before the great liberty rebase

PR #268, opened 2016-02-08
keystone: Add mention of CVE-2015-7546 in the migration for PKI->UUID

vuntz:
This helps understand why :-)


jsuchome:
+1

aplanas:
+1

vuntz:
Just a comment change, no need for gating

PR #272, opened 2016-02-09
neutron: Use correct hyperv package name on SUSE

toabctl:
Package was renamed and the correct name is now python-networking-hyperv.


vuntz:
+1

rsalevsky:
+1

vuntz:
Gating won't help here, so just merging.

PR #273, opened 2016-02-10
nova: Move from exception to log message when compute HA can't be setup

vuntz:
If the cluster is used for both controller and compute roles, then it
can happen that the nodes don't have the compute roles yet when
chef-client runs on the controller.

Here we don't do a search but really look at the roles attribute, and
this means that we need more than the role being set to the node through
the "crowbar-$node" role, which generally requires chef-client to run.
And that chef-client run would be in the second batch of apply_role
(compute roles), while the exception will be raised from the first batch
(controller role).


jsuchome:
+1

aplanas:
+1

PR #275, opened 2016-02-10
horizon: Add attribute to use new "Launch Instance" dialog

vuntz:
It's not the default upstream yet, but people might want to play with
it.

### DIFF
--- a/chef/data_bags/crowbar/template-horizon.schema
+++ b/chef/data_bags/crowbar/template-horizon.schema
@@ -52,6 +52,7 @@
                 "telemetry" : { "type": "str", "required": true }
               }
             },
+            "use_launch_instance_ng": { "type": "bool", "required": true },
             "can_set_mount_point": { "type": "bool", "required": false },
             "can_set_password": { "type": "bool", "required": false },
             "multi_domain_support": { "type": "bool", "required": true },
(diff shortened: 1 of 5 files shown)


chef/data_bags/crowbar/template-horizon.schema
```diff
@@ -52,6 +52,7 @@
                 "telemetry" : { "type": "str", "required": true }
               }
             },
+            "use_launch_instance_ng": { "type": "bool", "required": false },
```
toabctl: You add a migration and set a default in the .json file. So why not ```required: true``` ? Otherwise you don't need the migration.

vuntz: copy & paste error. Fixed.

aplanas:
+1

rsalevsky:
+1

dirkmueller:
+1

dirkmueller:
IMHO this deserves a UI visible option

toabctl:
+1 (even without UI)

vuntz:
@dirkmueller my concern is that upstream hasn't move to this UI by default, which makes me believe it's still experimental.

But maybe it's not experimental anymore?

toabctl:
> But maybe it's not experimental anymore?

It's beta since Kilo and I can't find any other info. And default is still ```False```.
So still +1

PR #276, opened 2016-02-10
Run flavor migration tool when upgrading from juno

rhafer:
When upgrading from juno to kilo, after running the db migrations the flavor
information for instance had be migrated using the "nova-manage db
migrate_flavor_data" tool. As we're upgrading directly from juno to liberty we
need execute this step else. Because of this the actually db migration step is
now split into three steps:
1. Run db migrations up to kilo
2. run flavor migration tool
3. Run the rest of the db migrations

To have this working a patched nova is needed (migrate_flavor_data forward
ported from kilo to juno).

### DIFF
--- a/chef/cookbooks/nova/recipes/database.rb
+++ b/chef/cookbooks/nova/recipes/database.rb
@@ -60,6 +60,52 @@ database_user "grant privileges to the nova database user" do
   only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
+# if we're upgrading from juno, we first need to upgrade to the kilo state,
+# then run nova-manage db migrate_flavor_data and after that, run the rest
+# of the db migrations.
+db_version_cmd = "nova-manage db version"
+last_juno_migration = 254
+last_kilo_migration = 290
+
+execute "nova-manage db sync (kilo migrations)" do
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage db sync --version #{last_kilo_migration}"
+  action :run
+  # We only do the sync the first time, and only if we're not doing HA or if we
+  # are the founder of the HA cluster (so that it's really only done once).
+  only_if do
+    if !node[:nova][:db_synced] &&
+        (!node[:nova][:ha][:enabled] || CrowbarPacemakerHelper.is_cluster_founder?(node))
+      cmd = Mixlib::ShellOut.new(db_version_cmd,
+                                 user: node[:nova][:user],
+                                 group: node[:nova][:group])
+      cmd.run_command
+      db_version = cmd.stdout.to_i
+      if db_version >= last_juno_migration && db_version < last_kilo_migration
+        true
+      else
+        false
+      end
+    else
+      false
+    end
+  end
+  notifies :run, "execute[nova-manage db migrate_flavor_data]", :immediately
+end
+
+# now run the db migrate_flavor tool if needed
+execute "nova-manage db migrate_flavor_data" do
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage db migrate_flavor_data --force"
+  # We only run this step on explicit notification of the previous
+  # "db sync (kilo migrations)" step to avoid another complex "only_if"
+  # condition.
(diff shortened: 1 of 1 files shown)


rhafer:
The patched nova package need for this can be found here:
https://build.opensuse.org/package/show/home:rhafer:branches:Liberty:Staging/openstack-nova

The important bit is:
https://build.opensuse.org/package/view_file/home:rhafer:branches:Liberty:Staging/openstack-nova/0001-Forward-port-flavor-migration-from-kilo.patch?expand=1

Any comments are welcome. I would be specifically interested in better ways to solve this (preferably without patching nova)

aplanas:
amazing job, IMHO is better if in the package we make some explicit reverts as patches, like this one[1], instead of a single patch (0001-Forward-port-flavor-migration-from-kilo.patch)

[1] https://github.com/openstack/nova/commit/ff7dd5b3bd51a6c570ca55eb2fe233d43e696a90

rhafer:
> IMHO is better if in the package we make some explicit reverts as patches, like this one[1], instead of a single patch (0001-Forward-port-flavor-migration-from-kilo.patch)

Actually there are already multiple commits in the single patch file (I didn't want to have that change spread multiple files). But I made it more obvious now that the first commit actually is just a revert of the commit you mentioned (minus the unit test changes). There is not other revert (or cherry-pick) involved.

rhafer:
Fixed Hound issues.

mjura:
Works for me +1

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -60,6 +60,62 @@
   only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
+# if we're upgrading from juno, we need to frist upgrade to the kilo state,
+# then run nova-manage db migrate_flavor_data and after that run the rest
```
vuntz: Nitpick: maybe a `,` after `after that` (had issues reading the line, sorry for being french)

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -60,6 +60,62 @@
...
+# of the db migrations.
+db_version_cmd = "nova-manage db version"
+last_juno_migration = 254
+last_kilo_migration = 290
```
vuntz: Empty line after this one?

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -60,6 +60,62 @@
...
+        false
+      end
+    else
+      false
+    end
+  end
+end
+
+# now run the db migrate_flavor tool if needed
+execute "nova-manage db migrate_flavor_data" do
```
vuntz: What about running this resource only as immediate notification of the previous one? That would remove the need for the `only_if`.

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -60,6 +60,62 @@
...
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage --debug db migrate_flavor_data --force"
```
vuntz: I've nothing against it, just wondering: do we want to keep `--debug` or was it just for you?

vuntz:
I'm +1 on the approach; just had a few comments (that can be fixed in a followup PR if wanted)

rhafer:
@vuntz adjusted as you suggested.

vuntz:
Thanks, +1

mjura:
Still +1

aplanas:
+1

rhafer:
Keeping the "do not merge" label until we decided about: https://build.opensuse.org/request/show/358825

rhafer:
Hm, ci is failing on applying nova (permission problem on nova-manage.log). I guess I need to run the only_if check as "nova" as well.

Investigating.

rhafer:
With the latest change (run ```nova manage db version``` as user nova) the ci issues should be resolved.
The nova package has also be merged meanwhile so I remove the do not merge label.

vuntz:
+1

mjura:
+1

PR #277, opened 2016-02-11
Switch to using the tenant_name for now in the neutron cookbook

dirkmueller:
All others were / are still using tenant_name, and there seems
to be a caching bug in the chef server with the tenant_id that
we're still investigating. This is a workaround until that bug
is fixed.


vuntz:
+1

rsalevsky:
+1

rhafer:
+1

PR #278, opened 2016-02-11
Switch from syncdb to migrate for Django database

rhafer:
"syncdb" has been deprecated in favor of "migrate" with recent Django releases.
Also "syncdb" didn't work correctly when run on a database migrated from an
older release. See https://docs.djangoproject.com/en/1.8/ref/django-admin/ for
details.


mjura:
I agree +1

vuntz:
+1

PR #279, opened 2016-02-11
neutron: Change order of resources in neutron-agents group

vuntz:
The L2 agent must start first, and at least before DHCP agent as DHCP
agent depends on it. Otherwise, this can result in port failing to bind.

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -30,15 +30,31 @@ group_members = []
 transaction_objects = []
 
 if use_l3_agent
-  l3_agent_primitive = "neutron-l3-agent"
-  pacemaker_primitive l3_agent_primitive do
-    agent node[:neutron][:ha][:network][:l3_ra]
+  case node[:neutron][:networking_plugin]
+  when "ml2"
+    ml2_mech_drivers = node[:neutron][:ml2_mechanism_drivers]
+    case
+    when ml2_mech_drivers.include?("openvswitch")
+      neutron_agent = node[:neutron][:platform][:ovs_agent_name]
+      neutron_agent_ra = node[:neutron][:ha][:network]["openvswitch_ra"]
+    when ml2_mech_drivers.include?("linuxbridge")
+      neutron_agent = node[:neutron][:platform][:lb_agent_name]
+      neutron_agent_ra = node[:neutron][:ha][:network]["linuxbridge_ra"]
+    end
+  when "vmware"
+    neutron_agent = ""
+    neutron_agent_ra = ""
+  end
+  neutron_agent_primitive = neutron_agent.sub(/^openstack-/, "")
+
+  pacemaker_primitive neutron_agent_primitive do
+    agent neutron_agent_ra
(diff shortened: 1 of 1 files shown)


vuntz:
Before the patch:

`group g-neutron-agents neutron-l3-agent neutron-dhcp-agent neutron-metadata-agent neutron-metering-agent neutron-lbaas-agent neutron-openvswitch-agent`

After the patch:

`group g-neutron-agents neutron-openvswitch-agent neutron-dhcp-agent neutron-l3-agent neutron-metadata-agent neutron-metering-agent neutron-lbaas-agent`

cc @rossella

toabctl:
How is this related to https://github.com/crowbar/crowbar-openstack/pull/236 ? Do we need both?

vuntz:
So I'm not really sure about https://github.com/crowbar/crowbar-openstack/pull/236 -- let's see gating. But this one is needed in general, as the order was just wrong.

rossella:
the logic makes sense, thank for fixing this. TBH just looking at the code I was not able to infer the order of agents startup

toabctl:
standard build suceeeded. Rebuild with hacloud=1

abel-navarro:
Although this change makes sense, I don't know if it'll fix the [problem](https://trello.com/c/30w2SCA0) with pings. In that case the agents were long started and lots of tests passed before the failure. It is worth though to check the results with 'hacloud=1'.

vuntz:
@abel-navarro the issue was that the dhcp agent was trying to start doing its magic for the fixed network, but br-int was not setup (since ovs agent was not started yet). So it failed. But all other networks created during tempest were fine since ovs agent was started then.

toabctl:
Error is ```[2016-02-11T15:18:38+00:00] ERROR: Report handler Chef::Handler::ErrorReport raised #<Encoding::UndefinedConversionError: "\xF2" from ASCII-8BIT to UTF-8>```

rossella:
@vuntz I'd like to be optimistic and hope that all the binding failures are now fixed. Anyway in most cases the race I saw was different. no error regarding br-int but I think that Neutron server received the message to create a dhcp port before processing the l2 agent report state message, that is Neutron server didn't know that an l2 agent was running on that host. Anyway this patch should improve that even if I think the final solution would be to wait for all agents to be up before wiring the network.

vuntz:
@toabctl was an issue with ceph taking too long to deploy, it seems.

vuntz:
So finally the gating is not failing with the ping test or port_binding failed. It still failed in tempest, with manila test though :/

vuntz:
> I'd like to be optimistic and hope that all the binding failures are now fixed. Anyway in most cases the race I saw was different.

@rossella Right, I don't think this will fix all failures. But it's fixing one that happens consistently with HA. The remaining ones might be fixed with https://github.com/crowbar/crowbar-openstack/issues/236 or with your suggestion to delay creation of networks.

toabctl:
+1

rossella:
+1 after investigating more with @toabctl we have the proof that the binding failed problem is due to the fact that Neutron server has no L2 agent registered for that host. This patch will surely help with that, even if it doesn't remove the race completely. It's a good step forward! Thanks a lot @vuntz !

PR #280, opened 2016-02-11
trove: Fix to work with insecure ssl

vuntz:
The keystone configuration in api-paste.ini was missing the insecure
bit.


jsuchome:
+1

aplanas:
+1

rsalevsky:
+1

PR #281, opened 2016-02-11
keystone, openstack: Invalidate caches for each chef run (bsc#966214)

vuntz:
We use a cache to avoid querying the same data again and again over the
network. However that cache is surviving chef-client runs inside the
chef-client daemon (since they're all in the same process).

So it needs to be invalidated for each chef-client run at some point. We
use the ohai time as a marker that lets us know if the cache is for the
current run or from a previous run (since ohai time is updated at the
beginning of each run, in the deployer cookbook).

https://bugzilla.suse.com/show_bug.cgi?id=966214

cc @dirkmueller


aplanas:
LGTM: +1

jsuchome:
LGTM

bmwiedemann:
could be made more DRY in a future PR...

PR #282, opened 2016-02-12
Fix ceilometer role renaming for HA

rhafer:
There can be multiple nodes with the ceilometer-cagent role. Update all of them
not only the first.


jsuchome:
+1 (is the role renaming in nova fixed the same way?)

vuntz:
+1

aplanas:
+1

rhafer:
> is the role renaming in nova fixed the same way

It was already accounting for multiple nodes.

PR #284, opened 2016-02-12
(!!WIP!!) Wait for L2 agents to start on the network nodes

rossella:
If the L2 agent on the network node is not detected as UP by
the Neutron server the port binding will fail (this happens
a lot with the dhcp and the gateway port). This patch
waits that all the L2 agent are UP on the network node before
starting creating the network resources

### DIFF
--- a/chef/cookbooks/neutron/recipes/post_install_conf.rb
+++ b/chef/cookbooks/neutron/recipes/post_install_conf.rb
@@ -103,6 +103,26 @@ else
   Chef::Log.error("networking plugin '#{networking_plugin}' invalid for creating provider networks")
 end
 
+missing_l2_nodes = search(:node, "roles:neutron-network")
+# wait for all L2 agents to be up
+ruby_block "wait for all L2 agents on network nodes to be started" do
+  block do
+    require "timeout"
+    begin
+      Timeout.timeout(20) do
+        missing_l2_nodes.delete_if {|agent| ::Kernel.system("#{neutron_cmd} agent-list | grep -q #{agent.name}")}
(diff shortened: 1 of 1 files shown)


rossella:
Not tested at all, I wanted to get some feedback first regarding the approach

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -103,6 +103,26 @@ def mask_to_bits(mask)
   Chef::Log.error("networking plugin '#{networking_plugin}' invalid for creating provider networks")
 end
 
+missing_l2_nodes = search(:node, "roles:neutron-network")
+# wait for all L2 agents to be up
+ruby_block "wait for all L2 agents on network nodes to be started" do
+  block do
+    require "timeout"
+    begin
+      Timeout.timeout(20) do
```
vuntz: Unfortunately, that won't help: you will just hit the timeout because what we do right now is:

 - deploy neutron-server
   - at the end of this, create the networks
 - then deploy neutron agents

If everything is running on the same node, then something like this approach could work.

But if it's on two different nodes, then the neutron agents will never be deployed until the end of the chef run on the neutron-server node, which means the timeout won't help :/

rossella: thanks for the explanation...I imagine that moving the resource creation after we have deployed the neutron agents is not a trivial change, is it?

vuntz:
Let's close this; I think we fixed this a while ago by changing the order of the resource creation when using HA.

PR #285, opened 2016-02-12
Fix postgresql restore only occuring on 2nd chef-client run (bsc#966575)

rhafer:
The file check in the condition for including the restore recipe happened in
the compile phase of chef-client and because of that it failed on the first
iteration in an HA setup (fs-postgresql isn't running at that time).
So we turn those checks into an only_if condition on the resource that actually
does the restore.

Also the ".restored-ok" file is now created only on explicit notification of
the restore resource, as now the recipe might be include also when the restore
step is not being executed.

https://bugzilla.suse.com/show_bug.cgi?id=966575

### DIFF
--- a/chef/cookbooks/postgresql/recipes/db_restore.rb
+++ b/chef/cookbooks/postgresql/recipes/db_restore.rb
@@ -20,6 +20,8 @@
 service_name = "postgresql"
 dump_location = node[:crowbar][:upgrade][:db_dump_location]
 
+dump_needed_cond = "[ ! -f \"#{dump_location}.restore-ok\" -a -f \"#{dump_location}\" ]"
+
 if node[:database][:ha][:enabled]
   # Checks that the service is available, if it's running on this node.
   service_available = "crm resource show #{service_name} | " \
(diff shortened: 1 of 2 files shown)


chef/cookbooks/postgresql/recipes/server.rb
```diff
@@ -129,7 +129,7 @@
 end
 
 db_dump = node["crowbar"].fetch("upgrade", {})["db_dump_location"]
-if !db_dump.nil? && File.exist?(db_dump) && !File.exist?("#{db_dump}.restored-ok")
+if !db_dump.nil? && !File.exist?("#{db_dump}.restored-ok")
```
vuntz: Is the node attribute only set on upgrade? (ie, not on new deployments)

rhafer: Good point. The answer is no :(. (It's even a proposal attribute it seems)

(It shouldn't really break anything,  but I guess it's ugly to have the recipe included all the time)

jsuchome:
It seems to work correctly for me...

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -27,16 +29,19 @@
...
     user "postgres"
+    only_if dump_needed_cond
+    notifies :create, "file[#{dump_location}.restored-ok]", :immediately
   end
 end
 
 # Creates file to indicate successful restore of database.
 file "#{dump_location}.restored-ok" do
-  action :create
+  action :nothing
```
sayalilunkad: Why do we need this at all then?

jsuchome:
+1

chef/cookbooks/postgresql/recipes/server.rb
vuntz: Actually, because the nodes wouldn't have the crowbar-config-default role, I guess it should just work as intended.

vuntz:
+1

PR #287, opened 2016-02-15
Escapes single quotes (bsc#953786)

sayalilunkad:


aplanas:
+1

dirkmueller:
+1

aplanas:
still +1

sayalilunkad:
I guess the gating failures are unrelated to this PR as they are manila client tests which are failing.

toabctl:
+1

mjura:
Rebuild gating and +1

vuntz:
I'm fine with it, but before merging, let me ask: wouldn't it make more sense to do the `gsub` in the template so that it's visible that we escape single quotes because the string is inside single quotes? (This might help the day someone decides to put the string in double quotes)

dirkmueller:
@vuntz I had the same comment to sayali previously in a out of band chat . In the end I don't care very much right now (this quoting stuff is a problem in 100s of places, we just need to add a generic helper for that and use it everywhere).

sayalilunkad:
@dirkmueller @vuntz so I tried out something else in the template file which also seems to work. If I just make the password_validator_help_text as translatable it fixes the problem and we need not use gsub. Do you think that would be a better option?
Lools like this:
"help_text": <%= @password_validator_help_text != "" ? "_('#{@password_validator_help_text}')" : '_("Your password does not meet the requirements.")' %>

aplanas:
Is %Q() a valid option here?

dirkmueller:
@sayalilunkad that just moves the problem from single quote to double quote. I would rather use alberto's suggestion in the config file.

sayalilunkad:
Ah you are right. I think I was referring to the problem with the non-ASCII characters but that is not relevant anymore.

sayalilunkad:
@aplanas I don't think %Q helps in escaping single quotes. It only works for double quotes AFAIK.

aplanas:
```ruby
help_text": <%= @password_validator_help_text != "" ? %Q(#{@password_validator_help_text}) : '_("Your password does not meet the requirements.")' %>
```

My test:
```
aplanas@lena:~> irb
irb(main):001:0> a = "a'b"
=> "a'b"
irb(main):002:0> b = 'a"b'
=> "a\"b"
irb(main):003:0> t = %Q(#{a})
=> "a'b"
irb(main):004:0> t = %Q(#{b})
=> "a\"b"
```

But maybe I am misunderstanding something.

chef/cookbooks/horizon/templates/default/local_settings.py.erb
```diff
@@ -89,7 +89,8 @@ OPENSTACK_KEYSTONE_MULTIDOMAIN_SUPPORT = <%= @multi_domain_support ? 'True' : 'F
 #}
 HORIZON_CONFIG["password_validator"] = {
     "regex": '<%= @password_validator_regex != "" ? @password_validator_regex : ".*" %>',
-    "help_text": <%= @password_validator_help_text != "" ? "'#{@password_validator_help_text}'" : '_("Your password does not meet the requirements.")' %>
+    "help_text": <%= @password_validator_help_text != "" ? "'#{@password_validator_help_text}'"
+    .gsub("'"){ "\\'" } : '_("Your password does not meet the requirements.")' %>
```
bmwiedemann: -1 as shown by CI, this breaks python syntax
```python
    { "help_text": \'Your password must be at least 8 characters long.\' }
                                                                         ^
SyntaxError: unexpected character after line continuation character

```

dirkmueller:
-1 , syntax error in config file

sayalilunkad:
So I needed to remove the gsub from the template file and add it back in the recipe because @password_validator_help_text with escaped single quotes needs to be passed as a whole in the template which is not possible when I add the gsub in the template string. If it is not done this way then the case where we use single quotes after a space in the helpt_text (eg: 'You're password should be 8 chars ' => \'You\' password should be 8 chars \') the last \ is interpreted as a line break in python and then throws an error.

vuntz:
> So I needed to remove the gsub from the template file and add it back in the recipe because @password_validator_help_text with escaped single quotes needs to be passed as a whole in the template which is not possible when I add the gsub in the template string

Hrm, I don't understand :-) Do you still have the old commit somewhere? It feels to me it simply had a bug.

sayalilunkad:
@vuntz  It was like this ```"help_text": <%= @password_validator_help_text != "" ? "'#{@password_validator_help_text}'".gsub("'"){ "\\'" } : '_("Your password does not meet the requirements.")' %>```

vuntz:
@sayalilunkad but it was on multiple lines, wasn't it?

Also, can you try changing `"'#{@password_validator_help_text}'".gsub("'"){ "\\'" }` to `"'#{@password_validator_help_text.gsub("'"){ "\\'" }}'"`

dirkmueller:
see https://github.com/crowbar/crowbar-openstack/pull/310 for an alternative implementation in the template itself, where it does feel more natural.

sayalilunkad:
@dirkmueller I have added it to the template now as per vuntz suggestion. It works.

dirkmueller:
@sayalilunkad right, but the quoting is still messed up in the else case.

sayalilunkad:
@dirkmueller that's right, I am fine with closing this PR and using #310 instead.

vuntz:
Using https://github.com/crowbar/crowbar-openstack/pull/310 instead.

PR #291, opened 2016-02-16
Adds encoding (bsc#953786)

sayalilunkad:
Adding this allows to use non-ASCII characters
in config file for openstack-dashboard.

### DIFF
--- a/chef/cookbooks/horizon/templates/default/local_settings.py.erb
+++ b/chef/cookbooks/horizon/templates/default/local_settings.py.erb
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 import os
 
 from django.utils.translation import ugettext_lazy as _


aplanas:
Reading the bug report I see why this fix works, but I find the message text a bit confusing. Actually this will allow UTF8 strings in the configuration file, not in Horizon itself.

aplanas:
+1 for the fix

toabctl:
Can you propose this upstream so the local_settings.py.sample has the encoding set?

toabctl:
but +1, too

aplanas:
@sayalilunkad thanks to adjusting the commit message: +1

sayalilunkad:
@toabctl thanks, have created the patch upstream https://review.openstack.org/#/c/281164/

toabctl:
@sayalilunkad thanks!

mjura:
+1

PR #297, opened 2016-02-19
nova: Also enable libvirtd debug logs when debug is true

toabctl:
When debugging nova problems and setting the debug attribute to true,
also enable libvirtd debugging which can be very useful.


toabctl:
Not sure if the current output with this logging settings is very useful...

vuntz:
+1

dirkmueller:
+1

aplanas:
+1

PR #298, opened 2016-02-19
ha-horizon: use different loadbalancer options

aplanas:
Depends on https://github.com/crowbar/crowbar-ha/pull/91

Description currently missing, but see also [Reduce haproxy health check intervals (Trello)](https://trello.com/c/Cwc43RXF/1500-reduce-haproxy-health-check-intervals)

### DIFF
--- a/chef/cookbooks/horizon/recipes/ha.rb
+++ b/chef/cookbooks/horizon/recipes/ha.rb
@@ -18,6 +18,8 @@ haproxy_loadbalancer "horizon" do
   port 80
   use_ssl false
   servers CrowbarPacemakerHelper.haproxy_servers_for_service(node, "horizon", "horizon-server", "plain")
+  options ["defaults", "httpchk"]
+  check ({ inter: 1000, downinter: 3000, rise: 3, fall: 1 })
   action :nothing
 end.run_action(:create)
 
@@ -27,6 +29,7 @@ if node[:horizon][:apache][:ssl]
     port 443
     use_ssl true
     servers CrowbarPacemakerHelper.haproxy_servers_for_service(node, "horizon", "horizon-server", "ssl")
+    check ({ inter: 1000, downinter: 3000, rise: 3, fall: 1 })
     action :nothing
   end.run_action(:create)
 end


chef/cookbooks/horizon/recipes/ha.rb
```diff
@@ -18,6 +18,8 @@
   port 80
   use_ssl false
   servers CrowbarPacemakerHelper.haproxy_servers_for_service(node, "horizon", "horizon-server", "plain")
+  options ["defaults", "httpchk"]
+  check ({ inter: 1000, downinter: 3000, rise: 3, fall: 1 })
```
aplanas: Well, this is how is in the rest of the examples.

abel-navarro:
+1

aplanas:
Oh, this also try to alleviate https://bugzilla.suse.com/show_bug.cgi?id=965858

chef/cookbooks/horizon/recipes/ha.rb
```diff
@@ -18,6 +18,8 @@
   port 80
   use_ssl false
   servers CrowbarPacemakerHelper.haproxy_servers_for_service(node, "horizon", "horizon-server", "plain")
+  options ["defaults", "httpchk"]
```
vuntz: `httpchk` won't work for setup using ssl. That's why we didn't add that.

aplanas: Yes, this is in the `horizon` section. Also you can see that `use_ssl` is set to `false`. IIRC this also is `mode http`, and `horizon_ssl` is `mode tcp`

abel-navarro: `httpchk` should be in the `listen` section for the service, not in the `defaults` which applies to all services.

aplanas: Is not in `defaults`. This `defaults` is to add the default options in the horizon section, together with `httpchk`. If we omit `defaults`, only the `httpchk` option is added.

abel-navarro: I'll have to believe you here, my chef understanding doesn't go that far.

aplanas:
* Use same checks for SSL

chef/cookbooks/horizon/recipes/ha.rb
abel-navarro: @aplanas showed me the related code and does make sense to me now.

abel-navarro:
+1

chef/cookbooks/horizon/recipes/ha.rb
vuntz: So what's the reason for not doing this by default for all services?

chef/cookbooks/horizon/recipes/ha.rb
vuntz: Also wondering out loud: since OpenStack APIs are HTTP-based, should we also have this for other OpenStack services?

chef/cookbooks/horizon/recipes/ha.rb
aplanas: IMHO a too drastic change. There is a card created by Abel to make sure that this options can be better than the current defaults. So far we are sure that this improve the HTTP and HTTPS situations for Horizon

chef/cookbooks/horizon/recipes/ha.rb
aplanas: Ahm... this makes sense to me. What do you think, @abel-navarro? Can we add `httpchk` in other services based on HTTP by default? (not HTTPS, of course)

abel-navarro: We cannot add `httpchk` to other services. `httpchk` expects an HTTP HELLO message that not all APIs do return. I tested this with all services and at least for (don't remember which) returned errors 400, 401 and 501. Haproxy will consider those services down.

chef/cookbooks/horizon/recipes/ha.rb
abel-navarro: The [card](https://trello.com/c/Cwc43RXF).

aplanas:
This PR needs more +1 : ))

aplanas:
This PR needs more +1, or be discarded

Itxaka:
Looks good to me, but if Im reading correct, once the service goes down it would take a minimum of 9 seconds to set it as up again (downinter 3 seconds * 3 tries to consider it working).

If the reason behind the change is to be more aggressive with the checks, shouldn't it make sense to try to reduce the downtime checks to have the service recovered as fast as possible (once the underlying service is up, of course) than 9 seconds?

I think we are being too aggressive on marking the service as down (only 1 check to consider it down) while being too late with the recovery checks..


In any case, this ios only a comment, non blocking, +1

abel-navarro:
@Itxaka you read it right: fast to fail, slow to recover. The reason for fast failing is that the time it takes haproxy to notice the service it's down, it is still sending requests there. Those requests fail, because the service is down even if haproxy hasn't noticed yet.

Since this is an ha environment, we can assume other service instance can take on the request. Failing fast should affect performance because other nodes have to take the load from the fallen one. I guess good dimensioning is important here. I find that preferable to rejected connections.

About slow recovery, we are not in a rush to recover the service. Taken that other nodes can handle the load, we prefer correctness here. In some connections it happened that the check may send and OK even if the service is not truly up. That leads to rejected connections while the service is getting up. But having a slow recovery we help the service to have a clean start and report OK when truly ready.

The values for the fast fail are really aggressive and I don't think should be modified. About the values for the slow recovery I'm up for discussion. The little experimentation I did back in the day proved those values worked well in my environment, but of course it may change in others.

Itxaka:
@abel-navarro Indeed, that is a good explanation. Thanks!

aspiers:
@abel-navarro I agree with @Itxaka, thanks for this good explanation. However it belongs in the commit message, not here :-/  Also there is one bit I don't understand:

>  In some connections it happened that the check may send and OK even if the service is not truly up.

How could this happen?  Wouldn't that be a bug in the check?

abel-navarro:
@aspiers in some checks (TCP checks) you only consider the socket being open to consider the service active. We found that even when the socket was alive, the service was not.

We are using HTTP checks here so that's less likely to happen since we require not only the socket to be alive buy also serving HTTP requests. Because the check is generic for all services, and some services may take longer to be alive (i.e. DB connections) I think it's ok to play safe and allow for slow recoveries.

About the commit message update, I'll let @aplanas take care of that.

aspiers:
> @aspiers in some checks (TCP checks) you only consider the socket being open to consider the service active. We found that even when the socket was alive, the service was not.
> We are using HTTP checks here so that's less likely to happen since we require not only the socket to be alive buy also serving HTTP requests.

@abel-navarro Thanks, that makes sense.  Also, given that TCP checks are a lot less heavyweight than HTTP checks, we should consider making `inter` default to something smaller for TCP checks than for HTTP checks.

> Because the check is generic for all services

Huh - which check do you mean?  The check in this PR is only for Horizon.

> and some services may take longer to be alive (i.e. DB connections) I think it's ok to play safe and allow for slow recoveries.

Agreed, we definitely need to allow for slow recoveries.

> About the commit message update, I'll let @aplanas take care of that.

Sure.

AbelNavarro:
@aplanas do we retake this? I guess there have been lots of changes to this code since last modification.

aplanas:
uau, 2 y.o lets drop it

PR #299, opened 2016-02-19
Fix config group of the ZVM related options

dirkmueller:
Those need to be in the [DEFAULT] section.

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -120,6 +120,14 @@ if neutron[:neutron][:networking_plugin] == "ml2"
   ml2_type_drivers = neutron[:neutron][:ml2_type_drivers]
 
   case
+  when ml2_mech_drivers.include?("zvm")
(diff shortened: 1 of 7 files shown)


vuntz:
+1 (although maybe it'd be simpler to just move the whole block of options to the existing section, since this bit is not upstream (iirc).

toabctl:
+1

vuntz:
@dirkmueller is it intended that the nova.conf change is huge? (github can't display it)

toabctl:
> (github can't display it)

@vuntz I see the change in github.

toabctl:
+1 (but untested)

vuntz:
> > (github can't display it)
> 
> @vuntz I see the change in github.

I can see the other changes, but not the ones to nova.conf.erb:

> 1,740 additions, 1,741 deletions not shown because the diff is too large. Please use a local Git client to view these changes. 

And when I do a local diff, the diff is indeed huge.

vuntz:
+1

toabctl:
+1

PR #300, opened 2016-02-20
glance: Do not create /etc/cron.d/glance-scrubber on SUSE platforms

vuntz:
We have /etc/cron.hourly/openstack-glance in the package already.


aplanas:
+1

toabctl:
+1

PR #302, opened 2016-02-22
swift: Fix check for running dispersion or not

vuntz:
It was not using the right API version nor the right auth URL.


aplanas:
+1

jsuchome:
That `populate-dispersion` call cannot read /etc/swift/dispersion.conf

jsuchome:
`group node[:swift][:group]` in `populate-dispersion` resource seems to help

jsuchome:
+1

aplanas:
+1

PR #305, opened 2016-02-22
Improve trove configuration

vuntz:
I guess rebasing the templates would actually bring many fixes, but for now, this is a small bit.


toabctl:
+1 . So the template is currently not the one from Liberty? Do we track this in a card?

vuntz:
> So the template is currently not the one from Liberty? Do we track this in a card?

We probably need to track the "let's fix trove cookbook" work in general ;-) But yeah, I'll create a card.

mjura:
+1

PR #306, opened 2016-02-23
glance: Fix glance-scrubber to work with insecure SSL for keystone

vuntz:


aplanas:
Is documented [1], so I guess that is also a bug in some of the CONF in glance, that is not exporting the parameter in the configuration file? Maybe submit an upstream bug report?

[1] http://docs.openstack.org/developer/glance/configuring.html

aplanas:
+1

vuntz:
> Maybe submit an upstream bug report?

Already did it: https://bugs.launchpad.net/glance/+bug/1548825

dirkmueller:
the glance cookbook isn't using oslo config generator yet.

dirkmueller:
+1

dirkmueller:
hmm, this should also be set in glance-cache and glance-api.conf for consistency

vuntz:
> hmm, this should also be set in glance-cache and glance-api.conf for consistency

Done in new commit.

mjura:
+1

vuntz:
@dirkmueller does the updated commit work for you?

rsalevsky:
+1

PR #308, opened 2016-02-25
Move db_dump_location attribute into database barclamp

jsuchome:
So it is not lost when we drop crowbar-config-default role.

This goes with https://github.com/crowbar/crowbar-core/pull/379

### DIFF
--- a/chef/cookbooks/postgresql/recipes/db_restore.rb
+++ b/chef/cookbooks/postgresql/recipes/db_restore.rb
@@ -18,7 +18,7 @@
 #
 
 service_name = "postgresql"
-dump_location = node[:crowbar][:upgrade][:db_dump_location]
+dump_location = node[:database][:upgrade][:db_dump_location]
 
 dump_needed_cond = "[ ! -f \"#{dump_location}.restore-ok\" -a -f \"#{dump_location}\" ]"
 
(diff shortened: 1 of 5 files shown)


jsuchome:
The problem with this solution is that user could actually change the value of location when reapplying database barclamp, which does not make sense

vuntz:
@jsuchome does it have to be in the proposal? Couldn't it be a default attribute in the cookbook instead? So all nodes with that cookbook would have the attribute.

jsuchome:
I actually like the alternative solution more https://github.com/crowbar/crowbar-core/pull/380 (still being tested)

rhafer:
> Couldn't it be a default attribute in the cookbook instead?

@jsuchome Yeah, I also think that this is the better solution.

jsuchome:
@rhafer The problem with having it as postgres/database cookbook default attribute is that it won't be available when I want to reach it from  crowbar-upgrade recipe (as no database recipe is run at the time I want to dump the database)

rhafer:
@jsuchome Argh, right. I forgot about that :(

rhafer:
@jsuchome I guess this and https://github.com/crowbar/crowbar-core/pull/379 and be closed, right?

PR #311, opened 2016-02-26
nova: Allow overcommitting disk when spawning instances

vuntz:
We allow setting the disk_allocation_ratio option, like we do for the
cpu and ram options.


jsuchome:
LGTM

toabctl:
+1

mjura:
+1

PR #312, opened 2016-02-28
neutron: Fix broken MTU for tunnel networks after upgrade (bsc#968426)

vuntz:
When upgrading from Juno, we move the MTU advertisement from
/etc/neutron/dnsmasq-neutron.conf (custom dnsmasq config file) to new
settings in configuration files.

However, the MTU for each network is saved in the database and the
database migration from Juno to Liberty just set this to NULL, resulting
in broken MTU.

This is only an issue for networks with a tunnel type. In tex/Juno, we
only had two such possibilities: gre and vxlan (and with ovs only).

We fix this by tweaking the MTU in the database directly and restarting
the DHCP agents, which results in restarting dnsmasq for each network
with the correct settings.

https://bugzilla.suse.com/show_bug.cgi?id=968426

### DIFF
--- a/chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
+++ b/chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
@@ -1,0 +1,43 @@
+#!/usr/bin/python
+
+import sqlalchemy
+import sys
+from neutron.plugins.common import constants as p_const
+from oslo_config import cfg
+
+_db_opts = [
+    cfg.StrOpt('connection',
+               deprecated_name='sql_connection',
+               default='',
+               secret=True,
+               help='URL to database'),
+]
+
+CONF = cfg.ConfigOpts()
+CONF.register_cli_opts(_db_opts, 'database')
+CONF(project='neutron')
+
+db_uri = CONF.database.connection
+
+db = sqlalchemy.create_engine(db_uri)
+try:
+    connection = db.connect()
+except sqlalchemy.exc.SQLAlchemyError as e:
+    print >>sys.stderr, 'Cannot connect to database: %s' % e
+    sys.exit(1)
+
+ret = connection.execute(
+    "UPDATE networks SET mtu=%s "
+    "FROM ml2_network_segments "
+    "WHERE ml2_network_segments.network_type=%s "
+    "AND networks.mtu IS NULL "
+    "AND networks.id=ml2_network_segments.network_id;",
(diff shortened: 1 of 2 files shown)


vuntz:
cc @abel-navarro @jsuchome @rhafer

chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -170,6 +170,34 @@
   provider Chef::Provider::CrowbarPacemakerService if ha_enabled
 end
 
+# Fix broken MTU on upgrade from Juno (bsc#968426)
+# This is only an issue on upgrade (hence the check for an existence of an old
+# file) and for tunnels (gre/vxlan -- was only available with ovs)
+if !node[:neutron][:network_mtu_fix] &&
```
abel-navarro: maybe this block should be before `service node[:neutron][:platform][:dhcp_agent_name]`. We want the dhcp agent to start after the MTU change.

vuntz: It actually doesn't matter because we need to restart it anyway (in case it was already started before).

chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
```diff
@@ -0,0 +1,43 @@
...
+db = sqlalchemy.create_engine(db_uri)
+try:
+    connection = db.connect()
+except sqlalchemy.exc.SQLAlchemyError as e:
+    print >>sys.stderr, 'Cannot connect to database: %s' % e
+    sys.exit(1)
+
+ret = connection.execute(
+    "UPDATE networks set mtu=%s "
+    "FROM ml2_network_segments "
```
aplanas: some nitpicks: `set` is lowercase and `FROM` is not part of the standard SQL (but works in PostgreSQL)

aplanas:
+1 from me

chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
vuntz: > some nitpicks: set is lowercase

fixed

> and FROM is not part of the standard SQL (but works in PostgreSQL)

Will see how to fix that (and will do it if it's trivial, otherwise, doesn't matter much)

aplanas:
+1 (the user will have PostgreSQL for sure)

abel-navarro:
+1

PR #313, opened 2016-02-28
neutron: Set path_mtu to 1500 and not 1400 for tunnel types

vuntz:
Neutron already substracts the overhead caused by tunnels (length of
tunnel headers), so no need to put a lower value.

cc @abel-navarro @rossella

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -179,7 +179,6 @@ when "ml2"
   physnet_map = NeutronHelper.get_neutron_physnets(network_node, external_networks)
   physnets = physnet_map.values
 
-  mtu_value = 0
   ml2_type_drivers = node[:neutron][:ml2_type_drivers]
   ml2_mechanism_drivers = node[:neutron][:ml2_mechanism_drivers].dup
   if use_hyperv
@@ -190,7 +189,6 @@ when "ml2"
   end
   if ml2_type_drivers.include?("gre") || ml2_type_drivers.include?("vxlan")
     ml2_mechanism_drivers.push("l2population") if node[:neutron][:use_dvr]
-    mtu_value = 1400
   end
 
   ml2_mech_drivers = node[:neutron][:ml2_mechanism_drivers]
@@ -215,7 +213,6 @@ when "ml2"
       vxlan_end: vni_end,
       vxlan_mcast_group: node[:neutron][:vxlan][:multicast_group],
       external_networks: physnets,
-      path_mtu: mtu_value
     )
   end
 when "vmware"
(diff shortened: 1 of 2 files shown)


abel-navarro:
+1

mjura:
> Neutron already substracts the overhead caused by tunnels (length of tunnel headers), so no need to put a lower value.

Could you provide some more details? Maybe some link?

abel-navarro:
Neutron MTU calculation for tunnel type GRE (openvswitch):
https://github.com/openstack/neutron/blob/deea2ba351ec30dd73f9a0462dbe1d43d325259a/neutron/plugins/ml2/drivers/type_gre.py#L93

mjura:
@abel-navarro : Thank you Abel. now I see

+1

dirkmueller:
Is it okay that this subtraction isn't done for type vlan?

abel-navarro:
VLAN header is considered part of the Ethernet header, so it doesn't change the total Ethernet payload size.

abel-navarro:
Notice how Neutron does not do overhead calculation for VLAN:
https://github.com/openstack/neutron/blob/deea2ba351ec30dd73f9a0462dbe1d43d325259a/neutron/plugins/ml2/drivers/type_vlan.py#L262

rossella:
@dirk @abel-navarro VLAN adds 4 bytes to the Ethernet frame and doesn't modify the L3 header. So the calculation should be OK. Neutron provides another parameter physnet_mtus to account for the mtu allowed by the physical network. That is used to calculate the maximum mtu also in the case of VLAN

rossella:
@vuntz in neutron there has been lots of discussion regarding mtu lately so I guess the code will change a lot during next release....we should keep an eye on it.

rhafer:
+1 

@vuntz, @rossella BTW, for getting some understanding/support for the more sophisticated mtu settings (segment_mtu, physical_network_mtu) I created https://trello.com/c/gPyrUXAh a few days ago.

rhafer:
@vuntz Wouldn't it make sense to always set path_mtu value. If at least for the benefit that we don't run into a similar upgrade problem again?

rossella:
thanks @rhafer I agree...we should devote some time to the mtu topic :)

vuntz:
> @vuntz Wouldn't it make sense to always set path_mtu value. If at least for the benefit that we don't run into a similar upgrade problem again?

I was wondering about that; my initial patch was doing that and then I thought we might want to be less intrusive. I'm happy to change it to 1500 for all setups.

@rossella any opinion?

rossella:
1500 should be the default and should be OK, 2 main bugs were just fixed, https://bugs.launchpad.net/neutron/+bug/1542475 and https://bugs.launchpad.net/neutron/+bug/1542108

mjura:
still +1

vuntz:
Updated to use 1500 all the time now.

aplanas:
+1

rhafer:
+1

abel-navarro:
many +1. CI was failing on unrelated manila tests. Time to merge?

vuntz:
@abel-navarro @rossella do you think this should be backported to the 3.0 branch (Cloud 6)?

PR #314, opened 2016-02-29
Do not set virt_type for docker (bnc#968656)

jsuchome:
So yes, booting docker instance works. Thanks to @vuntz for spotting this.

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -2548,7 +2548,9 @@ admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
 # Libvirt domain type (string value)
 # Allowed values: kvm, lxc, qemu, uml, xen, parallels
 #virt_type = kvm
+<% if %w(kvm lxc qemu uml xen parallels).include? @libvirt_type -%>
 virt_type = <%= @libvirt_type %>
+<% end -%>
 
 # Override the default libvirt URI (which is dependent on virt_type) (string
 # value)


chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -2547,7 +2547,9 @@ admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
 # Libvirt domain type (string value)
 # Allowed values: kvm, lxc, qemu, uml, xen, parallels
 #virt_type = kvm
+<% unless @libvirt_type.eql?('docker') -%>
```
vuntz: I guess we also need the same for vmware, zvm (and hyperv, although that one is less critical since it's not using this template)?

chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -2547,7 +2547,9 @@ admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
 # Libvirt domain type (string value)
 # Allowed values: kvm, lxc, qemu, uml, xen, parallels
 #virt_type = kvm
+<% if %w(kvm lxc qemu uml xen parallels).include? @libvirt_type -%>
```
mjura: It is still missing vmware and probably zvm!

jsuchome: See that I am only using it if it fits the list of allowed values described in the comment above.

mjura: Ok, sorry I've tested it

```
2016-03-01 07:44:41.553 28934 ERROR nova ConfigFileValueError: Value for option virt_type is not valid: Valid values are [kvm, lxc, qemu, uml, xen, parallels], but found 'zvm'
```

mjura:
+1

vuntz:
+1, not sure why gating is red.

vuntz:
This only failed in the manila cct test (see https://trello.com/c/6N1sZSxw/162-c6-scenario-manila-client-tests); so tempest was fine which is good enough.

PR #315, opened 2016-03-02
ceilometer: Fix RA for mongodb when using HA (bsc#968969)

vuntz:
The LSB init script is only on SLE11; move to systemd for anything else.

https://bugzilla.suse.com/show_bug.cgi?id=968969


aplanas:
+1

jsuchome:
+1

PR #316, opened 2016-03-02
horizon: Do not protect apache HA setup with wait/create sync marks

vuntz:
We include another recipe, and that recipe might need to do additional
setup that needs to happen on all nodes. So avoid the sync mark here,
and rely on sync marks there.

We keep our "sync" sync mark because it doesn't hurt.

**Goes with https://github.com/crowbar/crowbar-ha/pull/95**


aspiers:
+1

jsuchome:
+1

PR #318, opened 2016-03-03
Log which barclamp we're invalidating the caches on behalf of

aspiers:
This makes it more obvious why cache invalidation is being triggered...

and AFAICS, `ohai_time` gets updated more than once per run, so the cache gets invalidating too often, although this is obviously not a huge issue.  I even saw `ohai_time` go backwards within the run!?

Continuation of #281.

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -57,7 +57,10 @@ class CrowbarOpenStackHelper
     # daemon (which are all in the same process); so use the ohai time as a
     # marker for that.
     if @database_settings_cache_time != node[:ohai_time]
-      Chef::Log.info("Invalidating database settings cache") if @database_settings
+      if @database_settings
+        Chef::Log.info("Invalidating database settings cache " \
+                       "on behalf of #{barclamp}")
+      end
       @database_settings = nil
       @database_settings_cache_time = node[:ohai_time]
     end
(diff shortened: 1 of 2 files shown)


jsuchome:
+1

vuntz:
> AFAICS, ohai_time gets updated more than once per run

Hmm, do you have more details on this?

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -57,7 +57,9 @@ def self.database_settings(node, barclamp)
     # daemon (which are all in the same process); so use the ohai time as a
     # marker for that.
     if @database_settings_cache_time != node[:ohai_time]
-      Chef::Log.info("Invalidating database settings cache") if @database_settings
+      if @database_settings
+        Chef::Log.info("Invalidating database settings cache for #{barclamp}")
```
vuntz: The string is ambiguous: it's not the invalidating the cache dedicated to this barclamp, but it's invalidating the cache due to a request from this barclamp.

Can you improve the string somehow?

aspiers:
> Hmm, do you have more details on this?

Only that I added some debug to show `ohai_time` and it clearly showed the observed behaviour.  Sounds like a bug in ohai to me.

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
aspiers: Yeah, good point - fixed.

vuntz:
>> Hmm, do you have more details on this?
> 
> Only that I added some debug to show ohai_time and it clearly showed the observed behaviour. Sounds like a bug in ohai to me.

I'd really like to have more information on that. If you saw that in /var/log/chef/client.log but in different chef-client runs, then I think it's expected (because the cache invalidation was added for the chef-client daemon).

vuntz:
+1

PR #323, opened 2016-03-04
[3.0] nova: Allow overcommitting disk when spawning instances

vuntz:
We allow setting the disk_allocation_ratio option, like we do for the
cpu and ram options.

(cherry picked from commit 0613c81dcef4b5c04f6622536ce2914d24de8868)

Backport of https://github.com/crowbar/crowbar-openstack/pull/311


vuntz:
Gaaah. Wrong branch again :-)

PR #326, opened 2016-03-04
[3.0] horizon: Do not protect apache HA setup with wait/create sync marks

vuntz:
We include another recipe, and that recipe might need to do additional
setup that needs to happen on all nodes. So avoid the sync mark here,
and rely on sync marks there.

We keep our "sync" sync mark because it doesn't hurt.

(cherry picked from commit 99ebd719147db0b1da7ded054ca2bb3eb9e83d06)

Backport of https://github.com/crowbar/crowbar-openstack/pull/316

**Goes with https://github.com/crowbar/crowbar-ha/pull/96**


mjura:
+1

dirkmueller:
+1

PR #330, opened 2016-03-09
interleave Pacemaker clones to minimise disruption (bsc#965886)

aspiers:
~~**requires crowbar/crowbar-ha#172**~~

**NEEDS MORE TESTING**

By default, Pacemaker clones aren't interleaved.  This means that if Pacemaker wants to restart a dead clone instance, and there is an order constraint on that clone, it will do the same restart on all other nodes, even if all the others are healthy.

More details on interleaving are here:
- https://www.hastexo.com/resources/hints-and-kinks/interleaving-pacemaker-clones/

This behaviour is far more disruptive than we want.  For example, in
[bsc#965886 (L3: neutron-ha-tool not migrating dhcp networks during failover)](https://bugzilla.suse.com/show_bug.cgi?id=965886) we saw that when a network node dies and Pacemaker wants to stop the instance of cl-g-neutron-agents on that node, it also stops and restarts the same clone instances on the healthy nodes.  This means there is a small window in which there are no neutron agents running anywhere.  If neutron-ha-tool attempts a router migration during this window, it will fail, at which point things start to go badly wrong.

In general, the cloned (i.e. active/active) services on our controller and compute nodes should all behave like independent vertical stacks, so that a failure on one node should not cause ripple effects on other nodes.  So we interleave all our clones.

(There is [a corresponding commit to crowbar-ha for the Apache clone](https://github.com/crowbar/crowbar-ha/pull/98).)

### DIFF
--- a/chef/cookbooks/aodh/recipes/aodh_ha.rb
+++ b/chef/cookbooks/aodh/recipes/aodh_ha.rb
@@ -59,7 +59,7 @@ transaction_objects << "pacemaker_group[#{group_name}]"
 clone_name = "cl-#{group_name}"
 pacemaker_clone clone_name do
   rsc group_name
-  meta ({ "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node) })
+  meta CrowbarPacemakerHelper.clone_meta(node)
   action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
--- a/chef/cookbooks/barbican/recipes/ha.rb
+++ b/chef/cookbooks/barbican/recipes/ha.rb
@@ -78,7 +78,7 @@ transaction_objects << "pacemaker_group[#{group_name}]"
 clone_name = "cl-#{group_name}"
 pacemaker_clone clone_name do
   rsc group_name
-  meta("clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node))
+  meta CrowbarPacemakerHelper.clone_meta(node)
   action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
--- a/chef/cookbooks/ceilometer/recipes/mongodb.rb
+++ b/chef/cookbooks/ceilometer/recipes/mongodb.rb
@@ -60,7 +60,7 @@ if ha_enabled
   clone_name = "cl-#{service_name}"
   pacemaker_clone clone_name do
     rsc service_name
-    meta ({ "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node) })
+    meta CrowbarPacemakerHelper.clone_meta(node)
     action :update
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
--- a/chef/cookbooks/ceilometer/recipes/server_ha.rb
+++ b/chef/cookbooks/ceilometer/recipes/server_ha.rb
@@ -59,7 +59,7 @@ transaction_objects << "pacemaker_group[#{group_name}]"
 clone_name = "cl-#{group_name}"
 pacemaker_clone clone_name do
   rsc group_name
-  meta ({ "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node) })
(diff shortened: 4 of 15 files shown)


aspiers:
Suggestions welcome for the best way to keep HoundCI happy.

jsuchome:
Looks good, and the explanation makes sense. I did have some failures during failover, but maybe they are caused by something different.

vuntz:
+1

jsuchome:
+1

jsuchome:
gating fails in manila cct tests, which is not related (and being fixed elsewhere)

vuntz:
@aspiers is it still WIP? The trello card made me think it has been tested and validated.

aspiers:
Forced rebuild of Travis fixed a transient Rake `#last_comment` bug.  I also just forced a rebuild of `mkcloud` in the hope that the `manila` / `cct` thing is now fixed.

aspiers:
@vuntz Still waiting for suggestions on how to fix the HoundCI warnings.  Also in theory we shouldn't merge until we've had at least a fully successful `hacloud=1` run... but even that's of limited value without failover testing.  Basically our HA CI just isn't good enough yet, and given that this is `master`, I guess I'm OK with just merging if people don't care about the HoundCI thing.

aspiers:
Also the `mkcloud` gate has broken and can't be rebuilt in the normal way, since apparently someone either changed how test packages for gating are stored, or just wiped the package cache.

vuntz:
> Still waiting for suggestions on how to fix the HoundCI warnings.

I think they're invalid.

> Also the mkcloud gate has broken and can't be rebuilt in the normal way, since apparently someone either changed how test packages for gating are stored, or just wiped the package cache.

Let's just get the package rebuilt. I'll do that now.

jsuchome:
Hm, gating/ha fails in cct saying that "And the controller node is in "ready" state " ... I logged to the machine (crowbar.vd5) and all nodes are ready, plus the cluster looks healthy...

abel-navarro:
Retriggered CI job.

aplanas:
+1

aspiers:
The HA gate passed, so this is presumably OK, and we definitely want it in for SOC7.  However I know there's a code freeze - @vuntz and @rsalevsky what do you think?

rsalevsky:
@aspiers Let's wait till the milestone is out.

vuntz:
@aspiers can you rebase? We want this in, I think.

aspiers:
@vuntz Am doing, but I think it's worth having https://github.com/crowbar/crowbar-ha/pull/172 first, so we can eliminate the duplication.

aspiers:
Forced rebuild of gate now that crowbar/crowbar-ha#172 is merged.

dirkmueller:
@aspiers can you please rebase on top of the other fixes in crowbar-openstack? otherwise you're building a jenkins job on outdated code which no longer works due to the swift change

aspiers:
@dirkmueller Thanks - done.

aplanas:
Restarted the HA job. Not sure if it is related

dirkmueller:
@aspiers does this need to be adjusted given the recent changes for lbaas and aodh? Would you rebase again?

aspiers:
@dirkmueller rebased with that case included.  This still needs either a manual or automated test with compute HA enabled.  If automated, it needs to pass to the point where the clones on the compute nodes are created successfully, since AFAIK that code path (the call to `.clone_meta` with `remote: true`) didn't get tested yet.

aspiers:
Force-restarted the gate to workaround some weird issue with `mkcha` mentioned by @bmwiedemann on IRC.

aspiers:
@dirkmueller Rebased to fix conflict caused by keystone no longer using Pacemaker clones.

vuntz:
For the record, I've tested this as part of https://github.com/crowbar/crowbar-openstack/pull/799 and it works as intended.

vuntz:
Doh. chef/cookbooks/ec2-api/recipes/ec2api_ha.rb now needs to be updated for this too.

vuntz:
https://github.com/crowbar/crowbar-openstack/pull/799 contained this and got merged.

PR #331, opened 2016-03-10
Set rake to version to 10.3.2

rsalevsky:
This change is needed due to the incompatibility with rake 11.0.0.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,8 @@
 
 source "https://rubygems.org"
 
+gem "rake", "~> 10.3.2"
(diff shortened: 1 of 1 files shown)


tboerger:
:+1:

bmwiedemann:
+1

MaximilianMeister:
was this really necessary to merge? see https://github.com/crowbar/crowbar-core/pull/413

tboerger:
I think we should revert that as it seems to be not needed anymore.

PR #335, opened 2016-03-11
[3.0] glance: Fix glance-scrubber to work with insecure SSL for keystone

vuntz:
Backport of https://github.com/crowbar/crowbar-openstack/pull/306


dirkmueller:
+1

PR #338, opened 2016-03-18
Expose trusts_delegated_roles in Heat barclamp

jgrassler:
This commit exposes the previously hardwired trusts_delegated_roles
setting in the Heat barclamp. This allows setting
trusts_delegated_roles to `Member` which is more useful than the
previous hardwired setting of `heat_stack_owner` (the latter will have
to be assigned to each user manually, while most users already have
the former).

trusts_delegated_roles will default to:

* `Member` for new installations
* `heat_stack_owner` for existing installations, so as not to break
  SUSE Cloud installations that have assigned heat_stack_owner to all
  users manually.

In addition to exposing this setting it also uses its value for
assigning the roles required for using Heat to

* the admin user
* the unprivileged account created for Tempest's use

and sets the Tempest setting `stack_owner_role`.

(previously, all of these were also hardwired to
`heat_stack_owner`)

Note: the `stack_owner_role` setting in tempest.conf (relevant for
`tempest account-generator`, which we do not use currently) will only
work for single-entry values in trusts_delegated_roles until
https://bugs.launchpad.net/tempest/+bug/1559078 is fixed.

### DIFF
--- a/chef/cookbooks/heat/recipes/server.rb
+++ b/chef/cookbooks/heat/recipes/server.rb
@@ -145,28 +145,30 @@ keystone_register "add heat stack user role" do
   action :add_role
 end
 
-keystone_register "add heat stack owner role" do
-  protocol keystone_settings["protocol"]
-  insecure keystone_settings["insecure"]
-  host keystone_settings["internal_url_host"]
-  port keystone_settings["admin_port"]
-  token keystone_settings["admin_token"]
-  user_name keystone_settings["service_user"]
-  tenant_name keystone_settings["service_tenant"]
-  role_name "heat_stack_owner"
-  action :add_role
-end
+node[:heat][:trusts_delegated_roles].each do |role|
+  keystone_register "Create stack owner role #{role}" do
+    protocol keystone_settings["protocol"]
+    insecure keystone_settings["insecure"]
+    host keystone_settings["internal_url_host"]
+    port keystone_settings["admin_port"]
+    token keystone_settings["admin_token"]
+    user_name keystone_settings["service_user"]
+    tenant_name keystone_settings["service_tenant"]
+    role_name role
+    action :add_role
+  end
 
-keystone_register "give admin access to stack owner role" do
-  protocol keystone_settings["protocol"]
-  insecure keystone_settings["insecure"]
-  host keystone_settings["internal_url_host"]
-  port keystone_settings["admin_port"]
-  token keystone_settings["admin_token"]
-  user_name keystone_settings["admin_user"]
-  tenant_name keystone_settings["default_tenant"]
-  role_name "heat_stack_owner"
-  action :add_access
+  keystone_register "give admin access to stack owner role #{role}" do
+    protocol keystone_settings["protocol"]
+    insecure keystone_settings["insecure"]
+    host keystone_settings["internal_url_host"]
+    port keystone_settings["admin_port"]
+    token keystone_settings["admin_token"]
+    user_name keystone_settings["admin_user"]
+    tenant_name keystone_settings["default_tenant"]
+    role_name role
+    action :add_access
+  end
 end
 
 package "python-openstackclient" do
@@ -359,7 +361,8 @@ template "/etc/heat/heat.conf" do
     stack_user_domain: %x[ #{shell_get_stack_user_domain} ].chomp,
     stack_domain_admin: node[:heat]["stack_domain_admin"],
     stack_domain_admin_password: node[:heat]["stack_domain_admin_password"],
-    use_convergence_engine: node[:heat][:use_convergence_engine]
+    use_convergence_engine: node[:heat][:use_convergence_engine],
+    trusts_delegated_roles: node[:heat][:trusts_delegated_roles]
   )
 end
 
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -25,6 +25,9 @@ alt_comp_user = keystone_settings["default_user"]
 alt_comp_pass = keystone_settings["default_password"]
 alt_comp_tenant = keystone_settings["default_tenant"]
 
+# Will only be set if this cloud is actually running heat
+heat_trusts_delegated_roles = nil
+
 tempest_comp_user = node[:tempest][:tempest_user_username]
 tempest_comp_pass = node[:tempest][:tempest_user_password]
 tempest_comp_tenant = node[:tempest][:tempest_user_tenant]
@@ -69,7 +72,10 @@ users = [
         ]
 
 if use_heat
-  users.push({"name" => tempest_comp_user, "pass" => tempest_comp_pass, "role" => "heat_stack_owner"})
+  heat_trusts_delegated_roles = node[:heat][:trusts_delegated_roles]
+  heat_trusts_delegated_roles.each do |role|
+    users.push("name" => tempest_comp_user, "pass" => tempest_comp_pass, "role" => role)
+  end
 end
 
 users.each do |user|
@@ -465,6 +471,11 @@ template "/etc/tempest/tempest.conf" do
     object_versioning: swift_allow_versions,
     # orchestration settings
     heat_flavor_ref: heat_flavor_ref,
+    # FIXME: until https://bugs.launchpad.net/tempest/+bug/1559078 only the
+    # first element of this list will be used in the template (this works fine
+    # for all of our default settings for stack_delegated_roles (single values)
+    # but breaks in the (unlikely) case of anybody configuring multiple roles).
+    heat_trusts_delegated_roles: heat_trusts_delegated_roles,
     # scenario settings
     cirros_version: cirros_version,
     image_regex: image_regex,
--- a/chef/cookbooks/tempest/templates/default/tempest.conf.erb
+++ b/chef/cookbooks/tempest/templates/default/tempest.conf.erb
@@ -1064,6 +1064,9 @@ endpoint_type = internalURL
 
 # Role required for users to be able to manage stacks (string value)
 #stack_owner_role = heat_stack_owner
+<% if @heat_trusts_delegated_roles -%>
+stack_owner_role = <%= @heat_trusts_delegated_roles[0] %>
+<% end -%>
 
 # Time in seconds between build status checks. (integer value)
 #build_interval = 1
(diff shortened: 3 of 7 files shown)


chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -69,7 +72,10 @@
         ]
 
 if use_heat
-  users.push({"name" => tempest_comp_user, "pass" => tempest_comp_pass, "role" => "heat_stack_owner"})
+  heat_trusts_delegated_roles = node[:heat][:trusts_delegated_roles]
+  heat_trusts_delegated_roles.each { |role|
+    users.push({"name" => tempest_comp_user, "pass" => tempest_comp_pass, "role" => role})
```
MaximilianMeister: wondering why hound didn't complain about the hash syntax, but better use the new one here:

```ruby
users.push(name: tempest_comp_user, pass: tempest_comp_pass, role: role)
```

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -465,6 +471,11 @@
     object_versioning: swift_allow_versions,
     # orchestration settings
     heat_flavor_ref: heat_flavor_ref,
+    # FIXME: until https://bugs.launchpad.net/tempest/+bug/1559078 only the
+    # first element of this list will be used in the template (this works fine
+    # for all of our default settings for stack_delegated_roles (single values)
+    # but breaks in the (unlikely) case of anybody configuring multiple roles).
+    heat_trusts_delegated_roles: heat_trusts_delegated_roles,
```
MaximilianMeister: is it ok when the value is `nil` here?

chef/cookbooks/tempest/templates/default/tempest.conf.erb
```diff
@@ -1063,7 +1063,11 @@ region = <%= @keystone_settings['endpoint_region'] %>
 endpoint_type = internalURL
 
 # Role required for users to be able to manage stacks (string value)
-#stack_owner_role = heat_stack_owner
+<% if @use_heat -%>
+stack_owner_role = <%= @heat_trusts_delegated_roles[0] %>
+<% else -%>
+# stack_owner_role = heat_stack_owner
```
MaximilianMeister: MINOR: isn't it better to not use a space like the other commented out values?

`#stack_owner_role = heat_stack_owner`

MaximilianMeister:
other than the comments above, and the hound comments it looks good to me

chef/cookbooks/tempest/recipes/config.rb
jgrassler: Yes, the template will only set stack_owner_role if `use_heat` is not nil. But now that you mention it...I'll remove that extra layer of indirection and check for heat_trusts_delegated_roles being `nil`.

chef/cookbooks/tempest/templates/default/tempest.conf.erb
jgrassler: Done.

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -145,19 +145,20 @@
...
-  token keystone_settings["admin_token"]
-  user_name keystone_settings["service_user"]
-  tenant_name keystone_settings["service_tenant"]
-  role_name "heat_stack_owner"
-  action :add_role
-end
-
-keystone_register "give admin access to stack owner role" do
+node[:heat][:trusts_delegated_roles].each { |role|
+    keystone_register "Create stack owner role #{role}" do
```
vuntz: Can you fix this? :-)

chef/cookbooks/tempest/templates/default/tempest.conf.erb
vuntz: @jgrassler what we usually do here for templates is always leave the commented out value and then do our change. So that would be:

```
 # Role required for users to be able to manage stacks (string value)
#stack_owner_role = heat_stack_owner
<% if @use_heat -%>
stack_owner_role = <%= @heat_trusts_delegated_roles[0] %>
<% end -%>
```

chef/data_bags/crowbar/template-heat.json
```diff
@@ -15,6 +15,7 @@
       "service_password": "",
       "auth_encryption_key": "",
       "use_convergence_engine": false,
+      "trusts_delegated_roles": [ "Member" ],
```
vuntz: Hrm, just checking, but are we 100% sure that heat_stack_user users don't have the Member role too?

chef/cookbooks/tempest/templates/default/tempest.conf.erb
jgrassler: Ah...yes, I can see that coming in handy every once in a while :-)

Done.

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -69,7 +72,10 @@
         ]
 
 if use_heat
-  users.push({"name" => tempest_comp_user, "pass" => tempest_comp_pass, "role" => "heat_stack_owner"})
+  heat_trusts_delegated_roles = node[:heat][:trusts_delegated_roles]
+  heat_trusts_delegated_roles.each.do |role|
```
dirkmueller: should be 

```ruby
heat_trusts_delegated_roles.each do |role|
```

chef/data_bags/crowbar/template-heat.json
jgrassler: You'd have to explicitely asssign the `Member` role to one of these users. Our hybrid LDAP backend can assign default roles (such as `Member`, which is the whole point of changing the default for `stack_owner_role` to `Member`) won't assign them for users with role `heat_stack_user` since it only assigns default roles to users from LDAP (which users created by Heat are not).

Either way, it shouldn't matter, though: Heat's policy.json uses the `heat_stack_user` role (or whatever value is configured in the `heat_stack_user_role` setting) as a black mark to deny access to most Heat APIs, allowing only those required for passing feedback from a stack up to heat. Other APIs don't care about what's going on in the separate Keystone domain Heat is using for that purpose anyway.

chef/cookbooks/tempest/recipes/config.rb
jgrassler: Done.

rhafer:
+1

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -145,19 +145,21 @@
...
+    token keystone_settings["admin_token"]
+    user_name keystone_settings["service_user"]
+    tenant_name keystone_settings["service_tenant"]
+    role_name role
+    action :add_role
+  end
 end
 
-keystone_register "give admin access to stack owner role" do
+keystone_register "give admin access to stack owner role #{role}" do
```
MaximilianMeister: are you sure that `role` exists here? it seems to be outside scope if I look at the block above

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -165,7 +167,7 @@
   token keystone_settings["admin_token"]
   user_name keystone_settings["admin_user"]
   tenant_name keystone_settings["default_tenant"]
-  role_name "heat_stack_owner"
+  role_name role
```
MaximilianMeister: same here

chef/cookbooks/tempest/recipes/config.rb
jgrassler: I'll need to dig this one up again: all the other code in `chef/cookbooks/tempest/recipes/config.rb` uses the old hash syntax, and the old and new syntax don't mix. I.e. they will create differently named keys since the new syntax adds a leading and trailing ':' to the key. A `users.to_yaml` illustrates the problem:

```
---
- name: tempest-user-1FiNh4xr7Cwi
  pass: jkKJeUbcusrq
  role: Member
- name: tempest-adm-2HTqSZYieNn6
  pass: DGCFjQKXWCo3
  role: admin
- :name: tempest-user-1FiNh4xr7Cwi
  :pass: jkKJeUbcusrq
  :role: heat_stack_owner
- :name: tempest-user-1FiNh4xr7Cwi
  :pass: jkKJeUbcusrq
  :role: extrarole1
- :name: tempest-user-1FiNh4xr7Cwi
  :pass: jkKJeUbcusrq
  :role: extrarole2
```

[the first two entries were created with the old hash syntax, the remaining entries were created with the new hash syntax]

I changed this back to the old hash syntax to remain compatible with the surrounding code. I'd prefer doing this over moving all the surrounding code to the new hash syntax since that would modify more code with potential for more and subtler errors.

chef/cookbooks/heat/recipes/server.rb
jgrassler: Indeed, thanks for spotting this (ithappened when I substituted the curly braces by `do` ... `end`). Fixed.

chef/cookbooks/heat/recipes/server.rb
jgrassler: See above.

MaximilianMeister:
+1 (codewise)

vuntz:
Generally +1. Just one last question: should we add validation in the barclamp to check that the list of roles is not empty? Or do we want to allow this?

chef/data_bags/crowbar/template-heat.json
```diff
@@ -33,7 +34,7 @@
     "heat": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 23,
+      "schema-revision": 40,
```
vuntz: Actually, thinking about this, we might want to leave more empty revisions for Cloud 6. Will discuss on cloud-devel.

jgrassler: An empty list should be possible, I think, at least if it has the same semantics as a missing  `trusts_delegated_roles` setting. I'll try this quickly...

As for the revision. I can bump it to 100 as suggested on the mailing list.

jgrassler: Ok, just tested this: leaving `trusts_delegated_roles` empty has the same semantics as not setting it at all, i.e. it causes all of the user's roles to be delegated. Having an empty `trusts_delegated_roles` is not something we should encourage, of course (since it causes all the user's roles to be delegated), but it should be possible for a user to do this if they really want to. As far as I am concerned, our default of `Member` (or `heat_stack_owner` in existing installations) combined with this being a raw setting discourages an empty `trusts_delegated_roles` pretty effectively.

vuntz: Thanks for checking.

jgrassler:
Note: This change only bumped the schema revision for the Heat barclamp to `100` as discussed.

vuntz:
+1

rhafer:
+1

PR #336, opened 2016-03-17
nova: Do not setup NFS mount to controller when shared storage is manual (bsc#971546)

vuntz:
We specifically split the use_shared_instance_storage bit from
setup_shared_instance_storage to allow manually configured shared
storage, but we we were still configuring the NFS mount on the compute
nodes when the setup is manual.

PR #340, opened 2016-03-23
[3.0] nova: Do not setup NFS mount when shared storage is manual (bsc#971546)

vuntz:
We specifically split the use_shared_instance_storage bit from
setup_shared_instance_storage to allow manually configured shared
storage, but we were still configuring the NFS mount on the compute
nodes when the setup is manual.

https://bugzilla.suse.com/show_bug.cgi?id=971546
(cherry picked from commit bd748092f275bf02970fa8e8b919a33a180698db)

Backport of https://github.com/crowbar/crowbar-openstack/pull/336


dirkmueller:
+1

toabctl:
+1

PR #343, opened 2016-03-24
Update to python-PyMySQL.

dguitarbite:
Updating deprecated python-mysqldb to python-PyMySQL which should be
used by default for newer releases of OpenStack.

### DIFF
--- a/chef/cookbooks/openstack-common/attributes/database.rb
+++ b/chef/cookbooks/openstack-common/attributes/database.rb
@@ -205,7 +205,7 @@ when 'rhel'
   default['openstack']['db']['python_packages']['mysql'] = ['MySQL-python']
   default['openstack']['db']['python_packages']['db2'] = ['python-ibm-db', 'python-ibm-db-sa']
 when 'suse'
-  default['openstack']['db']['python_packages']['mysql'] = ['python-mysql']
+  default['openstack']['db']['python_packages']['mysql'] = ['python-PyMySQL']
 when 'debian'
   default['openstack']['db']['python_packages']['mysql'] = ['python-mysqldb']
 end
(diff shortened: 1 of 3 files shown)


chef/cookbooks/openstack-common/attributes/database.rb
```diff
@@ -205,7 +205,7 @@
   default['openstack']['db']['python_packages']['mysql'] = ['MySQL-python']
   default['openstack']['db']['python_packages']['db2'] = ['python-ibm-db', 'python-ibm-db-sa']
 when 'suse'
-  default['openstack']['db']['python_packages']['mysql'] = ['python-mysql']
+  default['openstack']['db']['python_packages']['mysql'] = ['python-PyMySQL']
```
vuntz: I don't think this is correct, according to the requirements.txt of trove.

dguitarbite: Sorry for the late reply. I will crawl through the repositories again and confirm this. I have a strong feeling that its not related to this change.

dguitarbite: @vuntz I verified the requirements of trove. The package name is MySQL-python, its definatly not python-mysql. Shall I add this dependency as well?

vuntz: Well what we have right now might be correct: make sure to not confuse what's in `requirements.txt` (which are python modules) and what's here which is either a package name or a name provided by a package. Did you try to install the python-mysql package and see which module it installs?

dguitarbite: That is a good point! I need to verify this. Not sure if this is the right way to do so but, the packages are pointing towards MySQL-python which is atleast from the pip's side different to python-PyMySQL. https://build.opensuse.org/package/view_file/Cloud:OpenStack:Mitaka/python-MySQL-python/python-MySQL-python.spec?expand=1
check line 26.

We should try to move to python-PyMySQL since this is where the OpenStack projects are moving to.

dguitarbite:
I have realized that this patch is not good change till we change the deps from the packages. I will hence abandon this patch and re-submit once we add the python-PyMySQL dep. for the same. At least there is no harm in keeping it as it is :).

PR #344, opened 2016-03-24
increase neutron-ha-tool start timeout to 120s (bsc#965886)

aspiers:
https://github.com/SUSE-Cloud/cookbook-openstack-network/pull/1 adds retry logic to `neutron-ha-tool` so that it will not fail immediately if there are problems talking to the `neutron-server` API.  Therefore we need to give neutron-ha-tool enough time to have a decent chance of succeeding if the `neutron-server` still needs a while to recover after some kind of failover.

https://bugzilla.suse.com/show_bug.cgi?id=965886


abel-navarro:
+1

jsuchome:
+1

jsuchome:
Gating failed in cct manila scenario, not related

dirkmueller:
+1

aspiers:
Unfortunately this was wrong and breaks HA.  Should have been `op start timeout=120s` not `interval=120s`.  Highlights the need for having automated HA gating for PRs outside the `crowbar-ha` repo ...

aspiers:
And also it was the wrong service.  But apart from that it was perfect.  Awesome!

PR #345, opened 2016-03-24
[3.0] increase neutron-ha-tool start timeout to 120s (bsc#965886)

jsuchome:
https://github.com/SUSE-Cloud/cookbook-openstack-network/pull/1
adds retry logic to neutron-ha-tool so that it will not fail
immediately if there are problems talking to the neutron-server
API.  Therefore we need to give neutron-ha-tool enough time
to have a decent chance of succeeding if the neutron-server
still needs a while to recover after some kind of failover.

https://bugzilla.suse.com/show_bug.cgi?id=965886
(cherry picked from commit 55504e2e8ccfb23abb8a6b5b31e4ff7ef99e911b)

backport of https://github.com/crowbar/crowbar-openstack/pull/344


vuntz:
+1

abel-navarro:
+1

PR #347, opened 2016-04-01
Set neutron-ha-tool start timeout to 120s (bsc#965886)

jsuchome:
Set the start timeout only for neutron-ha-tool resource,
not for neutron agent services.
This fixes incorrectly placement of this value in commit
55504e2e8ccfb23abb8a6b5b31e4ff7ef99e911b.

cc @aspiers @abel-navarro Adam, feel free to close this and use your own PR if you have something prepared.


aspiers:
Code change looks good but commit message is not quite right since it doesn't explain that the previous commit was also incorrectly using `start interval` instead of `start timeout`.

aspiers:
+1

abel-navarro:
+1

PR #348, opened 2016-04-01
cinder: Fix nova catalog info settings

vuntz:
They were not using the right service name for nova.


toabctl:
+1 .

jsuchome:
+1

toabctl:
@vuntz assigned you because it's ready for merging afais

PR #350, opened 2016-04-01
Set neutron-ha-tool start timeout to 120s (bsc#965886)

jsuchome:
Set the start timeout only for neutron-ha-tool resource,
not for neutron agent services.
This fixes change introduced by commit 55504e2e8ccfb23abb8a6b5b31e4ff7ef99e911b,
where the 120s value was incorrectly set as "start interval"
instead of "start timeout".

(cherry picked from commit ec264aa9b22dc919bb6d588e2dce13074f1c656b)


aspiers:
+1

abel-navarro:
+1
error in CI looks unrelated to this change (manila failures)

aspiers:
If there is an unrelated CI failure, please can you link to the card in the Jenkins Green board?  And if it's not there, create it?  I keep hearing about manila failures and I thought the manila CI was disabled but presumably not based on this.  But it is much harder to understand current status if we don't maintain the Jenkins Green board.

aspiers:
Retriggered mkcloud rebuild to try to clear CI gate.

PR #351, opened 2016-04-01
Expose trusts_delegated_roles in Heat barclamp

jgrassler:
This is a backport of https://github.com/crowbar/crowbar-openstack/pull/338 to stable/3.0. Apart from changing the Heat barclamp's schema revision to `24` to match the sequence in stable/3.0 it is identical to the source commit in `master`.


rhafer:
+1

PR #353, opened 2016-04-04
Enable validation runs for defcore compliance

dirkmueller:
Defcore guidelines 2016.1 or newer require run_validation to
be enabled for tests that are getting submitted.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -140,6 +140,7 @@ end.run_action(:add_access)
 end
 
 machine_id_file = node[:tempest][:tempest_path] + "/machine.id"
+alt_machine_id_file = node[:tempest][:tempest_path] + "/alt_machine.id"
 docker_image_id_file = node[:tempest][:tempest_path] + "/docker_machine.id"
 
 glance_node = search(:node, "roles:glance-server").first
@@ -196,6 +197,18 @@ RAMDISK_ID=$(glance #{insecure} --os-image-api-version 1 image-create \
     --disk-format ari < $(findfirst '*-initrd') | extract_id)
 echo "done."
 [ -n "$RAMDISK_ID" ] || exit 1
+
+echo -n "Adding alt image ... "
+ALT_MACHINE_ID=$(glance #{insecure} --os-image-api-version 1 image-create \
+    --name="$IMG_NAME-tempest-machine-alt" \
+    --is-public True --container-format ami --disk-format ami \
+    --property kernel_id=$KERNEL_ID \
+    --property ramdisk_id=$RAMDISK_ID < $(findfirst '*.img') | extract_id)
+echo "done."
+[ -n "$ALT_MACHINE_ID" ] || exit 1
+
+echo -n "Saving alt machine id ..."
+echo $ALT_MACHINE_ID > #{alt_machine_id_file}
 
 echo -n "Adding image ... "
 MACHINE_ID=$(glance #{insecure} --os-image-api-version 1 image-create \
@@ -421,6 +434,7 @@ template "/etc/tempest/tempest.conf" do
     # general settings
     keystone_settings: keystone_settings,
     machine_id_file: use_docker ? docker_image_id_file : machine_id_file,
+    alt_machine_id_file: alt_machine_id_file,
     tempest_path: node[:tempest][:tempest_path],
     use_swift: use_swift,
     use_horizon: use_horizon,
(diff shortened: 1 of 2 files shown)


chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -197,6 +198,18 @@
...
+    --name="$IMG_NAME-tempest-machine-alt" \
+    --is-public True --container-format ami --disk-format ami \
+    --property kernel_id=$KERNEL_ID \
+    --property ramdisk_id=$RAMDISK_ID < $(findfirst '*.img') | extract_id)
+echo "done."
+[ -n "$ALT_MACHINE_ID" ] || exit 1
+
+echo -n "Saving alt machine id ..."
+echo $ALT_MACHINE_ID > #{alt_machine_id_file}
+
```
vuntz: Doesn't really matter, but I would have put the code about the alt image after the code about the non-alt image.

chef/cookbooks/tempest/templates/default/tempest.conf.erb
```diff
@@ -821,7 +822,7 @@ flavor_regex = ^tempest-stuff$
 
 # List of user mapped to regex to matching image names. (string value)
 #ssh_user_regex = [["^.*[Cc]irros.*$", "cirros"]]
-ssh_user_regex = [["^.*[Cc]irros.*$", "root"]]
+ssh_user_regex = [["^.*[Cc]irros.*$", "cirros"]]
```
aplanas: maybe we can simply uncomment the original?

aplanas:
+1

chef/cookbooks/tempest/recipes/config.rb
dirkmueller: well, this way I had to change less code since the rest of the code checks whether image has been uploaded.. :)

kbaikov:
2 tempest tests are still skipped with this patch applied.

tempest.api.compute.servers.test_list_server_filters.ListServerFiltersTestJSON.test_list_servers_detailed_filter_by_image
tempest.api.compute.servers.test_list_server_filters.ListServerFiltersTestJSON.test_list_servers_filter_by_image


I suppose the alternative image was not uploaded.
```
glance image-list
+--------------------------------------+-------------------------------------+
| ID                                   | Name                                |
+--------------------------------------+-------------------------------------+
| bbe5f525-720d-4173-99fa-577de4666996 | cirros-0.3.4-x86_64-tempest-kernel  |
| 933d92dc-2acb-4533-8bea-e7304afe5a91 | cirros-0.3.4-x86_64-tempest-machine |
| 1a79babd-674a-4492-a73f-dbb924520e22 | cirros-0.3.4-x86_64-tempest-ramdisk |
+--------------------------------------+-------------------------------------+
```

kbaikov:
+1
retested on qa hardware. works as expected.

chef/cookbooks/tempest/templates/default/tempest.conf.erb
dirkmueller: thats what I'm doing. you mean I should remove the overwrite and leave the default?

dirkmueller:
@kbaikov I'm confused, so the previous comment is no longer relevant?

kbaikov:
@dirkmueller correct.

The confusion due to the fact that re-applying the tempest barclamp did not recreate the glance images because of the existing machine.id file. That's why I did not see the alt_image.
Removing the machine.id file and re-applying the barclamp re-created the standard glance images and the alt_image.

chef/cookbooks/tempest/templates/default/tempest.conf.erb
MaximilianMeister: if it's the default you don't have to declare it in the first place i guess

aplanas: remove 825 and uncomment 824

but is not important, for me is a +1 as stated in another comment

MaximilianMeister: > remove 825 and uncomment 824

the thing is you don't need to uncomment it, if it is the default. so you could just delete 824 and do nothing else

dirkmueller:
Thanks !

MaximilianMeister:
i still don't understand what is the difference between
```
#ssh_user_regex = [["^.*[Cc]irros.*$", "cirros"]]
ssh_user_regex = [["^.*[Cc]irros.*$", "cirros"]]
```
and
```
#ssh_user_regex = [["^.*[Cc]irros.*$", "cirros"]]
```

aplanas:
I think that `ssh_user_regex = [["^.*[Cc]irros.*$", "cirros"]]` is an example here, not ha default

MaximilianMeister:
> I think that ssh_user_regex = [["^.*[Cc]irros.*$", "cirros"]] is an example here, not ha default

okay, I thought that all commented keys are the default... hmm how can you then actually rely on any defaults in the config file?

PR #354, opened 2016-04-05
pass database connection uri instead of calculating it

MaximilianMeister:
avoids a bug when a node is neutron-network but not neutron-server

cc @vuntz

### DIFF
--- a/chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
+++ b/chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
@@ -3,21 +3,11 @@
 import sqlalchemy
 import sys
 from neutron.plugins.common import constants as p_const
-from oslo_config import cfg
 
-_db_opts = [
-    cfg.StrOpt('connection',
-               deprecated_name='sql_connection',
-               default='',
-               secret=True,
-               help='URL to database'),
-]
+if len(sys.argv) != 2:
+    raise ValueError('One and only one argument must be passed: database connection uri')
 
-CONF = cfg.ConfigOpts()
-CONF.register_cli_opts(_db_opts, 'database')
-CONF(project='neutron')
-
-db_uri = CONF.database.connection
+db_uri = sys.argv[1]
 
 db = sqlalchemy.create_engine(db_uri)
 try:
--- a/chef/cookbooks/neutron/recipes/network_agents.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents.rb
@@ -181,8 +181,15 @@ if !node[:neutron][:network_mtu_fix] &&
       mode "0755"
     end
 
+    if node.roles.include?("neutron-server")
+      neutron_server_node = node
+    else
+      neutron_server_node = search(:node, "roles:neutron-server").first
+    end
+
     execute "fix mtu of tunnel networks" do
-      command "/usr/bin/crowbar-fix-tunnel-networks-mtu"
+      command "/usr/bin/crowbar-fix-tunnel-networks-mtu \
+        #{neutron_server_node[:neutron][:db][:sql_connection]}"
       action :run
       # restart local dhcp agent to get dnsmasq restarted so it advertises the
       # correct mtu; this is why we run this on all nodes even with HA


MaximilianMeister:
here is the bug https://ci.suse.de/job/openstack-mkcloud/20561/console

chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
```diff
@@ -13,11 +13,10 @@ _db_opts = [
                help='URL to database'),
 ]
```
vuntz: If we remove the need from reading the config file, you can actually drop `_db_opts` and `from oslo_config import cfg`.

chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
```diff
@@ -13,11 +13,10 @@ _db_opts = [
...
 
-CONF = cfg.ConfigOpts()
-CONF.register_cli_opts(_db_opts, 'database')
-CONF(project='neutron')
+if len(sys.argv) < 2:
+    raise ValueError('Missing argument: database connection uri')
 
-db_uri = CONF.database.connection
+db_uri = sys.argv[1]
```
vuntz: FWIW, @MaximilianMeister and I quickly discussed about adding some real option parsing (`--database foo` instead of just `foo`). I think it's slightly nicer, but on the other hand, it's unlikely we'll need any other option. So we agreed it's okay the way it is.

chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
```diff
@@ -13,11 +13,10 @@ _db_opts = [
                help='URL to database'),
 ]
 
-CONF = cfg.ConfigOpts()
-CONF.register_cli_opts(_db_opts, 'database')
-CONF(project='neutron')
+if len(sys.argv) < 2:
```
vuntz: I guess you should check for `> 2` too, or just `!= 2`.

MaximilianMeister:
@vuntz thanks. done

mjura:
Looks ok for me +1

Let's wait for gating

rsalevsky:
+1

chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -182,7 +182,7 @@
     end
 
     execute "fix mtu of tunnel networks" do
-      command "/usr/bin/crowbar-fix-tunnel-networks-mtu"
+      command "/usr/bin/crowbar-fix-tunnel-networks-mtu #{node[:neutron][:db][:sql_connection]}"
```
vuntz: That's not correct: it should not be `node` but it should be a node which has the neutron-server role. It happens that, in our gating, this is true for `node`. But in the case where you had the bug, this was not true.

chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
```diff
@@ -3,21 +3,11 @@
...
 
-_db_opts = [
-    cfg.StrOpt('connection',
-               deprecated_name='sql_connection',
-               default='',
-               secret=True,
-               help='URL to database'),
-]
+if len(sys.argv) != 2:
+    raise ValueError('Missing argument: database connection uri')
```
vuntz: With `!= 2`, the error message should probably be something like "One and only one argument must be passed: database connection uri".

chef/cookbooks/neutron/recipes/network_agents.rb
MaximilianMeister: I have used now
```ruby
only_if node.roles.include?("neutron-server")
```

rhafer: Actually that is still wrong. The command needs to run on the network node (to be able to restart dhcp-agent). But it need the sql_connections attribute of the neutron-server node. To achieve that you'd need to find a neutron server node and use that instead of ```node``` similar to whats done here:
https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/neutron/recipes/common_agent.rb#L17

chef/cookbooks/neutron/files/default/crowbar-fix-tunnel-networks-mtu
```diff
@@ -3,21 +3,11 @@
...
 
-_db_opts = [
-    cfg.StrOpt('connection',
-               deprecated_name='sql_connection',
-               default='',
-               secret=True,
-               help='URL to database'),
-]
+if len(sys.argv) != 2:
+    raise ValueError('One and only one Argument must be passed: database connection uri')
```
vuntz: nitpick in case (and only if) you force-push again: s/Argument/argument/

chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -181,12 +181,23 @@
       mode "0755"
     end
 
+    neutron_server_node = nil
+    if node.roles.include?("neutron-network")
```
rhafer: I think this if conditions is not needed. It's always true, since this recipe is only part of the ```neutron-network``` run list.

chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -181,12 +181,23 @@
...
+
     execute "fix mtu of tunnel networks" do
-      command "/usr/bin/crowbar-fix-tunnel-networks-mtu"
+      command "/usr/bin/crowbar-fix-tunnel-networks-mtu \
+        #{neutron_server_node[:neutron][:db][:sql_connection]}"
       action :run
       # restart local dhcp agent to get dnsmasq restarted so it advertises the
       # correct mtu; this is why we run this on all nodes even with HA
       notifies :restart, "service[#{node[:neutron][:platform][:dhcp_agent_name]}]"
+      only_if neutron_server_node
```
rhafer: This is also unneeded now.

rhafer:
+1 (Feel free to fix that last hound complain though)

rsalevsky:
Still +1

PR #355, opened 2016-04-07
[stable/3.0] backport #354

MaximilianMeister:
pass database connection uri instead of calculating it
avoids a bug when a node is neutron-network but not neutron-server

https://github.com/crowbar/crowbar-openstack/pull/354

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents.rb
@@ -181,8 +181,15 @@ if !node[:neutron][:network_mtu_fix] &&
       mode "0755"
     end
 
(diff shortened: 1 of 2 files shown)


chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -181,8 +181,15 @@
       mode "0755"
     end
 
+    if node.roles.include?("neutron-server")
```
MaximilianMeister: @houndci-bot why didn't you say that in the other PR before? if I would change it here now it is not an exact cherry-pick. but it makes sense, so I will change it here and send a style fixup for master

rsalevsky:
+1

tboerger:
:+1:

PR #357, opened 2016-04-13
rabbitmq: increase ulimit nofile.

aplanas:
We use two different `ruby_block` to avoid two ugly calls to `rc.write_file` after every `insert_line_if_no_match`

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -21,6 +21,27 @@
 package "rabbitmq-server"
 package "rabbitmq-server-plugins" if node[:platform_family] == "suse"
 
+# Increase ulimit for file handlers
+ruby_block "ulimit hard nofile for rabbitmq" do
+  block do
+    rc = Chef::Util::FileEdit.new("/etc/security/limits.conf")
+    rc.insert_line_if_no_match(/^rabbitmq hard/,
+                               "rabbitmq hard nofile 65535")
+    rc.write_file
+  end
+  only_if { node[:platform_family] == "suse" }
+end
+
+ruby_block "ulimit soft nofile for rabbitmq" do
+  block do
+    rc = Chef::Util::FileEdit.new("/etc/security/limits.conf")
+    rc.insert_line_if_no_match(/^rabbitmq soft/,
+                               "rabbitmq soft nofile 65535")
+    rc.write_file
+  end
+  only_if { node[:platform_family] == "suse" }
+end
+
 directory "/etc/rabbitmq/" do
   owner "root"
   group "root"


chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -21,6 +21,19 @@
 package "rabbitmq-server"
 package "rabbitmq-server-plugins" if node[:platform_family] == "suse"
 
+# Increase ulimit for file handlers
+ruby_block "ulimit nofile for rabbitmq" do
+  block do
+    rc = Chef::Util::FileEdit.new("/etc/sysctl.conf")
+    rc.insert_line_if_no_match(/^rabbitmq hard/,
+                               "rabbitmq hard nofile 65535")
+    rc.insert_line_if_no_match(/^rabbitmq soft/,
```
nicolasbock: @aplanas, please split the two `insert_line_if_no_match` into two separate `ruby_block`s.

nicolasbock:
:+1:

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -21,6 +21,27 @@
 package "rabbitmq-server"
 package "rabbitmq-server-plugins" if node[:platform_family] == "suse"
 
+# Increase ulimit for file handlers
+ruby_block "ulimit hard nofile for rabbitmq" do
+  block do
+    rc = Chef::Util::FileEdit.new("/etc/security/limits.conf")
```
toabctl: wouldn't it be better to create a rabbitmq specific file in ```/etc/security/limits.d/``` ?

toabctl:
And are you sure this works in a systemd env? I think ```LimitNOFILE=``` should be used there.

chef/cookbooks/rabbitmq/recipes/default.rb
aplanas: This is a better approach because is compatible with CentOS and debian, but I do not see /etc/security/limits.d in any SLE12 (I see in the TW)

aplanas:
> And are you sure this works in a systemd env? I think LimitNOFILE= should be used there.

I tested in in SLE12 that is using systemd

nicolasbock:
@toabctl good point, thanks for pointing that out. I tested on tumbleweed with
~~~
$ cat /etc/systemd/system/limits.service
[Service]
Type=simple
ExecStart=/bin/bash -c 'ulimit -a'
LimitNOFILE=2000
$ journalctl -e
~~~

That this indeed has the desired effect.

tboerger:
There should be a resource for using this chef class within some crowbar-core cookbook. @ovisan used something similar within the hyperv barclamp.

toabctl:
If we don't care about non-systemd systems, you could just drop a static file into i.e. ```/etc/systemd/system/rabbitmq-server.service.d/30-limit.conf``` with the parameter set.

aplanas:
> If we don't care about non-systemd systems, you could just drop a static file into i.e. /etc/systemd/system/rabbitmq-server.service.d/30-limit.conf with the parameter set.

This works perfectly. But I am guessing that the correct solution here is to fix the package itself ...

aplanas:
@nicolasbock @toabctl @tboerger thanks a looot for the reviews. I am closing this to provide a better fix in the package!

PR #362, opened 2016-04-22
Add ssl options for heat barclamp

nicolasbock:


chef/cookbooks/heat/attributes/default.rb
```diff
@@ -81,7 +81,7 @@
 default[:heat][:service_user] = "heat"
 default[:heat][:service_password] = ""
 
-default[:heat][:api][:protocol] = "http"
+default[:heat][:api][:protocol] = "https"
```
aplanas: we want to change the default? If so maybe we need a migration?

nicolasbock: Ooops, maybe I should change that back. I am having some issues with the migration. For some reason it's not applied, i.e. if I updated the barclamp on an already deployed cloud, the database schema is not updated and the ssl fields are still missing after the migration.

chef/data_bags/crowbar/template-heat.schema
```diff
@@ -48,7 +48,15 @@
...
-            }
+            },
+            "ssl": { "type": "map", "required": true, "mapping": {
+              "certfile": { "type" : "str", "required" : true },
+              "keyfile": { "type" : "str", "required" : true },
+              "generate_certs": { "type" : "bool", "required" : true },
+              "insecure": { "type" : "bool", "required" : true },
+              "cert_required": { "type" : "bool", "required" : true },
+              "ca_certs": { "type" : "str", "required" : true }
+            }}
```
jsuchome: Could you indent it the way the other sections are, like
```
"ssl": { 
  "mapping": {
    "certfile": { ... }
  }
}
```

nicolasbock: Done in new revision.

nicolasbock:
Hound tests are all benign now.

chef/cookbooks/heat/templates/default/heat.conf.erb
```diff
@@ -944,10 +944,12 @@ bind_port=<%= @api_port %>
 # Location of the SSL certificate file to use for SSL mode. (string value)
 # Deprecated group/name - [DEFAULT]/cert_file
 #cert_file = <None>
+cert_file = <%= @heat_ssl[:certfile] %>
```
jsuchome: We should probably fill this only when ssl is required, right? For example in cinder conf we have `ssl_cert_file = <%= node[:cinder][:ssl][:certfile] if node[:cinder][:api][:protocol] == 'https' %>`

But maybe it does not hurt this way...

nicolasbock: Yes, true.

chef/cookbooks/heat/attributes/default.rb
```diff
@@ -78,6 +78,14 @@
 default[:heat][:db][:user] = "heat"
 default[:heat][:db][:password] = "" # Set by Recipe
 
+default[:heat][:ssl][:enabled] = false
```
jsuchome: I think you are not using this one elsewhere

nicolasbock: Fixed.

jsuchome:
I think this is quite good and already working for heat CLI. For Horizon integration, we need to add correct value of `heat_insecure` into `local_settings`: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/horizon/recipes/server.rb#L299

nicolasbock:
@jsuchome: @aplanas and I have been working the whole morning trying to debug this. We have added the insecure option to horizon, but are seeing difficult to reproduce errors between horizon and heat.

nicolasbock:
I have been testing this patch further together with @aplanas: When the heat barclamp is deployed with self-signed SSL enabled, the heat command line client works as expected

~~~
heat --insecure stack list
~~~

but horizon fails and times out. We have tracked this problem at this point to horizon using an incorrect URI protocol. Instead of `https` it still uses `http`. After switching the project in horizon from `admin` to `openstack`, horizon uses the correct URI. Switching back to the `admin` project works after this step as well, i.e. horizon somehow updates the URI it uses to talk to the heat server. We do not understand at this point why this is happening.

jsuchome:
+1

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -78,6 +78,72 @@
...
+            raise message
+          end
+
+          File.delete ssl_csr_file # Nobody should even try to use this
+        end # unless files exist
+      end # block
+    end # ruby_block
+  else # if generate_certs
+    unless ::File.size? node[:heat][:ssl][:certfile]
+      message = "Certificate \"#{node[:heat][:ssl][:certfile]}\" is not present or empty."
```
jgrassler: Just a nit: Since the quotes inside the quotes are just error message markup, I'd prefer single quotes or backticks instead of the escaped double quotes here - leaves less attack surface for accidental breakage in the future.

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -78,6 +78,72 @@
...
+      Chef::Log.fatal(message)
+      raise message
+    end
+    # we do not check for existence of keyfile, as the private key is allowed
+    # to be in the certfile
+  end # if generate_certs
+
+  if node[:heat][:ssl][:cert_required] && !::File.size?(node[:heat][:ssl][:ca_certs])
+    message = "Certificate CA \"#{node[:heat][:ssl][:ca_certs]}\" is not present or empty."
```
jgrassler: Same nitpick as above: better use single quotes or backticks.

jgrassler:
+1 once the two small style nitpicks are fixed.

nicolasbock:
@jgrassler I have fixed those two lines and replaced the double quotes with single quotes.

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -78,6 +78,72 @@
...
+            raise message
+          end
+
+          File.delete ssl_csr_file # Nobody should even try to use this
+        end # unless files exist
+      end # block
+    end # ruby_block
+  else # if generate_certs
+    unless ::File.size? node[:heat][:ssl][:certfile]
+      message = 'Certificate "#{node[:heat][:ssl][:certfile]}" is not present or empty.'
```
jgrassler: I was referring to the inner set of quotes in my previous comment. The outer quotes need to remain double quotes, otherwise interpolation of `#{node[:heat][:ssl][:ca_certs]}` won't work. I.e. this line should look something like this:

`message = "Certificate CA '#{node[:heat][:ssl][:ca_certs]}' is not present or empty."`

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -78,6 +78,72 @@
...
+      Chef::Log.fatal(message)
+      raise message
+    end
+    # we do not check for existence of keyfile, as the private key is allowed
+    # to be in the certfile
+  end # if generate_certs
+
+  if node[:heat][:ssl][:cert_required] && !::File.size?(node[:heat][:ssl][:ca_certs])
+    message = 'Certificate CA "#{node[:heat][:ssl][:ca_certs]}" is not present or empty.'
```
jgrassler: Same as above.

chef/cookbooks/heat/recipes/server.rb
aplanas: Yep, as a Python developer I find this also confusing : )

nicolasbock:
Sorry, as @aplanas already said, I was looking at this from the point of view of Python :smile:

jgrassler:
+1

Looks good now, thanks!

As for the quote thing: Python's `printf` style interpolation is definitely saner. If I didn't have a fair amount of Perl in my past, the double quote thing would probably weird me out, too :-)

jsuchome:
+1 (we might not need to merge this if the following - refactored - one is working correctly)

aplanas:
+1

chef/data_bags/crowbar/migrate/heat/101_add_ssl_attributes.rb
```diff
@@ -0,0 +1,15 @@
+def upgrade(ta, td, a, d)
+  a["ssl"] = {}
+  a["ssl"]["certfile"]       = ta["ssl"]["certfile"]
```
dirkmueller: this should / could have been simply

a["ssl"] = ta["ssl"]

PR #363, opened 2016-04-27
cinder: Fix nova catalog info settings

vuntz:
They were not using the right service name for nova.

(cherry picked from commit 3132cb133188a70982c4ec185de33bb616de9585)


jsuchome:
+1

aplanas:
+1

PR #364, opened 2016-04-28
Refactor ssl

nicolasbock:


nicolasbock:
This PR depends on https://github.com/crowbar/crowbar-openstack/pull/362

crowbar_framework/config/locales/horizon/en.yml
```diff
@@ -28,6 +28,7 @@ en:
           help_text: 'Text to display if the password does not pass validation'
         ssl_header: 'SSL Support'
         session_timeout: 'Session Timeout (in minutes)'
+        debug: 'Debug'
```
jsuchome: Why is this needed?

chef/cookbooks/heat/templates/default/heat.conf.erb
```diff
@@ -1579,9 +1597,18 @@ rabbit_virtual_host=<%= @rabbit_settings[:vhost] %>
 
 # CA certificate file to use to verify connecting clients. (string value)
 #ca_file = <None>
+<% if node[:heat][:api][:protocol] == "https" %>
```
jsuchome: Shouldn't we have a check for `cert_required` here?

crowbar_framework/config/locales/horizon/en.yml
nicolasbock: Sorry, a leftover from me debugging horizon yesterday. I'll remove it.

chef/cookbooks/heat/templates/default/heat.conf.erb
nicolasbock: Yes, good point.

jsuchome:
+1 (for the code; not tested)

aplanas:
+1

jsuchome:
OK, this is wrong - it failed for my keystone, saying that the keystone group was not present (and it wasn't, packages were not yet installed). 
I assume this means that the new function is executed in the compile phase (which is too soon), while the original `ruby_block` code was executed in the execution/converge phase.

-1

jsuchome:
Needs rebase after merging the original PR

chef/cookbooks/crowbar-openstack/libraries/ssl.rb
```diff
@@ -0,0 +1,64 @@
+module SSLHelper
```
tboerger: You know that this naming is not conventional for ruby?

tboerger:
-0.5, please build a resource instead of a class that gets invoked within a ruby block.

chef/cookbooks/crowbar-openstack/libraries/ssl.rb
nicolasbock: No. What's the convention for module names? Is there something like PEP8 for ruby that I can use as a reference?

nicolasbock:
@tboerger I'll give that a shot.

chef/cookbooks/crowbar-openstack/libraries/ssl.rb
nicolasbock: I looked at this [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide#camelcase-classes), which claims:

> Use CamelCase for classes and modules. (Keep acronyms like HTTP, RFC, XML uppercase.)

tboerger: The class name should match the filename, in your case it should be `ssl_helper.rb`

nicolasbock:
@tboerger and @jsuchome: Please have another look. I rebased and refactored the SSL helper code into a new resource.

jsuchome:
Yes, it works correctly. But please fix those fixable hound issues (Ruby hash syntax).

jsuchome:
+1

dirkmueller:
LGTM

nicolasbock:
I don't know what the problem with the gating check is. It fails already in the `setupadmin` step. Maybe we should trigger another gating test?

rsalevsky:
LGTM so far but I would suggest to add () at the function calls

toabctl:
+1 (but retriggering the tests now with ```want_all_ssl=1```)

chef/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb
```diff
@@ -0,0 +1,82 @@
...
+        @current_resource.keyfile(@new_resource.keyfile)
+        @current_resource.group(@new_resource.group)
+        @current_resource.fqdn(@new_resource.fqdn)
+        @current_resource.cert_required(@new_resource.cert_required)
+        @current_resource.ca_certs(@new_resource.ca_certs)
+        @current_resource
+      end
+
+      def action_setup
+        if @current_resource.generate_certs
```
nicolasbock: @rsalevsky, do you mean to add `()` here?

toabctl:
> LGTM so far but I would suggest to add () at the function calls

@rsalevsky can we teach Hound to complain about things like that? IIRC I had an opposite comment (remove ```()``` from the function calls) for some code...

toabctl:
Still +1 (and tests with ```want_all_ssl=1``` passed)

MaximilianMeister:
> @rsalevsky can we teach Hound to complain about things like that? IIRC I had an opposite comment (remove () from the function calls) for some code...

I also think the opposite should be the rule, or is this some new guideline now? when a function doesn't need arguments we just avoid using () everywhere in our ruby code

rsalevsky:
I will adapt the hound config and will fix all style issues according to that later.

PR #360, opened 2016-04-15
Initial Barclamp to deploy magnum

mmnelemane:
The barclamp is used to deploy magnum with SOC with basic features. It uses default Fedora Service image for container service. 

Needs https://github.com/crowbar/crowbar-openstack/pull/346 for testing on Mitaka

PR #365, opened 2016-04-29
Sahara: Add Initial barclamp

abonillasuse:
The barclamp was a fork from the magnum initial barclamp request #360

### DIFF
--- a/bin/crowbar_sahara
+++ b/bin/crowbar_sahara
@@ -1,0 +1,22 @@
+#!/usr/bin/env ruby
+#
+# Copyright 2016, SUSE LINUX GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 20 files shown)


bin/crowbar_sahara
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env ruby
```
tboerger: It doesn't make sense to add this file anymore, all those files will get dropped for next crowbar release in favor of crowbarctl.

crowbar_framework/app/views/barclamp/sahara/_edit_attributes.html.haml
```diff
@@ -0,0 +1,6 @@
+= attributes_for @proposal do
+  .panel-sub
+    = header @dep_raw, true
+
+  .panel-body
+    = select_field :foo, :collection => [['XKCD','xkcd'], ['Dilbert', 'dilbert']]
```
tboerger: This needs to be replaced with real fields or just a note that we currently don't provide any flag.

tboerger:
Please fix the naming of the classes, otherwise it's just broken and not usable.

dirkmueller:
superseded by #480

PR #366, opened 2016-05-03
Avoid usage of admin_token in cookbooks

mmnelemane:
Keystone register is modified to generate token from user name and password and all the cookbooks (except Trove and Keystone) use this logic instead of directly using the admin_token. There is a follow-up card to handle bootstrapping for Trove and Keystone cookbooks here https://trello.com/c/ipXUuRQl/1052-adapt-usage-of-keystone-manage-bootstrap-for-bootstrapping-keystone-service.

co-authored with @rhafer

### DIFF
--- a/chef/cookbooks/keystone/providers/register.rb
+++ b/chef/cookbooks/keystone/providers/register.rb
@@ -106,34 +106,26 @@ action :add_user do
     raise "Failed to talk to keystone in add_user"
   end
 
+  ret = false
   unless item_id
     # User does not exist yet
     body = _build_user_object(new_resource.user_name, new_resource.user_password, tenant_id)
     ret = _create_item(http, headers, path, body, new_resource.user_name)
-    new_resource.updated_by_last_action(ret)
   else
-    path = "/v2.0/tokens"
-    body = _build_auth(new_resource.user_name, new_resource.user_password, tenant_id)
-    resp = http.send_request("POST", path, JSON.generate(body), headers)
-    if resp.is_a?(Net::HTTPCreated) or resp.is_a?(Net::HTTPOK)
-      Chef::Log.info "User '#{new_resource.user_name}' already exists. No password change."
-      new_resource.updated_by_last_action(false)
-      data = JSON.parse(resp.read_body)
-      token_id = data["access"]["token"]["id"]
-      resp = http.delete("#{path}/#{token_id}", headers)
-      if !resp.is_a?(Net::HTTPNoContent) and !resp.is_a?(Net::HTTPOK)
-        Chef::Log.warn("Failed to delete temporary token")
-        Chef::Log.warn("Response Code: #{resp.code}")
-        Chef::Log.warn("Response Message: #{resp.message}")
-      end
+    auth_token = _get_token(http,
+                            new_resource.user_name,
+                            new_resource.user_password)
+    if auth_token
+      Chef::Log.debug "User '#{new_resource.user_name}' already exists and password still works."
+      http.delete("/v2.0/tokens/#{auth_token}", headers)
     else
-      Chef::Log.info "User '#{new_resource.user_name}' already exists. Updating password."
+      Chef::Log.debug "User '#{new_resource.user_name}' already exists. Updating Password."
       path = "/v2.0/users/#{item_id}/OS-KSADM/password"
       body = _build_user_password_object(item_id, new_resource.user_password)
       ret = _update_item(http, headers, path, body, new_resource.user_name)
-      new_resource.updated_by_last_action(ret)
     end
   end
+  new_resource.updated_by_last_action(ret)
 end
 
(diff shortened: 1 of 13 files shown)


rhafer:
I created another followup card for Trove. I guess the cleanest approach is if we ported the trove barclamp to use the resources of the keystone barclamp to create its endpoint(s) and user(s).

Other than that I think this PR is ok. But since I also did some parts of the coding on this I think we need somebody else as the primary reviewer. @jgrassler interested?

jgrassler:
Sure, I'll have a look...

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -110,30 +110,13 @@
     # User does not exist yet
     body = _build_user_object(new_resource.user_name, new_resource.user_password, tenant_id)
     ret = _create_item(http, headers, path, body, new_resource.user_name)
-    new_resource.updated_by_last_action(ret)
   else
-    path = "/v2.0/tokens"
-    body = _build_auth(new_resource.user_name, new_resource.user_password, tenant_id)
-    resp = http.send_request("POST", path, JSON.generate(body), headers)
-    if resp.is_a?(Net::HTTPCreated) or resp.is_a?(Net::HTTPOK)
-      Chef::Log.info "User '#{new_resource.user_name}' already exists. No password change."
```
jgrassler: Just curious: Why did you get rid of this check for an unchanged password? It's not really related to what the change is all about. Whatever the answer is, I'm in favour of removing it :-) 

(That check traded little value for a lot of code complexity after all)

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -356,8 +334,20 @@ def _build_connection(new_resource)
...
   http.verify_mode = OpenSSL::SSL::VERIFY_NONE if new_resource.insecure
 
+  auth_token = nil
+  if new_resource.token
+    auth_token = new_resource.token
+  elsif new_resource.auth
+    auth_token = _get_token(http,
+                            new_resource.auth[:user],
+                            new_resource.auth[:password],
+                            new_resource.auth[:tenant])
```
jgrassler: This won't cause an abort if _get_token` fails. Might be a good idea to check for `auth_token == nil` in this block.

jgrassler:
Looks good apart from the missing error handling for `_get_token()` . Why does it still fail gating, though?

chef/cookbooks/keystone/providers/register.rb
rhafer: Thanks for spotting this. Actually I think this check should stay. Having the logs full of password updates every 15min will probably make the admin nervous :)

chef/cookbooks/keystone/providers/register.rb
mmnelemane: @jgrassler : its handled now. thanks for spotting this.

chef/cookbooks/swift/templates/default/proxy-server.conf.erb
```diff
@@ -726,8 +726,9 @@ service_host = <%= @keystone_settings['internal_url_host'] %>
 auth_port = <%= @keystone_settings['admin_port'] %>
 auth_host = <%= @keystone_settings['internal_url_host'] %>
 auth_protocol = <%= @keystone_settings['protocol'] %>
-auth_token = <%= @keystone_settings['admin_token'] %>
-admin_token = <%= @keystone_settings['admin_token'] %>
+admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
+admin_user = <%= @keystone_settings['service_user'] %>
+admin_password = <%= @keystone_settings['service_password'] %>
```
rhafer: Looking through the s3token middleware code once more, it really seems, that these settings are not used anywhere. (That is: auth_token, admin_token, admin_tenant_name, admin_user, admin_password). Can you try to run the setup without them?

rhafer:
Last ci failure should be fixed by: https://github.com/SUSE-Cloud/automation/pull/992/files

retrying again.

rhafer:
@jgrassler Gating is fixed now. Are you fine with the latest changes?

jgrassler:
Looks good now, thanks :-)

+1

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -410,19 +412,39 @@ def _build_user_object(user_name, password, tenant_id)
...
+  else
+    Chef::Log.info "Failed to get token for User '#{user_name}' Tenant '#{tenant}'"
+    Chef::Log.info "Response Code: #{resp.code}"
+    Chef::Log.info "Response Message: #{resp.message}"
+    nil
+  end
+end
+
+private
+
```
abel-navarro: ugly newline here.

abel-navarro:
:+1:

PR #367, opened 2016-05-04
Add support for Neutron VPNaaS

toabctl:
Add a new "use_vpnaas" boolean parameter (default is enabled) to enable the
VPN as a service support in Neutron.


rhafer:
Closing this for now because of VPNaaS uncertain upstream future.

PR #368, opened 2016-05-10
Added support for Aodh, new stand-alone alarming service.

jsuchome:
Aodh replaces old ceilometer alarming services in Mitaka


jsuchome:
Seems to work so far, HA support still missing...

vuntz:
Wondering out loud: wouldn't it be cleaner to have a aodh cookbook? (not a barclamp, just a cookbook that would be used by the barclamp)

jsuchome:
I thought about it, but haven't decided yet. It's really tied to ceilometer (also ceilometer server node).

jsuchome:
... but it would solve the ugly hacks while fetching `network_settings`, and also with `CrowbarPacemakerHelper.haproxy_servers_for_service`

jsuchome:
Yes, https://github.com/crowbar/crowbar-openstack/pull/371 looks better

PR #372, opened 2016-05-18
Remove ceilometer-agent role as it should not be needed.

jsuchome:
ceilometer-agent-compute is obsoleted by usage of ceilometer-polling service.

- I'm not sure if this is right, but that's how I understand the introduction of polling service. 
- Also, I have no idea about agent-hyperv. 
- If moving polling agent to all nodes this way is correct, it's current HA setup is wrong

Relevant documentation: 
- http://specs.openstack.org/openstack/ceilometer-specs/specs/kilo/merge-compute-and-central-agents.html
- http://docs.openstack.org/developer/ceilometer/architecture.html
- http://docs.openstack.org/admin-guide/telemetry-data-collection.html


vuntz:
I think we needto run the polling agent on the compute nodes.

jsuchome:
Hm, right, looks like we got it wrong initially

jsuchome:
Hm, but according to http://docs.openstack.org/developer/ceilometer/architecture.html we should either use notification-agent or polling, right? We now configure both.

vuntz:
In the doc, I see:

> The polling agent daemon is configured to run one or more pollster plugins using either the ceilometer.poll.compute and/or ceilometer.poll.central namespaces.

I understand this as: we need a polling agent on at least a controller too, and the config is different on compute nodes and this controller. Am I getting this wrong?

jsuchome:
@vuntz I was also thinking about leaving central and agent roles as they were, remove polling role (but install polling package everywhere) and adapt their init scripts to run ceilometer-polling command, like it is here: http://docs.openstack.org/admin-guide/telemetry-data-collection.html

mmnelemane:
@jsuchome: I think `ceilometer-agent-compute` is still functional and sends hypervisor-specific data to `ceilometer-notification-agent` on controller. Earlier PR was to replace only the `ceilometer-agent-central` role.

jsuchome:
Yes, that's the way ubuntu seems to be doing it

jsuchome:
@mmnelemane How could it be? 
```
root@d52-54-77-77-77-02:~ # rpm -ql openstack-ceilometer-agent-compute
/etc/init.d/openstack-ceilometer-agent-compute
/usr/sbin/rcopenstack-ceilometer-agent-compute
/usr/share/polkit-1/rules.d/11-openstack-ceilometer.rules
root@d52-54-77-77-77-02:~ # grep startproc /etc/init.d/openstack-ceilometer-agent-compute
        /sbin/startproc -s -p $PIDFILE -u $USER -t ${STARTUP_TIMEOUT:-5} -q /usr/bin/ceilometer-$DAEMON --config-file=$CONFFILE --logfile=$LOGFILE
root@d52-54-77-77-77-02:~ # grep "^DAEMON" /etc/init.d/openstack-ceilometer-agent-compute
DAEMON="agent-compute"
```

mmnelemane:
@jsuchome: Sorry: Yes 'ceilometer-polling --polling-namespaces compute' should behave as `ceilometer-agent-compute` as per this doc.  Earlier it was only meant to replace `ceilometer-agent-central`.

jsuchome:
So we agree that this PR is also wrong. I'll prepare a new one.

jsuchome:
So, here's the alternative approach: https://github.com/crowbar/crowbar-openstack/pull/373

PR #373, opened 2016-05-19
Do not use polling as a role, bring central role back.

jsuchome:
ceilometer-polling service will be used as the one being executed by central/compute/ipmi services with different arguments.

This needs to go together with https://build.opensuse.org/request/show/397382

Relevant documentation: 
- http://specs.openstack.org/openstack/ceilometer-specs/specs/kilo/merge-compute-and-central-agents.html
- http://docs.openstack.org/developer/ceilometer/architecture.html
- http://docs.openstack.org/admin-guide/telemetry-data-collection.html

### DIFF
--- a/chef/data_bags/crowbar/migrate/ceilometer/101_bring-central-role-back.rb
+++ b/chef/data_bags/crowbar/migrate/ceilometer/101_bring-central-role-back.rb
@@ -1,0 +1,44 @@
+def upgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+  d["element_run_list_order"] = td["element_run_list_order"]
+
+  # ceilometer-central might be already present if the setup is an upgrade from Cloud6
+  unless d["elements"].key? "ceilometer-central"
+    d["elements"]["ceilometer-central"] = d["elements"]["ceilometer-polling"]
+    d["elements"].delete("ceilometer-polling")
+  end
+
+  # Update the run_list for controller node
+  nodes = NodeObject.find("run_list_map:ceilometer-polling")
+  nodes.each do |node|
+    node.add_to_run_list("ceilometer-central",
+                         td["element_run_list_order"]["ceilometer-central"],
+                         td["element_states"]["ceilometer-central"])
+    node.delete_from_run_list("ceilometer-polling")
+    node.save
+  end
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+  d["element_run_list_order"] = td["element_run_list_order"]
+
+  d["elements"]["ceilometer-polling"] = d["elements"]["ceilometer-central"]
+  d["elements"].delete("ceilometer-central")
+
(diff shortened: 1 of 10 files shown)


vuntz:
The (not so) good news is that we need to deal with setup that are using polling already, which means disabling the service, removing the HA resource, etc.

jsuchome:
Hm, right. But that we can save for Cloud6 port, I believe we can ignore it in master branch.

aplanas:
If this fix bsc#979410 we need to mention that in the commit message

Code-wise LGTM

mmnelemane:
+1

chef/data_bags/crowbar/template-ceilometer.json
```diff
@@ -36,25 +36,25 @@
     "ceilometer": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 23,
+      "schema-revision": 100,
       "element_states": {
         "ceilometer-server": [ "readying", "ready", "applying" ],
-        "ceilometer-polling": [ "readying", "ready", "applying" ],
+        "ceilometer-cagent": [ "readying", "ready", "applying" ],
```
vuntz: I'd take this opportunity to not use cagent, but central, or central-agent.

vuntz:
LGTM after a quick look.

chef/data_bags/crowbar/template-ceilometer.json
jsuchome: Right, good idea.

chef/data_bags/crowbar/migrate/ceilometer/101_bring-central-role-back.rb
```diff
@@ -0,0 +1,41 @@
+def upgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+  d["element_run_list_order"] = td["element_run_list_order"]
+
+  d["elements"]["ceilometer-central"] = d["elements"]["ceilometer-polling"]
```
vuntz: Can you protect this line with something like `unless d["elements"].key? "ceilometer-central"` so that we won't overwrite the attribute when upgrading from Cloud 6 while the migration already ran in Cloud 6.

aplanas:
+1

chef/data_bags/crowbar/migrate/ceilometer/101_bring-central-role-back.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+  d["element_run_list_order"] = td["element_run_list_order"]
+
+  d["elements"]["ceilometer-polling"] = d["elements"]["ceilometer-central"]
```
vuntz: Technically speaking, you should do a similar check here, but for all practical matters, it probably doesn't matter much.

vuntz:
+1

aspiers:
Corresponding doc bug filed: https://bugzilla.suse.com/show_bug.cgi?id=1027855

PR #376, opened 2016-05-23
Do not use polling as a role, bring central role back.

jsuchome:
ceilometer-polling service will be used as the one being executed
by central/compute/ipmi services with different arguments.

This should also fix bsc#979410.

Backport of https://github.com/crowbar/crowbar-openstack/pull/373. Compared to that one, we need to  take care of existing polling service and pacemaker resource....


aplanas:
s/ceilomter-polling/ceilometer-pooling/ (last commit message)

jsuchome:
I tested it with Cloud6 HA deployment and it seems to work well

chef/data_bags/crowbar/migrate/ceilometer/024_bring-central-role-back.rb
```diff
@@ -0,0 +1,41 @@
+def upgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
```
aplanas: this is to fix a missing migration?

jsuchome: The migration script is actually the same one as used with revision 023, when `cagent` role was renamed to `polling`

aplanas:
+1

chef/cookbooks/ceilometer/recipes/central_ha.rb
```diff
@@ -56,4 +49,23 @@
...
+# The correct setup should involve ceilometer-polling to be executed
+# by ceilometer-central service only.
+crowbar_pacemaker_order_only_existing "o-ceilometer-polling" do
+  action :delete
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+pacemaker_primitive "ceilometer-polling" do
+  agent node[:ceilometer][:ha][:polling][:agent]
+  op node[:ceilometer][:ha][:polling][:op]
```
vuntz: This doesn't exist anymore. That being said, you can just skip it for stop/delete.

chef/data_bags/crowbar/migrate/ceilometer/024_bring-central-role-back.rb
```diff
@@ -0,0 +1,41 @@
...
+  d["element_order"] = td["element_order"]
+  d["element_run_list_order"] = td["element_run_list_order"]
+
+  d["elements"]["ceilometer-central"] = d["elements"]["ceilometer-polling"]
```
vuntz: I just realized: we need to fix this migration in master, as it will run again (for 101) and break because we will overwrite the current element.

vuntz:
+1. Will you submit a fix for the migration in master, or do you want me to do it?

chef/data_bags/crowbar/migrate/ceilometer/024_bring-central-role-back.rb
jsuchome: The 101 migration is part of master PR, still not merged: https://github.com/crowbar/crowbar-openstack/pull/373. I'm not sure if I understand what should be different there; 101 is migrating from 100, which has polling role

vuntz: Ah, I'll reply there.

vuntz:
> (cherry picked from commit 95546e2)

Unfortunately, that commit id is now wrong :/

jsuchome:
@vuntz so what do I do, just change the commit message here? The PR looks correct..

vuntz:
> @vuntz so what do I do, just change the commit message here? The PR looks correct..

Yeah, it's just the commit message which needs to be changed. It's sad that it will re-trigger gating, though :/

vuntz:
@jsuchome mentions that only the commit message is different, so my +1 is still valid.

aplanas:
+1

PR #377, opened 2016-05-25
Fix Heat WaitCondition URL points to internalURL Endpoint (bug#1459414)

mjura:
When using the WaitCondition resource in Heat the

    { get_attr: ['wait_handle', 'curl_cli'] }

call will return a internal URL as endpoint not a public one.
Since those handles are used to trigger handles from inside the instance,
the internal IP in such cases is usually not usable.

https://bugs.launchpad.net/openstack-ansible/+bug/1459414


jgrassler:
+1

jsuchome:
+1

PR #384, opened 2016-05-31
Move to keystonemiddleware instead of keystoneclient.middleware

vuntz:


toabctl:
+1 (but if you need to update the patch, please mention that ```keystoneclient.middleware``` is deprecated)

aplanas:
+1

PR #386, opened 2016-05-31
Revert "neutron: Disable l2pop mechanism unless DVR is enabled"

vuntz:
This was a workaround for a HA issue, but the HA issue was actually
broken ordering of HA resources (see
f5e6477147533a4f6aeddcfcdbb6e6d0cf5d1f0c)

So we should be able to enable l2pop by default.

This reverts commit 92a316a639acc1004e930a371fe53b76c3fc95af.


abel-navarro:
+1

nkrinner:
+1

PR #388, opened 2016-05-31
Add "role recipes" with the goal of simplifying the roles

jsuchome:
We want the roles to only reference one recipe, so that we can put some
intelligence in that recipe:

 - in a first step, we will stop changing the run list of a node
   depending on its state, and instead have the role recipe "decide"
   whether something should be done or not, depending on the state. This
   will solve the issue that the search results for nodes are not
   reliable when a node is in some state that is not "readying",
   "applying" or "ready".

   Very concretely: many recipes are using searches to find attributes
   of another barclamp. The recipes do not expect the search to return
   an empty result (and that's okay, because we know that there must be
   a glance server when we deploy nova, due to barclamp dependencies),
   but that could happen in the past when the node was in "problem" or
   rebooting.

 - in a later step, we might want to have a new attribute that will tell
   us which role to apply instead of applying everything.

More generally speaking, it's actually wrong to pretend that a node
stops having a role because it's, say, rebooting. A glance server is
always a glance server.

(cherry picked from commit 1dcce8690cc0c0f73436a249e901671d813620bb)

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/role_ceilometer_agent.rb
+++ b/chef/cookbooks/ceilometer/recipes/role_ceilometer_agent.rb
@@ -14,13 +14,4 @@
 # limitations under the License.
 #
 
-barclamp = "ceilometer"
-role = "ceilometer-agent"
-
-# if nil, then this means all states are valid
-states_for_role = node[barclamp]["element_states"][role]
-
-if states_for_role.nil? || states_for_role.include?("all") || states_for_role.include?(node[:state])
-  include_recipe "ceilometer::agent"
-  include_recipe "ceilometer::common"
-end
+include_recipe "ceilometer::agent"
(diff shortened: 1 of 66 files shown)


aplanas:
+1

chef/cookbooks/ceilometer/recipes/role_ceilometer_agent.rb
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+include_recipe "ceilometer::agent"
+include_recipe "ceilometer::common"
```
vuntz: You don't need `common` here (was removed from the role recently).

vuntz:
Small detail, but +1. Feel free to add a small commit or do a followup PR for the detail.

chef/cookbooks/ceilometer/recipes/role_ceilometer_agent.rb
jsuchome: Ah, true, I missed the agent role

vuntz:
+1

PR #389, opened 2016-06-02
neutron: Switch to LBaaSv2

vuntz:
This is not experimental anymore since Liberty.

### DIFF
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -82,6 +82,8 @@ when "suse"
     zvm_agent_name: "openstack-neutron-zvm-agent",
     lbaas_agent_pkg: "openstack-neutron-lbaas-agent",
     lbaas_agent_name: "openstack-neutron-lbaas-agent",
+    lbaasv2_agent_pkg: "openstack-neutron-lbaas-agent",
+    lbaasv2_agent_name: "openstack-neutron-lbaasv2-agent",
     lbaas_haproxy_group: "haproxy",
     metadata_agent_name: "openstack-neutron-metadata-agent",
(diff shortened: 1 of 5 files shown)


toabctl:
commit msg: ```Switch to ...``` instead of ```Ship to``` ?

vuntz:
> commit msg: Switch to ... instead of Ship to ?

Oops, thanks :-)

mjura:
This change works fine!!! But as I have tested Magnum supports only lbaasv1 !!!
https://bugs.launchpad.net/puppet-magnum/+bug/1583024

vuntz:
@mjura gah. So I'll just use the patch I have for the stable/3.0 branch that adds an option to pick between lbaasv1 and lbaasv2...

mjura:
@vuntz: this is perfect idea

mjura:
Btw. upstream patches for Magnum and LBAAS v2 are still not ready.

toabctl:
> Btw. upstream patches for Magnum and LBAAS v2 are still not ready.

Is Magnum in Newton ready? Because LBaaSv1 was removed in Newton. See http://docs.openstack.org/releasenotes/neutron-lbaas/newton.html#new-features

toabctl:
Looks like Magnum supports LBaaSv2 (which is the default for devstack)

toabctl:
> @mjura gah. So I'll just use the patch I have for the stable/3.0 branch that adds an option to pick between lbaasv1 and lbaasv2...

@vuntz for SOC7 (on Newton) we don't need the possibility to select between v1 and v2. There is only v2.

vuntz:
As discussed on IRC: we should take the first few commits of https://github.com/crowbar/crowbar-openstack/pull/513 if we want lbaasv2 quickly. The code here is just incomplete and some old version of the commits there.

PR #391, opened 2016-06-02
swift: Correctly configure memcache for object-expirer (bsc#979067)

vuntz:
We were not defining the servers.

https://bugzilla.suse.com/show_bug.cgi?id=979067


chef/cookbooks/swift/recipes/storage.rb
```diff
@@ -44,6 +44,20 @@
...
+memcached_ips = ""
+env_filter = " AND swift_config_environment:#{node[:swift][:config][:environment]}"
+result = search(:node, "roles:swift-proxy #{env_filter}")
+if !result.nil? && !result.empty?
+  memcached_servers = result.map do |x|
+    "#{Swift::Evaluator.get_ip_by_type(x, :admin_ip_expr)}:11211 "
+  end
+  memcached_servers.sort!
+  log("memcached servers" + memcached_servers.join(",")) { level :debug }
+  memcached_ips = memcached_servers.join(",")
```
toabctl: change these 2 lines. then you can do ```log("memcached servers" + memcached_ips)```

toabctl: Beside that, +1

jsuchome:
DRY: couldn't you extract that functionality to some library?

vuntz:
> DRY: couldn't you extract that functionality to some library?

The code is really just bad, I rewrote it; don't think we need to move it to a library now. What do you think?

jsuchome:
Much better, thanks! I don't think the library is needed.

I think you might add that `sort` command directly after the  `end` of those `map` searches.

vuntz:
@jsuchome good point, done.

jsuchome:
+1

toabctl:
+1

PR #393, opened 2016-06-02
Use role recipe for magnum-server

mmnelemane:
Change magnum barclamp to use role_<> recipe for simplified role for magnum-server . Depends on the actual barclamp being merged : https://github.com/crowbar/crowbar-openstack/pull/360

### DIFF
--- a/chef/cookbooks/magnum/recipes/sql.rb
+++ b/chef/cookbooks/magnum/recipes/sql.rb
@@ -26,13 +26,14 @@ include_recipe "#{db_settings[:backend_name]}::python-client"
 
 crowbar_pacemaker_sync_mark "wait-magnum_database"
 
+is_cluster_founder = CrowbarPacemakerHelper.is_cluster_founder?(node)
 # Create the Magnum Database
 database "create #{node[:magnum][:db][:database]} database" do
   connection db_settings[:connection]
   database_name node[:magnum][:db][:database]
   provider db_settings[:provider]
   action :create
-  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  only_if { !ha_enabled || is_cluster_founder }
 end
 
 database_user "create magnum database user" do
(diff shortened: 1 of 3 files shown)


vuntz:
I guess only the last commit is worth reviewing -- rest comes from the other PR.

jsuchome:
For the last commit: we currently use only the simplest way, no checking for `element_states`. I think it would be better to do it in line with other role recipes.

vuntz:
Can you rebase, and add a commit for https://github.com/crowbar/crowbar-openstack/pull/360/files#r65667609 ?

chef/cookbooks/magnum/recipes/sql.rb
```diff
@@ -26,13 +26,14 @@
 
 crowbar_pacemaker_sync_mark "wait-magnum_database"
 
+is_cluster_founder = CrowbarPacemakerHelper.is_cluster_founder?(node)
```
jsuchome: Maybe I'm totally wrong, but are we sure this is executed at the same time as in that `only_if` section?

mmnelemane: @jsuchome: I actally expect it wo work on the similar lines as the `'ha_enabled'`flag inside `only_if`

jsuchome: It's different kind of check, but I believe it should work, yes.

mmnelemane: @jsuchome, @vuntz: any idea why the patch fails to apply. I verified that its up-to-date and rebased with master.

vuntz: The package needs to be updated to have the other PR. Doing that right now; feel free to restart the job in 5 minutes.

jsuchome:
+1

aplanas:
after that: +1

mmnelemane:
@aplanas : Done. Thanks

jsuchome:
Still +1

PR #394, opened 2016-06-03
trove: Fix configuration when other services are using HA (bsc#970648)

vuntz:
We should not reference the IP/hostname of a specific controller, but of
the VIP.

https://bugzilla.suse.com/show_bug.cgi?id=970648


nicolasbock:
+1

aplanas:
+1

PR #346, opened 2016-03-31
Mitaka porting

dirkmueller:
Various changes to get mitaka openstack deployable

PR #395, opened 2016-06-03
Add missing nova Api DB attributes

toabctl:
Since mitaka, nova has a second DB for the API. We already setup the DB
but the chef attributes were not there yet.
This is similar to what we already do for the other nova db.


toabctl:
Related: https://github.com/crowbar/crowbar-openstack/pull/346#discussion_r60558046

aplanas:
+1

vuntz:
Do we need to add the attributes to the proposal template too?

dirkmueller:
@vuntz the attributes are already in the schema defined as optional ones, so that should be good afaics. 

+1

dirkmueller:
(rerunning PR tests for this prior to merging)

PR #399, opened 2016-06-07
Fixed os-auth-url argument to openstack command

jsuchome:
It's a fix for current ugly code, which would rather require rewrite... but still we might want to have this as a quick fix.

UPDATE: Cloud6 is fine, it's using `--os-token` with `--os-url`


vuntz:
+1

aplanas:
+1

sayalilunkad:
+1

dirkmueller:
+1

PR #400, opened 2016-06-07
Save stack user domain id into node data.

jsuchome:
This makes the code bit more readable.

It is supposed to be better alternative to https://github.com/crowbar/crowbar-openstack/pull/399 . Not yet tested.

### DIFF
--- a/chef/data_bags/crowbar/template-heat.json
+++ b/chef/data_bags/crowbar/template-heat.json
@@ -10,6 +10,7 @@
       "database_instance": "none",
       "stack_domain_admin": "heat_domain_admin",
       "stack_domain_admin_password": "",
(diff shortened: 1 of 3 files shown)


chef/data_bags/crowbar/template-heat.json
```diff
@@ -42,7 +43,7 @@
     "heat": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 101,
+      "schema-revision": 102,
```
vuntz: You don't need to change anything in the template to add an attribute in the node from the cookbook, so you can avoid changing this file.

jsuchome:
Too old, not needed any more

PR #401, opened 2016-06-08
nova: Allow some SUSE specific vendordata

toabctl:
Create a vendordata.json file which will be served through the metadata
server to the guest VM.
This vedordata.json contains some information about a possible SMT server.
The provided information is then used by the SUSE guest images to automatically
configure the repositories for the guest.
This is based on a solution the public cloud team is already using.

### DIFF
--- a/chef/cookbooks/nova/recipes/config.rb
+++ b/chef/cookbooks/nova/recipes/config.rb
@@ -260,6 +260,18 @@ else
   bind_port_xvpvncproxy = node[:nova][:ports][:xvpvncproxy]
 end
 
+vendordata_jsonfile = "/etc/nova/suse-vendor-data.json"
+
+template vendordata_jsonfile do
+  source "suse-vendor-data.json.erb"
+  user "root"
+  group node[:nova][:group]
+  mode 0640
+  variables(
+    vendor_data: node[:nova][:metadata][:vendordata][:json]
+  )
+end
+
 template "/etc/nova/nova.conf" do
   source "nova.conf.erb"
   user "root"
@@ -290,6 +302,7 @@ template "/etc/nova/nova.conf" do
             glance_server_insecure: glance_server_insecure || keystone_settings["insecure"],
             metadata_bind_address: metadata_bind_address,
             vnc_enabled: node[:kernel][:machine] != "aarch64",
+            vendordata_jsonfile: vendordata_jsonfile,
             vncproxy_public_host: vncproxy_public_host,
             vncproxy_ssl_enabled: api[:nova][:novnc][:ssl][:enabled],
             vncproxy_cert_file: api_novnc_ssl_certfile,
--- a/chef/cookbooks/nova/templates/default/suse-vendor-data.json.erb
+++ b/chef/cookbooks/nova/templates/default/suse-vendor-data.json.erb
@@ -1,0 +1,1 @@
+<%= @vendor_data %>
--- a/chef/data_bags/crowbar/migrate/nova/102_nova_add_metadata_vendordata.rb
+++ b/chef/data_bags/crowbar/migrate/nova/102_nova_add_metadata_vendordata.rb
@@ -1,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "metadata"
+    a["metadata"] = ta["metadata"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless ta.key? "metadata"
+    a.delete("metadata")
+  end
+  return a, d
+end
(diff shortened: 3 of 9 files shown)


toabctl:
See also https://trello.com/c/oqPGr88I/8-enhance-user-plane-in-soc

nicolasbock:
+1

chef/cookbooks/nova/recipes/config.rb
```diff
@@ -260,6 +260,32 @@
   bind_port_xvpvncproxy = node[:nova][:ports][:xvpvncproxy]
 end
 
+# NOTE(toabctl): key names "fingerprint", "SMTserverName" and "SMTserverIP"
+# must match the names in regionserverclient
+# see https://github.com/SUSE/Enceladus/tree/master/regionServiceClient
```
dirkmueller: why this awkward conversion of the key names? is there a problem with like just copying the content from the barclamp config 1:1 ?

toabctl: @dirkmueller the keynames must match the expected names from the regionserverclient. we could rename ```[:nova][:metadata][:vendordata][:suse_smt_fingerprint]``` and the other 2 keys but I prefer the lower case names with the suse prefix

dirkmueller: Not sure if we talk about the same thing - my suggestion was to just allow an arbitrary hash to be injected in the barclamp and pass that down to the config file for nova 1:1

toabctl: @dirkmueller ah.ok. Fixed. The only downside now is, that you need to escape the json you want to add in the raw view.

nicolasbock:
+1

crowbar_framework/config/locales/nova/en.yml
```diff
@@ -90,3 +90,4 @@ en:
         assigned_node: 'Node %{key} has been assigned to a nova-compute role more than once.'
         assigned_remotes: 'Remotes %{key} has been assigned to a nova-compute role more than once.'
         assigned_node_and_remote: 'Node %{node} has been assigned to a nova-compute role as individual node and as remote node of cluster %{cluster}.'
+        vendor_data_invalid_json: "The provided vendorodata for the metadata server is invalid json."
```
aplanas: vendorodata/vendordata

aplanas:
after the typo: +1

crowbar_framework/config/locales/nova/en.yml
toabctl: done

toabctl:
Hm. Now with the possibility to add any json, the json needs to be escaped in the raw view (because the raw view itself is json). But then the escape chars are not removed so we end up with something like:
```
# curl 169.254.169.254/openstack/latest/vendor_data.json
"{\"key\": \"value\"}"
```

toabctl:
It's working now. The json in the rawview must be escaped but it's deescaped when it's written to the file.

toabctl:
Working now. The string in the raw view must be escaped but in the file there is the unescaped json:
```
# cat /etc/nova/suse-vendor-data.json 
{"key": "value"}
```
and from the metadata server:
```
# curl http://169.254.169.254/openstack/latest/vendor_data.json
{"key": "value"}
```

aplanas:
+1

nicolasbock:
+1

crowbar_framework/config/locales/nova/en.yml
```diff
@@ -90,3 +90,4 @@ en:
         assigned_node: 'Node %{key} has been assigned to a nova-compute role more than once.'
         assigned_remotes: 'Remotes %{key} has been assigned to a nova-compute role more than once.'
         assigned_node_and_remote: 'Node %{node} has been assigned to a nova-compute role as individual node and as remote node of cluster %{cluster}.'
+        vendor_data_invalid_json: "The provided vendordata for the metadata server is invalid json."
```
dirkmueller: nitpick :) I think its JSON not json

dirkmueller:
+1

chef/data_bags/crowbar/migrate/nova/102_nova_add_metadata_vendordata.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "metadata"
+    a["metadata"] = ta["metadata"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  if a.key? "metadata"
```
dirkmueller: unless ta.key? "metadata"

toabctl: It's in the downgrade path, ```if``` is correct imo

crowbar_framework/config/locales/nova/en.yml
toabctl: done

chef/cookbooks/nova/templates/default/suse-vendor-data.json.erb
```diff
@@ -0,0 +1 @@
+<%= @vendor_data %>
```
aspiers: Why use a template if the entire contents are determined by a variable?  In this case wouldn't the `file` resource be more appropriate?

PR #402, opened 2016-06-09
first attempt at setting up mention-bot

aspiers:
This is just a test setup for https://github.com/facebook/mention-bot but it's hopefully close to what we would want as a final configuration, although obviously we would probably want to add many more entries for all the OpenStack components and the team members working on them.

See https://github.com/facebook/mention-bot#configuration for more detail on the meaning of the configuration options.

Extended glob syntax is used; see https://github.com/facebook/mention-bot/issues/140

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -1,0 +1,43 @@
+{
+  "maxReviewers": 3,
+  "numFilesToCheck": 10,
+  "message": "@reviewers, please review this.",
+  "alwaysNotifyForPaths": [
+    {
+      "name": "toabctl",
+      "files": [
+        "{**/,}*+(cinder|manila)*{/**,}"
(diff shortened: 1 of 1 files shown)


aspiers:
Obviously we don't need to wait for the gate to pass before merging this.  If everyone's OK with merging now, I can set up the webhook and start testing this afternoon :)

aspiers:
BTW currently this is configured to only act on *new* PRs.  So my plan is simply to submit 2 or 3 test PRs modifying various files matching and not matching the globs configured, and check it's behaving as expected.

rhafer:
+1

toabctl:
+1

nicolasbock:
+1

PR #404, opened 2016-06-10
trigger mention-bot on PR "synchronize" events

aspiers:
A PR could be updated via a git push, in which case github will send a PR event of type `synchronize` to the webhook.  We want the mention-bot to react to this, because the push probably introduces new changes which might require additional reviewers.

This also makes testing the bot easier :-)

There appears to be little documentation covering exactly when `synchronize` is invoked, but there is this, at least:

  https://developer.github.com/v3/activity/events/types/#pullrequestevent

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -33,7 +33,7 @@
   "userBlacklist": [],
   "userBlacklistForPR": [],
   "requiredOrgs": ["crowbar"],
-  "actions": ["opened"],
+  "actions": ["opened", "synchronize"],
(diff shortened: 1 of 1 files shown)


aspiers:
Merging because obviously doesn't impact anything except for my testing, and everyone else went home ... part-timers ;-p

aspiers:
For some weird reason this did not work - still seeing this in the logs:

    Skipping because action is synchronize. We only care about: "opened"

PR #407, opened 2016-06-17
Switch from closure-compiler to uglifier

rsalevsky:
Since we switched to SLE12 we can use the faster uglifier which uses NodeJS.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@
 source "https://rubygems.org"
 
 group :development do
-  gem "closure-compiler", "~> 1.1.10"
+  gem "uglifier", "~> 2.7.2"
   gem "sass", "~> 3.2.19"
   gem "sprockets-standalone", "~> 1.2.1"
   gem "sprockets", "~> 2.11.0"
--- a/Rakefile
+++ b/Rakefile
@@ -32,7 +32,7 @@ begin
     task.compress = true
     task.digest = true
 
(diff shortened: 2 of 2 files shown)


rsalevsky:
The gating will fail due to dependencies issues which needs to be adapted when this PR is merged.

MaximilianMeister:
+1

nicolasbock:
+1

dirkmueller:
@rsalevsky  the reason that we didn't use uglifier before is because we didn't want to maintain nodejs. is it maintained now for us?

rsalevsky:
@dirkmueller It's currently not maintained but crowbar-init requires nodejs to run. crowbar-init requires rubygem-bootstrap-sass which has an dependency for execjs (nodejs). Removing rubygem-bootstrap-sass creates in my opinion more work then supporting nodejs. The plan is to use gem's for all our assets do maintain them just in one place instead of maintaining them in each package (crowbar and crowbar-init ) separate. Another point is that the design team is working on an angular js port of crowbar which also requires nodejs.

rsalevsky:
I discussed this with @vuntz, I will maintain nodejs for the cloud product and try to convince other projects to use the same nodejs for there product. If I get over the next months the feeling it is to much work to maintain it I will use nodejs just as build depencie like portus is doing it. But this means we can't use shared rubygems for assets between crowbar and crowbar-init.

PR #409, opened 2016-06-20
Remove unused oboleted crowbar_magnum script

rsalevsky:
The bin scripts are oboleted by the crowbar-client in Cloud 7 and
should not been used anymore.


MaximilianMeister:
Don't we still use them somewhere in mkcloud?
btw. there is a small typo in the commit message

dirkmueller:
-1, they're not unused, as gating shows. Also I think we should at least give a bit of grace time by printing out a warning and then calling crowbarctl.

rsalevsky:
Right, I overlooked a wrapper function in qa_crowbarsetup so this needs some more work then just removing. But can you define what do you mean with a bit more grace time?

rsalevsky:
Changed the PR to just remove crowbar_magnum for now to remove it before it get's used.

dirkmueller:
@rsalevsky I imagine replacing the wrappers with something like

```bash

echo "DEPRECATED - this command will dissapear by Cloud 8"
echo "forwarding request to 'crowbarctl proposal $2 $1  $3'
exec crowbarctl proposal $2 $1 $3
```

dirkmueller:
+1 for just removing magum (no point in keeping compatibility for new stuff)

tboerger:
> @rsalevsky I imagine replacing the wrappers with something like

Cloud 6 have already been released with a deprecation of the old crowbar scripts (just look at the changelog), so the plan have been to drop it entirely for cloud 7.

MaximilianMeister:
+1

but the PR title doesn't apply anymore

PR #405, opened 2016-06-10
[DO NOT MERGE] Mention bot test 1.1

aspiers:
Another test of #403.

PR #410, opened 2016-06-21
[mention-bot] only mention SUSE employees for now

MaximilianMeister:
as we don't have any contributors outside the SUSE org it does not
make sense to mention people that haven't contributed for a long
time.
This prevents spamming for now, and if crowbar grows to form an
infrastructure for a wider community we can adapt the list again.
Main reason for this change is to stop spamming non-active
contributors.

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -32,7 +32,7 @@
   "fileBlacklist": [],
   "userBlacklist": [],
   "userBlacklistForPR": [],
-  "requiredOrgs": ["crowbar"],
+  "requiredOrgs": ["SUSE"],
   "actions": ["opened", "synchronize"],
   "skipAlreadyAssignedPR": true,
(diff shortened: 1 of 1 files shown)


dirkmueller:
The suse org isn't public, right?

aspiers:
@dirkmueller I don't think there's such a thing as a private github organization. Public/private only applies to repos (and maybe teams, sort of).

This is :+1: from me, I think it should probably work, even though it's the opposite of what I said here: https://github.com/crowbar/crowbar-openstack/pull/405#issuecomment-225349159

dirkmueller:
I'm not against it, so lets merge this then..

aspiers:
Note that for some weird reason this will probably not take effect yet - the bot seems to be ignoring the updated config.  Some strange heroku quirk I didn't figure out yet.

PR #411, opened 2016-06-21
Add segment_mtu setting for neutron

nicolasbock:
For Layer 2 traffic, neutron needs to know the segment_mtu to correctly
calculate the the MTU of networks. The `mtu_value` has to match the MTU values
of the physical interfaces.

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -183,7 +183,9 @@ when "ml2"
     physnets.push(node[:neutron][:zvm][:zvm_xcat_mgt_vswitch])
   end
 
-  mtu_value = 0
+  os_sdn_net = Barclamp::Inventory.get_network_definition(node, "os_sdn")
+  mtu_value = os_sdn_net.nil? ? 1500 : os_sdn_net["mtu"].to_i
+
   ml2_type_drivers = node[:neutron][:ml2_type_drivers]
   ml2_mechanism_drivers = node[:neutron][:ml2_mechanism_drivers].dup
   if use_hyperv
@@ -194,7 +196,6 @@ when "ml2"
   end
   if ml2_type_drivers.include?("gre") || ml2_type_drivers.include?("vxlan")
     ml2_mechanism_drivers.push("l2population") if node[:neutron][:use_dvr]
-    mtu_value = 1400
   end
 
   ml2_mech_drivers = node[:neutron][:ml2_mechanism_drivers]
@@ -219,7 +220,7 @@ when "ml2"
       vxlan_end: vni_end,
       vxlan_mcast_group: node[:neutron][:vxlan][:multicast_group],
       external_networks: physnets,
-      path_mtu: mtu_value
+      mtu_value: mtu_value
     )
   end
 when "vmware"
--- a/chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
@@ -37,12 +37,13 @@ mechanism_drivers = <%= @ml2_mechanism_drivers.join(",") %>
 # path_mtu - max encap header size).  If <=0, the path MTU is
 # indeterminate and no calculation takes place.
 # path_mtu = 0
-path_mtu = <%= @path_mtu %>
+path_mtu = <%= @mtu_value %>
 
 # (IntOpt) Segment MTU.  The maximum permissible size of an
 # unfragmented packet travelling a L2 network segment.  If <=0,
 # the segment MTU is indeterminate and no calculation takes place.
 # segment_mtu = 0
+segment_mtu = <%= @mtu_value %>
 
 # (ListOpt) Physical network MTUs.  List of mappings of physical
 # network to MTU value.  The format of the mapping is


chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -180,7 +180,9 @@
     physnets.push(node[:neutron][:zvm][:zvm_xcat_mgt_vswitch])
   end
 
-  mtu_value = 0
+  # Make sure to match physical NIC MTU.
+  mtu_value = 1500
```
rhafer: To be really correct this should be set to the MTU value of the os_sdn network, when the "gre" and/or "vxlan" type drivers are enabled I think. Especially when https://github.com/crowbar/crowbar-core/pull/503 is getting merged.

Or am I missing something?

nicolasbock: Yes, I think you are right. I'll fix that.

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -180,7 +180,11 @@
     physnets.push(node[:neutron][:zvm][:zvm_xcat_mgt_vswitch])
   end
 
-  mtu_value = 0
+  mtu_value = 1500
```
dguitarbite: I would suggest using 1400 as the MTU value here. There were known issues especially with SSH access to the VM due to MTU size being 1500 I guess due to the GRE encapsulation overhead. The real number is around 1450~ but 1400 does not hurt and allows more room for other unforeseen issues.

nicolasbock: Thanks for the info. I'll go back down to 1400.

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -191,7 +195,6 @@
   end
   if ml2_type_drivers.include?("gre") || ml2_type_drivers.include?("vxlan")
     ml2_mechanism_drivers.push("l2population") if node[:neutron][:use_dvr]
-    mtu_value = 1400
```
dguitarbite: Does this ``if`` block make any sense if we remove mtu_value? And ofcourse I would suggest studying if the MTU value of the hardware is really going to be the usable value in case of GRE. What I can think of off the top of my head is something like ``mtu_value = mtu_value - GRE_overhead_offset``. I could be totally wrong but I get a funny feeling I am not.

chef/cookbooks/neutron/recipes/server.rb
rhafer: > There were known issues especially with SSH access to the VM due to MTU size being 1500 I guess due to the GRE encapsulation overhead.

Sorry, but that is wrong. Neutron will nowadays automatically account for the Tunneling overhead. I.e. when using GRE networks the dnsmasq will be configured to set the MTU option to 1450. You can keep the default to 1500 here.

nicolasbock: Thanks. That was my interpretation of the documentation as well.

chef/cookbooks/neutron/recipes/server.rb
rhafer: > I could be totally wrong but I get a funny feeling I am not, unless this logic has been covered in Neutron's code itself.

Yeah neutron already handles that. (IIRC since liberty)

dguitarbite: Ok, that makes it much simpler to write this :).

chef/cookbooks/neutron/recipes/server.rb
dguitarbite: Thanks for correcting this Ralf, I need to update my knowledge.

dguitarbite:
+1

chef/cookbooks/neutron/recipes/server.rb
dguitarbite: Although, do we need this if block? IMO the if block is probably not useful anymore.

nicolasbock: Why do you think it's not useful anymore? I don't quite understand what it does, but removing the if would change the logic of that piece of code, with or without the line that sets the `mtu_value`.

dguitarbite: ``ml2_mechanism_drivers.push("l2population") if node[:neutron][:use_dvr]`` I really think that the ``:use_dvr`` bit makes it obvious that its OVS/VXLAN so the if block kind of becomes redundant. We should confirm this part and may be this is another commit. But I think we should consider removing the if block.

The if block is trying to check which plugins are being configured which was required for the MTU issue previously.

nicolasbock: @dguitarbite Thanks for the explanation. I'll verify that.

vuntz:
LGTM, but I'm a bit confused as to why neutron has two settings for this (when would they differ?), so I'm not sure I fully understand all this.

vuntz:
cc @rossella

nicolasbock:
@vuntz I don't quite understand the difference either. I was using the description in [[1]](https://access.redhat.com/documentation/en/red-hat-enterprise-linux-openstack-platform/7/networking-guide/chapter-17-configure-mtu-settings) as a guide, and was assuming that the segment_mtu should be equal to the physical MTU coming from the network barclamp.

nicolasbock:
Since the reporter of [Bug 984128](https://bugzilla.suse.com/show_bug.cgi?id=984128) stated that their VM was using MTU 1500, and that they could fix the problem by setting their physical NICs to 1550, I figured that I should attempt the same. Unfortunately our physical NICs in `mkcloud1` don't allow for anything bigger than 1500, so I opted for 1200/1250 instead. I ran through the following steps:

- `mkcloud ... instnodes`
- Manually modify the `os_sdn_network` item in the `crowbar` data bag and added the `mtu` key with a value of 1250.
- `mkcloud proposal`
- Verify that the two networks `fixed` and `floating` have the expected MTU: `fixed` 1208, `floating` 1250.
- Create a `m1.tiny` instance, associate a floating IP and connect to console and verify MTU setting on `eth0`: 1208

Unfortunately I can't connect to the VM from the outside. I don't know whether that's due to a mismatch in the MTU values (as my laptop is using 1500) or some other problem. As far as I can tell at this point, the MTU value set in the data bag propagates all the way to neutron, however, I am not able to verify the customer's observed issue at this point.

rossella:
@vuntz agreed it's confusing. Let me try to explain, in Neutron you can set segment_mtu, path_mtu and you can set the mtu for every physical network you define.  When Neutron calculates the mtu for VLAN of flat network, it takes the smallest value between the segment_mtu and the mtu of the physical network used. If the network uses GRE or VXLAN as segmentation type, Neutron uses the min value between path_mtu and segment_mtu and automatically subtracts the size of the tunnel header. Hope it's more clear now.

chef/cookbooks/neutron/recipes/server.rb
nicolasbock: @dguitarbite You might be correct. I will have to dive deeper into the logic to understand this better before I feel comfortable to change this. What do you think of leaving it as it is for this PR?

dguitarbite: @nicolasbock Its fine by me to keep it as it is :). I agree with the fact that if something aint broken, dont fix it ...  although I see that statement to be the tip of the ice-berg ;). Ill create a noise card and address this in a different patch after your patch is merged.

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
```diff
@@ -43,6 +43,7 @@ path_mtu = <%= @path_mtu %>
 # unfragmented packet travelling a L2 network segment.  If <=0,
 # the segment MTU is indeterminate and no calculation takes place.
 # segment_mtu = 0
+#segment_mtu = <%= @segment_mtu %>
```
nicolasbock: @dguitarbite In 178427e08c6f6acc71d8e3ea004d6c441f22493d you enabled the `path_mtu` setting. Do you remember why you didn't also enable the `segment_mtu` setting?

dguitarbite:
+1 from my side. Just to address my paranoia, I would suggest trying to keep the segment_mtu value a bit lesser than the path mtu. As far as I understand from @rossella 's comment, I understood that for the case of GRE its not exactly trying to use the mtu value of the underlying physical network. I am sure that I may be missing the fact with GRE tunnels its the abstraction layer which may make my point irrelevant. Rossella, how does the GRE/VXLAN scenario consider the mtu values of the physical network??

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
dguitarbite: Yes, as far as I remember the segment_mtu was overridden if <= path_mtu! So
it made sense not to use it at the moment. As Rosella mentioned, now the
logic is more smarter so we should use it :).

nicolasbock: I found this explanation:
- Flat/VLAN: Minimum of (`segment_mtu` or `physical_network_mtus`)
- GRE: Minimum of (`segment_mtu`, `path_mtu`) minus (`GRE_ENCAP_OVERHEAD` which is 42 bytes)
- VXLAN: Minimum of (`segment_mtu`, `path_mtu`) minus (`VXLAN_ENCAP_OVERHEAD` which is 50 bytes)
which seems to suggest that for L3 routed traffic, both MTU values are considered. L2 routed traffic uses the `segment_mtu` value directly.

rossella: @dguitarbite @nicolasbock was faster than me in answering your question

rhafer:
> I would suggest trying to keep the segment_mtu value a bit lesser than the path mtu

That would mean that the ```path_mtu``` values would never be used anywhere since neutron will always take the lower value of ```segment_mtu``` and ```path_mtu``` when trying to figure out the mtu for a GRE/VXLAN based network.

> how does the GRE/VXLAN scenario consider the mtu values of the physical network?

When talking about "physical network" here I think we're referring to the physical networks that are defined in ```flat_networks``` and ```network_vlan_ranges``` in ml2_conf.ini and in the ```bridge_mappings``` in the ovs-agent configuration. GRE/VXLAN setups don't map to any of those physical networks so ```physical_network_mtus``` are not considered there.

rhafer:
+1 this looks good to me. Now we "just" need to figure out why the gating is failing.

It's also probably worth noting that for "master" we need a slightly different fix since ```segment_mtu``` has been deprecated there (I guess because of the confusion it caused). And a new value called ```global_physnet_mtu``` was introduced. (http://docs.openstack.org/releasenotes/neutron/mitaka.html#deprecation-notes)

rhafer:
Hm, IIRC this is addressing some bug in bugzilla. @nicolasbock please add a ```bsc#``` reference to the commit message's subject.

Also, do you already have this change forward ported to master somewhere?

nicolasbock:
@rhafer Thanks for the pointer regarding `master`. I haven't ported this yet to `master`, but will do that today. I'll also look into the gating problem.

rhafer:
Just for reference, the gating is currently failing because this one depends on https://github.com/crowbar/crowbar-core/pull/503, which is not merged yet.

rhafer:
> Just for reference, the gating is currently failing because this one depends on crowbar/crowbar-core#503, which is not merged yet.

Hm, thinking about this again. I think the code should handle the case where "mtu" is not present in the os_sdn network. (It's an optional attribute after all)

rhafer:
Unassigning myself here. As I will be on FTO for the next weeks.

nicolasbock:
Rebased

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -180,7 +180,11 @@
     physnets.push(node[:neutron][:zvm][:zvm_xcat_mgt_vswitch])
   end
 
-  mtu_value = 0
+  mtu_value = 1500
+  if node["crowbar"]["network"]["os_sdn"].key? "mtu"
```
vuntz: Gating fails here, probably because it's not the right attribute. I'd prefer if we use `get_network_by_type(node, "os_sdn")` to fetch the network data, anyway.

nicolasbock: ok, I will update it.

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -180,7 +180,9 @@
     physnets.push(node[:neutron][:zvm][:zvm_xcat_mgt_vswitch])
   end
 
-  mtu_value = 0
+  mtu_value = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "os_sdn").mtu
+  Chef::Log.info("setting os_sdn:mtu = #{mtu_value}")
```
vuntz: Do we really need the debug line?

Btw, we don't always have a os_sdn network, so this really be:
```ruby
os_sdn_net = Barclamp::Inventory.get_network_by_type(node, "os_sdn")
mtu_value = os_sdn_net.nil? ? 1500 : (os_sdn_net.mtu || 1500)
```

nicolasbock: I'll remove the debug line.

Unless I misunderstood how this works, the changes in https://github.com/crowbar/crowbar-core/pull/503/files#diff-d190f960c2b02d826739f1c7a8358960R314 should ensure that `os_sdn_net` will always have a value.

vuntz: There's simply no os_sdn network if we use linuxbridge+vlan :-)

nicolasbock: Hmm, ok, let me check that...

vuntz:
If we can `cherry-pick -x` the patch from master, I'll be happy ;-)

nicolasbock:
@vuntz cherry picking here would change `neutron.conf` as well and add some options that exist only for mitaka.

vuntz:
> @vuntz cherry picking here would change neutron.conf as well and add some options that exist only for mitaka.

As I said in another PR: "maybe mention the other commit even if it's not a cherry-pick then. The key point is that we want to have the link to the change in master."

vuntz:
+1, but requires the crowbar-core PR to be merged first for gating.

PR #412, opened 2016-06-22
Aarch64 Compute

dirkmueller:


chef/cookbooks/nova/templates/default/qemu.conf.erb
```diff
@@ -499,3 +516,43 @@ group = "<%= @group %>"
...
+# differently for different guest architectures. For instance, it's OVMF
+# for x86_64 and i686, but it's AAVMF for aarch64. The libvirt default
+# follows this scheme.
+#nvram = [
+#   "/usr/share/qemu/ovmf-x86_64-ms-code.bin:/usr/share/qemu/ovmf-x86_64-ms-vars.bin",
+#   "/usr/share/qemu/aavmf-aarch64-code.bin:/usr/share/qemu/aavmf-aarch64-vars.bin"
+#]
+
+nvram = [
+   "/usr/share/qemu/ovmf-x86_64-ms-code.bin:/usr/share/qemu/ovmf-x86_64-ms-vars.bin",
```
vuntz: AFAIK, we don't install any package for this one so we don't have it.

dirkmueller: Yeah, but thats a bug not a feature. we can install it though (it is part of SLES). do you want me to handle that in this PR?

vuntz: Ok, I'm fine leaving it then. And feel free to install the package for x86_64 too.

dirkmueller: added a commit for that

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -426,7 +426,7 @@
   use_interface_attach = true
   use_rescue = true
   use_suspend = true
-  use_vnc = true
+  use_vnc = node[:kernel][:machine] != "aarch64"
```
vuntz: Not sure it really matters, but I guess we should check the architecture of one of the compute node.

vuntz:
+1

chef/cookbooks/tempest/recipes/config.rb
dirkmueller: I was thinking about that, but decided against it. I think the implicit assumption is that we use tempest against the architecture environment where tempest is being deployed on. there is only one image to reference in the tempest.conf so running multi-architecture tempest doesn't work anyway..

This way you can choose to test aarch64 or x86_64 in a dual-architecture environment based on where you deploy tempest on. (yes, I know tempest doesn't handle aarch64 yet).

jsuchome:
+1. I do not know about Aarch64, but the code makes sense to me.

PR #413, opened 2016-06-22
Adds barbican as dependency for magnum

sayalilunkad:


sayalilunkad:
Depends on https://github.com/crowbar/crowbar-openstack/pull/392

mjura:
I'm wondering if we need this dependency, because it is recommended to use Barbican as Certificates Manager in Magnum, but it is not required. Magnum works as well also with 'local' driver for CertManager.

sayalilunkad:
As discussed we will be using barbican as cert-manager for magnum when barbican is deployed other wise use the local driver rather than having it as a dependency for magnum. So closing this PR.

PR #414, opened 2016-06-22
Sync hound configs with crowbar/crowbar

rsalevsky:


aspiers:
As per [my comment on the corresponding review in `crowbar-core`](https://github.com/crowbar/crowbar-core/pull/506#issuecomment-227902991), I might be OK with this change, but :-1: until a proper explanation and justification is given in the commit message.  Thanks in advance for the extra info ;-)

MaximilianMeister:
+1

rsalevsky:
Let's merge it, it's just a sync with the other crowbar repos.

PR #415, opened 2016-06-24
Access network info through API, instead of directly fetching attributes

vuntz:
This will help have the network info in only one location.

Should go with https://github.com/crowbar/crowbar-core/pull/510

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -91,6 +91,8 @@ directory "/var/cache/neutron" do
   only_if { node[:platform_family] == "debian" }
 end
 
+# accessing the network definition directly, since the node is not using this
+# network
 vlan_start = node[:network][:networks][:nova_fixed][:vlan]
 num_vlans = node[:neutron][:num_vlans]
 vlan_end = [vlan_start + num_vlans - 1, 4094].min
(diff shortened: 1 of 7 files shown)


chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -91,7 +91,8 @@
   only_if { node[:platform_family] == "debian" }
 end
 
-vlan_start = node[:network][:networks][:nova_fixed][:vlan]
+nova_fixed_net = Barclamp::Inventory.get_network_by_type(node, "nova_fixed")
```
rhafer: Don't you also need to access this directly  via the roles here? "nova_fixed" will not be enabled on the node when only the gre and/or vxlan drivers are enabled.

vuntz: You're right; I forgot to fix that back (I had the issue in `post_install_conf`).

vuntz: Fixed.

jsuchome:
Could you remove it from product, so the gating can pass?

vuntz:
> Could you remove it from product, so the gating can pass?

(for the record, since we discussed on irc) It's not in the package, I guess we're hitting a bug with jenkins that somehow reuses the package from the gating job :/

toabctl:
+1

rhafer:
+1, but AFAICS this needs to wait for https://github.com/crowbar/crowbar-core/pull/510 to be merged

PR #416, opened 2016-06-24
Support Cisco APIC Opflex integration

mmnelemane:
The changes here allow configuration of Cisco ACI as the ML2 backend for neutron. Cisco ACI uses Opflex to program the switches and expects the corresponding agents to be configured on the compute nodes. This PR does not do complete automation of the setup and expects some manual steps (Will be published in a separate PR) to be done on the compute nodes of the ACI before applying the barclamp to configure neutron services.  
Testing APIC support requires a number of packages to be manually installed and some manual steps on the Compute nodes:
The packages are temporarily available here: http://download.opensuse.org/repositories/Cloud:/OpenStack:/Liberty:/cisco-apic/SLE_12_SP1/

https://github.com/mfriesenegger/soc6-opflex-howto/blob/master/Installing%20and%20Configuring%20Cisco%20Opflex%20ML2%20configuration%20with%20SOC6%20barclamp.txt contains documentation for using the barclamp on ACI hardware.

UI changes are not updated yet, so the user cannot enable support for ACI from Crowbar UI . This needs to be taken up once the package issues are sorted out and can be part of a separate card.
(Branch for UI implementation: https://github.com/mmnelemane/crowbar-openstack/tree/apic_ui

Co-Authored with: @rhafer

### DIFF
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -88,6 +88,8 @@ when "suse"
     cisco_pkgs: ["python-networking-cisco"],
     cisco_apic_pkgs: ["python-apicapi",
                       "python-neutron-ml2-driver-apic"],
+    cisco_apic_gbp_pkgs: ["openstack-neutron-gbp",
+                          "python-gbpclient"],
     cisco_opflex_pkgs: ["agent-ovs",
                         "lldpd",
                         "openstack-neutron-opflex-agent"],
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -52,6 +52,11 @@ default[:neutron][:ssl][:insecure] = false
 default[:neutron][:ssl][:cert_required] = false
 default[:neutron][:ssl][:ca_certs] = "/etc/neutron/ssl/certs/ca.pem"
 
+default[:neutron][:apic][:system_id] = "SOC"
+default[:neutron][:apic][:hosts] = ""
+default[:neutron][:apic][:username] = "admin"
+default[:neutron][:apic][:password] = ""
+
 case node[:platform_family]
 when "suse"
   default[:neutron][:platform] = {
@@ -81,6 +86,13 @@ when "suse"
     nsx_pkgs: ["openvswitch-pki",
                    "ruby2.1-rubygem-faraday"],
     cisco_pkgs: ["python-networking-cisco"],
+    cisco_apic_pkgs: ["python-apicapi",
+                      "python-neutron-ml2-driver-apic"],
+    cisco_opflex_pkgs: ["agent-ovs",
+                        "openstack-neutron-opflex-agent"],
+    cisco_opflex_ovs_pkgs: ["openvswitch",
+                            "openvswitch-kmp-default",
+                            "openvswitch-switch"],
     user: "neutron",
     group: "neutron",
     neutron_rootwrap_sudo_template: "/etc/sudoers.d/openstack-neutron"
@@ -112,6 +124,7 @@ when "rhel"
     hyperv_pkg: "",
     nsx_pkgs: [""],
     cisco_pkgs: ["python-networking-cisco"],
+    cisco_apic_pkgs: [""],
     user: "neutron",
     group: "neutron",
     neutron_rootwrap_sudo_template: "/etc/sudoers.d/openstack-neutron"
@@ -144,6 +157,7 @@ else
     hyperv_pkg: "python-networking-hyperv",
     nsx_pkgs: [""],
     cisco_pkgs: [""],
+    cisco_apic_pkgs: [""],
     user: "neutron",
     group: "neutron",
     neutron_rootwrap_sudo_template: "/etc/sudoers.d/neutron-rootwrap"
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -89,6 +89,7 @@ when "suse"
     cisco_apic_pkgs: ["python-apicapi",
                       "python-neutron-ml2-driver-apic"],
     cisco_opflex_pkgs: ["agent-ovs",
+                        "lldpd",
                         "openstack-neutron-opflex-agent"],
     cisco_opflex_ovs_pkgs: ["openvswitch",
                             "openvswitch-kmp-default",
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -53,9 +53,19 @@ default[:neutron][:ssl][:cert_required] = false
 default[:neutron][:ssl][:ca_certs] = "/etc/neutron/ssl/certs/ca.pem"
 
 default[:neutron][:apic][:system_id] = "SOC"
-default[:neutron][:apic][:hosts] = ""
+default[:neutron][:apic][:hosts] = "10.105.1.10"
 default[:neutron][:apic][:username] = "admin"
-default[:neutron][:apic][:password] = ""
+default[:neutron][:apic][:password] = "cisco123"
+
+default[:neutron][:opflex][:peer_ip] = "10.0.0.30"
+default[:neutron][:opflex][:peer_port] = 8009
+default[:neutron][:opflex][:encap] = "vxlan"
+default[:neutron][:opflex][:vxlan][:uplink_iface] = "vlan.4093"
+default[:neutron][:opflex][:vxlan][:uplink_vlan] = 4093
+default[:neutron][:opflex][:vxlan][:encap_iface] = "br-int_vxlan0"
+default[:neutron][:opflex][:vxlan][:remote_ip] = "10.0.0.32"
+default[:neutron][:opflex][:vxlan][:remote_port] = 8472
+default[:neutron][:opflex][:vlan][:encap_iface] = ""
 
 case node[:platform_family]
 when "suse"
@@ -93,9 +103,6 @@ when "suse"
     cisco_opflex_pkgs: ["agent-ovs",
                         "lldpd",
                         "openstack-neutron-opflex-agent"],
-    cisco_opflex_ovs_pkgs: ["openvswitch",
-                            "openvswitch-kmp-default",
-                            "openvswitch-switch"],
     user: "neutron",
     group: "neutron",
     neutron_rootwrap_sudo_template: "/etc/sudoers.d/openstack-neutron"
@@ -127,7 +134,13 @@ when "rhel"
     hyperv_pkg: "",
     nsx_pkgs: [""],
     cisco_pkgs: ["python-networking-cisco"],
-    cisco_apic_pkgs: [""],
+    cisco_apic_pkgs: ["python-apicapi",
+                      "python-neutron-ml2-driver-apic"],
+    cisco_apic_gbp_pkgs: ["openstack-neutron-gbp",
+                          "python-gbpclient"],
+    cisco_opflex_pkgs: ["agent-ovs",
+                        "lldpd",
+                        "neutron-opflex-agent"],
     user: "neutron",
     group: "neutron",
     neutron_rootwrap_sudo_template: "/etc/sudoers.d/openstack-neutron"
@@ -161,6 +174,8 @@ else
     nsx_pkgs: [""],
     cisco_pkgs: [""],
     cisco_apic_pkgs: [""],
+    cisco_apic_gbp_pkgs: [""],
+    cisco_opflex_pkgs: [""],
     user: "neutron",
     group: "neutron",
     neutron_rootwrap_sudo_template: "/etc/sudoers.d/neutron-rootwrap"
--- a/chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
+++ b/chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
@@ -1,0 +1,72 @@
+neutron = nil
+if node.attribute?(:cookbook) && node[:cookbook] == "nova"
+  neutrons = search(:node, "roles:neutron-server AND \
+                    roles:neutron-config-#{node[:nova][:neutron_instance]}")
+  neutron = neutrons.first || raise("Neutron instance '#{node[:nova][:neutron_instance]}' \
+                                    for nova not found")
+else
+  neutron = node
+end
+
+ml2_mech_drivers = neutron[:neutron][:ml2_mechanism_drivers]
+ml2_type_drivers = neutron[:neutron][:ml2_type_drivers]
+
+if ml2_mech_drivers.include?("cisco_apic_ml2")
+  agent_config_path = "/etc/neutron/plugins/ml2/openvswitch_agent.ini"
+
+  node[:neutron][:platform][:cisco_apic_pkgs].each { |p| package p }
+  node[:neutron][:platform][:cisco_opflex_ovs_pkgs].each { |p| package p }
+  service node[:network][:ovs_service] do
+    supports status: true, restart: true
+    action [:start, :enable]
+  end
+  # apply configurations to compute node
+  if node.roles.include?("nova-compute-kvm")
+    node[:neutron][:platform][:cisco_opflex_pkgs].each { |p| package p }
+    # include neutron::common_config only now, after we've installed packages
+    include_recipe "neutron::common_config"
+
+    template agent_config_path do
+      cookbook "neutron"
+      source "openvswitch_agent.ini.erb"
+      owner "root"
+      group node[:neutron][:platform][:group]
+      mode "0640"
+      variables(
+        ml2_type_drivers: ml2_type_drivers,
+        tunnel_types: "",
+        use_l2pop: false,
+        dvr_enabled: false,
+        bridge_mappings: ""
+      )
+    end
+
+    # Update config file from template
+    opflex_agent_conf = "/etc/opflex-agent-ovs/conf.d/10-opflex-agent-ovs.conf"
+    template opflex_agent_conf do
+      cookbook "neutron"
+      source "10-opflex-agent-ovs.conf.erb"
+      mode "0755"
+      owner "root"
+      group neutron[:neutron][:platform][:group]
+      variables(
+        opflex_apic_domain_name: neutron[:neutron][:apic][:system_id],
+        hostname: node[:hostname],
+        socketgroup: neutron[:neutron][:platform][:group],
+        opflex_peer_ip: "10.0.0.30",
+        opflex_peer_port: "8009",
+        opflex_remote_ip: "10.0.0.32",
+        opflex_remote_port: "8472",
+        ml2_type_drivers: neutron[:neutron][:ml2_type_drivers]
+      )
+    end
+    service "neutron-opflex-agent" do
+      action [:enable, :start]
+      subscribes :restart, resources("template[#{agent_config_path}]")
+    end
+    service "agent-ovs" do
+      action [:enable, :start]
+      subscribes :restart, resources("template[#{opflex_agent_conf}]")
+    end
+  end
+end
--- a/chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
+++ b/chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
@@ -23,6 +23,11 @@ if ml2_mech_drivers.include?("cisco_apic_ml2")
   # apply configurations to compute node
   if node.roles.include?("nova-compute-kvm")
     node[:neutron][:platform][:cisco_opflex_pkgs].each { |p| package p }
+
+    service "lldpd" do
+      action [:enable, :start]
+    end
+
     # include neutron::common_config only now, after we've installed packages
     include_recipe "neutron::common_config"
 
--- a/chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
+++ b/chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
@@ -1,3 +1,19 @@
+#
+# Copyright 2016 SUSE Linux GmBH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 neutron = nil
 if node.attribute?(:cookbook) && node[:cookbook] == "nova"
   neutrons = search(:node, "roles:neutron-server AND \
@@ -11,67 +27,96 @@ end
 ml2_mech_drivers = neutron[:neutron][:ml2_mechanism_drivers]
 ml2_type_drivers = neutron[:neutron][:ml2_type_drivers]
 
-if ml2_mech_drivers.include?("cisco_apic_ml2")
-  agent_config_path = "/etc/neutron/plugins/ml2/openvswitch_agent.ini"
+return unless ml2_mech_drivers.include?("cisco_apic_ml2") ||
+    ml2_mech_drivers.include?("apic_gbp")
 
-  node[:neutron][:platform][:cisco_apic_pkgs].each { |p| package p }
-  node[:neutron][:platform][:cisco_opflex_ovs_pkgs].each { |p| package p }
-  service node[:network][:ovs_service] do
-    supports status: true, restart: true
-    action [:start, :enable]
+node[:neutron][:platform][:cisco_apic_pkgs].each { |p| package p }
+
+# We may need to review the ovs packages once we have better
+# clarity on what packages and kernel modules will be supported
+# for APIC integration.(eg: current version of APIC requires
+# openvswitch 2.4 with upstream kmp modules instead of kernel
+# provided modules). This will install default openvswitch
+# packages until the correct set is finalized.
+node[:network][:ovs_pkgs].each { |p| package p }
+
+service node[:network][:ovs_service] do
+  supports status: true, restart: true
+  action [:start, :enable]
+end
+
+if node.roles.include?("neutron-network")
+  # Explicitly stop and disable l3 and metadata agents if APIC is
+  # enabled on network node
+  service node[:neutron][:platform][:metadata_agent_name] do
+    action [:disable, :stop]
   end
-  # apply configurations to compute node
-  if node.roles.include?("nova-compute-kvm")
-    node[:neutron][:platform][:cisco_opflex_pkgs].each { |p| package p }
 
-    service "lldpd" do
-      action [:enable, :start]
-    end
+  service node[:neutron][:platform][:l3_agent_name] do
+    action [:disable, :stop]
+  end
+end
 
-    # include neutron::common_config only now, after we've installed packages
-    include_recipe "neutron::common_config"
+# apply configurations to compute node
+if node.roles.include?("nova-compute-kvm")
+  node[:neutron][:platform][:cisco_opflex_pkgs].each { |p| package p }
 
-    template agent_config_path do
-      cookbook "neutron"
-      source "openvswitch_agent.ini.erb"
-      owner "root"
-      group node[:neutron][:platform][:group]
-      mode "0640"
-      variables(
-        ml2_type_drivers: ml2_type_drivers,
-        tunnel_types: "",
-        use_l2pop: false,
-        dvr_enabled: false,
-        bridge_mappings: ""
-      )
-    end
+  service "lldpd" do
+    action [:enable, :start]
+  end
 
-    # Update config file from template
-    opflex_agent_conf = "/etc/opflex-agent-ovs/conf.d/10-opflex-agent-ovs.conf"
-    template opflex_agent_conf do
-      cookbook "neutron"
-      source "10-opflex-agent-ovs.conf.erb"
-      mode "0755"
-      owner "root"
-      group neutron[:neutron][:platform][:group]
-      variables(
-        opflex_apic_domain_name: neutron[:neutron][:apic][:system_id],
-        hostname: node[:hostname],
-        socketgroup: neutron[:neutron][:platform][:group],
-        opflex_peer_ip: "10.0.0.30",
-        opflex_peer_port: "8009",
-        opflex_remote_ip: "10.0.0.32",
-        opflex_remote_port: "8472",
-        ml2_type_drivers: neutron[:neutron][:ml2_type_drivers]
-      )
-    end
-    service "neutron-opflex-agent" do
-      action [:enable, :start]
-      subscribes :restart, resources("template[#{agent_config_path}]")
-    end
-    service "agent-ovs" do
-      action [:enable, :start]
-      subscribes :restart, resources("template[#{opflex_agent_conf}]")
-    end
+  # include neutron::common_config only now, after we've installed packages
+  include_recipe "neutron::common_config"
+
+  # Agent configurations for Cisco APIC driver
+  agent_config_path = "/etc/neutron/plugins/ml2/openvswitch_agent.ini"
+  template agent_config_path do
+    cookbook "neutron"
+    source "openvswitch_agent.ini.erb"
+    owner "root"
+    group node[:neutron][:platform][:group]
+    mode "0640"
+    variables(
+      ml2_type_drivers: ml2_type_drivers,
+      tunnel_types: "",
+      use_l2pop: false,
+      dvr_enabled: false,
+      bridge_mappings: ""
+    )
+  end
+
+  # Update config file from template
+  opflex_agent_conf = "/etc/opflex-agent-ovs/conf.d/10-opflex-agent-ovs.conf"
+  template opflex_agent_conf do
+    cookbook "neutron"
+    source "10-opflex-agent-ovs.conf.erb"
+    mode "0755"
+    owner "root"
+    group neutron[:neutron][:platform][:group]
+    variables(
+      opflex_apic_domain_name: neutron[:neutron][:apic][:system_id],
+      hostname: node[:hostname],
+      socketgroup: neutron[:neutron][:platform][:group],
+      opflex_peer_ip: neutron[:neutron][:opflex][:peer_ip],
+      opflex_peer_port: neutron[:neutron][:opflex][:peer_port],
+      opflex_vxlan_encap_iface: neutron[:neutron][:opflex][:vxlan][:encap_iface],
+      opflex_vxlan_uplink_iface: neutron[:neutron][:opflex][:vxlan][:uplink_iface],
+      opflex_vxlan_uplink_vlan: neutron[:neutron][:opflex][:vxlan][:uplink_vlan],
+      opflex_vxlan_remote_ip: neutron[:neutron][:opflex][:vxlan][:remote_ip],
+      opflex_vxlan_remote_port: neutron[:neutron][:opflex][:vxlan][:remote_port],
+      # TODO(mmnelemane) : update VLAN encapsulation config when it works.
+      # Currently set to VXLAN by default but can be modified from proposal.
+      ml2_type_drivers: neutron[:neutron][:ml2_type_drivers]
+    )
+  end
+
+  service "neutron-opflex-agent" do
+    action [:enable, :start]
+    subscribes :restart, resources("template[#{agent_config_path}]")
+  end
+
+  service "agent-ovs" do
+    action [:enable, :start]
+    subscribes :restart, resources("template[#{opflex_agent_conf}]")
   end
 end
--- a/chef/cookbooks/neutron/recipes/cisco_apic_support.rb
+++ b/chef/cookbooks/neutron/recipes/cisco_apic_support.rb
@@ -1,0 +1,24 @@
+node[:neutron][:platform][:cisco_apic_pkgs].each { |p| package p }
+
+if node[:neutron][:ml2_mechanism_drivers].include?("cisco_apic_ml2")
+  # Explicitly Stop and disable L3 and Metadata Agents if APIC is enabled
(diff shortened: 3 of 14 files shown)


chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
```diff
@@ -0,0 +1,67 @@
...
+    },
+
+    "renderers": {
+         "stitched-mode": {
+             "ovs-bridge-name": "br-int",
+             "encap": {
+                 "vxlan" : {
+                     "encap-iface": "br-int_vxlan0",
+                     "uplink-iface": "vlan.4093",
+                     "uplink-vlan": 4093,
```
rhafer: I guess we should make these two values ```uplink-vlan``` and ```uplink-iface``` configurable via proposal attributes.

chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
```diff
@@ -0,0 +1,67 @@
...
+                     "remote-ip": "<%= @opflex_remote_ip %>",
+                     "remote-port": <%= @opflex_remote_port %>
+                 }
+             },
+             "forwarding": {
+             "virtual-router": {
+             "enabled": true,
+                     "mac": "00:22:bd:f8:19:ff",
```
rhafer: Do we know if this is always true? Or it this specific to our test setup?

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,78 @@
...
+      owner "root"
+      group neutron[:neutron][:platform][:group]
+      variables(
+        opflex_apic_domain_name: neutron[:neutron][:apic][:system_id],
+        hostname: node[:hostname],
+        socketgroup: neutron[:neutron][:platform][:group],
+        opflex_peer_ip: "10.0.0.30",
+        opflex_peer_port: "8009",
+        opflex_remote_ip: "10.0.0.32",
+        opflex_remote_port: "8472",
```
rhafer: Those 4 attributes should likely also move to proposal attributes.

rhafer:
Gating currently fails with:
```
Recipe Compile Error in /var/chef/cache/cookbooks/neutron/recipes/role_neutron_server.rb
SyntaxError
-----------
/var/chef/cache/cookbooks/neutron/recipes/server.rb:406: syntax error, unexpected end-of-input, expecting keyword_end
```

Seems the patch is missing an ```end``` somewhere.

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -110,6 +110,16 @@
...
 end
+if neutron[:neutron][:networking_plugin] == "ml2"
+  if neutron[:neutron][:ml2_mechanism_drivers].include?("cisco_apic_ml2")
+    service_plugins = "cisco_apic_l3"
+  elsif neutron[:neutron][:ml2_mechanism_drivers].include?("apic_gbp")
+    service_plugins = "group_policy,servicechain,apic_gbp_l3"
+  else
+    service_plugins = "neutron.services.l3_router.l3_router_plugin.L3RouterPlugin, \
+                       #{service_plugins}"
+  end
```
vuntz: You're replacing `service_plugins` here, while you should append to it.

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -60,10 +60,17 @@
...
-    variables(
-      plugin_config_file: plugin_cfg_path +  " /etc/neutron/plugins/ml2/ml2_conf_cisco.ini"
-    )
+  if node[:neutron][:networking_plugin] == "ml2"
+    if node[:neutron][:ml2_mechanism_drivers].include?("cisco_nexus")
+      variables(
+        plugin_config_file: plugin_cfg_path + " /etc/neutron/plugins/ml2/ml2_conf_cisco.ini"
+      )
+    end
+    if node[:neutron][:ml2_mechanism_drivers].include?("cisco_apic_ml2")
```
vuntz: No need for `apic_gbp` test here?

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -252,6 +252,17 @@ def validate_proposal_after_save(proposal)
       validation_error I18n.t("barclamp.#{@bc_name}.validation.cisco_nexus")
     end
 
+    # cisco_apic_ml2 mech driver needs also openvswitch mech driver
+    if ml2_mechanism_drivers.include?("cisco_apic_ml2") &&
+        ml2_mechanism_drivers.include?("openvswitch")
```
vuntz: Hrm, missing `!` before the ovs part?

chef/data_bags/crowbar/template-neutron.schema
```diff
@@ -34,6 +34,27 @@
...
+                          "encap_iface": {"type": "str", "required": true },
+                          "uplink_iface": { "type": "str", "required": true },
+                          "uplink_vlan": { "type": "str", "required": true },
+                          "remote_ip": { "type": "str", "required": true },
+                          "remote_port": { "type": "int", "required": true }
+                      }},
+                      "vlan": { "type": "map", "required": false, "mapping": {
+                          "encap_iface": { "type": "str", "required": true }
+                      }}
+                    }},
```
vuntz: I know the attributes are not required, but does it make sense to still add them to the template, and add a migration?

chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
```diff
@@ -89,6 +89,8 @@ bridge_mappings = <%= @bridge_mappings %>
 # To enable the userspace datapath set this value to 'netdev'
 # datapath_type = system
 
+enable_tunneling = False
```
vuntz: Maybe add a comment saying that it's missing from the template?

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
```diff
@@ -80,8 +80,10 @@ flat_networks = <%= @external_networks.join(",") %>
 #
 # network_vlan_ranges =
 # Example: network_vlan_ranges = physnet1:1000:2999,physnet2
-<% if @ml2_type_drivers.include?("vlan") -%>
+<% unless @ml2_mechanism_drivers.include?("cisco_apic_ml2") || @ml2_mechanism_drivers.include?("apic_gbp") -%>
```
vuntz: Hrm, why? Do we disable vlan in that case?

chef/cookbooks/neutron/recipes/common_config.rb
rhafer: Hm, actually no. AFAICT "cisco_apic_l3" and "apic_gbp_l3" don't really work well together with some of those other plugins. (I am not sure which ones they are actually compatible with so it's safer to just replace here).

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
rhafer: It's a weird requirement of the "cisco_apic_ml2" and "apic_gbp" drivers. 

If ```network_vlan_ranges``` is present in the config. The driver will initialize the ACI incorrectly and configure it for "vlan" encapsulation. But up to now we only support "vxlan" encapsulation.

OTOH ```apic_gbp``` will refuse to load correctly if ```vlan``` is not present in the ```type_drivers```.

I know it's weird, but that's also the way how it is documentend in cisco's deployment docs.

chef/cookbooks/neutron/recipes/common_config.rb
mmnelemane: @vuntz: IIRC, other plugins (metering,l3,lbaas) are not required when using APIC and we dont run these agents for the APIC case. (https://github.com/crowbar/crowbar-openstack/pull/416/files/dab640f5d4376bfdc61ab375e40965328dfbc760#diff-91474df28ca9d038c83a65ad4c220591R58). @rhafer: please confirm if i missed something.

chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
rhafer: Actually I am pretty sure this one is wrong. ```enable_tunneling``` has been deprecated some releases ago. And nowadays setting ```tunnel_types``` to ```""```  (as we do) will have the same effect. I am also pretty sure that we successfully tested without this setting.

chef/cookbooks/neutron/recipes/common_config.rb
vuntz: Well, then at least for the non-apic case -- ie the l3_router plugin (last `else`)--, we should still append

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -33,11 +33,11 @@ def self.networking_plugins_valid
   end
 
   def self.networking_ml2_type_drivers_valid
-    ["vlan", "gre", "vxlan"]
+    ["vlan", "gre", "vxlan", "opflex"]
   end
 
   def self.networking_ml2_mechanism_drivers_valid
-    ["linuxbridge", "openvswitch", "cisco_nexus"]
+    ["linuxbridge", "openvswitch", "cisco_nexus", "cisco_apic_ml2"]
```
rhafer: I guess we're missing ```acpi_gbp``` here.

crowbar_framework/app/models/neutron_service.rb
rhafer: No, actually I think the comment is wrong. :)

chef/cookbooks/neutron/recipes/common_config.rb
rhafer: That's what we do, isn't it?

vuntz: Doh, yes. I read too fast :-)

vuntz: Still, the code is confusing. I would rewrite it like this:

```ruby
service_plugins = [ "neutron.services.metering.metering_plugin.MeteringPlugin", "neutron_fwaas.services.firewall.fwaas_plugin.FirewallPlugin"]
# L3 router goes first (not sure if it really matters?)
service.plugins.unshift("neutron.services.l3_router.l3_router_plugin.L3RouterPlugin") if neutron[:neutron][:networking_plugin] == "ml2"
service_plugins.push("neutron_lbaas.services.loadbalancer.plugin.LoadBalancerPlugin") if neutron[:neutron][:use_lbaas]
service_plugins = service_plugins.join(", ")

# In the case of APIC, we overwrite the whole set of service plugins
if neutron[:neutron][:networking_plugin] == "ml2"
  if neutron[:neutron][:ml2_mechanism_drivers].include?("cisco_apic_ml2")
    service_plugins = "cisco_apic_l3"
  elsif neutron[:neutron][:ml2_mechanism_drivers].include?("apic_gbp")
    service_plugins = "group_policy,servicechain,apic_gbp_l3"
   end
end

File: crowbar_framework/app/models/neutron_service.rb (hunk 10 above)
Comment (mmnelemane) [inline crowbar_framework/app/models/neutron_service.rb]: @rhafer: Added. Thanks. (Wonder how it got removed during rebase/merge)

File: chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb (hunk 8 above)
Comment (mmnelemane) [inline chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb]: @rhafer : Agreed.. will remove this.

File: chef/data_bags/crowbar/template-neutron.schema (hunk 7 above)
Comment (mmnelemane) [inline chef/data_bags/crowbar/template-neutron.schema]: @vuntz: My reasoning for excluding the attributes from the template is that this can only be tested against ACI hardware and not used otherwise. I followed the example of cisco_nexus switches which also does not show up in the template. But template entries do exist for zvm, vmware, infoblox. I can add migrations if you think its needed.

File: crowbar_framework/app/models/neutron_service.rb (hunk 6 above)
Comment (mmnelemane) [inline crowbar_framework/app/models/neutron_service.rb]: comment updated.

File: chef/cookbooks/neutron/recipes/server.rb (hunk 5 above)
Comment (mmnelemane) [inline chef/cookbooks/neutron/recipes/server.rb]: @vuntz: yes... its needed. Thanks.. Will update the same.

File: chef/cookbooks/neutron/recipes/common_config.rb (hunk 4 above)
Comment (mmnelemane) [inline chef/cookbooks/neutron/recipes/common_config.rb]: @vuntz: Thanks, updated it with a slight change.. Hope it would be better now.

File: chef/cookbooks/neutron/recipes/cisco_apic_agents.rb [hunk 11]
```diff
@@ -0,0 +1,83 @@
...
+    ml2_mech_drivers.include?("apic_gbp")
+  agent_config_path = "/etc/neutron/plugins/ml2/openvswitch_agent.ini"
+
+  node[:neutron][:platform][:cisco_apic_pkgs].each { |p| package p }
+  node[:neutron][:platform][:cisco_opflex_ovs_pkgs].each { |p| package p }
+  service node[:network][:ovs_service] do
+    supports status: true, restart: true
+    action [:start, :enable]
+  end
+  # apply configurations to compute node
```
vuntz: Can we have an empty line before this one?

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,83 @@
+neutron = nil
```
vuntz: You're missing the copyright/license header :-)

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,83 @@
...
+                                    for nova not found")
+else
+  neutron = node
+end
+
+ml2_mech_drivers = neutron[:neutron][:ml2_mechanism_drivers]
+ml2_type_drivers = neutron[:neutron][:ml2_type_drivers]
+
+if ml2_mech_drivers.include?("cisco_apic_ml2") ||
+    ml2_mech_drivers.include?("apic_gbp")
```
vuntz: Can we replace this with a `return unless ...` at the beginning of the recipe? That will remove one level of indentation.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,83 @@
...
+  agent_config_path = "/etc/neutron/plugins/ml2/openvswitch_agent.ini"
+
+  node[:neutron][:platform][:cisco_apic_pkgs].each { |p| package p }
+  node[:neutron][:platform][:cisco_opflex_ovs_pkgs].each { |p| package p }
+  service node[:network][:ovs_service] do
```
vuntz: Also an empty line before thie one.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,83 @@
...
+    node[:neutron][:platform][:cisco_opflex_pkgs].each { |p| package p }
+
+    service "lldpd" do
+      action [:enable, :start]
+    end
+
+    # include neutron::common_config only now, after we've installed packages
+    include_recipe "neutron::common_config"
+
+    template agent_config_path do
```
vuntz: Can you define the `agent_config_path` variable just before this line?

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,83 @@
...
+        opflex_vxlan_uplink_iface: neutron[:neutron][:opflex][:vxlan][:uplink_iface],
+        opflex_vxlan_uplink_vlan: neutron[:neutron][:opflex][:vxlan][:uplink_vlan],
+        opflex_vxlan_remote_ip: neutron[:neutron][:opflex][:vxlan][:remote_ip],
+        opflex_vxlan_remote_port: neutron[:neutron][:opflex][:vxlan][:remote_port],
+        # TODO(mmnelemane) : update VLAN encapsulation config when it works.
+        # Currently set to VXLAN by default but can be modified from proposal.
+        ml2_type_drivers: neutron[:neutron][:ml2_type_drivers]
+      )
+    end
+    service "neutron-opflex-agent" do
```
vuntz: Empty line before this one.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,83 @@
...
+      action [:enable, :start]
+      subscribes :restart, resources("template[#{agent_config_path}]")
+    end
+    service "agent-ovs" do
```
vuntz: Empty line before this one.

chef/cookbooks/neutron/recipes/cisco_apic_support.rb
```diff
@@ -0,0 +1,29 @@
+node[:neutron][:platform][:cisco_apic_pkgs].each { |p| package p }
```
vuntz: Missing copyright/license header.

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
vuntz: But then I would instead change Crowbar to disable the ability to have vlan in that case.

chef/data_bags/crowbar/template-neutron.schema
vuntz: I don't mind much; I just think it'll be harder for people to use it if it's not in the template.

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -252,6 +252,17 @@ def validate_proposal_after_save(proposal)
       validation_error I18n.t("barclamp.#{@bc_name}.validation.cisco_nexus")
     end
 
+    # openvswitch is not needed when cisco_apic_ml2 mechanism driver is used
+    if ml2_mechanism_drivers.include?("cisco_apic_ml2") &&
```
vuntz: No check for apic_gbp in this case?

Also, is it really "is not needed" or "should not be used"? Can we also update the javascript to unselect ovs when one of the APIC one is selected? (you might want to wait for the vxlan patch to land first, though)

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -252,6 +252,17 @@ def validate_proposal_after_save(proposal)
...
+        ml2_mechanism_drivers.include?("openvswitch")
+      validation_error I18n.t("barclamp.#{@bc_name}.validation.cisco_apic_ml2")
+    end
+
+    # cisco_apic_ml2 mech driver needs opflex as the type_driver
+    if ml2_mechanism_drivers.include?("cisco_apic_ml2") && ml2_type_drivers != ["opflex"]
```
vuntz: No check for apic_gbp in this case?

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -33,7 +33,7 @@ def self.networking_plugins_valid
   end
 
   def self.networking_ml2_type_drivers_valid
-    ["vlan", "gre", "vxlan"]
+    ["vlan", "gre", "vxlan", "opflex"]
```
vuntz: Hrm, I guess this should be removed also?

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -60,10 +60,18 @@
...
-    variables(
-      plugin_config_file: plugin_cfg_path +  " /etc/neutron/plugins/ml2/ml2_conf_cisco.ini"
-    )
+  if node[:neutron][:networking_plugin] == "ml2"
+    if node[:neutron][:ml2_mechanism_drivers].include?("cisco_nexus")
+      variables(
+        plugin_config_file: plugin_cfg_path + " /etc/neutron/plugins/ml2/ml2_conf_cisco.ini"
+      )
+    end
+    if node[:neutron][:ml2_mechanism_drivers].include?("cisco_apic_ml2") ||
```
rhafer: This is slightly wrong. It should be an ```elsif```

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -60,10 +60,18 @@
   mode 0640
   # whenever changing plugin_config_file here, keep in mind to change the call
   # to neutron-db-manage too
-  if node[:neutron][:networking_plugin] == "ml2" and node[:neutron][:ml2_mechanism_drivers].include?("cisco_nexus")
-    variables(
-      plugin_config_file: plugin_cfg_path +  " /etc/neutron/plugins/ml2/ml2_conf_cisco.ini"
-    )
+  if node[:neutron][:networking_plugin] == "ml2"
```
rhafer: We should work this if statement a bit. Currently it causes the gating to fail because, we don't update add ```/etc/neutron/plugins/ml2/ml2_conf.ini``` anymore to ```/etc/sysconfig/neutron``` in the non-apic case. I'd suggest something like this (untested):

```
if node[:neutron][:networking_plugin] == "ml2"
    plugin_cfgs = plugin_cfg_path
    if node[:neutron][:ml2_mechanism_drivers].include?("cisco_nexus")
      plugin_cfgs += " /etc/neutron/plugins/ml2/ml2_conf_cisco.ini"
    elsif node[:neutron][:ml2_mechanism_drivers].include?("cisco_apic_ml2") ||
        node[:neutron][:ml2_mechanism_drivers].include?("apic_gbp")
      plugin_cfgs += " /etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini"
    end
    variables(
      plugin_config_file: plugin_cfgs
    )
end
```

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
mmnelemane: Updated.

chef/cookbooks/neutron/recipes/cisco_apic_support.rb
mmnelemane: Updated.

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
mmnelemane: https://github.com/mmnelemane/crowbar-openstack/blob/apic_ui/crowbar_framework/app/models/neutron_service.rb#L263 (when merged) should take care vlan is not enabled when apic is used. So removing this condition for now.

vuntz:
This needs a rebase :/

chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
vuntz: It'd be surprising that this is always the same value.

chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
```diff
@@ -0,0 +1,67 @@
+{
```
vuntz: In general, the indentation in this file is not consistent. Could you fix this?

mmnelemane: @vuntz: hope its better now... and rebased too.

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -112,6 +124,7 @@
     hyperv_pkg: "",
     nsx_pkgs: [""],
     cisco_pkgs: ["python-networking-cisco"],
+    cisco_apic_pkgs: [""],
```
toabctl: what about ```cisco_opflex_pkgs``` and ```cisco_opflex_ovs_pkgs```?

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -144,6 +157,7 @@
     hyperv_pkg: "python-networking-hyperv",
     nsx_pkgs: [""],
     cisco_pkgs: [""],
+    cisco_apic_pkgs: [""],
```
toabctl: same here as in comment https://github.com/crowbar/crowbar-openstack/pull/416/commits/d5fb6e19eb5d4ce74c2e1be3d4f4407b39939638#r69571970

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,104 @@
...
+end
+
+ml2_mech_drivers = neutron[:neutron][:ml2_mechanism_drivers]
+ml2_type_drivers = neutron[:neutron][:ml2_type_drivers]
+
+return unless ml2_mech_drivers.include?("cisco_apic_ml2") ||
+    ml2_mech_drivers.include?("apic_gbp")
+
+# Agent configurations for Cisco APIC driver
+agent_config_path = "/etc/neutron/plugins/ml2/openvswitch_agent.ini"
```
toabctl: can you move this just one line before the ```template``` call where you use it the first time?

mmnelemane: Fixed

chef/cookbooks/neutron/attributes/default.rb
mmnelemane: updated the other packages too..

chef/cookbooks/neutron/attributes/default.rb
mmnelemane: fixed.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,104 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+neutron = nil
+if node.attribute?(:cookbook) && node[:cookbook] == "nova"
```
rhafer: Not sure either, but why? It supposed to catch the case where the recipe runs on the compute nodes. It's basically what we do in some other places as well.

chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
```diff
@@ -0,0 +1,66 @@
...
+        "virtual-router": {
+          "enabled": true,
+          "mac": "00:22:bd:f8:19:ff",
+          "ipv6": {
+            "router-advertisement": "false"
+          }
+        },
+        "virtual-dhcp": {
+          "enabled": "true",
+          "mac": "00:22:bd:f8:19:ff"
```
vuntz: Is hard-coding this fine? Or can it change on each setup?

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
toabctl: As long as we don't allow the recipe to run on anything else than the nova and neutron node, it's fine.

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -52,6 +52,21 @@
 default[:neutron][:ssl][:cert_required] = false
 default[:neutron][:ssl][:ca_certs] = "/etc/neutron/ssl/certs/ca.pem"
 
+default[:neutron][:apic][:system_id] = "SOC"
```
vuntz: Any reason to not keep openstack (as default upstream)?

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -81,6 +96,16 @@
...
+    cisco_apic_pkgs: ["python-apicapi",
+                      "python-neutron-ml2-driver-apic"],
+    cisco_apic_gbp_pkgs: ["openstack-neutron-gbp",
+                          "python-gbpclient"],
+    cisco_opflex_pkgs: ["agent-ovs",
+                        "lldpd",
+                        "openstack-neutron-opflex-agent"],
+    cisco_opflex_ovs_pkgs: ["openvswitch",
+                            "openvswitch-kmp-default",
+                            "openvswitch-switch"],
```
vuntz: Can't we drop this and use instead node[:network][:ovs_pkgs] in the recipe?

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
vuntz: Yes, it's fine to keep it that way.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,104 @@
...
+return unless ml2_mech_drivers.include?("cisco_apic_ml2") ||
+    ml2_mech_drivers.include?("apic_gbp")
+
+node[:neutron][:platform][:cisco_apic_pkgs].each { |p| package p }
+node[:neutron][:platform][:cisco_opflex_ovs_pkgs].each { |p| package p }
+
+service node[:network][:ovs_service] do
+  supports status: true, restart: true
+  action [:start, :enable]
+end
```
vuntz: In the neutron cookbook, we do:
```
  unless ::File.exist?("/sys/module/#{node[:network][:ovs_module]}")
    ::Kernel.system("modprobe #{node[:network][:ovs_module]}")
  end
```
Do we want this here too?

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,104 @@
...
+
+# apply configurations to compute node
+if node.roles.include?("nova-compute-kvm")
```
vuntz: So it only works with kvm compute nodes? Not for xen, or anything else?

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,104 @@
...
+  service "lldpd" do
+    action [:enable, :start]
+  end
+
+  # include neutron::common_config only now, after we've installed packages
+  include_recipe "neutron::common_config"
+
+  # Agent configurations for Cisco APIC driver
+  agent_config_path = "/etc/neutron/plugins/ml2/openvswitch_agent.ini"
+
```
vuntz: You can remove this empty line :-)

chef/cookbooks/neutron/attributes/default.rb
mmnelemane: No reason for any particular name. It can be any name and can be configured by the user. "soc" came from Mike's documentation.

chef/cookbooks/neutron/recipes/cisco_apic_support.rb
```diff
@@ -0,0 +1,46 @@
...
+end
+
+# Explicitly Stop and disable L3 and Metadata Agents if APIC is enabled
+service node[:neutron][:platform][:metadata_agent_name] do
+  action [:disable, :stop]
+end
+
+service node[:neutron][:platform][:l3_agent_name] do
+  action [:disable, :stop]
+end
```
vuntz: AFAICT, this recipe is included from neutron-server.rb, which means that these two services might not be installed (as they come from the other neutron role). So this might fail if neutron-server node doesn't have neutron-network role.

chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
mmnelemane: Currently, this is derived from Cisco documentation for ACI (http://www.cisco.com/c/en/us/td/docs/switches/datacenter/aci/apic/sw/1-x/openstack/b_ACI_with_OpenStack_OpFlex_Deployment_Guide_for_Red_Hat.pdf) and not sure if it will need to be configurable.

chef/cookbooks/neutron/attributes/default.rb
rhafer: Hm, good point. This was originally done because of the openvswitch-gbp package (which is no longer required at runtime). However, is seems that (especially in master) we have slightly different requirements for openvswitch soon. E.g. apic only works with 2.4 and requires the upstream provided kmps (not the once shipping in the kernel).
But that is currently not reflected in this attribute either :/. I guess we should at least add a note here until we figure out the correct packaging for the product.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
rhafer: Hm, at least in my tests the correct module was automatically loaded so we left that out.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
rhafer: I guess the simple answer is: We don't know. And we only tested with kvm.

mmnelemane: Cisco claims it can integrate with vmware/xen but I think they have significant config changes for the agents.

chef/cookbooks/neutron/recipes/cisco_apic_support.rb
mmnelemane: @vuntz, @rhafer: Do you recommend to remove these OR moved to network_agents.rb ?

vuntz: I'd move to the agents recipe.

rhafer: @mmnelemane Yeah, if possible move it to cisco_apic_agents.rb. network_agents is also wrong because l3 and metadata agent are actually started by  common_agents.rb (which includes cisco_apic_agents)

chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
```diff
@@ -0,0 +1,66 @@
...
+      "mode": "enabled",
+      "ca-store": "/etc/ssl/certs/"
+    },
+    "inspector": {
+    "enabled": true,
+      "socket-name": "/var/run/opflex-agent-ovs-inspect.sock"
+    },
+    "notif": {
+      "enabled": true,
+      "socket-name": "/var/run/opflex-agent-ovs-notif.sock",
```
toabctl: Should this be ```/var/run/neutron``` if the group is neutron?

chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
```diff
@@ -0,0 +1,66 @@
...
+    "peers": [
+      {"hostname": "<%= @opflex_peer_ip %>", "port": "<%= @opflex_peer_port %>"}
+    ],
+    "ssl": {
+      "mode": "enabled",
+      "ca-store": "/etc/ssl/certs/"
+    },
+    "inspector": {
+    "enabled": true,
+      "socket-name": "/var/run/opflex-agent-ovs-inspect.sock"
```
toabctl: Should this be  ```/var/run/neutron/``` ?

toabctl:
2 more questions but in general, +1

mmnelemane:
@toabctl : Looks like they are hard-coded: https://github.com/noironetworks/python-opflex-agent/blob/master/opflexagent/opflex_notify.py#L39. I could not find the relevant code for inspector.  Also i think /var/run/<sock> is not specific to neutron.. but may be used by other Northbound projects (eg: Opendaylight also can talk to opflex and somehow seems to use the same paths).

toabctl:
+1

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -0,0 +1,119 @@
...
+end
+
+# Explicitly stop and disable l3 and metadata agents if APIC is enabled
+service node[:neutron][:platform][:metadata_agent_name] do
+  action [:disable, :stop]
+end
+
+service node[:neutron][:platform][:l3_agent_name] do
+  action [:disable, :stop]
+end
```
vuntz: Now that we moved that here... will that actually work on compute nodes? I expect the services don't exist there... Should we protect this somehow with a test on whether the node has neutron-network role?

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
vuntz: If it's not needed, can we remove that code from the network cookbook?

chef/cookbooks/neutron/attributes/default.rb
vuntz: So if it's not needed like this, I'd use "openstack" by default, as SOC is SUSE-specific, which we should avoid in upstream Crowbar.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
mmnelemane: Yes.. we can check for neutron-network role.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
mmnelemane: @vuntz: Is it better to remove (if required) in a different PR ?

vuntz:
Let me put a +0.5: I'm fine merging this now, but the work is not complete as we need to change the network_agents_ha recipe to do the right thing (do nothing for metadata and l3, do something for the new services).

mmnelemane:
@vuntz: Can the HA changes be part of another PR ?
The code in this PR is tested to some extent (only the rails changes were removed) on physical hardware , though with some manual steps. And the commits here can be easily cherry-picked and used for future integration. Without hardware access, I doubt if I would be able to do any tests for the HA changes, in which case, it needs to be merged based only on reviews. So IMO, it should go in another PR. I have added a task on the card for this and can work on creating a fresh PR. Please let me know your opinion on this.

toabctl:
@mmnelemane can you create a card in the backlog for HA support? +1 for merging as it is now.

vuntz:
> @vuntz: Can the HA changes be part of another PR ?

Yes, that's what I mentioned. Don't know if @toabctl's +1 is still valid, if it is, you can merge.

PR #418, opened 2016-06-27
magnum: Allocate a public IP to magnum-server node

vuntz:
This is required to have the API reachable from outside.

The commit also contains the commented out code that is generally used
for HA.


jgrassler:
+1 as far as I understand the HA stuff. What about the Hound comments, though? They're tough to sort out in this case :-/

toabctl:
+1

vuntz:
> What about the Hound comments, though? They're tough to sort out in this case :-/

They'll be gone when we do the HA work, so let's ignore them.

PR #419, opened 2016-06-27
chef: Use new CrowbarRoleRecipe.node_state_valid_for_role helper

vuntz:
This is used in role recipes to decide whether the role should be run or
not. This will allow us to stop rebuilding the run_list based on the
state of the nodes, since we move the logic in the role recipes.


chef/cookbooks/ceilometer/recipes/role_ceilometer_swift_proxy_middleware.rb
```diff
@@ -14,4 +14,6 @@
 # limitations under the License.
 #
 
-include_recipe "ceilometer::swift"
+if CrowbarRoleRecipe.node_state_valid_for_role(node, "ceilometer", "ceilometer-swift-proxy-middleware")
```
jsuchome: I think it's nice, although long.

jsuchome:
+1

rsalevsky:
+1

PR #420, opened 2016-06-28
Make keystone deployable with mod_wsgi again

rhafer:
* Synch'ed vhost/mod_wsgi configuration to what's documented upstream. See:
  http://docs.openstack.org/developer/keystone/apache-httpd.html#mod-wsgi
* Removed custom wsgi httpd script. Use upstream variants keystone-wsgi-admin
  and keystone-wsgi-public now.
* Fixed cookbooks dependencies to require apache2.


rsalevsky:
LGTM

aplanas:
+1

PR #422, opened 2016-06-29
Start virtlogd service on compute nodes

jsuchome:
Without this, I'm getting http://pastebin.nue.suse.com/16785/src on SP2 compute nodes...
See also http://mailman.suse.de/mlarch/SuSE/cloud-devel/2016/cloud-devel.2016.06/msg00075.html

Requires https://github.com/crowbar/crowbar-openstack/pull/423

### DIFF
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -183,6 +183,15 @@ case node[:nova][:libvirt_type]
         notifies :restart, "service[libvirtd]"
       end
 
+      service "virtlogd" do
+        action [:enable, :start]
+        if node[:nova][:ha][:compute][:enabled]
+          provider Chef::Provider::CrowbarPacemakerService
+          supports no_crm_maintenance_mode: true
+        end
+        only_if { node[:platform_family] == "suse" && node[:platform_version].to_f > 12.1 }
+      end
+
       service "libvirtd" do
         action [:enable, :start]
         if node[:nova][:ha][:compute][:enabled]
--- a/chef/cookbooks/nova/recipes/compute_ha.rb
+++ b/chef/cookbooks/nova/recipes/compute_ha.rb
@@ -90,6 +90,18 @@ end
 compute_primitives = []
 compute_transaction_objects = []
 
+if node[:platform_family] == "suse" && node[:platform_version].to_f > 12.1
+  virtlogd_primitive = "virtlogd-compute"
+  pacemaker_primitive virtlogd_primitive do
+    agent "systemd:virtlogd"
+    op nova[:nova][:ha][:op]
+    action :update
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+  compute_primitives << virtlogd_primitive
+  compute_transaction_objects << "pacemaker_primitive[#{virtlogd_primitive}]"
+end
+
 libvirtd_primitive = "libvirtd-compute"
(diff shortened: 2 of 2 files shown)


chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -182,6 +182,15 @@
         notifies :restart, "service[libvirtd]"
       end
 
+      service "virtlogd" do
+        action [:enable, :start]
+        if node[:nova][:ha][:compute][:enabled]
+          provider Chef::Provider::CrowbarPacemakerService
+          supports no_crm_maintenance_mode: true
+        end
+        only_if { node[:platform_family] == "suse" && node[:platform_version].to_f > 12.1 }
```
aplanas: not sure that we want to check for `:platform_version`. IIUC this action will not do anything for 12.1, or I am wrong?

jsuchome: The service is not there for 12.1, I'm not sure if chef would exit with error, or ignore it in such case... checking the version seems safe.

aplanas: Oh right. For some reason I believed that was there and enabled. Thanks!

aplanas:
+1

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -182,6 +182,15 @@
         notifies :restart, "service[libvirtd]"
       end
 
+      service "virtlogd" do
+        action [:enable, :start]
+        if node[:nova][:ha][:compute][:enabled]
+          provider Chef::Provider::CrowbarPacemakerService
+          supports no_crm_maintenance_mode: true
+        end
```
vuntz: You should only do this if you add a pacemaker resource for this service (in compute_ha.rb, I assume). Which might make sense.

jsuchome: I've tried to add it

chef/cookbooks/nova/recipes/compute_ha.rb
```diff
@@ -90,6 +90,18 @@
 compute_primitives = []
 compute_transaction_objects = []
 
+if node[:platform_family] == "suse" && node[:platform_version].to_f > 12.1
```
vuntz: So we can't really do the test on `node` as it's not a compute node. We could look if one of the compute node is SLES 12 SP2, but what if we have a mix of platforms among the compute nodes?

I'm tempted to just leave it the way it is, but with a comment explaining that the test is factually wrong, but there's no good way to deal with this.

jsuchome: Once we switch to SP2, we won't allow SP1 compute nodes, right? So the check is mostly to let it build until we do the switch...

vuntz: Right, that's my thought also.

jsuchome:
Hm, I wonder why it's still failing even when I have https://github.com/crowbar/crowbar-openstack/pull/423 in updaterepos...

chef/cookbooks/nova/recipes/compute.rb
dguitarbite: IMO, we should fix this at the package level rather than here. Restarting libvirtd service should trigger virtlogd service. Not sure if there my thinking is flawed.

nicolasbock:
+1

jsuchome:
Finally, green. Can anyone merge it?

PR #424, opened 2016-06-30
neutron: Enable vxlan support to linuxbridge

vuntz:


vuntz:
For the record, when running tempest with vxlan, it sometimes fails like in http://artifacts.opnfv.org/logs/functest/opnfv-jump-2/2016-03-18_01-43-02/master/tempest/tempest.log

It's actually a race in nova/tempest :/

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -181,34 +159,52 @@ def validate_proposal_after_save(proposal)
...
     end
 
-    # only allow valid ml2 type drivers
-    ml2_type_drivers.each do |drv|
-      unless ml2_type_drivers_valid.include? drv
+    # only allow valid ml2 mechanism drivers
+    ml2_mechanism_drivers.each do |drv|
+      unless ml2_mechanism_drivers_valid.include? drv
         validation_error I18n.t(
           "barclamp.#{@bc_name}.validation.no_vaild_ml2_type_driver",
```
toabctl: Should be "validation.no_vaild_ml2_mechanism_driver" .

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -175,17 +175,17 @@ def validate_ml2 proposal
...
     end
 
-    # cisco_nexus mech driver needs also openvswitch mech driver
+    # cisco_nexus mech driver needs also openvswitch mech driver and vlan type driver
     if ml2_mechanism_drivers.include? "cisco_nexus" and not ml2_mechanism_drivers.include? "openvswitch"
-      validation_error I18n.t("barclamp.#{@bc_name}.validation.cisco_nexus")
+      validation_error I18n.t("barclamp.#{@bc_name}.validation.cisco_nexus_ovs")
+    end
+
+    if ml2_mechanism_drivers.include?("cisco_nexus") && !ml2_type_drivers.include?("vlan")
```
toabctl: is ```ml2_type_drivers.include?("vlan")``` enough (the ```include```)? Doesn't ```cisco_nexus``` **only** need ```vlan``` as type driver?

crowbar_framework/app/models/neutron_service.rb
aplanas: /vaild/valid

vuntz: Interesting, this was already broken before. Added a commit for this.

crowbar_framework/app/models/neutron_service.rb
vuntz: That's the same validation as before. I think the idea is that we still want to allow gre/vxlan; networks with gre/vxlan  just won't use the cisco_nexus features.

toabctl:
+1

nicolasbock:
+1

PR #426, opened 2016-07-01
aodh: Fix aodh deployment if mongodb is used

vuntz:
The db connection string was not correct.


jsuchome:
+1 Looks correct (coming from ceilometer/common.rb).

aplanas:
+1

PR #427, opened 2016-07-01
glance: Put swift config in /etc/glance/glance-swift.conf

vuntz:
This seems to be required now with keystone v3, as glance is unable to
authenticate with swift without specifying the domain.

### DIFF
--- a/chef/cookbooks/glance/templates/default/glance-swift.conf.erb
+++ b/chef/cookbooks/glance/templates/default/glance-swift.conf.erb
@@ -1,0 +1,16 @@
+# glance-swift.conf
+#
+# Specify the reference name in []
+# For each section, specify the auth_address, user and key.
+#
+# WARNING:
+# * If any of auth_address, user or key is not specified,
+# the glance-api's swift store will fail to configure
+
+[crowbar]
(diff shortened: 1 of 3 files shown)


jsuchome:
+1 It looks correct, I did not test it

chef/cookbooks/glance/templates/default/glance-swift.conf.erb
```diff
@@ -0,0 +1,16 @@
+# glance-swift.conf
+#
+# Specify the reference name in []
+# For each section, specify the auth_address, user and key.
+#
+# WARNING:
+# * If any of auth_address, user or key is not specified,
+# the glance-api's swift store will fail to configure
+
+[crowbar]
```
dirkmueller: should this really be the crowbar group?

jsuchome: `default_swift_reference` in glance-api.conf

vuntz: Yeah, the name of the group is only a reference. It can be anything.

nicolasbock:
+1

PR #429, opened 2016-07-05
nova: Only install openvswitch-kmp-xen on SLE 12 <= SP1

vuntz:
openSUSE and SLES12SP2 use the module shipped with upstream kernel.

See https://github.com/crowbar/crowbar-core/pull/517


aplanas:
+1

toabctl:
+1

PR #431, opened 2016-07-05
Ceilometer tempest

toabctl:


toabctl:
This will fail for now (package update on the way and there is a race in one tempest test)

aplanas:
re-trigger the gating, but codewise +1

toabctl:
> re-trigger the gating, but codewise +1

it can not pass. The packages need to pass cleanvm first.

vuntz:
+1, let's just wait for gating :-)

toabctl:
I still had a race condition with deleting the security groups in the tearDown function. Let's see if we get this in the gating, too.

toabctl:
@aplanas @vuntz tempest passed (with 2 ceilometer tests, wow! :)) . Can be merged imo....

vuntz:
> with 2 ceilometer tests

Hrm, shouldn't we expect to have more tests?

toabctl:
> Hrm, shouldn't we expect to have more tests?

There are 4 api tests in total and only 2 of them are smoke tests. See https://github.com/openstack/ceilometer/blob/master/ceilometer/tests/tempest/api/test_telemetry_notification_api.py . And there's anotherr scenario test (non-smoke). So looks ok.

aplanas:
+1 then

PR #434, opened 2016-07-06
set retries explicitly from 0 to 5

MaximilianMeister:
there was a zypper 4 error after an upgrade from 6 to 7:
4 - ZYPPER_EXIT_ERR_ZYPP
  A problem reported by ZYPP library.  E.g.  another  instance  of
  ZYPP is running.

NOTE: I am not sure if this is the right solution, or if it is better to investigate why zypper was locking at that moment

the error happens during the upgrade job: https://ci.suse.de/job/openstack-mkcloud/26697

### DIFF
--- a/chef/cookbooks/aodh/recipes/aodh.rb
+++ b/chef/cookbooks/aodh/recipes/aodh.rb
@@ -1,7 +1,14 @@
 ha_enabled = node[:ceilometer][:ha][:server][:enabled]
 
 node[:aodh][:platform][:packages].each do |p|
-  package p
+  package p do
+    action :install
+    retries 5
+    retry_delay 2
+    package_name p
+    cookbook_name "aodh"
+    recipe_name "aodh"
+  end
 end
 
 # mongdb setup has been already done by server.rb recipe


aplanas:
Gating failing with: `FATAL: ArgumentError: wrong number of arguments (1 for 0)`

vuntz:
This sounds like the wrong fix; we should try to find out what zypper was doing at this point in time that can explain the issue.

toabctl:
I agree that this is not the right fix. -1

toabctl:
@MaximilianMeister can we close this?

MaximilianMeister:
> This sounds like the wrong fix; we should try to find out what zypper was doing at this point in time that can explain the issue.

I think it could even have to do with the old cloud 6 repos, because a `zypper ref` fails already... it's a bit difficult to debug, I couldn't find any log (maybe need to set chef to verbose, and watch the zypper plugin)...

@toabctl I think we can close this in favor of https://github.com/crowbar/crowbar-core/pull/545 if I am right with my assumption. but lets wait for the upgrade gating there, https://ci.suse.de/job/openstack-mkcloud/27159/ and see if it fixes the issue

toabctl:
See https://github.com/crowbar/crowbar-openstack/pull/434#issuecomment-231763216

PR #436, opened 2016-07-06
tempest: Disable aodh and magnum tempest tests for now

toabctl:
This needs to be fixed when the tempest tests pass. But
for now we want to add the tests to the packages but don't run them.


aplanas:
I would split this in two: one for magnum and other for aodh.

But not big deal for me, so if not ones complains: +1

vuntz:
+1. Not merging in case you want to do the change that @aplanas suggested, but feel free to merge if not.

PR #438, opened 2016-07-07
added toggle to disable creation of non-admin user

skazi0:
This adds the option "Create Regular User" to the barclamp UI and respective handling in the keystone cookbook. See: https://trello.com/c/lF2pdHYl/21-3-make-creation-of-non-admin-user-in-keystone-barclamp-optional

### DIFF
--- a/chef/cookbooks/keystone/libraries/helpers.rb
+++ b/chef/cookbooks/keystone/libraries/helpers.rb
@@ -47,6 +47,8 @@ module KeystoneHelper
                                                 node[:keystone][:api][:service_port],
                                                 node[:keystone][:api][:version])
 
+      has_default_user = node["keystone"]["default"]["create_user"]
+
       @keystone_settings ||= Hash.new
       @keystone_settings[cookbook_name] = {
         "api_version" => node[:keystone][:api][:version],
@@ -72,8 +74,8 @@ module KeystoneHelper
         "admin_password" => node["keystone"]["admin"]["password"],
         "default_tenant" => node["keystone"]["default"]["tenant"],
         "default_tenant_id" => node["keystone"]["default"]["tenant_id"],
-        "default_user" => node["keystone"]["default"]["username"],
-        "default_password" => node["keystone"]["default"]["password"],
+        "default_user" => has_default_user ? node["keystone"]["default"]["username"] : nil,
+        "default_password" => has_default_user ? node["keystone"]["default"]["password"] : nil,
         "service_tenant" => node["keystone"]["service"]["tenant"],
         "service_tenant_id" => node["keystone"]["service"]["tenant_id"]
       }
--- a/chef/data_bags/crowbar/template-keystone.json
+++ b/chef/data_bags/crowbar/template-keystone.json
@@ -47,6 +47,7 @@
         "token": "999888777666"
       },
       "default": {
+        "create_user": true,
         "tenant": "openstack",
         "username": "crowbar",
         "password": "crowbar"
@@ -139,7 +140,7 @@
     "keystone": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 35,
+      "schema-revision": 100,
       "element_states": {
         "keystone-server": [ "readying", "ready", "applying" ]
       },
--- a/crowbar_framework/app/views/barclamp/keystone/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/keystone/_edit_attributes.html.haml
@@ -16,8 +16,15 @@
       = string_field %w(default tenant)
       = string_field %w(admin username)
       = password_field %w(admin password)
-      = string_field %w(default username)
-      = password_field %w(default password)
+
+      = boolean_field %w(default create_user),
+        "data-showit" => "true",
+        "data-showit-target" => "#default_user_container",
+        "data-showit-direct" => "true"
+
+      #default_user_container
+        = string_field %w(default username)
+        = password_field %w(default password)
 
     %fieldset
       %legend
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -476,9 +476,21 @@ openstack_command << " --insecure" if keystone_insecure
 end
 
 # Create users
-[[node[:keystone][:admin][:username], node[:keystone][:admin][:password], node[:keystone][:admin][:tenant]],
-  [node[:keystone][:default][:username], node[:keystone][:default][:password], node[:keystone][:default][:tenant]]
-].each do |user_data|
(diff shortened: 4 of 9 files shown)


toabctl:
@skazi0 Can you add a note to the commit msg **why** you want to do that? Beside that I'm not sure if we need that because we want to switch soon to ```keystone bootstrap``` to handle the initial user creation (see https://trello.com/c/haB7x3kz/27-3-use-keystone-manage-bootstrap-to-initialize-keystone )

skazi0:
I added the link to the trello card to the PR description.

chef/data_bags/crowbar/migrate/keystone/036_non_admin_user_create_flag.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["default"]["create_user"] = true
```
vuntz: We usually do `a["default"]["create_user"] = ta["default"]["create_user"]` (to avoid putting the default value here again -- we just re-use the one from the template)

chef/data_bags/crowbar/template-keystone.json
```diff
@@ -139,7 +140,7 @@
     "keystone": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 35,
+      "schema-revision": 36,
```
vuntz: It was agreed to bump to 100 for Cloud 7 (to leave space for migrations in Cloud 6). Can you do this here?

vuntz:
So the tempest cookbook is using this user (see `alt_comp_user` variable in `chef/cookbooks/tempest/recipes/config.rb`). Maybe it'd be nice to have some validation in the tempest proposal to check that this option is enabled in keystone? That'd be the `validate_proposal_after_save` method in `crowbar_framework/app/models/tempest_service.rb`, and you probably would need to look at the role for Keystone (with `RoleObject.find_role_by_name`)

skazi0:
The only place where this alt_comp_user is actually used in tempest seems to be in `templates/default/tempest.conf.erb`:
```erb
# Username of alternate user to use for Nova API requests. (string
# value)
# This option is deprecated for removal.
# Its value may be silently ignored in the future.
#alt_username = <None>
alt_username = <%= @alt_comp_user %>
```
which is marked as *deprecated for removal*. Do you think it makes sense to add more code on top of this (besides being an exercise for me)?

nicolasbock:
+1

vuntz:
> which is marked as deprecated for removal. Do you think it makes sense to add more code on top of this (besides being an exercise for me)?

So it's marked as deprecated, but I think it's still used for the time being, and it might be that it'll be replaced by something else still using that user anyway. So I think it's still useful to add the validation.

skazi0:
OK, so if this option is disabled in keystone, tempest proposal should cause validation_error, right?

mjura:
LGTM, works fine +1

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -72,6 +72,7 @@ def self.keystone_settings(current_node, cookbook_name)
         "admin_password" => node["keystone"]["admin"]["password"],
         "default_tenant" => node["keystone"]["default"]["tenant"],
         "default_tenant_id" => node["keystone"]["default"]["tenant_id"],
+        "default_create_user" => node["keystone"]["default"]["create_user"],
         "default_user" => node["keystone"]["default"]["username"],
         "default_password" => node["keystone"]["default"]["password"],
```
vuntz: Actually, I wouldn't put a `default_create_user` entry here, but I would make sure that `default_user` and `default_password` are `nil` if we don't create the user.

skazi0:
Added setting the cached "default_user/default_password" to nil when the user is not created.

crowbar_framework/app/views/barclamp/keystone/_edit_attributes.html.haml
```diff
@@ -16,8 +16,15 @@
       = string_field %w(default tenant)
       = string_field %w(admin username)
       = password_field %w(admin password)
-      = string_field %w(default username)
-      = password_field %w(default password)
+
+      = boolean_field %w(default create_user),
+        'data-showit' => 'true',
+        'data-showit-target' => '#default_user_container',
+        'data-showit-direct' => 'true'
```
vuntz: I hadn't noticed before, but we should have double quotes instead of single quotes here, according to our coding style.

vuntz:
Minor detail, but +1 after it's fixed.

crowbar_framework/app/views/barclamp/keystone/_edit_attributes.html.haml
skazi0: This was changed to solve Hound's violation comments.

vuntz: Ah, fun.

@rsalevsky what's the true style here? :-)

skazi0: @vuntz, @rsalevsky
The default Hound config from https://github.com/houndci/linters/blame/master/config/haml.yml has   UnnecessaryInterpolation: -> enabled: true and it was never changed.
Our local version only overrides LineLength: -> max: 150.
So we need to decide which version is the preferred one and update our local config if needed.

rsalevsky: We always use double quotes. I will adapt the config according to this.

rsalevsky:
Is the "do not merge yet" still valid?

skazi0:
@rsalevsky: we can keep the label until we have the Hound string quoting stuff fixed.

I did some more testing/digging on this problem and the `UnnecessaryInterpolation` doesn't seem to fix this.

Hound uses [haml-lint](https://github.com/brigade/haml-lint) for testing haml files which uses RuboCop internally to check the ruby parts of haml templates. The problem is that Hound doesn't pass custom ruby checking config down to the haml-lint checker (see: https://github.com/houndci/hound/issues/1121).

At this moment, the only option I found which actually works is to disable `Style/StringLiterals` check in haml files altogether. This can be done by adding it to the `ignored_caps` in `.hound.haml.yml` (see: https://github.com/brigade/haml-lint/blob/master/lib/haml_lint/linter/README.md#rubocop). Note that you need to copy the defaults from https://github.com/brigade/haml-lint/blob/master/config/default.yml to keep the default `ignored_caps`.

skazi0:
it was agreed that we don't want to hold this PR just because of Hound so I'm removing the "do not merge yet" label.

vuntz:
+1.

@mjura, is your +1 still valid?

mjura:
yes, I'm still +1

PR #439, opened 2016-07-08
tempest: Configure and enable magnum tests

toabctl:
The tempest tests for magnum need a couple of configuration options.
Add theses options and use sane defaults.

### DIFF
--- a/chef/data_bags/crowbar/template-tempest.json
+++ b/chef/data_bags/crowbar/template-tempest.json
@@ -17,6 +17,11 @@
         "image_username": "root",
         "image_password": "",
         "default_share_type_name": "default"
+      },
+      "magnum": {
+        "image_id": "magnum-service-image",
+        "flavor_id": "m1.small",
+        "master_flavor_id": "m1.medium"
       }
     }
   },
@@ -24,7 +29,7 @@
     "tempest": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 12,
+      "schema-revision": 100,
       "element_states": {
         "tempest": [ "ready", "applying" ]
       },
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -38,6 +38,9 @@ tempest_adm_pass = node[:tempest][:tempest_adm_password]
 # manila (share)
 tempest_manila_settings = node[:tempest][:manila]
 
+# magnum (container)
+tempest_magnum_settings = node[:tempest][:magnum]
+
 register_auth_hash = { user: keystone_settings["admin_user"],
                        password: keystone_settings["admin_password"],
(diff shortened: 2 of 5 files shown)


chef/data_bags/crowbar/template-tempest.json
```diff
@@ -17,14 +17,19 @@
         "image_username": "root",
         "image_password": "",
         "default_share_type_name": "default"
+      },
+      "magnum": {
+        "image_id": "magnum-service-image",
+        "flavor_id": "m1.small",
+        "master_flavor_id": "m1.medium"
```
aplanas: we are not going to use this `m1.magnum` flavor?

toabctl: we don't have it. But we can also create ```m1.magnum``` and use that.

toabctl:
We want to merge first https://github.com/SUSE-Cloud/automation/pull/1061 to be able to test this PR

chef/data_bags/crowbar/template-tempest.json
jgrassler: I played with this in the course of trying to replicate the gating failures of https://review.openstack.org/#/c/326428/ locally and found the following settings work too:

```
[magnum]
nic_id = floating
image_id = magnum-service-image
flavor_id = tempest-stuff
master_flavor_id = m1.small
```

`tempest-stuff` is a lot smaller than `m1.small` but will still do the trick for most tests. Likewise, `m1.small` is big enough for the master. Sadly these two...

* `magnum.tests.functional.api.v1.test_bay.BayTest.test_certificate_sign_and_show`
* `magnum.tests.functional.api.v1.test_bay.BayTest.test_create_list_and_delete_bays`

...continue to run into `No valid host was found` trouble. I haven't taken a
closer look, yet, but going by its name at least the latter test creates
multiple bays until there are too many bays for a two node cloud to cope.

jgrassler: What I forgot to mention earlier: `nic_id` is currently missing from the magnum configuration dict in the data bag. That one needs to be set, otherwise `magnum` will fall back to its (useless to us) default of `public`.

dirkmueller:
+1

aplanas:
+1

chef/data_bags/crowbar/template-tempest.json
toabctl: > What I forgot to mention earlier: nic_id is currently missing from the magnum configuration dict in the data bag. That one needs to be set, otherwise magnum will fall back to its (useless to us) default of public.

```nic_id``` is hardcoded to floating for us. AFAIC we are not allowing to change the public network name

jgrassler: Sorry, missed the bit where it's hardwired in the `tempest.conf.erb`. In that case it'll work of course.

PR #440, opened 2016-07-11
[magnum] Fix register Magnum domain

mjura:


chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,31 +43,47 @@
...
+            --description "Owns users and projects created by magnum" \
+            #{trustee_domain_name})
+        MAGNUM_DOMAIN_ID=$id
+    fi
+
+    [ -n "$MAGNUM_DOMAIN_ID" ] || exit 1
 
     id=
-    eval $(openstack #{insecure} user create \
+    eval $(openstack #{insecure} \
```
toabctl: Can't we use ```add_user``` from ```keystone_register``` ?

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,31 +43,47 @@
...
+        domain show \
+        -f shell --variable id \
+        #{trustee_domain_name})
+
+    MAGNUM_DOMAIN_ID=$id
+
+    if [ -z "$MAGNUM_DOMAIN_ID" ]; then
+        id=
+        eval $(openstack #{insecure} \
+            domain create \
```
toabctl: What about adding a ```:add_domain``` action to ```keystone_register``` and use that?

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,31 +43,47 @@
...
         --password #{trustee_domain_admin_password} \
         --description "Manages users and projects created by magnum" \
-        #{trustee_domain_admin})
-    ret=
-    eval $(openstack #{insecure} role add \
-        --user $TRUSTEE_DOMAIN_ADMIN_ID \
-        --domain $TRUSTEE_DOMAIN_ID admin)
+        #{trustee_domain_admin}
+
+    openstack #{insecure} \
```
toabctl: What about adding a ```:add_role``` action to ```keystone_register``` and use that?

dirkmueller: afaik keystone_register doesn't use v3 so you can't do that

chef/cookbooks/magnum/recipes/setup.rb
mjura: @toabctl: I've checked `keystone_register` resource and the reason, why we have to do this in this way is, that we are creating user with `trustee_domain` . Adding domains is not supported by this resource and better would to keep this in one logical block like for Heat (in Hear we have similar solution)

chef/cookbooks/magnum/recipes/setup.rb
mjura: `:add_domain` doesn't exist for `keystone_register`

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,31 +43,47 @@
...
   code <<-EOF
+
+    # Find domain ID
     id=
-    eval $(openstack #{insecure} domain create \
-        #{trustee_domain_name} \
-        --description 'Owns users and projects created by magnum' \
-        --or-show -f value -c id)
-    TRUSTEE_DOMAIN_ID=$id
+    eval $(openstack #{insecure} \
```
dirkmueller: is anything actually installing openstack client? doesn't look like..

dirkmueller:
I wonder if we shouldn't just do a "only_if { .. not already there .. }" clause instead of that complicated logic

chef/cookbooks/magnum/recipes/setup.rb
mjura: Adding new action is completely new feature, I agree with that but this is new card which has also influence on Heat for example.

chef/cookbooks/magnum/recipes/setup.rb
mjura: Adding new action is completely new feature, I agree with that but this is new card which has also influence on Heat for example.

chef/cookbooks/magnum/recipes/setup.rb
mjura: Thx, I will check it

mjura: I've added `python-openstackclient` package to the list of packages

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -23,9 +23,11 @@
 trustee_domain_admin = node["magnum"]["trustee"]["domain_admin_name"]
 trustee_domain_admin_password = node["magnum"]["trustee"]["domain_admin_password"]
 
-# Install magnumclient package
+# Install openstack and magnu mclient packages
```
sayalilunkad: s/magnu mclient/magnum client

mjura:
I fixed logic and I did tests, works fine

chef/cookbooks/magnum/recipes/setup.rb
dirkmueller: thanks

dirkmueller:
+1

toabctl:
+1

PR #441, opened 2016-07-11
[magnum] Enable global_index search in Heat policy

mjura:
Enable stacks global_index search in Heat policy.json,
so that Magnum can use list(global_tenant=True)

https://bugs.launchpad.net/magnum/+bug/1589955

### DIFF
--- a/chef/cookbooks/heat/recipes/server.rb
+++ b/chef/cookbooks/heat/recipes/server.rb
@@ -390,6 +390,23 @@ template "/etc/heat/heat.conf" do
   )
 end
 
+magnum_env_filter = "magnum-server AND magnum_config_environment:magnum-config-default"
+magnum_servers = search(:node, "roles:#{magnum_env_filter}") || []
+unless magnum_servers.empty?
+
+  # https://bugs.launchpad.net/magnum/+bug/1589955
+  # enable stacks global_index search so that magnum can use
+  # list(global_tenant=True)
+  ruby_block "update_heat_policy" do
+    block do
+      file = Chef::Util::FileEdit.new("/etc/heat/policy.json")
+      file.search_file_replace_line("stacks:global_index",
+                                    '    "stacks:global_index": "role:admin",')
+      file.write_file
+    end
+  end
+end
+
 service "heat-engine" do
   service_name node[:heat][:engine][:service_name]
   supports status: true, restart: true


vuntz:
cc @jgrassler

chef/cookbooks/magnum/recipes/heat.rb
```diff
@@ -0,0 +1,29 @@
...
+# Recipe:: heat
+#
+
+# https://bugs.launchpad.net/magnum/+bug/1589955
+# enable stacks global_index search so that magnum can use
+# list(global_tenant=True)
+ruby_block "update_heat_policy" do
+  block do
+    file = Chef::Util::FileEdit.new("/etc/heat/policy.json")
+    file.search_file_replace_line("stacks:global_index", '    "stacks:global_index": "role:admin",')
```
toabctl: If we want that, can we just update policy file in the package?

jgrassler: It'd rather have this in in Crowbar rather than the package: otherwise we'd always be shipping a questionable `policy.json` for Heat, even in cases where that is not required (i.e. clouds that do not even run Magnum).

One could argue about whether this should really go into the Magnum barclamp or whether it should be some code in the Heat barclamp checking whether Magnum is running as well. Personally, I'd prefer the Heat barclamp, since there's less potential for surprises that way.

mjura: I moved this code to Heat cookbook, right now it should be visible.

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -390,6 +390,22 @@
...
+magnum_servers = search(:node, "roles:#{magnum_env_filter}") || []
+if magnum_servers.length > 0
+
+  # https://bugs.launchpad.net/magnum/+bug/1589955
+  # enable stacks global_index search so that magnum can use
+  # list(global_tenant=True)
+  ruby_block "update_heat_policy" do
+    block do
+      file = Chef::Util::FileEdit.new("/etc/heat/policy.json")
+      file.search_file_replace_line("stacks:global_index", '    "stacks:global_index": "role:admin",')
```
mjura: I would like to merge it, how it is

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -390,6 +390,22 @@
...
+      file.write_file
+    end
+  end
+end
```
vuntz: You need to restart heat after this change.

mjura: @vuntz: No, restart is not needed. This file will be loaded and policy will be checked again during creating Heat stack.

jgrassler:
+1 Looks good now. thanks for moving it to the Heat barclamp!

mjura:
I fixed hound violations about `Too long line`

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -390,6 +390,23 @@
   )
 end
 
+magnum_env_filter = "magnum-server AND magnum_config_environment:magnum-config-default"
+magnum_servers = search(:node, "roles:#{magnum_env_filter}") || []
+unless magnum_servers.length.empty?
```
jsuchome: remove `.length`

mjura: Done

jsuchome:
+1 (not tested, but looks correct) Hopefully we can remove that once @jgrassler 's fix is approved upstream.

toabctl:
+1

toabctl:
Retrigger gating with ```want_magnum``` (see https://ci.suse.de/job/openstack-mkcloud/27233/ )

PR #442, opened 2016-07-12
Add tunnel_csum attribute to neutron barclamp

Itxaka:
As per http://openvswitch.org/pipermail/dev/2015-August/059335.html enabling tunnel_csum on the openvswitch agents may boost the network performance by a huge margin.

### DIFF
--- a/chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
@@ -186,6 +186,9 @@ enable_distributed_routing = True
 # carrying GRE/VXLAN tunnel. The default value is False.
 #
 # tunnel_csum = False
+<% if @tunnel_csum -%>
+tunnel_csum = True
+<% end -%>
 
 # (StrOpt) agent_type to report.
 # This config entry allows configuration of the neutron agent type reported
--- a/chef/data_bags/crowbar/migrate/neutron/100_add_tunnel_csum.rb
+++ b/chef/data_bags/crowbar/migrate/neutron/100_add_tunnel_csum.rb
@@ -1,0 +1,14 @@
+def upgrade(ta, td, a, d)
+  a["ovs"] ||= {}
+  a["ovs"]["tunnel_csum"] = ta["ovs"]["tunnel_csum"]
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  if a.key?("ovs")
+    a["ovs"].delete("tunnel_csum")
+  end
+
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-neutron.json
+++ b/chef/data_bags/crowbar/template-neutron.json
@@ -29,6 +29,9 @@
         "vni_end": 99999,
         "multicast_group": "239.1.1.1"
       },
+      "ovs": {
+        "tunnel_csum": false
+      },
       "allow_overlapping_ips": true,
       "use_syslog": false,
       "database_instance": "none",
@@ -117,7 +120,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 48,
+      "schema-revision": 100,
       "element_states": {
         "neutron-server": [ "readying", "ready", "applying" ],
         "neutron-network": [ "readying", "ready", "applying" ]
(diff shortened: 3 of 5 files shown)


chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
```diff
@@ -186,7 +186,9 @@ enable_distributed_routing = True
 # carrying GRE/VXLAN tunnel. The default value is False.
 #
 # tunnel_csum = False
-
+<% if @tunnel_csum -%>
+tunnel_csum = True
+<% end -%>
```
aplanas: Can we have the removed new line here?

aplanas:
LGTM +1

Itxaka:
Btw, this only adds the option, and sets the default value to false. Unfortunately I could not find a lot of information regarding what could be affected by setting it to true by default.

toabctl:
you need to update the schema-revision in the template-neutron.json file

bmwiedemann:
Enabling offloading to network devices has often broken network in subtle ways, because much of the offloading-code in the various drivers did not handle the extra 4 byte added by tagged VLANs correctly. And apparently very few people use tagged VLANs.
So it is good that it is off by default.

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -118,7 +118,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 48,
+      "schema-revision": 49,
```
toabctl: Sorry for not mentioning it - we agreed to start with 100 for Cloud7 :)

Itxaka:
I also bumped the name of the migration. Should they go together? I actually have no idea how the migration is applied and based on what, but it makes sense that would run the migration up tot he schema number?

vuntz:
Can you squash all commits into one commit? (you can force push, to not have to reopen another PR) After that, +1.

nkrinner:
@Itxaka As far as I know the migration number has to match the schema revision. Thanks!

toabctl:
> I actually have no idea how the migration is applied and based on what, but it makes sense that would run the migration up tot he schema number?

you can have a look at the rake tasks on the admin node: 
```
cd /opt/dell/crowbar_framework/bin/
./rake  -T|grep migrate
```

To get an overview:
```
RAILS_ENV=production rake crowbar:schema_migrate_status
```

To migrate all barclamps:
```
RAILS_ENV=production rake crowbar:schema_migrate
```

toabctl:
+1

rsalevsky:
+1

sayalilunkad:
+1

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -13,6 +13,7 @@
       "dhcp_domain": "openstack.local",
       "use_lbaas": true,
       "use_dvr": false,
+      "tunnel_csum": false,
```
vuntz: Could we put the attribute inside a ovs attribute, to make it clear that it's ovs-specific?

Itxaka: done

chef/data_bags/crowbar/migrate/neutron/100_add_tunnel_csum.rb
```diff
@@ -0,0 +1,16 @@
+def upgrade(ta, td, a, d)
+  unless a.key?("ovs")
+    a["ovs"] = {}
+  end
```
vuntz: `a["ovs"] ||= {}`

vuntz:
Thanks for the change. There's one minor code style comment from me, but generally +1.

Btw, you can squash all commits together.

chef/data_bags/crowbar/migrate/neutron/100_add_tunnel_csum.rb
Itxaka: done!

vuntz:
+1

Itxaka:
@aspiers @vuntz @nkrinner @toabctl I have the possibility of merging this myself, but not sure if there is some key members that are in charge of merging stuff or  all team members can just merge any patches (as long as everything is green!)

vuntz:
@Itxaka you can merge your PR yourself, it's fine. The only policy is that it should not be merged unless there are two +1 for the latest status of your PR (here, there's only one). So I'd recommend asking one of the persons who gave an earlier +1 to confirm their +1 :-)

toabctl:
+1

PR #443, opened 2016-07-12
trove: Replace openstack-identity with keystone_register

toabctl:
This is a first step to unify the trove cookbook with the rest of the
cookbooks.
Remove the openstack-identity cookbook and replace it with the keystone_register
methods to have a single resource provider handling the user/service/endpoint
registration.
Note: The reason why the code to register the endpoints and users moved to api.rb
is that this is an ongoing effort to also get rid of openstack-database and
have the chef recipies all in the trove cookbook. And all the other services are
doing the endpoint/service/user handling in the api.rb recipe.


toabctl:
note: completly untested....

chef/data_bags/crowbar/template-trove.schema
```diff
@@ -47,6 +54,7 @@
             "crowbar-status": { "type": "str" },
             "crowbar-failed": { "type": "str" },
             "crowbar-queued": { "type": "bool" },
+            "schema-revision": { "type": "int" },
```
aplanas: I guess that this is the reason why the migration of Trove fails. Maybe this will cause problems in the upgrade

toabctl: @aplanas hm. where does it fail?

aplanas: In the barclamp PoV: Key 'api' needed error. Checking this `rake` command that I see from you:

```
# RAILS_ENV=production rake crowbar:schema_migrate_status
(in /opt/dell/crowbar_framework)
*barclamp*           *latest revision*    *proposals revision*
ceilometer           101                  default:101
...
tempest              12                   default:12
trove                100                  default:
updater                                   
```

When I try to run the migrations:

``` # RAILS_ENV=production rake crowbar:schema_migrate
(in /opt/dell/crowbar_framework)
rake aborted!
Failed to migrate proposal default for trove: error while executing migration script /opt/dell/chef/data_bags/crowbar/migrate/trove/001_add_configurable_db_attrs.rb:
undefined method `[]=' for nil:NilClass
/opt/dell/crowbar_framework/lib/schema_migration.rb:214:in `rescue in migrate_proposal'
/opt/dell/crowbar_framework/lib/schema_migration.rb:211:in `migrate_proposal'
/opt/dell/crowbar_framework/lib/schema_migration.rb:43:in `block in run_for_bc'
/opt/dell/crowbar_framework/lib/schema_migration.rb:42:in `run_for_bc'
/opt/dell/crowbar_framework/lib/schema_migration.rb:23:in `block in run'
/opt/dell/crowbar_framework/lib/schema_migration.rb:22:in `each'
/opt/dell/crowbar_framework/lib/schema_migration.rb:22:in `run'
/opt/dell/crowbar_framework/lib/tasks/crowbar.rake:27:in `block (2 levels) in <top (required)>'
Tasks: TOP => crowbar:schema_migrate
(See full trace by running task with --trace)
```

Deactivating and deleting the barclamp, allows the installation of trove. Now I am running tempest just in case. But I need to check how to enable tempests tests for Trove, if there is any

sayalilunkad: I still get the same error as well.

aplanas:
Tempest fails in the database service with `error: [Errno 111] Connection refused` (tempest.api.database.*) Checking why.

aplanas:
The tempest problem was that the password in mysql for the Trove user and the Trove database was not set properly. This issue is easy to reproduce: remove the barclamp and remove the trove database (do not touch the trove user in the mysql database). Reapplying the barclamp will create the db, but the password is incorrect.

But this issue is not related with this PR. I only see the migration problem.

toabctl:
@aplanas @sayalilunkad hopefully fixed now. I'm retesting...

toabctl:
@sayalilunkad @aplanas Tested and works for me now.

aplanas:
+1

sayalilunkad:
I am still getting some error to start the trove-api service but that could be due to the stale db, still checking.

Failed to apply the proposal to: d52-54-77-77-77-01.c14.cloud.suse.de 
[2016-07-15T13:54:50+00:00] INFO: template[/etc/trove/trove.conf] updated content 
[2016-07-15T13:54:50+00:00] INFO: template[/etc/trove/trove.conf] owner changed to 467 
[2016-07-15T13:54:50+00:00] INFO: template[/etc/trove/trove.conf] group changed to 471 
[2016-07-15T13:54:50+00:00] INFO: template[/etc/trove/trove.conf] mode changed to 640 
[2016-07-15T13:54:50+00:00] INFO: template[/etc/trove/trove.conf] sending restart action to service[trove-api] (immediate) 
[2016-07-15T13:54:50+00:00] INFO: Processing service[trove-api] action restart (openstack-database::api line 41) 

Error executing action `restart` on resource 'service[trove-api]' 

sayalilunkad:
This seems to happen only when we try to replace and already existing trove barclamp. IMO this PR works when a fresh install is done with this code so the PR does what it is supposed to do. We can probably file a bug for trove to address the database problem. So  +1.

toabctl:
> This seems to happen only when we try to replace and already existing trove barclamp. IMO this PR works when a fresh install is done with this code so the PR does what it is supposed to do. We can probably file a bug for trove to address the database problem. So +1.

@sayalilunkad Confirmed. I installed a cloud with mkcloud (trove was deployed then) and added this patch and redeployed trove. Works fine for me.
I then deactivated and deleted the proposal and then recreated/reapply and trove-api did not start. The logs says:
```
2016-07-18 06:39:27.414 19231 ERROR root     connection = self.__pool._invoke_creator(self)
2016-07-18 06:39:27.414 19231 ERROR root   File "/usr/lib64/python2.7/site-packages/sqlalchemy/engine/strategies.py", line 97, in connect
2016-07-18 06:39:27.414 19231 ERROR root     return dialect.connect(*cargs, **cparams)
2016-07-18 06:39:27.414 19231 ERROR root   File "/usr/lib64/python2.7/site-packages/sqlalchemy/engine/default.py", line 385, in connect
2016-07-18 06:39:27.414 19231 ERROR root     return self.dbapi.connect(*cargs, **cparams)
2016-07-18 06:39:27.414 19231 ERROR root   File "/usr/lib64/python2.7/site-packages/MySQLdb/__init__.py", line 81, in Connect
2016-07-18 06:39:27.414 19231 ERROR root     return Connection(*args, **kwargs)
2016-07-18 06:39:27.414 19231 ERROR root   File "/usr/lib64/python2.7/site-packages/MySQLdb/connections.py", line 187, in __init__
2016-07-18 06:39:27.414 19231 ERROR root     super(Connection, self).__init__(*args, **kwargs2)
2016-07-18 06:39:27.414 19231 ERROR root OperationalError: (_mysql_exceptions.OperationalError) (1045, "Access denied for user 'trove'@'localhost' (using password: YES)")
```

sayalilunkad:
@toabctl yes but i think fixing that might be out of the scope of this PR so it is still a +1 from my side, just that we should file a bug for this problem.

aplanas:
I had the same problem, described in a previous comment. The workaround the manually set the password in the mysql db, or remove the travis user.

Still for me is a different issue that the one that this PR resolve. So we need a different one.

chef/cookbooks/trove/recipes/default.rb
```diff
@@ -115,56 +115,8 @@ class ::Chef::Recipe
 node.set["openstack"]["database"]["insecure"] = keystone_settings["insecure"]
 node.set["openstack"]["region"] = keystone_settings["endpoint_region"]
 node.set["openstack"]["database"]["region"] = keystone_settings["endpoint_region"]
```
vuntz: Do we still need everything that is done above? This was really only to enable the openstack-* cookbooks.

chef/cookbooks/trove/recipes/taskmanager.rb
```diff
@@ -0,0 +1,53 @@
...
+    rabbit_trove_settings: rabbitmq_trove_settings,
+    nova_url: nova_url,
+    nova_insecure: nova_insecure,
+    cinder_url: cinder_url,
+    cinder_insecure: cinder_insecure,
+    object_store_url: object_store_url,
+    object_store_insecure: object_store_insecure
+  )
+
+  notifies :restart, "service[trove-taskmanager]", :immediately
```
vuntz: We usually don't do the `:immediately` unless there's a real need for it.

chef/cookbooks/trove/recipes/conductor.rb
```diff
@@ -0,0 +1,40 @@
...
+  owner node[:trove][:user]
+  group node[:trove][:group]
+  mode 00640
+  variables(
+    keystone_settings: KeystoneHelper.keystone_settings(node, :trove),
+    sql_connection: sql_connection,
+    rabbit_default_settings: fetch_rabbitmq_settings,
+    rabbit_trove_settings: rabbitmq_trove_settings
+  )
+  notifies :restart, "service[trove-conductor]", :immediately
```
vuntz: Same thing here.

chef/cookbooks/trove/recipes/api.rb
```diff
@@ -0,0 +1,168 @@
...
+    nova_insecure: nova_insecure,
+    cinder_url: cinder_url,
+    cinder_insecure: cinder_insecure,
+    object_store_url: object_store_url,
+    object_store_insecure: object_store_insecure,
+    bind_host: bind_host,
+    bind_port: bind_port
+    )
+
+  notifies :restart, "service[trove-api]", :immediately
```
vuntz: And here.

chef/cookbooks/trove/definitions/trove_service.rb
```diff
@@ -0,0 +1,17 @@
+define :trove_service, use_pacemaker_provider: false do
+  trove_service_name = "trove-#{params[:name]}"
+  trove_name = trove_service_name
+  trove_name = "openstack-trove-#{params[:name]}"\
+                if ["rhel", "suse"].include? node[:platform_family]
+
+  package trove_name if ["rhel", "suse"].include? node[:platform_family]
```
vuntz: With this line, I guess we can remove some of the `package` line in the recipes.

chef/cookbooks/trove/definitions/trove_service.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  service trove_service_name do
+    service_name trove_name
+    supports status: true, restart: true
+    action [:enable, :start]
+    subscribes :restart, resources(template: "/etc/trove/trove.conf")
```
vuntz: Maybe replace `resources(template: "/etc/trove/trove.conf")` with `"template[/etc/trove/trove.conf]"` as it allows to define the template later on in the code, which might be useful.

chef/cookbooks/trove/libraries/helpers.rb
```diff
@@ -0,0 +1,127 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+class Chef
+  class Recipe
+    def get_sql_connection
```
vuntz: Hrm, I'm not a big fan of putting this in `Chef::Recipe`. It means that other cookbooks can call this method which has no trove in its name, and that would be confusing.

chef/cookbooks/trove/libraries/helpers.rb
```diff
@@ -0,0 +1,127 @@
...
+      # get Database data
+      db_password = ""
+      if node.roles.include? "trove-server"
+        db_password = node[:trove][:db][:password]
+      else
+        # pickup password to database from trove-server node
+        node_servers = search(:node, "roles:trove-server") || []
```
vuntz: s/`search`/`search_env_filtered`/
s/`|| []`//

chef/cookbooks/trove/libraries/helpers.rb
```diff
@@ -0,0 +1,127 @@
...
+      # running on the same node
+      "mysql://#{node[:trove][:db][:user]}:"\
+      "#{db_password}@127.0.0.1/"\
+      "#{node[:trove][:db][:database]}"
+    end
+
+    def get_rabbitmq_trove_settings
+      # get rabbitmq-server information
+      # NOTE: Trove uses it's own vhost instead of the default one
+      rabbitmq_servers = search(:node, "roles:rabbitmq-server") || []
```
vuntz: Same thing.

chef/cookbooks/trove/libraries/helpers.rb
vuntz: Actually, `get_instance` would be better.

In general for all the searches in this file, use the same method than in the old code from the default.rb recipe.

chef/cookbooks/trove/recipes/api.rb
```diff
@@ -0,0 +1,168 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Cookbook Name:: trove
+# Recipe:: api
+#
+
+keystone_settings = KeystoneHelper.keystone_settings(node, :trove)
```
vuntz: I might be wrong, but I think you can skip `:trove` as it will be guessed correctly.

chef/cookbooks/trove/recipes/api.rb
```diff
@@ -0,0 +1,168 @@
...
+
+ha_enabled = false
+
+# address/port binding
+my_ipaddress = Chef::Recipe::Barclamp::Inventory.get_network_by_type(
```
vuntz: You can remove the `Chef::Recipe::` prefix.

chef/cookbooks/trove/recipes/default.rb
toabctl: Ups. I'll remove all of theses.

toabctl: done

chef/cookbooks/trove/recipes/taskmanager.rb
toabctl: done

chef/cookbooks/trove/recipes/conductor.rb
toabctl: done

chef/cookbooks/trove/recipes/api.rb
toabctl: done

chef/cookbooks/trove/definitions/trove_service.rb
toabctl: done

chef/cookbooks/trove/definitions/trove_service.rb
toabctl: done

chef/cookbooks/trove/libraries/helpers.rb
toabctl: @vuntz so where should I move the code? Or should I prefix the function names with ```trove_``` ?

chef/cookbooks/trove/recipes/api.rb
toabctl: done

chef/cookbooks/trove/recipes/api.rb
toabctl: that's how other cookbooks are doing. I'll leave it here for now. Doesn't hurt.

chef/cookbooks/trove/libraries/helpers.rb
toabctl: done

chef/cookbooks/trove/libraries/helpers.rb
toabctl: done

chef/cookbooks/trove/libraries/helpers.rb
toabctl: done

sayalilunkad:
So this still failed for me with the same problem.
I found a workaround to this problem. Seems like the way mysql interprets '%' in grants statement excludes localhost which causes the permission denied problem. So explicitly granting permission to locahost seems to help. 
>> #grant privs to user
 +database_user node[:trove][:db][:user] do
 +  provider ::Chef::Provider::Database::MysqlUser
 +  connection conn
 +  password node[:trove][:db][:password]
 +  database_name node[:trove][:db][:database]
 +  host "localhost"
 +  privileges [:all]
 +  action :grant
 +end

See this blog post for more detailed explanation: http://www.tikalk.com/devops/solution-mysql-error-1045-access-denied-userlocalhost-breaks-openstack/

chef/cookbooks/trove/libraries/helpers.rb
```diff
@@ -0,0 +1,106 @@
...
+      if ceph_radosgws.empty?
+        object_store_url = nil
+        object_store_insecure = false
+      else
+        radosgw = ceph_radosgws[0]
+        radosgw_api_host = CrowbarHelper.get_host_for_admin_url(
+          radosgw, (radosgw[:ceph][:ha][:radosgw][:enabled] rescue false)
+        )
+        radosgw_api_protocol = radosgw[:ceph][:radosgw][:ssl][:enabled] ? "https" : "http"
+        radosgw_api_port = radosgw[:ceph][:radosgw][:ssl][:enabled] ?
```
toabctl: Tsts. I had a ```if / else``` before and hound was also complaining...

toabctl:
> So this still failed for me with the same problem.
> I found a workaround to this problem. Seems like the way mysql interprets '%' in grants statement excludes localhost which causes the permission denied problem. So explicitly granting permission to locahost seems to help.

@sayalilunkad I think the problem is (after deleting the trove barclamp) that the db password is deleted and then a new password is created. And after that the permissions (see  ```show grants for 'trove'@'%';```) are not updated. That's why we get a ```Access denied for user 'trove'@'localhost' (using password: YES)```. But this problem is unrelated to this PR because the problem already exits.

chef/cookbooks/trove/libraries/helpers.rb
rhafer: Not sure what you had before, but I guess it wants you to do this:

```
radosgw_api_port = if radosgw[:ceph][:radosgw][:ssl][:enabled]
  node[:ceph][:radosgw][:rgw_port_ssl]
else
  node[:ceph][:radosgw][:rgw_port]
end
```

chef/cookbooks/trove/recipes/api.rb
```diff
@@ -0,0 +1,176 @@
...
+    nova_insecure: nova_insecure,
+    cinder_url: cinder_url,
+    cinder_insecure: cinder_insecure,
+    object_store_url: object_store_url,
+    object_store_insecure: object_store_insecure,
+    bind_host: bind_host,
+    bind_port: bind_port
+    )
+
+  notifies :restart, "service[trove-api]"
```
rhafer: Isn't this somewhat redundant since the service is already subscribed to the template?

chef/cookbooks/trove/recipes/default.rb
```diff
@@ -17,154 +17,7 @@
...
-end
-
-include_recipe "openstack-database::identity_registration"
-include_recipe "openstack-database::api"
-include_recipe "openstack-database::conductor"
-include_recipe "openstack-database::taskmanager"
+include_recipe "trove::sql"
+include_recipe "trove::api"
+include_recipe "trove::conductor"
+include_recipe "trove::taskmanager"
```
rhafer: Nitpick: should we update the years in the copyright header here?

chef/data_bags/crowbar/template-trove.schema
rhafer: For me the migration seems to have worked. But I guess the change to 001_add_configurable_db_attrs.rb did that.

rhafer:
In general this looks good to me. I only tested it on a already deployed system up to now, will give it a try on a clean install tomorrorw.

@toabctl  What are your plans regarding the remaining hound complaints?

toabctl:
> @toabctl What are your plans regarding the remaining hound complaints?

@rhafer I'll ignore Hound

chef/cookbooks/trove/recipes/default.rb
toabctl: done

chef/cookbooks/trove/recipes/api.rb
toabctl: done

rhafer:
> @rhafer I'll ignore Hound

@toabctl Hm, you should at least address those complaints about the empty lines, line length and indentation depth. The conditional assignment stuff should also be easy to address.

rhafer:
Apart from the hound related issues I am +1 on this. Since @toabctl  is away for the rest of the week. I reopen another PR with those fixes on top.

rhafer:
Hm, not sure why this one shows up as merged here. I didn't merge this but I merged #481, which had a commit on top of this PR.

aplanas:
@rhafer all the PR of this are merged, because this is a subset of #481, so is quite right.

rhafer:
Yeah, I guess that makes sense somehow.

PR #444, opened 2016-07-13
Enable role deployment for SLES12SP2 only

jsuchome:
This will be needed once we switch to SP2 as default architecture for nodes


aplanas:
+1 (for when this can be merged)

nicolasbock:
+1

jsuchome:
Rebased, and restarted gating with want_sles12sp2=1

PR #445, opened 2016-07-13
Add  ssl to manila barclamp

Itxaka:


crowbar_framework/app/helpers/manila_barclamp_helper.rb
```diff
@@ -45,4 +45,14 @@ def netapp_transports_for_manila(selected)
       selected.to_s
     )
   end
+
+  def api_protocols_for_heat(selected)
```
skazi0: the _heat part should be changed to _manila to match the reference in the haml file

nicolasbock:
What about a migration?

toabctl:
@Itxaka Can you create a PR in parallel for our CI? See https://github.com/SUSE-Cloud/automation/blob/master/scripts/qa_crowbarsetup.sh#L2619 .

Itxaka:
@nicolasbock Godd point!

@toabctl thanks, I will have a look

Itxaka:
@nicolasbock I didnt do a migration because the ssl attributes were already in the schema+json since the start of the barclamp (https://github.com/crowbar/crowbar-openstack/commit/bc28b2305a97b84f4ac3cb88b44cae56a781b3d1) so I didnt think it was necessary.

Is it?

nicolasbock:
@itxaka Yes you are right. I should have looked more carefully :) No
migration necessary.

toabctl:
+1 (after you squashed to 2 commits)

nicolasbock:
+1

toabctl:
Let's wait for milestone 2 before we merge this.

vuntz:
Wondering out loud: there's code in horizon and tempest cookbooks, iirc, that checks if one of the openstack service requires the `--insecure` bit. Don't we need to add manila to these checks? (and possibly other services which might be missing)

Itxaka:
@vuntz that is a very good point, let me check that just in case.

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -253,7 +253,7 @@
 end
 
 manilas = search(:node, "roles:manila-server") || []
-if manilas.length > 0
+if !manilas.empty?
```
nicolasbock: Can you squash this commit into the previous one?

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -252,6 +252,14 @@
   heat_insecure = false
 end
 
+manilas = search(:node, "roles:manila-server") || []
```
toabctl: can you use ```search_env_filtered``` from the openstack-core utils recipe? or ```get_instance```

Itxaka: done

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -252,6 +252,13 @@
   heat_insecure = false
 end
 
+manila = get_instance("roles:manila-server")
```
vuntz: Actually, I think @toabctl's suggestion is not fully correct: you can use `get_instance` or `search_env_filtered` only when the manila proposal is referenced in the attributes of the horizon proposal, which is not the case. So from that perspective, we should do the same thing as what we do for heat a couple of lines above.

Itxaka: Indeed, after a local test `search_env_filtered` checks in the local node for [barclamp][:config][:environment]

Im guessing it may work if manila is installed on the same node as horizon, but if its not, it wont work. Reverting the change, thanks @vuntz for the catch.

chef/cookbooks/manila/recipes/common.rb
```diff
@@ -142,4 +142,20 @@
   )
 end
 
+# ssl
+if node[:manila][:api][:protocol] == "https"
+  if node[:manila][:ssl][:generate_certs]
+    package "openssl"
+  end
```
vuntz: Unrelated to this PR, but I believe this should be done automatically by `ssl_setup`. @nicolasbock any reason why this wasn't done that way?

vuntz:
+1

chef/cookbooks/manila/recipes/common.rb
nicolasbock: Hmm, an oversight on my part :( I'll fix it.

Itxaka:
waiting for https://github.com/crowbar/crowbar-openstack/pull/465 to be merged so I can change this pr

Itxaka:
Updated to be in sync with https://github.com/crowbar/crowbar-openstack/pull/465 changes

nicolasbock:
+1

vuntz:
+1

PR #446, opened 2016-07-13
[stable/3.0] backport #416

mmnelemane:
Support Cisco APIC integration on Cloud6.


rsalevsky:
Does this require new packages?

mmnelemane:
@rsalevsky: yes.. it requires new packages. That should not cause problem because we have not yet enabled it from crowbar_framework. Some ovs fixes are required (from Cisco) before we enable the feature from crowbar_framework. This PR is just to make sure the code is available for demos on Cloud6 whenever some customer needs.

dirkmueller:
+1

rsalevsky:
@mmnelemane Thanks for clarification, please ping me when you enable it in crowbar.

+1

PR #447, opened 2016-07-14
bsc# 984128: Match MTU values of VLAN and Neutron ml2

nicolasbock:
Neutron correctly subtracts the GRE or VXLAN overheads and the MTU for Neutron
routers can be set to the MTU of the underlying NIC.

Forward port of https://github.com/crowbar/crowbar-openstack/pull/411.

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_config.rb
+++ b/chef/cookbooks/neutron/recipes/common_config.rb
@@ -126,6 +126,9 @@ if neutron[:neutron][:elements_expanded]
   network_nodes_count = neutron[:neutron][:elements_expanded]["neutron-network"].count
 end
 
+os_sdn_net = Barclamp::Inventory.get_network_definition(node, "os_sdn")
+mtu_value = os_sdn_net.nil? ? 1500 : os_sdn_net["mtu"].to_i
+
 template "/etc/neutron/neutron.conf" do
     cookbook "neutron"
     source "neutron.conf.erb"
@@ -157,7 +160,8 @@ template "/etc/neutron/neutron.conf" do
       allow_overlapping_ips: neutron[:neutron][:allow_overlapping_ips],
       dvr_enabled: neutron[:neutron][:use_dvr],
       network_nodes_count: network_nodes_count,
-      dns_domain: neutron[:neutron][:dhcp_domain]
+      dns_domain: neutron[:neutron][:dhcp_domain],
+      mtu_value: mtu_value
     }.merge(nova_notify))
 end
 
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -133,6 +133,9 @@ when "ml2"
     physnets.push(node[:neutron][:zvm][:zvm_xcat_mgt_vswitch])
   end
 
+  os_sdn_net = Barclamp::Inventory.get_network_definition(node, "os_sdn")
+  mtu_value = os_sdn_net.nil? ? 1500 : os_sdn_net["mtu"].to_i
+
   ml2_type_drivers = node[:neutron][:ml2_type_drivers]
   ml2_mechanism_drivers = node[:neutron][:ml2_mechanism_drivers].dup
   if use_hyperv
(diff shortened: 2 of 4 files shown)


dirkmueller:
I wonder if we shouldn't set segment_mtu instead of path_mtu. in general it  looks good to me, +1

nicolasbock:
As far as I understand it, the use of `segment_mtu` is being phased out. I will double check that though.

nicolasbock:
@dirkmueller I got this mixed up: It's `path_mtu` that's being phased out. In addition, a new config variable was introduced, `global_physnet_mtu`. I updated the patch.

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -126,6 +126,9 @@
   network_nodes_count = neutron[:neutron][:elements_expanded]["neutron-network"].count
 end
 
+os_sdn_net = Barclamp::Inventory.get_network_by_type(node, "os_sdn")
+mtu_value = os_sdn_net.nil? ? 1500 : (os_sdn_net.mtu || 1500)
```
vuntz: I think you don't need the `|| 1500` bit since you did that in the other PR already.

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -133,6 +133,9 @@
     physnets.push(node[:neutron][:zvm][:zvm_xcat_mgt_vswitch])
   end
 
+  os_sdn_net = Barclamp::Inventory.get_network_by_type(node, "os_sdn")
```
vuntz: Hrm, we really need my work from https://github.com/crowbar/crowbar-core/pull/510. This won't work in some cases here: if neutron-server node doesn't have neutron-network node, then it will not be on the os_sdn network, while the os_sdn network might still exist. So you will get unexpected results...

In the PR I reference, there's a new `get_network_definition` method which is what we would need here.

nicolasbock: Do you want us to wait until your PR is merged then?

vuntz: I don't know; it might take time until it gets reviewed :/

The alternative is to cherry-pick https://github.com/crowbar/crowbar-core/pull/510/commits/36815d71cbc57712ed6292e2c776c8a1f1b0efb9 in your PR.

nicolasbock: @vuntz I cherry picked that commit into crowbar-core (https://github.com/crowbar/crowbar-core/pull/548). I'll rework this PR now to use the new method.

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -126,6 +126,9 @@
   network_nodes_count = neutron[:neutron][:elements_expanded]["neutron-network"].count
 end
 
+os_sdn_net = Barclamp::Inventory.get_network_by_type(node, "os_sdn")
```
vuntz: Needs to be changed to use the new method we added.

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -133,6 +133,9 @@
     physnets.push(node[:neutron][:zvm][:zvm_xcat_mgt_vswitch])
   end
 
+  os_sdn_net = Barclamp::Inventory.get_network_defition(node, "os_sdn")
```
vuntz: s/defition/definition/

nicolasbock: sorry, fixing.... :(

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -126,6 +126,9 @@
   network_nodes_count = neutron[:neutron][:elements_expanded]["neutron-network"].count
 end
 
+os_sdn_net = Barclamp::Inventory.get_network_definition(node, "os_sdn")
+mtu_value = os_sdn_net.nil? ? 1500 : os_sdn_net["mtu"]
```
vuntz: Should we do something like `os_sdn_net["mtu"].to_i`?

vuntz:
Small question, but other than that, +1

chef/cookbooks/neutron/recipes/common_config.rb
nicolasbock: Fixed.

vuntz:
Still +1

abel-navarro:
+1

PR #448, opened 2016-07-14
Re-write recipe for creating Magnum trustee domain

mjura:
During first run of chef-client options `trustee_domain_id`
and `trustee_domain_admin_id` were not saved in magnum.conf.
This was bacuse of order chef resources for ruby_block,
these attributes were saved after chef-client run ends.

### DIFF
--- a/chef/cookbooks/magnum/recipes/setup.rb
+++ b/chef/cookbooks/magnum/recipes/setup.rb
@@ -17,11 +17,13 @@
 # Recipe:: setup
 #
 
+return if node["magnum"]["trustee"]["domain_id"] && node["magnum"]["trustee"]["domain_admin_id"]
+
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
-trustee_domain_name = node["magnum"]["trustee"]["domain_name"]
-trustee_domain_admin = node["magnum"]["trustee"]["domain_admin_name"]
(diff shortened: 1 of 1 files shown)


toabctl:
@mjura are you saying that https://github.com/crowbar/crowbar-openstack/pull/440 doesn't work?

mmnelemane:
@mjura:  Just a couple of questions: 
Doesn't the template changes alone fix the problem of saving the trustee domain parameters into magnum.conf ? 
Are the changes in setup.rb (moving openstack commands out of the bash block) also  part of the fix OR is it a just a cleanup of the recipe ?

mjura:
@toabctl: Changes https://github.com/crowbar/crowbar-openstack/pull/440 works too, problem is only with first run and because normally mkcloud is applying tempest after Magnum, there is second run which is fixing this issue. That's way this problem wasn't discovered before.

mjura:
@mmnelemane: Moving out from bash block is also fix for this bug. Probably this same problem apply to Heat, but it requires more tests.

mjura:
@mmnelemane: You are right changes in template `magnum.conf.rb` can be omitted.

mjura:
@mmnelemane, @toabctl: Any comments from you? I would like to merge this change

mmnelemane:
+1

toabctl:
+1

PR #452, opened 2016-07-18
Raw entry support to configure APIC from crowbar UI

mmnelemane:
Adding this PR only to extract a build (rpm) for customer demos on Cisco ACI. This is not meant for merging into SOC product. The PR can be reviewed and updated when we have finalized packages in the cloud media to run this integration. Currently, for tests, we install the packages from a temp repository. 

To be merged after https://github.com/crowbar/crowbar-openstack/pull/885


mmnelemane:
The UI message for absence of APIC configurations when one of "cisco_apic_ml2" or "apic_gbp" selected  appears as seen in the screen shot. 
screenshot_20170419_184440

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -65,15 +65,15 @@
...
-default[:neutron][:opflex][:vlan][:encap_iface] = ""
+default[:neutron][:apic][:opflex][:peer_ip] = "10.0.0.30"
+default[:neutron][:apic][:opflex][:peer_port] = 8009
+default[:neutron][:apic][:opflex][:encap] = "vxlan"
+default[:neutron][:apic][:opflex][:vxlan][:uplink_iface] = "vlan.4093"
+default[:neutron][:apic][:opflex][:vxlan][:uplink_vlan] = 4093
+default[:neutron][:apic][:opflex][:vxlan][:encap_iface] = "br-int_vxlan0"
+default[:neutron][:apic][:opflex][:vxlan][:remote_ip] = "10.0.0.32"
+default[:neutron][:apic][:opflex][:vxlan][:remote_port] = 8472
+default[:neutron][:apic][:opflex][:vlan][:encap_iface] = ""
```
rhafer: I am not really sure if this a relevant here (as we might have nobody currently using this code). But as you're renaming all these attributes, we're loosing any custom values that might have been set for these. I.e. we might be breaking existing setups. Is this something we need to handle here? Or is it ok to address that manually?

mmnelemane:
Covered in https://github.com/crowbar/crowbar-openstack/pull/638

PR #453, opened 2016-07-19
Check carefully if the role we are renaming is actually present

jsuchome:
When we are reinstalling packages from cloud5 to cloud6, ceilometer-polling role is not even installed, thus the migration fails.

I did not test it, and it might be completely wrong...


vuntz:
Do we need fixes for downgrade?

jsuchome:
I don't know; do we actually have a use case for downgrade?

vuntz:
> I don't know; do we actually have a use case for downgrade?

TBH, not really, but we usually try to still be nice citizens ;-) I don't mind much, tbh.

aplanas:
+1

vuntz:
+1

PR #457, opened 2016-07-20
Always enable force_config_drive for ZVM

mmnelemane:


vuntz:
Do the options need to be set only on the compute node, or are they needed on the controller?

dirkmueller:
+1

dirkmueller:
@vuntz I only tested it on all nodes (e.g. by changing the crowbar attribute), which works, but thinking about it I guess only doing compute node should be fine. it is the local configuration only that the driver is reading anyway.

mmnelemane:
https://github.com/crowbar/crowbar-openstack/pull/458 should do the backport

chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -1038,7 +1038,7 @@ linuxnet_interface_driver = ""
 
 # Whether to attempt to inject network setup into guest (boolean value)
 #flat_injected = false
-<% if @force_config_drive %>
+<% if @force_config_drive or @libvirt_type.eql?('zvm') %>
```
vuntz: Our guidelines say to use `||` instead of `or`, but minor detail

vuntz:
+1. Would be nice to fix the minor detail before merging, though

mmnelemane:
@vuntz: fixed

PR #459, opened 2016-07-21
cinder: Allow to use a HA cluster for cinder-volume in active/passive

vuntz:
We cannot do active/active at the moment because there needs to be some
synchonization between the various volume services in that case.

But it turns out that active/passive is relatively easy to achieve: it's
only a matter of making sure that the cinder-volume service on the
various cluster members register with the same host. We use the
vhostname of the cluster for this.

Of course, it doesn't work with the LVM backend, so we have validation
checks for this in Crowbar.

Another restriction is that we cannot use the same cluster for
cinder-controller and cinder-volume: due to the use of the same host in
the config, if we were to allow cinder-controller there too, only one
cinder-scheduler service would be register (with the vhostname of the
cluster), which is wrong in this case.

### DIFF
--- a/crowbar_framework/app/models/cinder_service.rb
+++ b/crowbar_framework/app/models/cinder_service.rb
@@ -42,6 +42,7 @@ class CinderService < PacemakerServiceObject
         "cinder-volume" => {
           "unique" => false,
           "count" => -1,
+          "cluster" => true,
           "admin" => false,
           "exclude_platform" => {
             "suse" => "< 12.2",
@@ -91,6 +92,7 @@ class CinderService < PacemakerServiceObject
 
     volume_names = {}
     local_file_names = {}
+    local_count = 0
     raw_count = 0
     raw_want_all = false
     rbd_crowbar = false
@@ -113,6 +115,8 @@ class CinderService < PacemakerServiceObject
 
         file_name = volume["local"]["file_name"]
         local_file_names[file_name] = (local_file_names[file_name] || 0) + 1
+
+        local_count += 1
       end
 
       if backend_driver == "raw"
@@ -166,11 +170,16 @@ class CinderService < PacemakerServiceObject
       end
     end
 
-    if raw_count > 0
+    volume_elements = proposal["deployment"][@bc_name]["elements"]["cinder-volume"]
+    volume_clusters = volume_elements.select { |n| is_cluster? n }
+
+    if (local_count > 0 || raw_count > 0) && volume_clusters.any?
+      validation_error I18n.t("barclamp.#{@bc_name}.validation.lvm_ha")
+    elsif raw_count > 0
         if raw_count > 1 && raw_want_all
           validation_error I18n.t("barclamp.#{@bc_name}.validation.raw_device_backend")
         else
-          nodes_without_suitable_drives = proposal["deployment"][@bc_name]["elements"]["cinder-volume"].select do |node_name|
+          nodes_without_suitable_drives = volume_elements.select do |node_name|
             node = NodeObject.find_node_by_name(node_name)
             if node.nil?
               false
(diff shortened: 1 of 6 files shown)


vuntz:
This was tested with ceph. Added "do not merge yet" as there's some code cleanup to do and I'd like to discuss this with potential users.

But some early feedback might be welcome.

dirkmueller:
I wonder if we want to do a white list than a blacklist.

aspiers:
@dirkmueller wrote:
> I wonder if we want to do a white list than a blacklist.

If you're referring to the mention-bot pinging non-SUSE people, the answer is no because it's already in the config file to restrict pings in that way, but as I mentioned before I left for vacation, the bot isn't yet honouring the modified config for reasons I didn't yet figure out.

aspiers:
@vuntz wrote:
> Another restriction is that we cannot use the same cluster for cinder-controller and cinder-volume: due to the use of the same host in the config, if we were to allow cinder-controller there too, only one cinder-scheduler service would be register (with the vhostname of the cluster), which is wrong in this case.

Sorry for the stupid question, but why would only one `cinder-scheduler` service be registered as a result of this?

vuntz:
> Sorry for the stupid question, but why would only one cinder-scheduler service be registered as a result of this?

Because the registration is based on the host configured in cinder.conf. So if it's the same host, then there's only one scheduler visible (even though multiple schedulers are running -- which might actually cause issues).

vuntz:
@aspiers updated to have a split cinder-volume.conf

crowbar_framework/config/locales/cinder/en.yml
```diff
@@ -144,6 +144,7 @@ en:
         invalid_filename: 'Invalid file name \"%{file_name}\" for local file-based LVM: file name must be an absolute path.'
         invalid_whitespaces_in_filename: 'Invalid file name \"%{file_name}\" for local file-based LVM: file name cannot contain whitespaces.'
         invalid_backend_filename: '%{count} backends are using \"%{file_name}\" for local file-based LVM.'
+        isci_ha: 'The Raw Devices and Local File backends cannot be used with High Availability clusters.'
```
aspiers: What is `isci`? Typo?

crowbar_framework/app/models/cinder_service.rb
```diff
@@ -166,11 +170,16 @@ def validate_proposal_after_save(proposal)
       end
     end
 
-    if raw_count > 0
+    volume_elements = proposal["deployment"][@bc_name]["elements"]["cinder-volume"]
+    volume_clusters = volume_elements.select { |n| is_cluster? n }
+
+    if (local_count > 0 || raw_count > 0) && volume_clusters.any?
+      validation_error I18n.t("barclamp.#{@bc_name}.validation.lvm_ha")
```
aspiers: I'd prefer something less cryptic here, e.g. `raw_local_backend_ha_conflict`, but it's a minor point.

aspiers:
Assuming the gate passes and it actually works (I haven't tested it) then other than my minor nitpick I'm +1 on this.

aspiers:
Tested with NFS backend and it works great! +1

aspiers:
https://build.opensuse.org/request/show/423359 is merged so I think this can be. Maybe @dirkmueller or @mjura could give the final +1?

mjura:
@aspiers : I read this change and looks great to me! Thanks for demo, wondering how does it work with Ceph backend.

+1

aspiers:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;QkpcjlzL&#x2F;16-p1-investigate-cinder-volume-active-passive-for-sap"> P1: Investigate cinder-volume active&#x2F;passive for SAP</a>

PR #462, opened 2016-07-22
Keystone bootstrap

sayalilunkad:


chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -443,6 +443,21 @@
   action :wakeup
 end
 
+# Creates admin user, admin role and admin project
+execute "keystone-manage bootstrap" do
+  command "keystone-manage bootstrap \
```
jgrassler: The `command` parameter can be a list (better than having a multiline string). See the Nova cookbook's [availability_zones](https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/nova/recipes/availability_zones.rb#L24) recipe and the [helper function it uses]( https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/nova/libraries/availability_zone.rb#L18) for an example list style command.

jgrassler:
+1 with the list style command parameter and passing gating (currently fails due to unrelated problems, the cloud got past the bootstrapping stage after all).

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -87,6 +87,28 @@ def self.keystone_settings(current_node, cookbook_name)
     @keystone_settings[cookbook_name]
   end
 
+  def self.keystone_bootstrap(node)
+    command = ["keystone-manage bootstrap"]
+    command << "--bootstrap-password"
```
rsalevsky: Can you please `command.push("--bootstrap-password")`? Also for the others...

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -443,6 +443,13 @@
   action :wakeup
 end
 
+# Creates admin user, admin role and admin project
+bootstrap = KeystoneHelper.keystone_bootstrap(node)
+execute "keystone-manage bootstrap" do
+  command bootstrap
```
rsalevsky: `command KeystoneHelper.keystone_bootstrap(node)` should also work and you can safe the extra variable

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -87,6 +87,21 @@ def self.keystone_settings(current_node, cookbook_name)
     @keystone_settings[cookbook_name]
   end
 
+  def self.keystone_bootstrap(node)
+    [
+      "/usr/bin/keystone-manage bootstrap",
+      "--bootstrap-password #{node[:keystone][:admin][:password]}",
+      "--bootstrap-username admin",
```
vuntz: I think the username can be specified, so should not be harcoded.

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -87,6 +87,21 @@ def self.keystone_settings(current_node, cookbook_name)
...
+      "--bootstrap-project-name admin",
+      "--bootstrap-role-name admin",
+      "--bootstrap-service-name keystone",
+      "--bootstrap-region-id #{node[:keystone][:api][:region]}",
+      "--bootstrap-admin-url #{node[:keystone][:api][:versioned_admin_URL]}",
+      "--bootstrap-public-url #{node[:keystone][:api][:versioned_public_URL]}",
+      "--bootstrap-internal-url #{node[:keystone][:api][:versioned_internal_URL]}"
+    ]
```
vuntz: I'm a bit confused as to why we're using a helper here, while it's something that is used in only one location.

But maybe people find it cleaner?

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -87,6 +87,21 @@ def self.keystone_settings(current_node, cookbook_name)
     @keystone_settings[cookbook_name]
   end
 
+  def self.keystone_bootstrap(node)
+    [
+      "/usr/bin/keystone-manage bootstrap",
+      "--bootstrap-password #{node[:keystone][:admin][:password]}",
+      "--bootstrap-username admin",
+      "--bootstrap-project-name admin",
```
vuntz: I'm also thinking that the project name can be changed in the attributes.

chef/cookbooks/keystone/libraries/helpers.rb
sayalilunkad: So initially I had just put the command in the recipe but @jgrassler  commented 
> The command parameter can be a list (better than having a multiline string). See the Nova cookbook's availability_zones recipe and the helper function it uses for an example list style command.

So we thought maybe it would be cleaner and to do it that way. But I would not mind keeping it just in the recipe to keep it simple.

aplanas:
I see a tempest fail in relation with authentication in the deletion of a user. Can you run tempest locally to makes sure that this is not happening too?

chef/cookbooks/keystone/libraries/helpers.rb
toabctl: I'm also in favor of moving it back. Things from the library should be used more than once imo.

rhafer: > I'm also in favor of moving it back.

Yeah, I agree.

rhafer:
Is it save to run the bootstrap command on every chef-client run? And even if it is, should we still make sure to only run it once similar to what we do for the various db_sync calls in all the other cookbooks? (e.g. like here: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/keystone/recipes/server.rb#L285)

If the admin password changes in the attributes, will the bootstrap update the password in the db as well? (IIRC we did that when still using the admin token)

sayalilunkad:
@aplanas It was related to the PR as it does not happen without the change. But the refactored code seems to be working.

sayalilunkad:
@rhafer 
Is it save to run the bootstrap command on every chef-client run? And even if it is, should we still make sure to only run it once similar to what we do for the various db_sync calls in all the other cookbooks? (e.g. like here: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/keystone/recipes/server.rb#L285)
> I agree it should not be run each chef run. updated the PR.

If the admin password changes in the attributes, will the bootstrap update the password in the db as well? (IIRC we did that when still using the admin token)
>good point. I will need to check that.

rhafer:
> > If the admin password changes in the attributes, will the bootstrap update the password in the db as well? (IIRC we did that when still using the admin token)

> good point. I will need to check that.

I just had a look at the code. The answer is no. If the user already exist during the bootstrap call it will not touch the existing user. I guess that means we need to document the correct procedure to update the admin password in the manual. I guess it would be something like this:

1. Change the admin password using the ```openstack``` cli
2. Update the ```keystone.admin.password``` attribute in the keystone proposal
3. reapply keystone

Would you mind creating a Documentation bug report for that?

sayalilunkad:
@rhafer ok, I can do that. Thanks for looking into that!

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -427,36 +426,69 @@
...
+  --bootstrap-username #{node[:keystone][:admin][:username]} \
+  --bootstrap-project-name #{node[:keystone][:admin][:tenant]} \
+  --bootstrap-role-name admin \
+  --bootstrap-service-name keystone \
+  --bootstrap-region-id #{node[:keystone][:api][:region]} \
+  --bootstrap-admin-url #{node[:keystone][:api][:versioned_admin_URL]} \
+  --bootstrap-public-url #{node[:keystone][:api][:versioned_public_URL]} \
+  --bootstrap-internal-url #{node[:keystone][:api][:versioned_internal_URL]}"
+  action :run
+  only_if { !node[:keystone][:bootstrap] && (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)) }
```
rhafer: Just turn this into:

```
only_if do
  !node[:keystone][:bootstrap] &&
    (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node))
end
```

To make hound happy

rhafer:
+1 after the remaining hound issue is fixed and ci is green.

aplanas:
+1

sayalilunkad:
@rhafer 
Would you mind creating a Documentation bug report for that?
>Done.  https://bugzilla.suse.com/show_bug.cgi?id=992076

PR #464, opened 2016-07-25
Add hitachi HUSVM config to cinder barclamp

Itxaka:
Adds config to the UI, locale, modifies volume_backend_name
to be added if the hitachi backend is in use, add hitachi
config to cinder.conf, update cinder schema, add migration


chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -248,6 +248,14 @@ def make_volume(node, backend_id, volume)
 
     when volume[:backend_driver] == "vmware"
 
+    when volume[:backend_driver] == "hitachi"
+      # create dir
```
Itxaka: Wondering if I should do this steps here as they seems to be part of the openstack config.

http://docs.openstack.org/mitaka/config-reference/block-storage/drivers/hitachi-storage-volume-driver.html#set-up-hitachi-storage-volume-driver

crowbar_framework/app/views/barclamp/cinder/_edit_attributes.html.haml
```diff
@@ -161,6 +161,38 @@
               = boolean_field %w(volumes {{@index}} vmware insecure)
 
           {{/if_eq}}
+          {{#if_eq backend_driver 'hitachi'}}
+          %li.list-group-item
+            %fieldset
+              %legend
+                = t('.volumes.hitachi_parameters')
+
```
Itxaka: Not sure if we want to expose everything?

crowbar_framework/config/locales/cinder/en.yml
```diff
@@ -127,6 +129,30 @@ en:
               file_name: 'Volume File Name'
               file_size: 'Maximum File Size (GB)'
               volume_name: 'Name of Volume'
+            hitachi:
+              use_iscsi: 'Use iscsi driver'
```
vuntz: s/iscsi/iSCSI/

crowbar_framework/app/views/barclamp/cinder/_edit_attributes.html.haml
vuntz: We usually expose the ones that are most useful, and leave the other ones hidden. Do you have any idea which ones are most useful? :-)

crowbar_framework/app/helpers/barclamp/cinder_helper.rb
```diff
@@ -28,7 +28,8 @@ def volume_driver_for_cinder(selected)
           [t(".volumes.rbd_volume_driver"), "rbd"],
           [t(".volumes.vmware_volume_driver"), "vmware"],
           [t(".volumes.local_volume_driver"), "local"],
-          [t(".volumes.manual_volume_driver"), "manual"]
+          [t(".volumes.manual_volume_driver"), "manual"],
+          [t(".volumes.hitachi_volume_driver"), "hitachi"]
```
vuntz: The list is alphabetically sorted (except for the first item and the last two items, on purpose). Can you keep it that way?

chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -653,7 +653,7 @@ iscsi_ip_address = <%= volume["eternus"]["iscsi_ip"] -%>
 
 # The backend name for a given driver implementation (string value)
 #volume_backend_name = <None>
-<% if @use_multi_backend -%>
+<% if @use_multi_backend || node['volume']['backend_driver'] == 'hitachi' -%>
```
vuntz: Why do you need this here?

chef/data_bags/crowbar/template-cinder.json
```diff
@@ -89,6 +89,31 @@
           "ca_file": "",
           "insecure": false
         },
+        "hitachi": {
+          "use_iscsi": false,
```
vuntz: Look at what we do for NetApp: we have a storage_protocol attribute; I think it's nicer.

chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -248,6 +264,25 @@ def make_volume(node, backend_id, volume)
 
     when volume[:backend_driver] == "vmware"
 
+  when volume[:backend_driver] == "hitachi"
+    # create dir
```
vuntz: No need for this comment :-)

chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -248,6 +264,25 @@ def make_volume(node, backend_id, volume)
 
     when volume[:backend_driver] == "vmware"
 
+  when volume[:backend_driver] == "hitachi"
```
vuntz: Indentation seems wrong; also, probably best to not add at the end, but try to keep the list of cases sorted (even though I'm awayre it's not the case right now...)

chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -248,6 +264,25 @@ def make_volume(node, backend_id, volume)
...
+      block do
+        `#{cinder_command} type-create husvm`
+      end
+    end
+
+    ruby_block "set cinder volume backend name" do
+      block do
+        `#{cinder_command} type-key HUSVM set volume_backend_name=#{volume["backend_name"]}`
+      end
+    end
```
vuntz: I don't think we need these two commands; or actually, if we want to do that, we would want them for all backends.

chef/cookbooks/cinder/recipes/volume.rb
Itxaka: Indentation is under the case as hound complains otherwise :(

Was thinking of moving all `when` to the "proper" indentation at the end of the PR so they nare all properly indented.

chef/cookbooks/cinder/templates/default/cinder.conf.erb
Itxaka: As per the docs[0] we need to set up the backend name when using the hitachi driver. If I understood correctly, this will only set the backend name if we use multi_backend.


[0] http://docs.openstack.org/mitaka/config-reference/block-storage/drivers/hitachi-storage-volume-driver.html#set-up-hitachi-storage-volume-driver

crowbar_framework/app/views/barclamp/cinder/_edit_attributes.html.haml
Itxaka: No clue ¯\_(ツ)_/¯

chef/cookbooks/cinder/recipes/volume.rb
Itxaka: done

chef/cookbooks/cinder/recipes/volume.rb
Itxaka: done

crowbar_framework/app/helpers/barclamp/cinder_helper.rb
Itxaka: done

chef/cookbooks/cinder/recipes/volume.rb
vuntz: Hrm, okay. Maybe worth another pull request just to fix that :-)

chef/cookbooks/cinder/templates/default/cinder.conf.erb
vuntz: I understand, but I really believe it's not needed. The doc were only mentioning this because of the type that they create (and that you removed from your earlier patch).

chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -2385,6 +2385,13 @@ volume_driver = cinder.volume.drivers.zfssaiscsi.ZFSSAISCSIDriver
 <% if volume['backend_driver'] == 'glusterfs' -%>
 volume_driver = cinder.volume.drivers.glusterfs.GlusterfsDriver
 <% end -%>
+<% if volume['backend_driver'] == 'hitachi' -%>
+  <% if volume['hitachi']['hitachi_storage_protocol'] == 'iscsi' -%>
+volume_driver = cinder.volume.drivers.hitachi.hbsd_iscsi.HBSDISCSIDriver
+  <% else -%>
```
vuntz: Should not be a blank `else` statement, I'd prefer to check if the value is `fc`

crowbar_framework/app/views/barclamp/cinder/_edit_attributes.html.haml
```diff
@@ -178,6 +213,8 @@
             %span.help-block
               = t(".volumes.manual.config_hint")
           {{/if_eq}}
+
+
```
vuntz: Why the new lines here? :-)

crowbar_framework/config/locales/cinder/en.yml
```diff
@@ -127,6 +129,30 @@ en:
...
+              hitachi_horcm_password: 'Password of storage system for HORCM'
+              hitachi_horcm_resource_lock_timeout: 'Timeout until a resource lock is released, in seconds. The value must be between 0 and 7200'
+              hitachi_horcm_user: 'Username of storage system for HORCM'
+              hitachi_ldev_range: 'Range of logical device of storage system'
+              hitachi_pool_id: 'Pool ID of storage system'
+              hitachi_serial_number: 'Serial number of storage system'
+              hitachi_target_ports: 'Control port names for HostGroup or iSCSI Target'
+              hitachi_thin_pool_id: 'Thin pool ID of storage system'
+              hitachi_unit_name: 'Name of an array unit'
+              hitachi_zoning_request: 'Request for FC Zone creating HostGroup'
```
vuntz: I guess some strings might need to be reworked, but I don't have a good suggestion for now -- would need to check the doc for each of them.

chef/cookbooks/cinder/templates/default/cinder.conf.erb
Itxaka: done

chef/cookbooks/cinder/recipes/volume.rb
Itxaka: nicolas fixed it on https://github.com/crowbar/crowbar-openstack/pull/469

chef/cookbooks/cinder/templates/default/cinder.conf.erb
Itxaka: done

crowbar_framework/app/views/barclamp/cinder/_edit_attributes.html.haml
Itxaka: woops

crowbar_framework/config/locales/cinder/en.yml
Itxaka: Yeah, as I have no idea about them I just copied the descriptions from the openstack config page:/

toabctl:
According to http://docs.openstack.org/liberty/config-reference/content/hitachi-reqs.html python-pecpect needs to be installed on the nodes where the cinder-volume service is running. This is not forced via a package dependency so you need to add it to the recipe.

Itxaka:
good catch @toabctl !

chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -2385,6 +2385,13 @@ volume_driver = cinder.volume.drivers.zfssaiscsi.ZFSSAISCSIDriver
 <% if volume['backend_driver'] == 'glusterfs' -%>
 volume_driver = cinder.volume.drivers.glusterfs.GlusterfsDriver
 <% end -%>
+<% if volume['backend_driver'] == 'hitachi' -%>
+  <% if volume['hitachi']['hitachi_storage_protocol'] == 'iscsi' -%>
+volume_driver = cinder.volume.drivers.hitachi.hbsd_iscsi.HBSDISCSIDriver
+  <% elsif volume['hitachi']['hitachi_storage_protocol'] == 'fc'  -%>
```
dirkmueller: extra whitespace here between 'fc' and  '-%>'

chef/data_bags/crowbar/migrate/cinder/100_add_hitachi_options.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["volume_defaults"]["hitachi"] = ta["volume_defaults"]["hitachi"]
```
dirkmueller: need to check if a["volume_defaults"]["hitachi"] doesn't exist already.

dirkmueller:
pexpect needs to be added then..

chef/cookbooks/cinder/templates/default/cinder.conf.erb
Itxaka: done

chef/data_bags/crowbar/migrate/cinder/100_add_hitachi_options.rb
Itxaka: done

Itxaka:
added pexpect package

aplanas:
LGTM +1

chef/cookbooks/cinder/recipes/volume.rb
vuntz: @Itxaka I'm confused as @nicolasbock's PR got merged, but we still have some inconsistency here. Can we have a followup PR to fix it?

vuntz:
I'd be in favor of merging this now, and iterate on the UI once we know better what to do.

But this needs a rebase first...

Itxaka:
rebased + migration number fixed

chef/data_bags/crowbar/migrate/cinder/100_add_hitachi_options.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  unless a["volume_defaults"].key?("hitachi")
+    a["volume_defaults"]["hitachi"] = ta["volume_defaults"]["hitachi"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  if a["volume_defaults"].key?("hitachi")
```
nicolasbock: @Itxaka I guess I still don't fully get it :) If this barclamp gets installed in cloud 7 it will have the `hitachi` key because of the attribute template. If it is upgrade from 6 and then downgraded again, it will have the key because the upgrade process added it. Am I forgetting something?

nicolasbock:
+1

chef/data_bags/crowbar/migrate/cinder/100_add_hitachi_options.rb
vuntz: FWIW, it's fine to call `delete` for a key that is missing; it won't crash. So I guess @nicolasbock's point is that we can just simply remove the test.

chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -238,6 +238,8 @@ def make_volume(node, backend_id, volume)
 
   when volume[:backend_driver] == "vmware"
 
+  when volume[:backend_driver] == "hitachi"
+    package "python-pexpect"
   end
```
vuntz: In case you push again to the PR (and only if you do this): can you add an empty line before the `end` (it was there before, and it's highlighting the fact that it goes with the `case`, not with the `when`)

chef/data_bags/crowbar/migrate/cinder/100_add_hitachi_options.rb
vuntz: Ah, the issue is that the test should be on `ta`, not `a`.

nicolasbock: @vuntz Yes, that's a good point. The upgrade/downgrade pattern would be
~~~
def upgrade(ta, td, a, d)
  a["new_key"] = ta["new_key"] unless a.key?("new_key")
end

def downgrade(ta, td, a, d)
  a.delete("new_key") unless ta.key?("new_key")
end
~~~

Itxaka: ahh, very good point, done!

chef/cookbooks/cinder/recipes/volume.rb
Itxaka: done

vuntz: Unfortunately, you put it before the second `end`, I was asking before the first one ;-) But no big deal.

vuntz:
@Itxaka I don't see the schema revision bump anymore. Was it intentionally left out?

Itxaka:
@vuntz Yes, its left out. Actually its set to 101, which is what its on master as https://github.com/crowbar/crowbar-openstack/pull/469 already bumped it :)

PR #465, opened 2016-07-26
Move openssl package check into ssl provider

nicolasbock:


chef/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb
```diff
@@ -19,10 +19,13 @@ def action_setup
         if @current_resource.generate_certs
           unless ::File.exist?(@current_resource.certfile) \
             && ::File.exist?(@current_resource.keyfile)
+
             require "fileutils"
 
             Chef::Log.info("Generating SSL certificate...")
 
+            package "openssl"
```
vuntz: You probabl need something like:
```ruby
package "openssl" do
  action :nothing
end.run_action(:install)
```

(or something similar)

nicolasbock: I tested this (and since `openssl` was already installed on the nodes, I added a line `package 'emacs'` just to be on the safe side :smile:) and it works as in the PR. I get log output stating that:
~~~
[2016-07-26T11:16:13+00:00] INFO: Processing package[openssl] action install (/var/chef/cache/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb line 27)
[2016-07-26T11:16:13+00:00] INFO: Processing package[emacs] action install (/var/chef/cache/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb line 28)
~~~
I also found the `emacs` package to be installed after the proposal was done.

Itxaka:
lgtm +1

abel-navarro:
+1

chef/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb
vuntz: Interesting; thanks for the test. I'm a bit surprised, but that's fine :-)

PR #466, opened 2016-07-26
nova: Allow defining on which network live migration happens

vuntz:
Right now, live migration is being done only on the admin network.
However, the admin network might be on a 1G interface, while some other
network might be defined with 10G interface. It'd be better to use such
a network for live migration, to speed up things.

We add a new attribute to specify which network to use.

It's important to keep in mind that the public network is not
appropriate here, as it would make the compute nodes publicly reachable,
and libvirt listen on the public IP. So it's better to define a new
private network, or use an existing one like the storage network.

### DIFF
--- a/chef/data_bags/crowbar/template-nova.json
+++ b/chef/data_bags/crowbar/template-nova.json
@@ -20,6 +20,9 @@
       "setup_shared_instance_storage": false,
       "use_shared_instance_storage": false,
       "use_migration": false,
+      "migration": {
+        "network": "admin"
+      },
       "use_syslog": false,
       "neutron_url_timeout": 30,
       "force_config_drive": false,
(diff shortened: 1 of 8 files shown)


chef/data_bags/crowbar/template-nova.json
```diff
@@ -20,6 +20,9 @@
       "setup_shared_instance_storage": false,
       "use_shared_instance_storage": false,
       "use_migration": false,
+      "migration": {
+        "network": "admin"
+      }
```
bhdn: Isn't there a missing comma here?

vuntz: Yes, fixed. Thanks!

aplanas:
+1

nicolasbock:
+1

PR #468, opened 2016-07-27
Adds cert_manager option to magnum

sayalilunkad:
* Adds dropdown to select cert_manager_type in magnum.
* Makes barbican the default cert_manager_type if barbican
  is present else defaults to local.

### DIFF
--- a/chef/cookbooks/magnum/templates/default/magnum.conf.erb
+++ b/chef/cookbooks/magnum/templates/default/magnum.conf.erb
@@ -76,7 +76,7 @@ verbose=<%= node[:magnum][:verbose] %>
 # is ignored if log_config_append is set. (string value)
 # Deprecated group/name - [DEFAULT]/logdir
 #log_dir = <None>
-log_dir=/var/log/magnum 
+log_dir=/var/log/magnum
 
 # Uses logging handler designed to watch file system. When log file is moved or
 # removed this handler will open a new log file with specified path
@@ -386,7 +386,7 @@ region_name = <%= @keystone_settings['endpoint_region'] %>
 
 # Certificate Manager plugin. Defaults to barbican. (string value)
 #cert_manager_type = barbican
-cert_manager_type = local
+cert_manager_type = <%= "#{node[:magnum][:cert][:cert_manager_type]}" %>
 
 # Absolute path of the certificate storage directory. Defaults to
 # /var/lib/magnum/certificates/. (string value)
--- a/chef/data_bags/crowbar/migrate/magnum/100_add_cert_manager_attribute.rb
+++ b/chef/data_bags/crowbar/migrate/magnum/100_add_cert_manager_attribute.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["cert"]["cert_manager_type"] = "local"
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["cert"]["cert_manager_type"].delete("local")
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-magnum.json
+++ b/chef/data_bags/crowbar/template-magnum.json
@@ -28,6 +28,9 @@
         "password": "",
         "user": "magnum",
         "database": "magnum"
+      },
+      "cert": {
+        "cert_manager_type": "local"
       }
     }
   },
@@ -35,6 +38,7 @@
     "magnum": {
       "crowbar-revision": 1,
       "crowbar-applied": false,
+      "schema-revision": 100,
       "element_states": {
         "magnum-server": [ "readying", "ready", "applying" ]
       },
--- a/crowbar_framework/app/helpers/barclamp/magnum_helper.rb
+++ b/crowbar_framework/app/helpers/barclamp/magnum_helper.rb
@@ -1,10 +1,12 @@
-# Copyright 2015, SUSE, Inc.
+#
+# Copyright 2011-2013, Dell
+# Copyright 2013-2014, SUSE LINUX Products GmbH
 #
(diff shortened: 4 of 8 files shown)


chef/cookbooks/magnum/attributes/default.rb
```diff
@@ -23,3 +23,9 @@
 default[:magnum][:api][:protocol] = "http"
 
 default[:magnum][:ha][:enabled] = false
+
+if roles.include?("barbican-server")
+  default[:magnum][:cert][:cert_manager_type] = "barbican"
+else
+  default[:magnum][:cert][:cert_manager_type] = "local"
+end
```
vuntz: I think in this file, you should just default to `local`. The logic about auto-detection should be only in the rails app.

chef/cookbooks/magnum/templates/default/magnum.conf.erb
```diff
@@ -386,7 +386,7 @@ region_name = <%= @keystone_settings['endpoint_region'] %>
 
 # Certificate Manager plugin. Defaults to barbican. (string value)
 #cert_manager_type = barbican
-cert_manager_type = local
+cert_manager_type=<%= node[:magnum][:cert][:cert_manager_type] %>
```
vuntz: Can you put spaces around `=`?

vuntz: Also, it's slightly better to pass `node[:magnum][:cert][:cert_manager_type]` as a variable to the template, instead of referencing it directly here.

chef/data_bags/crowbar/template-magnum.json
```diff
@@ -28,6 +28,9 @@
         "password": "",
         "user": "magnum",
         "database": "magnum"
+      },
+      "cert": {
+        "cert_manager_type": ""
```
vuntz: You should put a valid default value here, `local`

vuntz: Also you need a schema migration :-)

crowbar_framework/app/helpers/barclamp/barbican_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+#
+
+module Barclamp
+  module BarbicanHelper
+    def cert_manager_types(selected)
+      options_for_select(
+        [
+          ["local", "local"],
+          ["barbican", "barbican"]
```
vuntz: One of the items should be a localized string, which explains what the values are (so we don't want "local" to appear, we want a string that says what it means).

chef/cookbooks/magnum/attributes/default.rb
vuntz: In case you wonder where in the rails app it should be: it should be in the `create_proposal` method of magnum_service.rb.

rsalevsky:
+1

chef/data_bags/crowbar/migrate/magnum/001_add_cert_manager_attribute.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["cert_manager_type"] = "local"
```
toabctl: isn't it ```a["cert"]["cert_manager_type"]``` ?

chef/data_bags/crowbar/migrate/magnum/001_add_cert_manager_attribute.rb
```diff
@@ -0,0 +1,9 @@
...
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["cert_manager_type"].delete("local")
```
toabctl: same here

crowbar_framework/app/helpers/barclamp/magnum_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+module Barclamp
+  module MagnumHelper
+    def cert_manager_types(selected)
+      options_for_select(
+        [
+          ["Local driver", "local"],
```
toabctl: Imo just "Local" would be better

toabctl:
@sayalilunkad you need to bump the schema-revision in the template-magnum.json file.

chef/data_bags/crowbar/template-magnum.json
```diff
@@ -28,13 +28,17 @@
...
+      "cert": {
+        "cert_manager_type": "local"
       }
     }
   },
   "deployment": {
     "magnum": {
       "crowbar-revision": 1,
       "crowbar-applied": false,
+      "schema-revision": 1,
```
toabctl: we decided to start with 100 for Cloud 7. I also wonder if we need a migration to add the schema-revision key? Not sure...

toabctl: Also you need to add the ```schema-revision``` key to template-magnum.schema

toabctl:
+1

chef/data_bags/crowbar/migrate/magnum/100_add_cert_manager_attribute.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["cert"]["cert_manager_type"] = "local"
```
nkrinner: I think you have to create a["cert"] first before accessing a["cert"]["..."]

sayalilunkad: That is already defined in the attributes if I understand correctly.

nkrinner:
+1

PR #469, opened 2016-07-27
Add support for NFS backend

nicolasbock:


nicolasbock:
At this point a cherry-pick from @vuntz: https://github.com/crowbar/barclamp-cinder/compare/release/tex/master...vuntz:nfs-backend

Itxaka:
This will collide with https://github.com/crowbar/crowbar-openstack/pull/464, at least on the migrations part.

nicolasbock:
@Itxaka Thanks for pointing this out. I could go with revision 101, which would resolve the potential collision. I didn't see any other obvious places of collision potential, did you?

Itxaka:
@nicolasbock I also didnt see anything that could not be merged cleanly so we should be ok in that aspect

Itxaka:
tested locally and it works as exected

toabctl:
+1

Itxaka:
+1

aplanas:
@nicolasbock is this still wip? do this needs more reivew?

nicolasbock:
@aplanas I am currently testing it, but soon it should be good to go.

Itxaka:
damn, I cant believe I missed that. Thats why you test it with real stuff :fire: 

Still lgtm +1

Itxaka:
tested it thanks to @nicolasbock and seems to work as expected, creating volumes work and I can see them on the nfs server dir.
+1

dirkmueller:
+1

chef/data_bags/crowbar/migrate/cinder/101_nfs_backend.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["volume_defaults"]["nfs"] = ta["volume_defaults"]["nfs"]
```
dirkmueller: I think we need to also check here if a["volume_defaults"]["nfs"] doesn't already exist (think about this being backported to cloud 6)

chef/data_bags/crowbar/migrate/cinder/101_nfs_backend.rb
```diff
@@ -0,0 +1,9 @@
...
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["volume_defaults"].delete("nfs")
```
dirkmueller: same here

nicolasbock:
Rebased to trigger gating.

PR #470, opened 2016-07-27
Improves barbican barclamp description

sayalilunkad:


barbican.yml
```diff
@@ -27,7 +27,7 @@
 barclamp:
   name: 'barbican'
   display: 'Barbican'
-  description: 'OpenStack key and management service'
+  description: 'OpenStack Key Manager: Secure storage, provisioning and management of secret data'
```
vuntz: I think I would do s/Manager/Management/ if it's still correct.

sayalilunkad: I got the OpenStack key manager from http://docs.openstack.org/developer/barbican/ so maybe it's better to keep it the same?

jgrassler:
+1 for "Key Manager". That's also the term we use in the service catalog.

barbican.yml
vuntz: Thanks, agreed.

vuntz:
+1

PR #471, opened 2016-07-28
Port SUSE OpenStack Cloud to aarch64

dirkmueller:
Various needed features for an SUSE OpenStack Cloud demo on AArch64


chef/data_bags/crowbar/template-nova.json
```diff
@@ -14,6 +14,7 @@
       "trusted_flavors": false,
       "libvirt_type": "kvm",
       "use_novnc": true,
+      "use_serial": false,
```
vuntz: Any reason to not always enable it? Or to not default to `true`?

vuntz: (btw, you'll need a schema migration)

chef/cookbooks/nova/recipes/config.rb
```diff
@@ -302,11 +317,13 @@
             glance_server_insecure: glance_server_insecure || keystone_settings["insecure"],
             metadata_bind_address: metadata_bind_address,
             vnc_enabled: node[:kernel][:machine] != "aarch64",
+            serial_enabled: node[:kernel][:machine] == "aarch64",
```
vuntz: I guess you need to be consistent for the condition, and use `node[:nova][:use_serial]` here.

dguitarbite:
There are some security concerns from my side. Could we configure a VNC password and have it as a configuration option for the same?

chef/data_bags/crowbar/template-nova.json
dirkmueller: yes, its wip for a reason ;)

dirkmueller:
@dguitarbite you mean the iptables blocking thing that you implemented for vnc? otherwise I don't understand what you mean by security concerns. the serial console proxy is behind keystone..

dguitarbite:
@dirkmueller Yes, the code I wrote using iptables for restricting serial console (not just the proxy) ports from the Compute Nodes (libvirt hosts) access. Either we set the password in the ``/etc/libvirt/qemu.conf`` and then make the serial console inject this password to connect or use something like my iptables code (strongly suggest not to use my code ...).

chef/cookbooks/nova/recipes/config.rb
```diff
@@ -285,6 +299,7 @@
             bind_port_objectstore: bind_port_objectstore,
             bind_port_novncproxy: bind_port_novncproxy,
             bind_port_xvpvncproxy: bind_port_xvpvncproxy,
+            bind_port_serialproxy: bind_port_xvpvncproxy,
```
dguitarbite: Why is the variable set to ``bind_port_xvpvncproxy`` here where as earlier it was set to another value. I may be missing something here, just skimming through the code.

chef/cookbooks/nova/recipes/config.rb
dguitarbite: I recommend you to change either the commit message or make a short comment explaining why ``non-x86`` is being used in the commit message where as here the conditional is just checking ``aarch64``. 

``aarch64`` ⊂ ``!x86`` is not true the other way around.

chef/data_bags/crowbar/template-nova.json
dguitarbite: @dirkmueller , I could partially confirm that NoVNC is a dumb server limited by the dumbness of [websockify](https://github.com/kanaka/websockify). Keeping this always on should not be necessary, opening more ports additionally opens up more potential security issues.

@vuntz Please create an additional (noise?) followup card if required for addressing the security concern's w.r.t. enabling ``serial console``.

dguitarbite:
@dirkmueller Small nitpick, is there a possibility to change the 'non-x86' part to more appropriate one in the commit message?

toabctl:
@dirkmueller commit msg needs ```$component: ...``` prefix

dirkmueller:
@toabctl done

chef/cookbooks/nova/recipes/config.rb
```diff
@@ -320,12 +336,14 @@
...
             metadata_bind_address: metadata_bind_address,
-            vnc_enabled: node[:kernel][:machine] != "aarch64",
+            vnc_enabled: node[:nova][:use_novnc],
+            serial_enabled: node[:nova][:use_serial],
             vendordata_jsonfile: vendordata_jsonfile,
             vncproxy_public_host: vncproxy_public_host,
             vncproxy_ssl_enabled: api[:nova][:novnc][:ssl][:enabled],
             vncproxy_cert_file: api_novnc_ssl_certfile,
             vncproxy_key_file: api_novnc_ssl_keyfile,
+            serialproxy_public_host: serialproxy_public_host,
```
toabctl: add a var ```serialproxy_ssl_enabled: node[:nova][:serial][:ssl][:enabled]``` and use it later like we do with ```vncproxy_ssl_enabled``` ?

toabctl: Also what about ```serialproxy_cert_file``` and ```serialproxy_key_file``` ? you have theses vars in the .schema file

dirkmueller: I have removed  ```serialproxy_cert_file``` and ```serialproxy_key_file```

dirkmueller:
@dguitarbite so I looked a bit into this. it seems there is no way to secure serial over ssl or shared secret as far as I can see.

dguitarbite:
@dirkmueller I guess that is the reason for isolating this traffic over a
protected dedicated private network. Solved over architectural design.

+1

dguitarbite:
Hope that is considered or remembered in case of potential request for the
same in future.

PR #472, opened 2016-07-28
Fixes barbican barclamp description

sayalilunkad:
As pointed out by Andreas Jaeger after todays review
meeting that the description for barbican should not be
prefixed with OpenStack as the offical way is Key Manager Service.

### DIFF
--- a/barbican.yml
+++ b/barbican.yml
@@ -27,7 +27,7 @@
 barclamp:
   name: 'barbican'
   display: 'Barbican'
-  description: 'OpenStack Key Manager: Secure storage, provisioning and management of secret data'
+  description: 'Key Manager Service: Secure storage, provisioning and management of secret data'
   version: 0
   requires:
     - '@crowbar'


jgrassler:
+1

vuntz:
Hrm, I thought it was fine with the big tent... I'll defer to @toabctl as he always knows this stuff :-)

toabctl:
> Hrm, I thought it was fine with the big tent... I'll defer to @toabctl as he always knows this stuff :-)

I don't know that :-) But looking at our other descriptions i.e. at http://ananas.suse.de:27771/crowbar/openstack/1.0 , all our OpenStack barclamps start with ```OpenStack ...```. So I would a) not merge this or b) change all other descriptions, too. Imo we should leave it as it is.

sayalilunkad:
@toabctl I guess it makes sense to leave it like it is for now then. So I will close this PR.

PR #473, opened 2016-07-28
Nova/neutron: Enable Daemon mode

dirkmueller:


chef/cookbooks/nova/recipes/config.rb
```diff
@@ -67,6 +67,9 @@
   api = node
 end
 
+# use nova-rootwrap daemon on compute nodes
+use_rootwrap_daemon = !node["roles"].include?("nova-controller")
```
toabctl: Comment not valid for the code in case controller and compute are on the same node. Maybe something like ```!node["roles"].grep(/^nova-compute-.*/).empty?``` works here?

dirkmueller: I changed the comment instead

nicolasbock:
I don't get how this works: You remove the sudoers file and then add a `sudo ...` command. What am I missing?

aplanas:
@nicolasbock I think that the logic is this:

* Remove the wrapper scripts. Those scripts are used by OpenStack services to execute some commands as a different user (root)

^ Use a new feature in OpenStack based in a wrapper service, that take the command and execute it as root.

nicolasbock:
@aplanas After some digging I realized that the sudoers file is in the package [package](https://build.suse.de/package/show/Devel:Cloud:7/openstack-neutron) itself. Now I get it :smile:.

nicolasbock:
+1

toabctl:
+1

PR #474, opened 2016-07-30
Stop overwriting auth_url with v2 endpoint

dirkmueller:
glance-scrubber can just use normal keystone client methods for
determining access, it doesn't have to fall back to a specific /v2.0
endpoint. this avoids a lot of pointless queries on the /v2.0 endpoint
(which is in my deployment the most often triggered keystone query)


dirkmueller:
hmm, maybe wrong.

PR #476, opened 2016-08-01
Check addresses only when they exist

jsuchome:
It looks like some network interfaces might miss addresses key on SP2.


abel-navarro:
+1

Itxaka:
+1

jsuchome:
BTW, https://github.com/crowbar/crowbar-openstack/pull/477 might be a better solution...

PR #477, opened 2016-08-01
Look into crowbar wall for network interfaces.

jsuchome:
We are only interested in interfaces managed by crowbar

An alternative to https://github.com/crowbar/crowbar-openstack/pull/476

### DIFF
--- a/chef/cookbooks/swift/libraries/rack_awareness.rb
+++ b/chef/cookbooks/swift/libraries/rack_awareness.rb
@@ -37,9 +37,8 @@ class Rack_awareness
             storage_ip = Swift::Evaluator.get_ip_by_type(n, :storage_ip_expr)
             #get storage iface
             iface=""
-            n[:network][:interfaces].keys.each do |ifn|
-              addresses = n[:network][:interfaces][ifn.to_s][:addresses]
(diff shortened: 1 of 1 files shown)


rhafer:
+1

Though I think the whole IP address handling in the swift cookbook would benefit from some cleanup. To me it seems as if ```Swift::Evaluator.get_ip_by_type``` is not needed any longer and could be replaced with ```Chef::Recipe::Barclamp::Inventory.get_network_by_type``` . That would also make the interface lookup a lot easier as well. However this would require changes in a few other places. I'll add a card to the back log for it.

abel-navarro:
@jsuchome since #476 was already merged, can we close this one?

jsuchome:
Yes... unless this is the better solution. What do you think? And @rhafer, do you remember?

rhafer:
@jsuchome I think this is the better solution. Because ```crowbar_wall``` only contains the interfaces that are managed by crowbar. And AFAIK the interfaces there always have the ```addresses``` key.

jsuchome:
@abel-navarro can you approve this so we can merge it?

abel-navarro:
+1

PR #478, opened 2016-08-02
Fixes bind_port in barbican

sayalilunkad:


jsuchome:
Yes, this seems to be the way we are using in horizon and apache cookbooks.

+1

aspiers:
+1

PR #480, opened 2016-08-02
Add Sahara barclamp

Itxaka:
This is basically https://github.com/crowbar/crowbar-openstack/pull/365 fixed and improved upon.

### DIFF
--- a/chef/cookbooks/sahara/attributes/default.rb
+++ b/chef/cookbooks/sahara/attributes/default.rb
@@ -1,0 +1,41 @@
+# Copyright 2016, SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+default[:sahara][:debug] = false
+default[:sahara][:verbose] = false
+override[:sahara][:user] = "sahara"
+override[:sahara][:group] = "sahara"
+default[:sahara][:max_header_line] = 16384
+default[:sahara][:api][:protocol] = "http"
+default[:sahara][:api][:bind_port] = 8386
+default[:sahara][:use_syslog] = false
+
+default[:sahara][:ha][:enabled] = false
+# Ports to bind to when haproxy is used for the real ports
+default[:sahara][:ha][:ports][:api_port] = 5573
+
+default[:sahara][:ha][:api][:ra] = if ["rhel", "suse"].include? node[:platform_family]
+  "lsb:openstack-sahara-api"
+else
+  "lsb:sahara-api"
+end
+
+default[:sahara][:ha][:engine][:ra] = if ["rhel", "suse"].include? node[:platform_family]
+  "lsb:openstack-sahara-engine"
+else
+  "lsb:sahara-engine"
+end
+
+default[:sahara][:ha][:op][:monitor][:interval] = "10s"
--- a/chef/cookbooks/sahara/metadata.rb
+++ b/chef/cookbooks/sahara/metadata.rb
@@ -1,0 +1,28 @@
+# Copyright 2016, SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name "sahara"
+maintainer "Crowbar project"
+maintainer_email "crowbar@googlegroups.com"
+license "Apache 2.0"
+description "Installs/Configures sahara"
+long_description IO.read(File.join(File.dirname(__FILE__), "README.md"))
+version "0.1"
+
+depends "database"
+depends "keystone"
+depends "crowbar-openstack"
+depends "crowbar-pacemaker"
+depends "utils"
--- a/chef/cookbooks/sahara/recipes/api.rb
+++ b/chef/cookbooks/sahara/recipes/api.rb
@@ -1,0 +1,98 @@
+# Copyright 2016 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Cookbook Name:: sahara
+# Recipe:: api
+#
+
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+
+sahara_port = node[:sahara][:api][:bind_port]
+sahara_protocol = node[:sahara][:api][:protocol]
+
+ha_enabled = node[:sahara][:ha][:enabled]
+
+my_admin_host = CrowbarHelper.get_host_for_admin_url(node, ha_enabled)
+my_public_host = CrowbarHelper.get_host_for_public_url(
+  node, node[:sahara][:api][:protocol] == "https", ha_enabled
+)
+
+register_auth_hash = { user: keystone_settings["admin_user"],
+                       password: keystone_settings["admin_password"],
+                       tenant: keystone_settings["admin_tenant"] }
+
+crowbar_pacemaker_sync_mark "wait-sahara_register"
+
+keystone_register "sahara api wakeup keystone" do
+  protocol keystone_settings["protocol"]
+  insecure keystone_settings["insecure"]
+  host keystone_settings["internal_url_host"]
+  port keystone_settings["admin_port"]
+  auth register_auth_hash
+  action :wakeup
+end
+
+keystone_register "register sahara user" do
+  protocol keystone_settings["protocol"]
+  insecure keystone_settings["insecure"]
+  host keystone_settings["internal_url_host"]
+  port keystone_settings["admin_port"]
+  auth register_auth_hash
+  user_name keystone_settings["service_user"]
+  user_password keystone_settings["service_password"]
+  tenant_name keystone_settings["service_tenant"]
+  action :add_user
+end
+
+keystone_register "give sahara user access" do
+  protocol keystone_settings["protocol"]
+  insecure keystone_settings["insecure"]
+  host keystone_settings["internal_url_host"]
+  port keystone_settings["admin_port"]
+  auth register_auth_hash
+  user_name keystone_settings["service_user"]
+  tenant_name keystone_settings["service_tenant"]
+  role_name "admin"
+  action :add_access
+end
+
+keystone_register "register sahara service" do
+  protocol keystone_settings["protocol"]
+  insecure keystone_settings["insecure"]
+  host keystone_settings["internal_url_host"]
+  port keystone_settings["admin_port"]
+  auth register_auth_hash
+  service_name "sahara"
+  service_type "data-processing"
+  service_description "Openstack Sahara - Data Processing"
+  action :add_service
+end
+
+keystone_register "register sahara endpoint" do
+  protocol keystone_settings["protocol"]
+  insecure keystone_settings["insecure"]
+  host keystone_settings["internal_url_host"]
+  port keystone_settings["admin_port"]
+  auth register_auth_hash
+  endpoint_service "sahara"
+  endpoint_region keystone_settings["endpoint_region"]
+  endpoint_publicURL "#{sahara_protocol}://#{my_public_host}:#{sahara_port}/v1.1/%(tenant_id)s"
+  endpoint_adminURL "#{sahara_protocol}://#{my_admin_host}:#{sahara_port}/v1.1/%(tenant_id)s"
+  endpoint_internalURL "#{sahara_protocol}://#{my_admin_host}:#{sahara_port}/v1.1/%(tenant_id)s"
+  action :add_endpoint_template
+end
+
+crowbar_pacemaker_sync_mark "create-sahara_register"
+
+sahara_service "api"
--- a/chef/cookbooks/sahara/recipes/common.rb
+++ b/chef/cookbooks/sahara/recipes/common.rb
@@ -1,0 +1,77 @@
+# Copyright 2016 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+package "openstack-sahara"
+
+network_settings = SaharaHelper.network_settings(node)
+db_settings = fetch_database_settings
+
+include_recipe "database::client"
+include_recipe "#{db_settings[:backend_name]}::client"
+include_recipe "#{db_settings[:backend_name]}::python-client"
+
+# get Database data
+db_password = node[:sahara][:db][:password]
+sql_connection = "#{db_settings[:url_scheme]}://#{node[:sahara][:db][:user]}:"\
+                 "#{db_password}@#{db_settings[:address]}/"\
+                 "#{node[:sahara][:db][:database]}"
+
+# cinder insecure?
+cinder = get_instance("roles:cinder-controller")
+cinder_insecure = cinder[:cinder][:ssl][:insecure]
+
+# heat insecure?
+heat = get_instance("roles:heat-server")
+heat_insecure = heat[:heat][:ssl][:insecure]
+
+# neutron insecure?
+neutron = get_instance("roles:neutron-server")
+neutron_insecure = neutron[:neutron][:ssl][:insecure]
+
+# nova insecure?
+nova = get_instance("roles:nova-controller")
+nova_insecure = nova[:nova][:ssl][:insecure]
+
+# use ceilometer?
+ceilometers = search_env_filtered(:node, "roles:ceilometer-server")
+use_ceilometer = !ceilometers.empty?
+
+template "/etc/sudoers.d/openstack-sahara" do
+  source "sahara-rootwrap.erb"
+  owner "root"
+  group "root"
+  mode "0440"
+end
+
+template "/etc/sahara/sahara.conf" do
+  source "sahara.conf.erb"
+  owner "root"
+  group node[:sahara][:group]
+  mode "0640"
+  variables(
+    bind_host: network_settings[:api][:bind_host],
+    bind_port: network_settings[:api][:bind_port],
+    sql_connection: sql_connection,
+    rabbit_settings: fetch_rabbitmq_settings,
+    keystone_settings: KeystoneHelper.keystone_settings(node, :sahara),
+    cinder_insecure: cinder_insecure,
+    heat_insecure: heat_insecure,
+    neutron_insecure: neutron_insecure,
+    nova_insecure: nova_insecure,
+    use_ceilometer: use_ceilometer
+  )
+end
+
+node.save
--- a/chef/cookbooks/sahara/recipes/ha.rb
+++ b/chef/cookbooks/sahara/recipes/ha.rb
@@ -1,0 +1,82 @@
+# Copyright 2016, SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+unless node[:sahara][:ha][:enabled]
+  Chef::Log.info("HA support for sahara is disabled")
+  return
+end
+
+network_settings = SaharaHelper.network_settings(node)
+
+haproxy_loadbalancer "sahara-api" do
+  address network_settings[:api][:ha_bind_host]
+  port network_settings[:api][:ha_bind_port]
+  use_ssl node[:sahara][:api][:protocol] == "https"
+  servers CrowbarPacemakerHelper.haproxy_servers_for_service(node, "sahara",
+                                                             "sahara-server", "api_port")
+  action :nothing
+end.run_action(:create)
+
+# Wait for all nodes to reach this point so we know that all nodes will have
+# all the required packages installed before we create the pacemaker
+# resources
+crowbar_pacemaker_sync_mark "sync-sahara before_ha"
+
+# Avoid races when creating pacemaker resources
+crowbar_pacemaker_sync_mark "wait-sahara_ha_resources"
+
+transaction_objects = []
+primitives = []
+
+["api", "engine"].each do |service|
+  primitive = "sahara-#{service}"
+  pacemaker_primitive primitive do
+    agent node[:sahara][:ha][service.to_sym][:ra]
+    op node[:sahara][:ha][:op]
+    action :update
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+  primitives << primitive
+  transaction_objects << "pacemaker_primitive[#{primitive}]"
+end
+
+group_name = "g-sahara"
+pacemaker_group group_name do
+  members primitives
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects << "pacemaker_group[#{group_name}]"
+
+clone_name = "cl-#{group_name}"
+pacemaker_clone clone_name do
+  rsc group_name
+  meta "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node)
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects << "pacemaker_clone[#{clone_name}]"
+
+location_name = openstack_pacemaker_controller_only_location_for clone_name
+transaction_objects << "pacemaker_location[#{location_name}]"
+
+pacemaker_transaction "sahara server" do
+  cib_objects transaction_objects
+  # note that this will also automatically start the resources
+  action :commit_new
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+crowbar_pacemaker_sync_mark "create-sahara_ha_resources"
--- a/chef/cookbooks/sahara/templates/default/sahara-rootwrap.erb
+++ b/chef/cookbooks/sahara/templates/default/sahara-rootwrap.erb
@@ -1,0 +1,1 @@
+<%= @node[:sahara][:user] %> ALL = (root) NOPASSWD: /usr/bin/sahara-rootwrap /etc/sahara/rootwrap.conf *
--- a/chef/cookbooks/sahara/templates/default/sahara.conf.erb
+++ b/chef/cookbooks/sahara/templates/default/sahara.conf.erb
@@ -1,0 +1,1463 @@
+[DEFAULT]
+
+#
+# From oslo.log
+#
+
+# If set to true, the logging level will be set to DEBUG instead of the default
+# INFO level. (boolean value)
+# Note: This option can be changed without restarting.
+#debug = false
+debug = <%= node[:sahara][:debug] %>
+
+# DEPRECATED: If set to false, the logging level will be set to WARNING instead
+# of the default INFO level. (boolean value)
+# This option is deprecated for removal.
+# Its value may be silently ignored in the future.
+#verbose = true
+verbose = <%= node[:sahara][:verbose] %>
+
+# The name of a logging configuration file. This file is appended to any
+# existing logging configuration files. For details about logging configuration
+# files, see the Python logging module documentation. Note that when logging
+# configuration files are used then all logging configuration is set in the
+# configuration file and other logging configuration options are ignored (for
+# example, logging_context_format_string). (string value)
+# Note: This option can be changed without restarting.
+# Deprecated group/name - [DEFAULT]/log_config
+#log_config_append = <None>
+
+# Defines the format string for %%(asctime)s in log records. Default:
+# %(default)s . This option is ignored if log_config_append is set. (string
+# value)
+#log_date_format = %Y-%m-%d %H:%M:%S
+
+# (Optional) Name of log file to send logging output to. If no default is set,
+# logging will go to stderr as defined by use_stderr. This option is ignored if
+# log_config_append is set. (string value)
+# Deprecated group/name - [DEFAULT]/logfile
+#log_file = <None>
+
+# (Optional) The base directory used for relative log_file  paths. This option
+# is ignored if log_config_append is set. (string value)
+# Deprecated group/name - [DEFAULT]/logdir
+log_dir = /var/log/sahara
+
+# Uses logging handler designed to watch file system. When log file is moved or
+# removed this handler will open a new log file with specified path
+# instantaneously. It makes sense only if log_file option is specified and
+# Linux platform is used. This option is ignored if log_config_append is set.
+# (boolean value)
+#watch_log_file = false
+
+# Use syslog for logging. Existing syslog format is DEPRECATED and will be
+# changed later to honor RFC5424. This option is ignored if log_config_append
+# is set. (boolean value)
+#use_syslog = false
+use_syslog=<%= node[:sahara][:use_syslog] ? "True" : "False" %>
+
+# Syslog facility to receive log lines. This option is ignored if
+# log_config_append is set. (string value)
+#syslog_log_facility = LOG_USER
+
+# Log output to standard error. This option is ignored if log_config_append is
+# set. (boolean value)
+#use_stderr = true
+
+# Format string to use for log messages with context. (string value)
+#logging_context_format_string = %(asctime)s.%(msecs)03d %(process)d %(levelname)s %(name)s [%(request_id)s %(user_identity)s] %(instance)s%(message)s
+
+# Format string to use for log messages when context is undefined. (string
+# value)
+#logging_default_format_string = %(asctime)s.%(msecs)03d %(process)d %(levelname)s %(name)s [-] %(instance)s%(message)s
+
+# Additional data to append to log message when logging level for the message
+# is DEBUG. (string value)
+#logging_debug_format_suffix = %(funcName)s %(pathname)s:%(lineno)d
+
+# Prefix each line of exception output with this format. (string value)
+#logging_exception_prefix = %(asctime)s.%(msecs)03d %(process)d ERROR %(name)s %(instance)s
+
+# Defines the format string for %(user_identity)s that is used in
+# logging_context_format_string. (string value)
+#logging_user_identity_format = %(user)s %(tenant)s %(domain)s %(user_domain)s %(project_domain)s
+
+# List of package logging levels in logger=LEVEL pairs. This option is ignored
+# if log_config_append is set. (list value)
+#default_log_levels = amqplib=WARN,qpid.messaging=INFO,stevedore=INFO,eventlet.wsgi.server=WARN,sqlalchemy=WARN,boto=WARN,suds=INFO,keystone=INFO,paramiko=WARN,requests=WARN,iso8601=WARN,oslo_messaging=INFO,neutronclient=INFO
+
+# Enables or disables publication of error events. (boolean value)
+#publish_errors = false
+
+# The format for an instance that is passed with the log message. (string
+# value)
+#instance_format = "[instance: %(uuid)s] "
+
+# The format for an instance UUID that is passed with the log message. (string
+# value)
+#instance_uuid_format = "[instance: %(uuid)s] "
+
+# Enables or disables fatal status of deprecations. (boolean value)
+#fatal_deprecations = false
+
+#
+# From oslo.messaging
+#
+
+# Size of RPC connection pool. (integer value)
+# Deprecated group/name - [DEFAULT]/rpc_conn_pool_size
+#rpc_conn_pool_size = 30
+
+# The pool size limit for connections expiration policy (integer value)
+#conn_pool_min_size = 2
+
+# The time-to-live in sec of idle connections in the pool (integer value)
+#conn_pool_ttl = 1200
+
+# ZeroMQ bind address. Should be a wildcard (*), an ethernet interface, or IP.
+# The "host" option should point or resolve to this address. (string value)
+#rpc_zmq_bind_address = *
+
+# MatchMaker driver. (string value)
+# Allowed values: redis, dummy
+#rpc_zmq_matchmaker = redis
+
+# Number of ZeroMQ contexts, defaults to 1. (integer value)
+#rpc_zmq_contexts = 1
+
+# Maximum number of ingress messages to locally buffer per topic. Default is
+# unlimited. (integer value)
+#rpc_zmq_topic_backlog = <None>
+
+# Directory for holding IPC sockets. (string value)
+#rpc_zmq_ipc_dir = /var/run/openstack
+
+# Name of this node. Must be a valid hostname, FQDN, or IP address. Must match
+# "host" option, if running Nova. (string value)
+#rpc_zmq_host = localhost
+
+# Seconds to wait before a cast expires (TTL). The default value of -1
+# specifies an infinite linger period. The value of 0 specifies no linger
+# period. Pending messages shall be discarded immediately when the socket is
+# closed. Only supported by impl_zmq. (integer value)
+#rpc_cast_timeout = -1
+
+# The default number of seconds that poll should wait. Poll raises timeout
+# exception when timeout expired. (integer value)
+#rpc_poll_timeout = 1
+
+# Expiration timeout in seconds of a name service record about existing target
+# ( < 0 means no timeout). (integer value)
+#zmq_target_expire = 300
+
+# Update period in seconds of a name service record about existing target.
+# (integer value)
+#zmq_target_update = 180
+
+# Use PUB/SUB pattern for fanout methods. PUB/SUB always uses proxy. (boolean
+# value)
+#use_pub_sub = true
+
+# Use ROUTER remote proxy. (boolean value)
+#use_router_proxy = true
+
+# Minimal port number for random ports range. (port value)
+# Minimum value: 0
+# Maximum value: 65535
+#rpc_zmq_min_port = 49153
+
+# Maximal port number for random ports range. (integer value)
+# Minimum value: 1
+# Maximum value: 65536
+#rpc_zmq_max_port = 65536
+
+# Number of retries to find free port number before fail with ZMQBindError.
+# (integer value)
+#rpc_zmq_bind_port_retries = 100
+
+# Default serialization mechanism for serializing/deserializing
+# outgoing/incoming messages (string value)
+# Allowed values: json, msgpack
+#rpc_zmq_serialization = json
+
+# Size of executor thread pool. (integer value)
+# Deprecated group/name - [DEFAULT]/rpc_thread_pool_size
+#executor_thread_pool_size = 64
+
+# Seconds to wait for a response from a call. (integer value)
+#rpc_response_timeout = 60
+
+# A URL representing the messaging driver to use and its full configuration.
+# (string value)
+#transport_url = <None>
+
+# DEPRECATED: The messaging driver to use, defaults to rabbit. Other drivers
+# include amqp and zmq. (string value)
+# This option is deprecated for removal.
+# Its value may be silently ignored in the future.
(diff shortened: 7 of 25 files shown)


Itxaka:
tempest tests are now passing:

```
======
Totals
======
Ran: 40 tests in 12.0000 sec.
 - Passed: 40
 - Skipped: 0
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 0
Sum of execute time for each test: 5.6090 sec.
```

Itxaka:
retriggered the last jenkins job with `want_sahara=1` to run the tempests tests for sahara

Itxaka:
rebased against master becuase manila keeps failing on jenkins and I cannot see why :cry:

Itxaka:
gating passing :grinning:

chef/cookbooks/sahara/attributes/default.rb
```diff
@@ -0,0 +1,35 @@
+# Copyright 2015, SUSE, Inc.
```
toabctl: ```2016, SUSE Linux GmbH```

chef/cookbooks/sahara/recipes/common.rb
```diff
@@ -0,0 +1,94 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+package "openstack-sahara"
+
+# FIXME: sahara policy.json file should be copied to /etc/sahara on package install!
```
toabctl: That's fixed, right? So this part can be removed I think

chef/cookbooks/sahara/recipes/common.rb
```diff
@@ -0,0 +1,94 @@
...
+execute "Copy policy.json file" do
+  user    "root"
+  command "cp /usr/share/sahara/policy.json /etc/sahara/policy.json"
+end
+# FIXME: sahara api-paste.ini file should be copied to /etc/sahara on package install!
```
toabctl: same here

chef/cookbooks/sahara/recipes/common.rb
```diff
@@ -0,0 +1,94 @@
...
+  bind_port = node[:sahara][:ha][:ports][:api_port]
+  bind_host = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+else
+  bind_port = node[:sahara][:api][:bind_port]
+  bind_host = node[:sahara][:api][:bind_host]
+  bind_host = "0.0.0.0" if node[:sahara][:api][:bind_open_address]
+end
+
+# Get the name from the default keystone domain
+insecure = keystone_settings["insecure"] ? "--insecure" : ""
```
toabctl: Hm. Shouldn't the default domain be part of the keystone_settings? @rhafer  ?

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+#node_domain = novalocal
+
+# Use Neutron Networking (False indicates the use of Nova networking). (boolean
+# value)
+#use_neutron = false
+use_neutron = true
+
+# Use network namespaces for communication (only valid to use in conjunction
+# with use_neutron=True). (boolean value)
+#use_namespaces = false
```
toabctl: hm. do we need this?

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+# computes and racks. (string value)
+#compute_topology_file = etc/sahara/compute.topology
+
+# File with Swift topology.It should contain mapping between Swift nodes and
+# racks. (string value)
+#swift_topology_file = etc/sahara/swift.topology
+
+# Enables Sahara to use Keystone API v3. If that flag is disabled, per-job
+# clusters will not be terminated automatically. (boolean value)
+#use_identity_api_v3 = true
```
toabctl: I think we still allow v2 deployments. I think it's from the keystone cookbook the ```['api']['version']``` flag

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+#cluster_remote_threshold = 70
+
+# Proxy command used to connect to instances. If set, this command should open
+# a netcat socket, that Sahara will use for SSH and HTTP connections. Use
+# {host} and {port} to describe the destination. Other available keywords:
+# {tenant_id}, {network_id}, {router_id}. (string value)
+#proxy_command =
+
+# Region name used to get services endpoints. (string value)
+#os_region_name = <None>
```
toabctl: ```os_region_name = <%= @keystone_settings['endpoint_region'] %>```

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+
+# A method for Sahara to execute commands on VMs. (string value)
+#remote = ssh
+
+# Number of workers for Sahara API service (0 means all-in-one-thread
+# configuration). (integer value)
+#api_workers = 1
```
toabctl: ```api_workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>```

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+# Minimum value: 1
+#ssh_timeout_interactive = 1800
+
+# Overrides timeout for ssh operations with files, in seconds (integer value)
+# Minimum value: 1
+#ssh_timeout_files = 120
+
+# Enable the usage of the OpenStack Key Management service provided by
+# barbican. (boolean value)
+#use_barbican_key_manager = false
```
toabctl: Should be enabled when barbican is deployed. I think we have something similar in the magnum barclamp.

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+#
+# From sahara.config
+#
+
+# Version of the Cinder API to use. (integer value)
+# Deprecated group/name - [cinder]/cinder_api_version
+#api_version = 2
+
+# Allow to perform insecure SSL requests to cinder. (boolean value)
+#api_insecure = false
```
toabctl: Needs to be set if cinder is deployed with ssl. See other barclamps

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+
+
+[heat]
+
+#
+# From sahara.config
+#
+
+# Allow to perform insecure SSL requests to heat. (boolean value)
+#api_insecure = false
```
toabctl: Needs to be set when heat is deployed with ssl

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+
+
+[keystone]
+
+#
+# From sahara.config
+#
+
+# Allow to perform insecure SSL requests to keystone. (boolean value)
+#api_insecure = false
```
toabctl: needs to be set when keystone is deployed with ssl

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+
+# A PEM encoded Certificate Authority to use when verifying HTTPs connections.
+# Defaults to system CAs. (string value)
+#cafile = <None>
+
+# Verify HTTPS connections. (boolean value)
+#insecure = false
+
+# The region in which the identity server can be found. (string value)
+#region_name = <None>
```
toabctl: ```region_name = <%= @keystone_settings['endpoint_region'] %>```

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+
+# Directory used to cache files related to PKI tokens. (string value)
+#signing_dir = <None>
```
toabctl: ```signing_dir = /var/cache/sahara/keystone-signing``` (and the dir should be created in the package. can you check that please?)

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+
+
+[neutron]
+
+#
+# From sahara.config
+#
+
+# Allow to perform insecure SSL requests to neutron. (boolean value)
+#api_insecure = false
```
toabctl: needs to be set when neutron is deployed with ssl

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+
+
+[nova]
+
+#
+# From sahara.config
+#
+
+# Allow to perform insecure SSL requests to nova. (boolean value)
+#api_insecure = false
```
toabctl: needs to be set when nova is deployed with ssl

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1446 @@
...
+# Enables or disables inter-process locks. (boolean value)
+# Deprecated group/name - [DEFAULT]/disable_process_locking
+#disable_process_locking = false
+
+# Directory to use for lock files.  For security, the specified directory
+# should only be writable by the user running the processes that need locking.
+# Defaults to environment variable OSLO_LOCK_PATH. If external locks are used,
+# a lock path must be set. (string value)
+# Deprecated group/name - [DEFAULT]/lock_path
+#lock_path = <None>
```
toabctl: ```lock_path = /var/run/sahara``` (and should be created in the pkg)

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -462,6 +462,9 @@
   horizon_protocol = horizon[:horizon][:apache][:ssl] ? "https" : "http"
 end
 
+saharas = search(:node, "roles:sahara-server") || []
+use_sahara = saharas.empty? ? false : true
+
```
toabctl: We usually do:
```
`#{openstackcli_adm} service show data-processing &> /dev/null`
use_sahara = $?.success?
```

crowbar_framework/app/controllers/sahara_controller.rb
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright 2011-2013, Dell
```
toabctl: Just the ```SUSE Linux GmbH``` copyright

crowbar_framework/app/models/cinder_service.rb
```diff
@@ -80,6 +80,7 @@ def create_proposal
 
     base["attributes"][@bc_name]["service_password"] = random_password
     base["attributes"][@bc_name][:db][:password] = random_password
+    @logger.debug("Attributes: #{base}")
```
toabctl: I guess that's left from a debugging session?

crowbar_framework/app/models/sahara_service.rb
```diff
@@ -0,0 +1,118 @@
+#
+# Copyright 2011-2013, Dell
+# Copyright 2013-2014, SUSE LINUX Products GmbH
+# Copyright 2016, SUSE LINUX GmbH
```
toabctl: Just the ```SUSE Linux GmbH``` copyright

chef/cookbooks/sahara/recipes/common.rb
rhafer: It should, but it currently isn't :(. I guess we should address that.

But as we currently still create the admin user using the v2 API. The name will always be  "Default".

rhafer: @toabctl Should we quickly add something to the keystone helper. To avoid hardcoding it here?

toabctl: @rhafer you mean just adding a domain name which is always ```Default```? sounds good to me for now.

chef/cookbooks/sahara/templates/default/sahara.conf.erb
rhafer: Hm, but the v3 endpoint is always there. Not sure if there's a good reason to set this to ```false```

chef/cookbooks/sahara/recipes/common.rb
rhafer: @toabctl Yeah for now it would always give "Default" (until we make the keystone barclamp more domain-aware). I'll prepare something.

chef/cookbooks/sahara/attributes/default.rb
Itxaka: done

chef/cookbooks/sahara/recipes/common.rb
Itxaka: done

chef/cookbooks/sahara/recipes/common.rb
Itxaka: done

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done

chef/cookbooks/tempest/recipes/config.rb
Itxaka: done

crowbar_framework/app/models/cinder_service.rb
Itxaka: done

crowbar_framework/app/models/sahara_service.rb
Itxaka: done

chef/cookbooks/sahara/recipes/common.rb
rhafer: https://github.com/crowbar/crowbar-openstack/pull/509 should do it

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: Isn't /var/run cleared on reboot? 

I thougth that anything under /var/run for services would be created by the statup scripts of those instead?

from /etc/init.d/openstack-sahara-api

```
RUNDIR="/var/run/sahara"

# $RUNDIR can be tmpfs, thus we have to create/own it here:
mkdir -p $RUNDIR && chown $USER. $RUNDIR

```

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: This seems to be created by the keystone middleware itself:

`
2016-08-18 09:51:59.442 3870 INFO keystonemiddleware.auth_token [-] Using /var/cache/sahara/keystone-signing as cache directory for signing certificate
`

Removing the dir and relaunching recreates the dir automatically :)

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: Very good question, Im not sure.

Checking the node attributes for the neutron-server shows that there is a key `allow_overlapping_ips` in the neutron attributes.
`root@crowbar:~ # knife node show d52-54-77-77-77-01.g10.cloud.suse.de -a neutron.allow_overlapping_ips
neutron.allow_overlapping_ips:  true`


As far as I know, to allow for overlapping IPs, then namespaces should be enabled. And I think thats the default so different tenants can create the same overlapping networks.

Would be good to have a neutron expert here to confirm or denied this...

toabctl: @rhafer @rossella could you help here?

rhafer: Skimming through the sahara docs I think it makes sense to set it to ```true```. In neutron we always enable the use_namespaces (there's really no good reason to disable it there).

There's a very good reason to enable it stated in the sahara docs:
"Sahara accesses instances at several stages of cluster spawning through SSH and HTTP. Floating IPs and network namespaces (see Networking configuration) will be automatically used for access when present. When floating IPs are not assigned to instances and namespaces are not being used, sahara will need an alternative method to reach them." (http://docs.openstack.org/developer/sahara/userdoc/advanced.configuration.guide.html#custom-network-topologies)


But for sahara there are a few nodes about additional configuration that this requires when running sahara as non root: http://docs.openstack.org/developer/sahara/userdoc/configuration.guide.html#neutron-nova-network

toabctl:
Typo in commit msg ```keysotne api version```

toabctl:
Also can you prefix the commits with ```sahara:``` or something to make it clear that it's about sahara?

chef/cookbooks/sahara/metadata.rb
```diff
@@ -0,0 +1,31 @@
...
+
+name "sahara"
+maintainer "Crowbar project"
+maintainer_email "crowbar@googlegroups.com"
+license "Apache 2.0"
+description "Installs/Configures sahara"
+long_description IO.read(File.join(File.dirname(__FILE__), "README.md"))
+version "0.1"
+
+depends "database"
```
toabctl: Not sure about this but doesn't Sahara depend on ```cinder``` also? Maybe even ```heat``` ?

chef/cookbooks/sahara/recipes/common.rb
toabctl: @Itxaka #509 is merged. Can you rebase and fix the code?

chef/cookbooks/sahara/templates/default/sahara.conf.erb
toabctl: @Itxaka with the comment from @rhafer , can you revert this and just leave it always set to ```true``` ?

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1459 @@
...
+#wsgi_default_pool_size = 100
+
+# Maximum line size of message headers to be accepted. max_header_line may need
+# to be increased when using large tokens (typically those generated when
+# keystone is configured to use PKI tokens with big service catalogs). (integer
+# value)
+#max_header_line = 16384
+
+# If False, closes the client socket connection explicitly. (boolean value)
+#wsgi_keep_alive = true
```
toabctl: We set this to ```false``` in other configs. See bb7420313010995c0ba62bad70e9ecb4d99df04a

chef/cookbooks/sahara/recipes/common.rb
```diff
@@ -0,0 +1,131 @@
...
+
+# keystone api version
+use_identity_api_v3 = keystone_settings["api_version"].to_i == 3 ? true : false
+
+# Use barbican
+`#{openstack_command} service show key-manager &> /dev/null`
+use_barbican = $?.success? && node[:sahara][:use_barbican]
+
+# cinder insecure?
+cinders = search(:node, "roles:cinder-controller") || []
```
toabctl: iirc we want to get rid of ```search```. Can you use ```search_env_filtered()``` instead. @vuntz is that the way to go?

chef/cookbooks/sahara/templates/default/sahara.conf.erb
toabctl: ok. That's fine. The ```and should be created in the pkg``` was more a question if that's done. thx

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1459 @@
...
+# Deprecated group/name - [DEFAULT]/notification_level
+#level = INFO
+
+# Notification publisher_id for outgoing notifications (string value)
+# Deprecated group/name - [DEFAULT]/notification_publisher_id
+#publisher_id = <None>
+
+# Enables sending notifications to Ceilometer (boolean value)
+# Deprecated group/name - [DEFAULT]/enable_notifications
+#enable = false
```
toabctl: I think we want to enable this if we have ceilometer deployed.

chef/cookbooks/sahara/recipes/common.rb
Itxaka: done!

chef/cookbooks/sahara/metadata.rb
Itxaka: The recipes dont use any resources from those cookbooks (nova, heat, glance) so unless there is a policy on having those cookbooks as depends, I will just remove them. For the other cookbooks (database, keystone, crowbar-*, utils) we use resources from there so they should be included.

FYI: cinder/heat/nova interaction from this barclamp is only for searching for which those cookbooks are not required. No resources are used from them.

Itxaka:
> Also can you prefix the commits with sahara: or something to make it clear that it's about sahara?

Im waiting until everything is moreless stable to squash the commits together and will change the commit message :)

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done!

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done!

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: done!

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: Good catch! I addded the use_namespaces + use_rootwrap and the sudoers file so sahara can work as non-root.

chef/cookbooks/sahara/recipes/common.rb
Itxaka: done!

As far as I understand this function reduces the search to the environment, but I havent seen different environments anywhere? Any tips on what those environments come to be?

chef/roles/sahara-server.rb
```diff
@@ -0,0 +1,10 @@
+name "sahara-server"
+description "sahara Role - Node registered as a sahara server"
+run_list(
+  "recipe[sahara::common]",
+  "recipe[sahara::sql]",
+  "recipe[sahara::api]",
+  "recipe[sahara::engine]"
```
vuntz: See https://github.com/crowbar/crowbar-openstack/pull/510/files#r76246115; we should use a role recipe.

Itxaka: done!

toabctl:
@Itxaka I retriggered gating with the ```want_sahara=1``` parameter.

chef/cookbooks/sahara/metadata.rb
toabctl: ok. thx for clarification.

chef/cookbooks/sahara/templates/default/sahara-rootwrap.erb
```diff
@@ -0,0 +1 @@
+<%= @node[:sahara][:user] %> ALL = (root) NOPASSWD: /usr/bin/sahara-rootwrap /etc/sahara/rootwrap.conf *
```
toabctl: We usually have the sudoers file in the package. Can you add it there? See the openstack-cinder package for an example.

Itxaka: Done! but as in the cinder barclamp we want to overwrite it with ours, Im guessing to set the proper username? Doesnt seem like we are gonna change the user at anytime but ¯\_(ツ)_/¯

toabctl:
@Itxaka the barclamp is failing when applying. See logs from the ```want_sahara``` run. I reserved the host where this happes so just ping me on irc.

chef/cookbooks/sahara/templates/default/sahara-rootwrap.erb
toabctl: Done where? I don't see ```sahara-sudoers``` in https://build.opensuse.org/package/show/Cloud:OpenStack:Master/openstack-sahara . Can you create a SR for that? Also for Mitka:Staging? Or am I missing something?

Itxaka: I got it ready but the openstack-sahara seems to be failing to build due to missing dependencies (outdated openstacksdk, missing osc-lib) so I didnt created a request yet, will try to fix those packages first before pushing the requests against Cloud:Openstack:Master|Mitaka:Staging

Sorry!

Itxaka:
> @Itxaka the barclamp is failing when applying. See logs from the want_sahara run. I reserved the host where this happes so just ping me on irc.

Damn, was missing a semicolon :see_no_evil:

Itxaka:
00:00:03.997 all dirs (/root/pool/3 /root/pool/1) are currently in use at /root/github.com/SUSE-Cloud/automation/scripts/mkcloudhost/allocpool line 44.
00:00:03.997 PID file had 30874
00:00:03.997 
00:00:03.997 someone else owns /root/pool/3/tmpqueuesched.pid and is alive
00:00:03.997 PID file had 32269
00:00:03.997 
00:00:03.997 someone else owns /root/pool/1/tmpqueuesched.pid and is alive
00:00:03.997 + ret=25

Itxaka:
working now, I did a mess while testing the context stuff, sorry :open_mouth:

crowbar_framework/app/models/sahara_service.rb
```diff
@@ -0,0 +1,115 @@
...
+    base["attributes"][@bc_name][:trustee][:domain_admin_password] = random_password
+
+    @logger.debug("sahara create_proposal: exiting")
+    base
+  end
+
+  def validate_proposal_after_save(proposal)
+    validate_one_for_role proposal, "sahara-server"
+    validate_at_least_n_for_role proposal, "sahara-server", 1
+    super
```
vuntz: Can be done in a followup PR: you should validate that barbican is deployed when the use_barbican attribute is set.

crowbar_framework/app/models/sahara_service.rb
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  # Turn off multi proposal support till it really works and people ask for it
+  def self.allow_multiple_proposals?
+    false
+  end
+
+  def proposal_dependencies(role)
+    answer = []
+    deps = ["database", "rabbitmq", "keystone", "nova", "glance", "neutron", "heat"]
```
vuntz: Aren't you missing cinder here? Or is it optional? If optional, see the code for ceilometer where we check for the database only when needed.

We should do this for barbican too.

chef/data_bags/crowbar/template-sahara.schema
```diff
@@ -0,0 +1,128 @@
...
+            "verbose": { "type": "bool", "required": true },
+            "use_syslog": { "type": "bool", "required": true },
+            "use_barbican": { "type": "bool", "required": true },
+            "database_instance": { "type": "str", "required": true },
+            "keystone_instance": { "type": "str", "required": true },
+            "glance_instance": { "type": "str", "required": true },
+            "nova_instance": { "type": "str", "required": true },
+            "heat_instance": { "type": "str", "required": true },
+            "neutron_instance": { "type": "str", "required": true },
+            "rabbitmq_instance": { "type": "str", "required": true },
```
vuntz: Probably missing the attributes for cinder_instance and barbican_instance here.

chef/cookbooks/sahara/recipes/common.rb
```diff
@@ -0,0 +1,127 @@
...
+cinders = search_env_filtered(:node, "roles:cinder-controller") || []
+if cinders.empty?
+  raise "Need a cinder controller to deploy sahara"
+else
+  cinder = cinders[0]
+  cinder_insecure = cinder[:cinder][:ssl][:insecure]
+end
+
+# heat insecure?
+heats = search_env_filtered(:node, "roles:heat-server") || []
```
vuntz: Because you've put heat (and friends like neutron, nova, etc.) in the attributes (with heat_instance, etc.), you can directly do `heat = get_instance("roles:heat-server")` and assume that it will not be `nil`.

If you add such attributes for barbican and cinder, then do the same for these too.

Itxaka: Ohhh, I didnt knew that, thats handy.

Althougth, as Barbican is not required but optional, will that make Barbican a dependency of the barclamp? Or will get_instance just return nil in that case because there is no barbican instance? Or will the barclamp proposal view just die?

chef/data_bags/crowbar/template-sahara.schema
Itxaka: Added cinder_instance, not sure about barbican.

Itxaka:
RE:Barbican

 - Is optional
 - The barclamp does not default to using barbican if it exists, it also allows the user to select if they want to use barbican or not `use_barbican`

To deploy it:
 - Should Barbican be added to the attributes on the proposal?
   - What happens if there is no barbican available? Does `get_instance` returns nil?
   - Do we need to add it to the json/schema if its optional?
 - Should Barbican be only searched on chef run?
   - How to communicate to the user that `use_barbican` will only work if there is a barbican proposal deployed? Help text on the view?
 - Should the attribute `use_barbican` only be displayed if we find a deployed proposal for Barbican?

Itxaka:
@vuntz @toabctl @aspiers Maybe I should just remove the barbican use in this barclamp and create a new card to add it later as I dont want to drag the barclamp merge any further :/

Itxaka:
I have removed the barbican stuff. I really want to get this in and then afterwards we can enable the use of barbican for the cookbook.

sahara.yml
```diff
@@ -0,0 +1,37 @@
...
+barclamp:
+  name: 'sahara'
+  display: 'Sahara'
+  description: 'OpenStack Sahara Data Processing'
+  version: 0
+  user_managed: true
+  requires:
+    - 'keystone'
+    - 'heat'
+    - 'neutron'
```
dguitarbite: Small nitpick, glance should be required and AFAIK, it is not really quite fruitful to run M-R batches without HDD :dart: ... so may be should be considered w.r.t. even the way Hadoop works for the given impact on the dependent OpenStack services. They need not be implicitly mentioned as dependencies but logically they are required for sure.

dguitarbite:
+1
Small nitpick w.r.t. the dependency thingy.

sahara.yml
Itxaka: Good point, I had the dependency on the proposal and totally forget about the yml file!

Itxaka:
@vuntz @aspiers @toabctl All tests passing, just missing a +1 :D

Itxaka:
oh wow, you can +1 your own patches and lgtm will take them as good :S

toabctl:
+1

chef/cookbooks/sahara/templates/default/sahara-rootwrap.erb
aplanas: is this now in the pkg? does this template still needed?

Itxaka: Its is but we need to overwrite it for the user, in case we change it down the line (I know, thats practically impossible but who knows) ¯\_(ツ)_/¯

Pushing a change for the name as it has changed in the package to /etc/sudoers.d/openstack-sahara

aplanas: ah I see, thanks!

toabctl:
+1 (still)

sahara.yml
```diff
@@ -0,0 +1,39 @@
...
+    - 'nova'
+    - 'rabbitmq'
+    - 'cinder'
+    - 'heat'
```
vuntz: Heat is listed twice here.

crowbar_framework/app/views/barclamp/sahara/_edit_attributes.html.haml
```diff
@@ -0,0 +1,18 @@
...
+    = instance_field :heat
+    = instance_field :glance
+    = instance_field :neutron
+
+    %fieldset
+      %legend
+        = t(".logging_header")
+      = boolean_field :verbose
+      = boolean_field :debug
+      = boolean_field :use_syslog
```
vuntz: Do we usually expose the use_syslog setting? I think it's hidden for all other barclamps.

crowbar_framework/app/models/sahara_service.rb
```diff
@@ -0,0 +1,115 @@
...
+    base["attributes"][@bc_name]["service_password"] = random_password
+    base["attributes"][@bc_name][:db][:password] = random_password
+
+    @logger.debug("sahara create_proposal: exiting")
+    base
+  end
+
+  def validate_proposal_after_save(proposal)
+    validate_one_for_role proposal, "sahara-server"
+    validate_at_least_n_for_role proposal, "sahara-server", 1
```
vuntz: I believe you don't need that second line, as we're already doing some check on the number of elements for that role in the line earlier.

chef/cookbooks/sahara/templates/default/sahara.conf.erb
```diff
@@ -0,0 +1,1463 @@
...
+
+
+[swift]
+
+#
+# From sahara.config
+#
+
+# Allow to perform insecure SSL requests to swift. (boolean value)
+#api_insecure = false
```
vuntz: Do we have swift integration enabled? If yes, we should set this option to the correct value.

chef/cookbooks/sahara/templates/default/sahara.conf.erb
dguitarbite: @Itxaka Small nitpick: I am not sure if we need `use_rootwrap` for the scope of sahara's use-cases.

chef/data_bags/crowbar/template-sahara.schema
dguitarbite: Barbican is not a hard requirement. If I understand this JSON block's purpose.

dguitarbite:
+1 Of course with the CI and other reviewer's comments addressed. Thanks.

sahara.yml
Itxaka: done!

crowbar_framework/app/views/barclamp/sahara/_edit_attributes.html.haml
Itxaka: Indeed, other barclamps do not expose this. But we still have the switch in the configuration to enable or disable it based on node[barclamp][:use_syslog], so why not expose it to the user and set it to a default value? And if we want it to be always false (because of supportconfig?) then whyset a switch instead of setting it to false on the template directly?

Knowing that we do not have a central logging system in place for crowbar, enabling this means that there is a cheap and easy way for customers to use syslog as forwarder to set up their own centralized log management.

chef/cookbooks/sahara/templates/default/sahara.conf.erb
Itxaka: No, we dont have swift integration. I would like to have that as a follow up patch as it requires more general changes like users and swift proxy, domain delegation trust and so on.

crowbar_framework/app/models/sahara_service.rb
Itxaka: Indeed, fixed!

Itxaka:
Ummm. Hound finding stuff now that was already in there since the start? was hound broken?

toabctl:
> Ummm. Hound finding stuff now that was already in there since the start? was hound broken?

Maybe the Hound creators changed something. I would just ignore hound as long as I'm not able to run it locally with a reproducible result ( but I guess I'm the only one with that opinion).

toabctl:
+1

toabctl:
And while I'm already complaining - LGTM is also useless - you updated the PR and it still shows that there are 2 +1's. tsts (end of complain mode)

chef/cookbooks/sahara/recipes/api.rb
```diff
@@ -0,0 +1,98 @@
...
+
+keystone_register "register sahara service" do
+  protocol keystone_settings["protocol"]
+  insecure keystone_settings["insecure"]
+  host keystone_settings["internal_url_host"]
+  port keystone_settings["admin_port"]
+  auth register_auth_hash
+  service_name "sahara"
+  service_type "data-processing"
+  service_description "Openstack sahara - Data Processing"
```
vuntz: Can be done in followup PR, but I'd use "OpenStack Sahara"

chef/cookbooks/sahara/recipes/common.rb
```diff
@@ -0,0 +1,101 @@
...
+nova = get_instance("roles:nova-controller")
+nova_insecure = nova[:nova][:ssl][:insecure]
+
+# use ceilometer?
+ceilometers = search_env_filtered(:node, "roles:ceilometer-server")
+use_ceilometer = false
+
+unless ceilometers.empty?
+  use_ceilometer = true
+end
```
vuntz: Can be done in followup PR: change the last 5 lines to `use_ceilometer = !ceilometers.empty?`

vuntz: Actually, reading the code: I'd just always send notifications, and not depend on whether ceilometer is deployed or not. The notifications won't hurt, I guess.

chef/cookbooks/sahara/templates/default/sahara-rootwrap.erb
vuntz: To be fair, if we change the user, we'd change it also in the package, so we shouldn't need to do anything in the cookbook :-)

The only reason we have an attribute for the user is to make sure we use the right user on each distro.

So I'd be in favor of killing that bit too

(can be done in followup PR)

chef/cookbooks/sahara/recipes/ha.rb
```diff
@@ -0,0 +1,71 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+haproxy_loadbalancer "sahara-api" do
+  address "0.0.0.0"
```
vuntz: This should probably be different in case of `bind_open_port == false`; see what we do for glance there.

chef/cookbooks/sahara/recipes/ha.rb
```diff
@@ -0,0 +1,71 @@
...
+api_primitive = "sahara-api"
+pacemaker_primitive api_primitive do
+  agent node[:sahara][:ha][:api_ra]
+  op node[:sahara][:ha][:op]
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects << "pacemaker_primitive[#{api_primitive}]"
+
+group_name = "s-sahara"
```
vuntz: The group name should be `g-$something`, so `g-sahara` here.

vuntz: That being said, since you only have one element, there's no need for a group :-)

vuntz: Actually, I guess we want to have a pacemaker resource for sahara-engine too, so we neeD the group.

chef/cookbooks/sahara/recipes/common.rb
```diff
@@ -0,0 +1,101 @@
...
+    heat_insecure: heat_insecure,
+    neutron_insecure: neutron_insecure,
+    nova_insecure: nova_insecure,
+    use_ceilometer: use_ceilometer
+  )
+end
+
+if ha_enabled
+  Chef::Log.info("HA support for sahara is enabled")
+  include_recipe "sahara::ha"
```
vuntz: I would include it from the role recipe, as this needs to be run after the engine bit too.

dguitarbite:
+1 from my side. Thanks for updating the PR.

chef/cookbooks/sahara/recipes/ha.rb
Itxaka: @vuntz I thougth that while we need the API to be HA with a VIP and everything, we dont really need the engine to have a VIP as we can deploy it to X nodes without any repercussions so there is really no need to have it checked out and controlled by pacemaker?

I may be wrong in this, but as far as I undertand the arch of sahara, there should be no need for the engine to be controlled under pacemaker.

chef/cookbooks/sahara/recipes/common.rb
Itxaka: I guess that with my comment above regarding the engine on pacemaker, this should not change.

chef/cookbooks/sahara/recipes/api.rb
Itxaka: done

chef/cookbooks/sahara/recipes/common.rb
Itxaka: Wouldnt that fill up a MQ queue with useless messages?

Seems like a recipe for an issue down the road IMO...

Itxaka: > Can be done in followup PR: change the last 5 lines to use_ceilometer = !ceilometers.empty?

done!

chef/cookbooks/sahara/recipes/ha.rb
Itxaka: done!

chef/cookbooks/sahara/recipes/ha.rb
Itxaka: > That being said, since you only have one element, there's no need for a group :-)

done!

Itxaka:
goddammit hound

Itxaka:
waiting for https://github.com/crowbar/crowbar-openstack/pull/523

chef/cookbooks/sahara/recipes/ha.rb
```diff
@@ -0,0 +1,65 @@
...
+# all the required packages installed before we create the pacemaker
+# resources
+crowbar_pacemaker_sync_mark "sync-sahara before_ha"
+
+# Avoid races when creating pacemaker resources
+crowbar_pacemaker_sync_mark "wait-sahara_ha_resources"
+
+transaction_objects = []
+
+api_primitive = "sahara-api"
```
vuntz: Continuing the discussion from https://github.com/crowbar/crowbar-openstack/pull/480#discussion_r77850812:

> @vuntz I thougth that while we need the API to be HA with a VIP and everything, we dont really need the engine to have a VIP as we can deploy it to X nodes without any repercussions so there is really no need to have it checked out and controlled by pacemaker?
> 
> I may be wrong in this, but as far as I undertand the arch of sahara, there should be no need for the engine to be controlled under pacemaker.

So that's based on a wrong assumption: we don't put services in pacemaker for the VIP, but because we want pacemaker to manage the service, restart it when needed and do something when things are awfully wrong.

So yes, we want sahara-engine to be handled in pacemaker (but not part of haproxy).

Btw, that's what we do in other barclamps too. Look at nova-scheduler, for instance.

Itxaka: > So that's based on a wrong assumption: we don't put services in pacemaker for the VIP, but because we want pacemaker to manage the service, restart it when needed and do something when things are awfully wrong.

I see. So pacemaker is our service watchdog _only_ when the cloud is on HA.

Will change this to get the engine in pacemaker and run the HA recipe from the role instead.

toabctl:
@Itxaka ```Error: Committing the crowbar 'default' proposal for 'sahara' failed (1).```.:
```
[2016-09-08T18:13:30+00:00] WARN: Current  execute[a2enmod deflate]: /var/chef/cache/cookbooks/apache2/definitions/apache_module.rb:28:in `block in from_file'
Recipe Compile Error in /var/chef/cache/cookbooks/sahara/recipes/role_sahara_server.rb
NameError
---------
uninitialized constant Chef::Recipe::SaharaHelper
Cookbook Trace:
---------------
  /var/chef/cache/cookbooks/sahara/recipes/common.rb:18:in `from_file'
  /var/chef/cache/cookbooks/sahara/recipes/role_sahara_server.rb:18:in `from_file'
Relevant File Content:
----------------------
/var/chef/cache/cookbooks/sahara/recipes/common.rb:
 11:  # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 12:  # See the License for the specific language governing permissions and
 13:  # limitations under the License.
 14:  #
 15:  
 16:  package "openstack-sahara"
 17:  
 18>> network_settings = SaharaHelper.network_settings(node)
 19:  db_settings = fetch_database_settings
 20:  
 21:  include_recipe "database::client"
 22:  include_recipe "#{db_settings[:backend_name]}::client"
 23:  include_recipe "#{db_settings[:backend_name]}::python-client"
 24:  
 25:  # get Database data
 26:  db_password = node[:sahara][:db][:password]
 27:  sql_connection = "#{db_settings[:url_scheme]}://#{node[:sahara][:db][:user]}:"\
[2016-09-08T18:13:30+00:00] ERROR: Running exception handlers
```

Itxaka:
Thanks @toabctl should be fixed. That was a very dumb mistake :D

toabctl:
I retriggered gating with ```want_sahara=1```...

nicolasbock:
+1

Itxaka:
Retriggered with want_sahara. Did a mistake and set the port to the host and the host to the port so the service wasnt coming up :D

dguitarbite:
+1

Itxaka:
@vuntz @aspiers  Any last comments before I merge? I think I covered all your requests :)

PR #481, opened 2016-08-03
trove: Replace openstack-identity with keystone_register

rhafer:
This is a first step to unify the trove cookbook with the rest of the
cookbooks.
Remove the openstack-identity cookbook and replace it with the keystone_register
methods to have a single resource provider handling the user/service/endpoint
registration.
Note: The reason why the code to register the endpoints and users moved to api.rb
is that this is an ongoing effort to also get rid of openstack-database and
have the chef recipies all in the trove cookbook. And all the other services are
doing the endpoint/service/user handling in the api.rb recipe.


rhafer:
BTW, this is just https://github.com/crowbar/crowbar-openstack/pull/443 with the additional hound fixes (and nothing else).
So I guess my +1 on the old PR is still mostly valid.

sayalilunkad:
+1

rhafer:
@aplanas Would mind having another look at this. It's #443 with just some hound fixes on top.

aplanas:
@rhafer sure

aplanas:
As I can see the MySQL grant problem is still there, but I think that is a different issue. If we create a but report for the MySQL thingy, I would be very happy

+1

rhafer:
> As I can see the MySQL grant problem is still there, but I think that is a different issue. If we create a but report for the MySQL thingy, I would be very happy

Yeah, that's something that needs to be fixed by either switching trove to use postgresql like every other service. Or by fixing the mysql_user_provider in the database cookbook (https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/database/libraries/provider_database_mysql_user.rb) like we did for postgresql some years ago: https://github.com/crowbar/crowbar-openstack/commit/49b149e563f0054ad2eaef830e0df280d729a899

rhafer:
I'll create a card for that.

PR #482, opened 2016-08-04
Updating keystone template for Mitaka

sayalilunkad:


sayalilunkad:
Note: Not tested yet.

mjura:
Why do you remove Ldap support configuration?

sayalilunkad:
@mjura If I understand correctly the ldap configuration that I removed was not present in the mitaka sample config template and you can see the comment here https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/keystone/templates/default/keystone.conf.erb#L1090 which say it would be deprecated.

rhafer:
> If I understand correctly the ldap configuration that I removed was not present in the mitaka sample config

Yeah, AFAIK the assignment backend for LDAP was removed in mitaka. The only thing that's left for LDAP is the identity backend(users and groups).

chef/cookbooks/keystone/attributes/default.rb
```diff
@@ -128,7 +101,7 @@
 default[:keystone][:ldap][:group_allow_create] = true
 default[:keystone][:ldap][:group_allow_update] = true
 default[:keystone][:ldap][:group_allow_delete] = true
-default[:keystone][:ldap][:use_pool] = false
+default[:keystone][:ldap][:use_pool] = true
```
toabctl: Why are you changing the default here?

chef/data_bags/crowbar/migrate/keystone/101_update_default_attributes.rb
```diff
@@ -0,0 +1,63 @@
+def upgrade(ta, td, a, d)
+  if a["ldap"]["use_pool"] == false
```
toabctl: Why? If a user set has set this to "false" we don't want to override this decision.

chef/data_bags/crowbar/migrate/keystone/101_update_default_attributes.rb
```diff
@@ -0,0 +1,63 @@
...
+  a["ldap"].delete("role_member_attribute")
+  a["ldap"].delete("role_attribute_ignore")
+  a["ldap"].delete("role_allow_create")
+  a["ldap"].delete("role_allow_update")
+  a["ldap"].delete("role_allow_delete")
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  if a["ldap"]["use_pool"] == true
```
toabctl: just use the default. No need for the if

chef/data_bags/crowbar/migrate/keystone/101_update_default_attributes.rb
```diff
@@ -0,0 +1,63 @@
...
+    a["ldap"]["use_pool"] = ta["ldap"]["use_pool"]
+  end
+  a["ldap"]["project_tree_dn"] = ""
```
toabctl: Should be ```a["ldap"]["project_tree_dn"] = ta["ldap"]["project_tree_dn"]```
And same thing for all following lines

chef/cookbooks/keystone/templates/default/keystone.conf.erb
```diff
@@ -94,83 +103,97 @@ admin_endpoint = <%= @admin_endpoint %>
...
+# instantaneously. It makes sense only if log_file option is specified and
+# Linux platform is used. This option is ignored if log_config_append is set.
+# (boolean value)
+#watch_log_file = false
+
 # Use syslog for logging. Existing syslog format is DEPRECATED and will be
-# changed later to honor RFC5424. (boolean value)
+# changed later to honor RFC5424. This option is ignored if log_config_append
+# is set. (boolean value)
 #use_syslog = false
```
toabctl: Why not ```use_syslog = <%= @use_syslog ? "True" : "False" %>``` ?

chef/data_bags/crowbar/template-keystone.json
```diff
@@ -6,7 +6,6 @@
       "debug": false,
       "frontend": "native",
       "verbose": true,
-      "use_syslog": false,
```
toabctl: Why was this removed? It's still available in the config.

chef/data_bags/crowbar/template-keystone.schema
```diff
@@ -3,14 +3,13 @@
   "mapping": {
     "id": { "type": "str", "required": true, "pattern": "/^keystone-|^template-keystone$/" },
     "description": { "type": "str", "required": true },
-    "attributes": { "type": "map", "required": true, 
+    "attributes": { "type": "map", "required": true,
```
toabctl: whiltespace fixes are unrelated

chef/cookbooks/keystone/attributes/default.rb
sayalilunkad: Because the default have changed. You can check here http://docs.openstack.org/mitaka/config-reference/tables/conf-changes/keystone.html

chef/data_bags/crowbar/migrate/keystone/101_update_default_attributes.rb
sayalilunkad: So because the default value changed for `[:ldap][:use_pool]` I thought we need to do the migration in this way. Am I missing something?

chef/cookbooks/keystone/templates/default/keystone.conf.erb
sayalilunkad: It is actually a deprecated option if you see the config reference http://docs.openstack.org/mitaka/config-reference/tables/conf-changes/keystone.html so I thought it's best if we stop using it.

chef/data_bags/crowbar/template-keystone.json
sayalilunkad: it has been deprecated.

chef/cookbooks/keystone/templates/default/keystone.conf.erb
toabctl: for the record: the option is not deprecated. The release notes are wrong

chef/data_bags/crowbar/template-keystone.schema
sayalilunkad: I can revert the change but wouldn't  it be better like this?

toabctl: It's better but not what you describe in the commit message. just do this in an extra commit

chef/data_bags/crowbar/migrate/keystone/101_update_default_attributes.rb
rhafer: While I think it's ok to change the default for new deployments to true (that's what the changes to ```attributes/default.rb``` and ```template-keystone.json``` are good for).

I think we should not touch that value for existing deployments. So I'd prefer to leave ```[:ldap][:use_pool]``` untouched in the migration.

chef/cookbooks/keystone/templates/default/keystone.conf.erb
```diff
@@ -1362,14 +1268,19 @@ use_pool = <%= node[:keystone][:ldap][:use_pool] %>
...
 # End user auth connection pool size. (integer value)
 #auth_pool_size = 100
 
 # End user auth connection lifetime in seconds. (integer value)
 #auth_pool_connection_lifetime = 60
 
+# If the members of the group objectclass are user IDs rather than DNs, set
+# this to true. This is the case when using posixGroup as the group objectclass
+# and OpenDirectory. (boolean value)
+#group_members_are_ids = false
```
rhafer: I think we should make this configurable via attributes.

chef/cookbooks/keystone/templates/default/keystone.conf.erb
```diff
@@ -1014,6 +1072,9 @@ user_id_attribute = <%= node[:keystone][:ldap][:user_id_attribute] %>
 #user_name_attribute = sn
 user_name_attribute = <%= node[:keystone][:ldap][:user_name_attribute] %>
 
+# LDAP attribute mapped to user description. (string value)
+#user_description_attribute = description
```
rhafer: Also a canditate for a new attribute

chef/cookbooks/keystone/templates/default/keystone.conf.erb
```diff
@@ -1081,195 +1154,16 @@ user_enabled_emulation = <%= node[:keystone][:ldap][:user_enabled_emulation] %>
 #user_enabled_emulation_dn = <None>
 user_enabled_emulation_dn = <%= node[:keystone][:ldap][:user_enabled_emulation_dn] %>
 
+# Use the "group_member_attribute" and "group_objectclass" settings to
+# determine membership in the emulated enabled group. (boolean value)
+#user_enabled_emulation_use_group_config = false
```
rhafer: Also a canditate for a new attribute

chef/data_bags/crowbar/migrate/keystone/101_update_default_attributes.rb
```diff
@@ -1,4 +1,7 @@
 def upgrade(ta, td, a, d)
+  a["ldap"]["user_description_attribute"] = "description"
+  a["ldap"]["group_members_are_ids"] = false
+  a["ldap"]["user_enabled_emulaton_use_group_config"] = false
```
rhafer: Please use ```ta["ldap"]["<attributename>"]``` as the default for the new attributes here. Instead of hardcoded values.

sayalilunkad: Done.

chef/cookbooks/keystone/templates/default/keystone.conf.erb
sayalilunkad: Done.

chef/cookbooks/keystone/templates/default/keystone.conf.erb
sayalilunkad: Done.

chef/data_bags/crowbar/migrate/keystone/101_update_default_attributes.rb
```diff
@@ -0,0 +1,64 @@
+def upgrade(ta, td, a, d)
+  a["ldap"]["user_description_attribute"] = ta["ldap"]["user_description_attribute"]
+  a["ldap"]["group_members_are_ids"] = ta["ldap"]["group_members_are_ids"]
+  a["ldap"]["user_enabled_emulaton_use_group_config"] = \
```
rhafer: I think you don't need the backslash here.

rhafer:
I think this is good. After the nitpick about the backslash is fixed.

rhafer:
+1

toabctl:
+1

rhafer:
+1 again to make lgtm happy

jsuchome:
Looks like some keys were not correctly deleted on migration. When I'm trying to migrate schemas from cloud6 to cloud7, I get:
```
Failed to validate migrated proposal default for keystone:
key 'project_enabled_attribute:' is undefined.
key 'project_enabled_emulation:' is undefined.
key 'project_enabled_emulation_dn:' is undefined.
key 'user_enabled_emulaton_use_group_config:' is undefined.
```

PR #483, opened 2016-08-04
Fixes cinder-volumes check

sayalilunkad:
`not_if` executes ruby code so the previous condition always returns true.

### DIFF
--- a/chef/cookbooks/cinder/recipes/volume.rb
+++ b/chef/cookbooks/cinder/recipes/volume.rb
@@ -55,11 +55,9 @@ def make_loopback_volume(backend_id, volume)
   volname = volume[:local][:volume_name]
   fname = volume[:local][:file_name]
 
-  return if volume_exists(volname)
-
   bash "Create volume group #{volname}" do
     code "vgcreate #{volname} `losetup -j #{fname} | cut -f1 -d:`"
-    not_if "vgs #{volname}"
+    not_if { `vgs "#{volname}"`.include?(volname) }
   end
 end
 


chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -55,11 +55,9 @@ def make_loopback_volume(backend_id, volume)
   volname = volume[:local][:volume_name]
   fname = volume[:local][:file_name]
 
-  return if volume_exists(volname)
-
   bash "Create volume group #{volname}" do
     code "vgcreate #{volname} `losetup -j #{fname} | cut -f1 -d:`"
-    not_if "vgs #{volname}"
+    not_if { `vgs "#{volname}"`.include?(volname) }
```
jsuchome: Can't you use `volume_exists` function here?

rhafer:
> not_if executes ruby code so the previous condition always returns true.

Actually that's not quite true. If the argument to ```not_if``` is a string chef tries to execute that as a shell command. It's only evaluating ruby code if the argument is a ruby block. (https://docs.chef.io/resource_common.html#guards)

sayalilunkad:
@rhafer but does `not_if` actually translate exit status 0 to true and the rest to false? Because doesn't seem so here https://ci.suse.de/job/openstack-mkcloud/29707/artifact/.artifacts/mkcloud_short_stdout.log

rsalevsky:
@rhafer We had a similar problem in the cookbook for the crowbar postgresql database. I think the problem is that chef can always execute the bash so it's always true. For executing the `not_if` code, chef uses it's build in ShellOut which is based on Open3. And as long Open3 can execute the code it returns true so the exit code is ignored.

See:
```ruby
irb(main):008:0> puts "true" if Open3.popen3("vgs not_existing")
true
```

rhafer:
> We had a similar problem in the cookbook for the crowbar postgresql database. I think the problem is that chef can always execute the bash so it's always true.

Which would be a pretty severe bug in chef (the docs clearly state otherwise). I somehow doubt, be because we rely on the documented behaviour in many other places.

rhafer:
> We had a similar problem in the cookbook for the crowbar postgresql database. I think the problem is that chef can always execute the bash so it's always true

Actually the conditional.rb code looks for the exit code of the programm. This is pretty much what it does:

```
require 'chef/mixin/shell_out'
include Chef::Mixin::ShellOut
shell_out("/bin/false").status.success?
=> false
shell_out("/bin/true").status.success?                                                                                                                             
=> true
```

So there must be something else that's going wrong. Not sure yet what.

MaximilianMeister:
It seems that the exit code is just ignored here:
https://github.com/chef/chef/blob/master/lib/chef/mixin/command.rb#L133 (not sure if that is in our old chef version as well)

but as @rsalevsky mentioned we had the issue in crowbar-init where I had to extract the exit code like this https://github.com/crowbar/crowbar-init/blob/master/lib/crowbar/init/application.rb#L92-L100

rhafer:
> It seems that the exit code is just ignored here:
> https://github.com/chef/chef/blob/master/lib/chef/mixin/command.rb#L133 (not sure if that is in our old chef version as well)

But ```only_if``` uses a completely different code path. It's not using chef/mixin/command.rb. It using Chef::Mixin::ShellOut which itself will endup calling this: https://github.com/chef/mixlib-shellout/blob/master/lib/mixlib/shellout/unix.rb#L92

sayalilunkad:
Confirmed that the error is due to some other problem and not the `not_if` check so closing this PR. Thanks @rhafer for confirming this!

PR #484, opened 2016-08-05
ActiveRecord::Base.save doesn't take any arguments

MaximilianMeister:
this failed in one of the CI jobs with:
(wrong number of arguments (1 for 0)

this came to light after introducing:
  https://github.com/crowbar/crowbar-core/pull/582
which is not yet merged but present as a patch in the package

### DIFF
--- a/crowbar_framework/app/models/cinder_service.rb
+++ b/crowbar_framework/app/models/cinder_service.rb
@@ -233,12 +233,10 @@ class CinderService < PacemakerServiceObject
       end
     end
     if dirty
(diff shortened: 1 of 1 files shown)


jsuchome:
So do we add this one to package as well, or rather revert the previous one and wait until both pass gating and get properly merged?

MaximilianMeister:
@jsuchome actually there is a bug in the package patch... it should be `save(*args)`

http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-save

PR #485, opened 2016-08-08
Set the trove service password to random

s-t-e-v-e-n-k:
Force the trove service password to be random_password in the
model, to match other services.

### DIFF
--- a/chef/data_bags/crowbar/migrate/trove/101_add_service_details.rb
+++ b/chef/data_bags/crowbar/migrate/trove/101_add_service_details.rb
@@ -1,0 +1,16 @@
+def upgrade(ta, td, a, d)
+  # Use a class variable, since migrations are run twice.
+  unless defined?(@@trove_service_password)
+    service = ServiceObject.new "fake-logger"
+    @@trove_service_password = service.random_password
+  end
+  a["service_user"] = ta["service_user"]
+  a["service_password"] = @@trove_service_password
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("service_user")
+  a.delete("service_password")
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-trove.json
+++ b/chef/data_bags/crowbar/template-trove.json
@@ -11,6 +11,7 @@
       "cinder_instance": "none",
       "rabbitmq_instance": "none",
       "volume_support": false,
+      "service_user": "trove",
       "db": {
         "password": "",
         "user": "trove",
@@ -27,7 +28,7 @@
     "trove": {
       "crowbar-revision": 1,
       "crowbar-applied": false,
-      "schema-revision": 100,
+      "schema-revision": 101,
       "element_states": {
         "trove-server": [ "readying", "ready", "applying" ]
       },
--- a/chef/data_bags/crowbar/template-trove.schema
+++ b/chef/data_bags/crowbar/template-trove.schema
@@ -20,6 +20,8 @@
             "swift_instance": { "type": "str", "required": true },
             "rabbitmq_instance": { "type": "str", "required": true },
             "volume_support": { "type": "bool", "required": true },
+            "service_user": { "type": "str", "required": true },
+            "service_password": { "type": "str" },
             "db": {
               "type": "map",
               "required": true,
--- a/crowbar_framework/app/models/trove_service.rb
+++ b/crowbar_framework/app/models/trove_service.rb
@@ -45,6 +45,7 @@ class TroveService < ServiceObject
     base["attributes"][@bc_name]["swift_instance"] = find_dep_proposal("swift", true)
     base["attributes"][@bc_name]["rabbitmq_instance"] = find_dep_proposal("rabbitmq")
     base["attributes"][@bc_name]["db"]["password"] = random_password
+    base["attributes"][@bc_name]["service_password"] = random_password
 
     # assign a default node to the trove-server role
     nodes = NodeObject.all


rhafer:
I think we should also create a schema migration in https://github.com/crowbar/crowbar-openstack/tree/master/chef/data_bags/crowbar/migrate/trove that, if the current pw is set to ```trove``` resets it to a random password.

tboerger:
And the bot should exclude former team mates.

s-t-e-v-e-n-k:
@rhafer I will investigate how feasible that is.

chef/data_bags/crowbar/template-trove.json
```diff
@@ -25,7 +26,7 @@
   },
   "deployment": {
     "trove": {
-      "crowbar-revision": 1,
+      "crowbar-revision": 2,
       "crowbar-applied": false,
       "schema-revision": 100,
```
dirkmueller: schema-revision needs to be bumped to 101

chef/data_bags/crowbar/template-trove.json
```diff
@@ -25,7 +26,7 @@
   },
   "deployment": {
     "trove": {
-      "crowbar-revision": 1,
+      "crowbar-revision": 2,
```
dirkmueller: that doesn't exist, it should stay with 1

chef/data_bags/crowbar/migrate/trove/101_add_service_user.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["service_user"] = ta["service_user"]
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("service_user")
+  return a, d
```
dirkmueller: service_password needs to be deleted as well

chef/data_bags/crowbar/migrate/trove/101_add_service_user.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["service_user"] = ta["service_user"]
+  return a, d
```
dirkmueller: service_password needs to be initialized to something sensible here

chef/data_bags/crowbar/migrate/trove/101_add_service_user.rb
```diff
@@ -0,0 +1,9 @@
...
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("service_user")
```
dirkmueller: do we need to back port this to cloud 6 as this smells like a security bug?

rhafer:
Please add the bsc reference to the summary line of the commit message. That way our scripts will catch it for the changelog of the RPM packages.

rhafer:
BTW, for the generation of the password in the schema migration feel free to get some inspiration from: https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/migrate/glance/004_add_db.rb

It's a little tricky since the migration is usually run two times on systems where trove is already deployed (1st for the Proposal, stored in crowbar's datbase, 2nd for the chef-role that holds the currently deployed configuration). The password needs to be set to the same value for both runs, that's why in class variable is needed there. (I think you can ignore the stuff dealing with node attribute from the above reference migration)

chef/data_bags/crowbar/migrate/trove/101_add_service_user.rb
rhafer: > do we need to back port this to cloud 6 as this smells like a security bug?

Yes, definitely.

dirkmueller:
@s-t-e-v-e-n-k ping? :)

chef/data_bags/crowbar/template-trove.schema
```diff
@@ -20,6 +20,8 @@
             "swift_instance": { "type": "str", "required": true },
             "rabbitmq_instance": { "type": "str", "required": true },
             "volume_support": { "type": "bool", "required": true },
+            "service_user": { "type": "str", "required": true },
+            "service_password": { "type": "str" },
```
Itxaka: service_password in on the schema but not on the json. Shouldn't it be?

Itxaka:
Apart from the small nit, looks good.

Itxaka:
retriggering jenkins as it seems that the job never posted back :/

nicolasbock:
+1

chef/data_bags/crowbar/migrate/trove/101_add_service_details.rb
```diff
@@ -0,0 +1,22 @@
...
+  unless defined?(@@trove_service_password)
+    service = ServiceObject.new "fake-logger"
+    @@trove_service_password = service.random_password
+  end
+  Chef::Search::Query.new.search(:node) do |node|
+    unless defined?(node[:trove])
+      node[:trove][:service_password] = @@trove_service_password
+      node.save
+    end
+  end
```
vuntz: I don't think we need to change the nodes; the migration will change the role for the proposals, so the nodes will automatically get the attribute.

chef/data_bags/crowbar/template-trove.schema
s-t-e-v-e-n-k: We can't set it to a reasonable value in the JSON, since we want it to a random password.

s-t-e-v-e-n-k:
Thanks for the review @vuntz. You're right that the node setting isn't required. I added the code since during my testing, I saw the updated password in the crowbar database, but not in the output of crowbar machine show. I've re-tested it without that chunk of code, and confirmed that it does show up in crowbar machine show after chef-client has run, so I've removed it.

vuntz:
Thanks, +1

Itxaka:
Looks good, thanks for addressing my comment +1

Itxaka:
retriggered the job :)

PR #486, opened 2016-08-08
glance: Update glance config templates for Mitaka

toabctl:
Update the configuration templates for glance

### DIFF
--- a/chef/cookbooks/glance/templates/default/glance-api.conf.erb
+++ b/chef/cookbooks/glance/templates/default/glance-api.conf.erb
@@ -67,8 +67,8 @@
 # For example, if using the file system store a URL of
 # "file:///path/to/image" will be returned to the user in the
 # 'direct_url' meta-data field. Revealing storage location can be a
-# security risk, so use this setting with caution!  The overrides
-# show_image_direct_url. (boolean value)
+# security risk, so use this setting with caution! Setting this to
+# true overrides the show_image_direct_url option. (boolean value)
 #show_multiple_locations = false
 show_multiple_locations = <%= node[:glance][:show_storage_location] %>
 
@@ -95,10 +95,6 @@ show_multiple_locations = <%= node[:glance][:show_storage_location] %>
 # Deploy the v2 OpenStack Images API. (boolean value)
 #enable_v2_api = true
 
-# Deploy the v3 OpenStack Objects API. (boolean value)
-#enable_v3_api = false
-enable_v3_api = <%= node[:glance][:enable_v3_api] %>
-
 # Deploy the v1 OpenStack Registry API. (boolean value)
 #enable_v1_registry = true
 
@@ -110,8 +106,8 @@ enable_v3_api = <%= node[:glance][:enable_v3_api] %>
 #pydev_worker_debug_host = <None>
 
 # The port on which a pydev process is listening for connections.
-# (integer value)
-# Minimum value: 1
+# (port value)
+# Minimum value: 0
 # Maximum value: 65535
 #pydev_worker_debug_port = 5678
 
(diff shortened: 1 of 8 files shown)


sayalilunkad:
Looks good, except maybe you want to remove the `use_syslog` as well since it is deprecated?

toabctl:
> Looks good, except maybe you want to remove the use_syslog as well since it is deprecated?

It's not deprecated. See https://git.openstack.org/cgit/openstack/oslo.log/tree/oslo_log/_options.py?h=master#n87 .

toabctl:
for the record: the option ```use_syslog``` is not deprecated. The release notes are wrong

sayalilunkad:
@toabctl yup, thanks for confirming that. +1

chef/cookbooks/glance/templates/default/glance-api.conf.erb
```diff
@@ -266,39 +257,80 @@ registry_port = <%= @registry_bind_port %>
...
+# Reason: This option was considered harmful and has been deprecated
+# in M release. It will be removed in O release. For more information
+# read OSSN-0060. Related functionality with uploading big images has
+# been implemented with Keystone trusts support.
 #auth_region = <None>
 
 # The protocol to use for communication with the registry server.
 # Either http or https. (string value)
 #registry_client_protocol = http
-registry_client_protocol = http
```
rhafer: I know it's nothing new, but shouldn't we set this to ```https``` when ssl is enabled? Would of course also mean to enable SSL support for the glance registry service. Is that something we should track in the backlog?

chef/cookbooks/glance/templates/default/glance-api.conf.erb
```diff
@@ -226,15 +216,16 @@ key_file = <%= node[:glance][:ssl][:keyfile] if node[:glance][:api][:protocol] =
 #image_cache_driver = sqlite
 
 # The upper limit (the maximum size of accumulated cache in bytes)
-# beyond which pruner, if running, starts cleaning the images cache.
-# (integer value)
+# beyond which the cache pruner, if running, starts cleaning the image
+# cache. (integer value)
 #image_cache_max_size = 10737418240
```
rhafer: I think this needs to be ```<%= node[:glance][:image_cache_max_size] %>``` (the same value as in glance-cache.conf), see http://docs.openstack.org/developer/glance/cache.html#configuration-options-for-the-image-cache

chef/cookbooks/glance/templates/default/glance-api.conf.erb
```diff
@@ -226,15 +216,16 @@ key_file = <%= node[:glance][:ssl][:keyfile] if node[:glance][:api][:protocol] =
...
 
-# The amount of time to let an image remain in the cache without being
-# accessed. (integer value)
+# The amount of time to let an incomplete image remain in the cache,
+# before the cache cleaner, if running, will remove the incomplete
+# image. (integer value)
 #image_cache_stall_time = 86400
```
rhafer: This also needs to have the same value as in glance-cache.conf

toabctl: @rhafer No. The value is not used when running the glance-api service.

chef/cookbooks/glance/templates/default/glance-api.conf.erb
toabctl: Done

chef/cookbooks/glance/templates/default/glance-api.conf.erb
toabctl: Done

chef/cookbooks/glance/templates/default/glance-api.conf.erb
toabctl: I created https://trello.com/c/nMBPFtws/1129-enable-https-for-glance-registry

rhafer:
+1

sayalilunkad:
Still +1 for me. Gating fail was unrelated so I rebuild it. Let's wait to see if it passes.

PR #487, opened 2016-08-09
Fix lvm loopback related race (upgrade related)

rhafer:
This should fix the current CI failures in the 6->7 upgrade jobs


aplanas:
+1

toabctl:
+1

rhafer:
Please ignore the result of the upgrade6to7 ci job, this can't currently work as:

1. There is a race in mkcloud that sometimes causes the backup to fail after installing updates because the crowbar service is not restarted fast enough (patch for that is on it's way).

2. mkcloud installs the patched packages from this PR already into the cloud 6 deployment. It would need to be installed only on the cloud 7 deployment before re-applying any proposals, which is currently not possible, I'll create a card on the backlog for that.

toabctl:
+1

aplanas:
+1

PR #488, opened 2016-08-10
horizon: Enable password retrieve option

toabctl:
When launching windows instances, there is no ssh and no public key can be
injected via cloud-init. But cloud-init for windows can generate a password
and expose that. The nova CLI already can get th password via
"nova get-password" and the same option is now enabled via the web UI.
Note: The password is sill encrypted with the public ssh key so when getting the
password you need to provide the private ssh key to decrypt the password.


aplanas:
/th/the, /sill/still in the commit message

toabctl:
> /th/the, /sill/still in the commit message

fixed :)

jsuchome:
Have you thought about making it configurable?

aplanas:
+1

toabctl:
> Have you thought about making it configurable?

@jsuchome Yes, but I don't see a reason why. The same thing is already doable via the CLI (```nova get-password```). Or why would we have this configurable?

jsuchome:
+1

Itxaka:
+1

sayalilunkad:
+1

dirkmueller:
+1 :-)

PR #491, opened 2016-08-10
horizon: Enable password retrieve option

toabctl:
When launching windows instances, there is no ssh and no public key can be
injected via cloud-init. But cloud-init for windows can generate a password
and expose that. The nova CLI already can get the password via
"nova get-password" and the same option is now enabled via the web UI.
Note: The password is still encrypted with the public ssh key so when getting the
password you need to provide the private ssh key to decrypt the password.

(cherry picked from commit f1d0687aaa49ee484bc2a9ac3aba40be404d7c32)


dirkmueller:
+1

PR #494, opened 2016-08-11
Updating magnum template for Mitaka

sayalilunkad:


jgrassler:
+1

chef/cookbooks/magnum/templates/default/magnum.conf.erb
```diff
@@ -658,18 +651,6 @@ region_name = <%= @keystone_settings['endpoint_region'] %>
...
-
-# Optional PEM-formatted certificate chain file. (string value)
-#cert_file = <None>
-
-# Optional PEM-formatted file that contains the private key. (string value)
-#key_file = <None>
-
-# If set, then the server's certificate will not be verified. (boolean value)
-#insecure = false
-
```
rhafer: @sayalilunkad What happened to those options? AFAICS they're still valid for mitaka.

chef/cookbooks/magnum/templates/default/magnum.conf.erb
```diff
@@ -1340,6 +1320,9 @@ trustee_domain_admin_password = <%= @trustee["domain_admin_password"] %>
 # The roles which are delegated to the trustee by the trustor (list value)
 #roles =
 
+# If set, then the server's certificate will not be verified. (boolean value)
+#insecure = false
+
```
rhafer: I guess this one needs be setting depending on keystone's ssl usage?

chef/cookbooks/magnum/templates/default/magnum.conf.erb
jgrassler: They might in fact be, as I also learned yesterday night when I was looking for magnum-api's SSL setting in vain:

```
18:15 < hieulq_> jgrassler: the Oslo config generator at master branch and Mitaka branch are not 
                 configured in the right way
18:15 -!- zenoway [~zenoway@46.39.36.133] has joined #openstack-containers
18:16 < hieulq_> jgrassler: so tox -e genconfig will generate the sample file that contain not enough config options
```

Which is to say: they're gone from the sample configuration file generated using `tox -egenconfig`, but that may not signify much given that IRC quote...

jgrassler: No, these settings are gone indeed:

https://github.com/openstack/magnum/blob/stable/mitaka/magnum/common/clients.py#L62

I'll try and figure out why...

jgrassler: Correction. They're not gone, but they are now being registered in a manner that hides them from oslo-config-generator:

https://github.com/openstack/magnum/commit/07285d24247bac55146a18f21263441d9389d02c

sayalilunkad: Is it intentional that it is hidden or is it a bug?  And I can just add it back for now.

sayalilunkad: Also @rhafer @jgrassler these haven't been set in the past so I guess they haven't been used, so do we still want to have then in the file?

jgrassler: Definitely a bug in Magnum's `oslo-config-generator` configuration and/or the way Magnum registers these options since https://github.com/openstack/magnum/commit/07285d24247bac55146a18f21263441d9389d02c . So yes, please add

* `ca_file`
* `cert_file`
* `key_file`
* `insecure`

for the following sections:

`[heat_client]`
`[glance_client]`
`[neutron_client]`
`[nova_client]`

jgrassler: If our CI ever runs Magnum tempest tests in a cloud with `want_ssl=1` (I don't think we're currently doing that) we'll probably need these: that CI cloud will have self-signed certificates and HTTPS URLs in the service catalog. For that to work, Magnum will need the CA certificates the service catalog URLs use in its catalog. So basically we'll need these settings in any testing environment that involves SSL for the services Magnum talks to.

jgrassler: At a minimum we'll need the `insecure` setting to disable certificate validation in testing environments.

rhafer:
+1

jgrassler:
+1

PR #495, opened 2016-08-12
Remove the postgresql dump restore step when upgrading from 3.0

rhafer:
There is no postgresql version upgrade needed currently (so we can just
continue to work with the old database files). And the previous code was very
SLE11 specific. (e.g. the check for postgresql 9.1, broke the restore).


rhafer:
This should go in together with https://github.com/crowbar/crowbar-core/pull/592

jsuchome:
Should we remove `db_dump_location` from the schemas? Or rather leave it there for possible future reuse...

chef/cookbooks/postgresql/recipes/db_restore.rb
```diff
@@ -1,47 +0,0 @@
-#
```
sayalilunkad: I think we should leave this recipe as it is right now till there is another way of doing this. For now it should be enough not to include it.

rhafer: What for? If we ever need it again it's easy to restore it from git. Leaving it lying around unused will just cause confusion IMO.

rhafer:
> Should we remove db_dump_location from the schemas? Or rather leave it there for possible future reuse...

db_dump_location is an attribute of the crowbar barclamp. So if it needs to go away it should happen there.

jsuchome:
+1

rsalevsky:
+1

chef/cookbooks/postgresql/recipes/db_restore.rb
sayalilunkad: ok, fair enough. +1

PR #497, opened 2016-08-12
Cinder nfs backend 3.0

nicolasbock:
Backport to Cloud 6.

https://github.com/crowbar/crowbar-openstack/pull/469

### DIFF
--- a/chef/data_bags/crowbar/template-cinder.json
+++ b/chef/data_bags/crowbar/template-cinder.json
@@ -72,6 +72,10 @@
           "pool": "",
           "iscsi_ip": ""
         },
+        "nfs": {
+          "nfs_shares": "",
+          "nfs_mount_options": ""
+        },
         "rbd": {
           "use_crowbar": true,
           "config_file": "/etc/ceph/ceph.conf",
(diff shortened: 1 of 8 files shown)


toabctl:
@nicolasbock we usually wait for 2 +1 comments before we merge things.

nicolasbock:
Sorry about that. I only looked at the "passed tests" which was green. Doesn't that usually count the number of +1's?

chef/data_bags/crowbar/template-cinder.json
```diff
@@ -131,7 +135,7 @@
     "cinder": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 47,
+      "schema-revision": 101,
```
Itxaka: @nicolasbock Is this correct? As its a backport to cloud6, shouldnt it follow the cloud6 numbering and be 48?

toabctl: @Itxaka right. the commit is wrong. Let's revert it and create a new PR.

toabctl:
> Sorry about that. I only looked at the "passed tests" which was green. Doesn't that usually count the number of +1's?

I think there is the lgtm service which counts the +1's but tbh I wouldn't trust that. It also depends when somebody added the +1 (maybe before another force push or so).

nicolasbock:
Yes, good point.

PR #499, opened 2016-08-12
nova: Start virtlogd also for SLES 12.1 platform

toabctl:
We recently added newer libvirt packages to our CI and we still use
SLES12SP1 so we need to start virtlogd now also for suse-12.1.

This commit should be reverted once we did the switch to SLE12SP2.


dirkmueller:
wow, good point :(

dirkmueller:
+1

jsuchome:
I'm surprised that the new package affects SP1, I thought PTF is for SP2 only.

Anyway, +1 for now

PR #502, opened 2016-08-15
Updates nova config template for Mitaka

sayalilunkad:


toabctl:
# nova.conf.erb
## bindir

should be ```/usr/bin``` instead of ```/usr/local/bin```

## use_neutron
should be set to ```true```

## network_api_class
default is ok (default is ```nova.network.api.API```) because ```use_neutron``` is the value we need to set

## memcached_servers
Moved to ```[cache]memcache_servers``` . Maybe the format changed...

## cinder_insecure
This is no longer in the config file. Maybe a upstream bug and not in the config sample? Needs to be verified

sayalilunkad:
>bindir

>should be /usr/bin instead of /usr/local/bin

>use_neutron

>should be set to true

>network_api_class

>default is ok (default is nova.network.api.API) because use_neutron is the value we need to set

Done
>memcached_servers

>Moved to [cache]memcache_servers . Maybe the format changed...

Added to [cache] section
>cinder_insecure

>This is no longer in the config file. Maybe a upstream bug and not in the config sample? Needs to be >verified

This needs to be fixed upstream. Created a bug report https://bugs.launchpad.net/nova/+bug/1615666

toabctl:
@sayalilunkad You need to rebase this patch...

toabctl:
+1

aplanas:
[cinder]
missing "insecure = <%= @cinder_insecure ? 'True' : 'False' %>"

sayalilunkad:
@aplanas thanks, done!

aplanas:
The s3 and ec2 information where is stored in mitaka? Or is in no use anymore?

sayalilunkad:
@aplanas that has it's own package now, openstack-ecp-api iirc.

aplanas:
@sayalilunkad makes sense, thanks!

+1

PR #504, opened 2016-08-16
cinder: Allow to use a HA cluster for cinder-volume in active/passive

Itxaka:
We cannot do active/active at the moment because there needs to be some
synchonization between the various volume services in that case.

But it turns out that active/passive is relatively easy to achieve: it's
only a matter of making sure that the cinder-volume service on the
various cluster members register with the same host. We use the
vhostname of the cluster for this.

Of course, it doesn't work with the iSCSI backend, so we have validation
checks for this in Crowbar.

Another restriction is that we cannot use the same cluster for
cinder-controller and cinder-volume: due to the use of the same host in
the config, if we were to allow cinder-controller there too, only one
cinder-scheduler service would be register (with the vhostname of the
cluster), which is wrong in this case.

(cherry picked from commit 0c87c97f8ed9dec479102ceb6e965c7208a5ae4f)

### DIFF
--- a/chef/cookbooks/cinder/attributes/default.rb
+++ b/chef/cookbooks/cinder/attributes/default.rb
@@ -49,9 +49,11 @@ default[:cinder][:ha][:enabled] = false
 if %w(rhel suse).include? node[:platform_family]
   default[:cinder][:ha][:api_ra] = "lsb:openstack-cinder-api"
   default[:cinder][:ha][:scheduler_ra] = "lsb:openstack-cinder-scheduler"
+  default[:cinder][:ha][:volume_ra] = "lsb:openstack-cinder-volume"
 else
   default[:cinder][:ha][:api_ra] = "lsb:cinder-api"
   default[:cinder][:ha][:scheduler_ra] = "lsb:cinder-scheduler"
+  default[:cinder][:ha][:volume_ra] = "lsb:cinder-volume"
 end
 default[:cinder][:ha][:op][:monitor][:interval] = "10s"
 # Ports to bind to when haproxy is used for the real ports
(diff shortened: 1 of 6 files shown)


Itxaka:
synced to latest commit from https://github.com/crowbar/crowbar-openstack/pull/459

Itxaka:
ceilometer proposal failed on ci, retriggering test job

vuntz:
+

aspiers:
+1

Itxaka:
rebased to stable/3.0 to see if CI now passes :/

aspiers:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;QkpcjlzL&#x2F;16-p1-investigate-cinder-volume-active-passive-for-sap"> P1: Investigate cinder-volume active&#x2F;passive for SAP</a>

aplanas:
I guess that we can merge this, and refactor the overwride of the config file together with the rest

+1

aspiers:
Gated failed with

    /bin/systemctl start openstack-neutron-metadata-agent returned 1, expected 0

Itxaka:
Service `openstack-neutron-metadata-agent` seems to time out while starting :/

aplanas:
@aspiers The fix was merged this morning

Itxaka:
@aplanas still failing. Was the fix backported? Can you link to the fix/card?

aplanas:
@Itxaka In the case of ceilometer, this was the fix: https://build.opensuse.org/request/show/424201

There is not PIDFile. Most probably gating is taking the old package (maybe we need to wait until the ISO is updated?)

Itxaka:
thanks @aplanas, so lets wait until monday for triggering this again :D

PR #507, opened 2016-08-17
Cinder migrate 3.0

nicolasbock:


chef/data_bags/crowbar/migrate/cinder/048_nfs_backend.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["volume_defaults"]["nfs"] = ta["volume_defaults"]["nfs"]
```
Itxaka: FYI: As discussed on https://github.com/crowbar/crowbar-core/pull/598 migrations that are backported should be modified on master to protect against being run several times and overwriting data :)

nicolasbock: Yes, It seemed a bit odd at first, but now I get it :smile:

Itxaka:
LGTM +1

jsuchome:
Why was it reverted earlier?

nicolasbock:
In the original PR, https://github.com/crowbar/crowbar-openstack/pull/497, I had gotten the revision for migration wrong. @toabctl suggested to revert the PR, fix the revision and resubmit.

jsuchome:
Ah, true, this is 3.0 branch. +1

PR #508, opened 2016-08-17
Neutron: Update config template for Mitaka

sayalilunkad:


chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -1,850 +1,1012 @@
...
+use_ssl = <%= @ssl_enabled ? "True" : "False" %>
+
+# Seconds between running periodic tasks (integer value)
+#periodic_interval = 40
+
+# Number of separate API worker processes for service. If not specified, the
+# default is equal to the number of CPUs available for best performance.
+# (integer value)
+#api_workers = <None>
+api_workers = 0
```
rhafer: I wonder if we should try running multiple workers again (like we do for all the other services). AFAICS a couple of thing related to that have been fixed meanwhile. But we could as well handle that later. Could you create an item on the backlog for that.

(For ```rpc_workers``` and ```rpc_state_report_works``` we might want to stay with the default value 1 in that case)

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -1,850 +1,1012 @@
...
-# How frequently to retry connecting with RabbitMQ. (integer
-# value)
-# rabbit_retry_interval=1
 
-# How long to backoff for between retries when connecting to
-# RabbitMQ. (integer value)
-# rabbit_retry_backoff=2
+[agent]
+root_helper = sudo neutron-rootwrap /etc/neutron/rootwrap.conf
+root_helper=sudo neutron-rootwrap /etc/neutron/rootwrap.conf
```
rhafer: Duplicate line. Please remove the 2nd one

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -871,9 +1033,9 @@ endpoint_type = internal
 # Authentication URL (unknown value)
 #auth_url = <None>
 auth_url = <%= KeystoneHelper.versioned_service_URL(@keystone_settings["protocol"],
-               @keystone_settings["internal_url_host"],
-               @keystone_settings["service_port"],
-               "2.0") %>
+                                    @keystone_settings["internal_url_host"],
+                                    @keystone_settings["service_port"],
+                                    "2.0") %>
```
rhafer: AFAICS neutron is now able to use the v3 API.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -1,850 +1,1012 @@
...
+# root endpoint e.g. https://localhost:35357/ (string value)
+#identity_uri = <None>
 identity_uri = <%= @keystone_settings['admin_auth_url'] %>
 
-auth_version = <%= @keystone_settings['api_version_for_middleware'] %>
-admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
+# This option is deprecated and may be removed in a future release. Single
+# shared secret with the Keystone configuration used for bootstrapping a
+# Keystone installation, or otherwise bypassing the normal authentication
+# process. This option shou
```
rhafer: The diff between the packaged config files and the template has again a lot those ```(unknown value)``` things in it.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
sayalilunkad: https://trello.com/c/gX8NK4yn/1907-run-multiple-api-worker-in-neutron

sayalilunkad:
@rhafer I changed the auth_url to un-versioned, that should use v3 by default now AFAIK.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
sayalilunkad: Ah that's probably because we added the patch to oslo.config after I created this template from our package. Let me see if I can fix that easily.

sayalilunkad: Fixed.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -1,850 +1,1011 @@
...
 
-# Send notifications to nova when port data (fixed_ips/floatingips) change
-# so nova can update it's cache.
-# notify_nova_on_port_data_changes = True
+# Maximum number of allowed address pairs (integer value)
+#max_allowed_address_pair = 10
 
-# URL for connection to nova (Only supports one nova region currently).
-# nova_url = http://127.0.0.1:8774/v2
-nova_url = <%= @nova_url if @nova_url %>
```
rhafer: It seems ```@nova_url¸¸``` is not used anymore. It should be removed from the template resource in ```common_config.rb```

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -912,194 +1070,163 @@ insecure = <%= @nova_insecure %>
...
 
-# User id (unknown value)
+# User id (string value)
 #user_id = <None>
 
-# Username (unknown value)
+# Username (string value)
 # Deprecated group/name - [DEFAULT]/username
 #username = <None>
 username = <%= @nova_admin_username if @nova_admin_username %>
```
rhafer: BTW, I think we can make the code in common_config.rb a lot simpler by just using the neutron service credentials here (instead of nova's). Since neutron now looks up the nova endpoint in keystone (instead of needing it supplied in the config) we could get rid of the search for the nova node (which is a bit ugly since it only really work after nova is deployed)

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -1,850 +1,1011 @@
...
+#min_pool_size = 1
+min_pool_size = <%= @sql_min_pool_size %>
 
-# Number of security group rules allowed per tenant. A negative value means
-# unlimited.
-# quota_security_group_rule = 100
+# Maximum number of SQL connections to keep open in a pool. (integer value)
+# Deprecated group/name - [DEFAULT]/sql_max_pool_size
+# Deprecated group/name - [DATABASE]/sql_max_pool_size
+#max_pool_size = <None>
```
rhafer: It seems the default was change here from 10 to unlimited. Is this true? And if yes, we might want to make this configurable.

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
```diff
@@ -1,126 +1,226 @@
...
+#
+# From neutron.dhcp.agent
+#
 
-# Allow overlapping IP (Must have kernel build with CONFIG_NET_NS=y and
-# iproute2 package that supports namespaces). This option is deprecated and
-# will be removed in a future release, at which point the old behavior of
-# use_namespaces = True will be enforced.
-# use_namespaces = True
-use_namespaces = <%= @use_namespaces %>
```
rhafer: Seems that ```@use_namespace``` can be removed from the  template resource in the recipe.

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
```diff
@@ -1,148 +1,272 @@
...
+
+# Iptables mangle mark used to mark ingress from external network. This mark
+# will be masked with 0xffff so that only the lower 16 bits will be used.
+# (string value)
+#external_ingress_mark = 0x2
+
+# Name of bridge used for external network traffic. This should be set to an
+# empty value for the Linux Bridge. When this parameter is set, each L3 agent
+# can be associated with no more than one external network. This option is
+# deprecated and will be removed in the M release. (string value)
```
rhafer: Hm, interesting comment in a config file supposed to be for the M release :). But I guess that's an upstream issue?

chef/cookbooks/neutron/templates/default/metering_agent.ini.erb
```diff
@@ -1,24 +1,115 @@
...
 
-# interface_driver = neutron.agent.linux.interface.OVSInterfaceDriver
+# The driver used to manage the virtual interface. (string value)
+#interface_driver = <None>
 interface_driver = <%= @interface_driver %>
 
-# This option is deprecated and will be removed in a future release,
-# at which point the old behavior of use_namespaces = True will be enforced.
-# use_namespaces = True
-use_namespaces = <%= @use_namespaces %>
```
rhafer: ```use_namespaces``` is gone so let's remove it from the recipe.

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
sayalilunkad: Done.

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
```diff
@@ -1,76 +1,210 @@
+[DEFAULT]
+
```
rhafer: Since we now introduce the ```[DEFAULT]``` section here shouldn't we set all the bit that we set for the other configs in the default section ```debug```, ```verbose```, ```use_syslog``` and so on?

OTOH, the agents are also including ```neutron.conf``` so in theory this shouldn't be needed. But then I'd prefer to not have this section at all if possible.

@toabctl Do you have any opinion on this?

rhafer: BTW, this is also true for the ovs-agent's config file and ml2_conf.ini it seems.

chef/cookbooks/neutron/templates/default/metering_agent.ini.erb
sayalilunkad: done.

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -81,28 +81,15 @@
...
 
-#TODO: nova should depend on neutron, but neutron also depends on nova
-# so we have to do something like this
-novas = search(:node, "roles:nova-controller") || []
-if novas.length > 0
-  nova = novas[0]
-  nova = node if nova.name == node.name
+# Get Nova's insecure setting
+nova = search(:node, "roles:nova-controller") || []
+if !nova.length.empty?
```
nicolasbock: Shouldn't that simply be `nova.empty?`?

chef/data_bags/crowbar/migrate/neutron/101_add_max_pool_size.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  a["sql"]["max_pool_size"] = ta["sql"]["max_pool_size"]
```
nicolasbock: We should protect here in case this key already exists
~~~
a["sql"]["max_pool_size"] = ta["sql"]["max_pool_size"] unless a["sql"].has_key?("max_pool_size")
~~~

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
```diff
@@ -1,76 +1,210 @@
...
-# Example: firewall_driver = neutron.agent.linux.iptables_firewall.IptablesFirewallDriver
-firewall_driver = neutron.agent.linux.iptables_firewall.IptablesFirewallDriver
-
-# Controls if neutron security group is enabled or not.
-# It should be false when you use nova security group.
-# enable_security_group = True
+# Enable local ARP responder which provides local responses instead of
+# performing ARP broadcast into the overlay. Enabling local ARP responder is
+# not fullycompatible with the allowed-address-pairs extension. (boolean value)
+#arp_responder = false
```
rhafer: This is a new setting in Mitaka. Older releases behaved as if it was set to true, so this will result in a changed behaviour for people upgrading from liberty.

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
```diff
@@ -1,99 +1,184 @@
...
-# metadata_proxy_group is agent effective group,
-# 'all': set metadata proxy socket mode to 0o666, to use otherwise.
-# metadata_proxy_socket_mode = deduce
-
-# Number of separate worker processes for metadata server. Defaults to
-# half the number of CPU cores
-# metadata_workers =
+# Number of separate worker processes for metadata server (defaults to half of
+# the number of CPUs) (integer value)
+#metadata_workers = 4
```
rhafer: Default in our package in 8, or does that depend on the machine the package was build on?

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
```diff
@@ -1,99 +1,184 @@
...
-# Otherwise default_ttl specifies time in seconds a cache entry is valid for.
-# No cache is used in case no value is passed.
-# cache_url = memory://?default_ttl=5
+#
+# From oslo.log
+#
+
+# If set to true, the logging level will be set to DEBUG instead of the default
+# INFO level. (boolean value)
+#debug = falsee
```
rhafer: typo ```falsee```

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
```diff
@@ -1,99 +1,184 @@
...
-# Show debugging output in log (sets DEBUG log level output)
-# debug = True
-debug = <%= @debug ? "True" : "False" %>
-
-# The Neutron user information for accessing the Neutron API.
-# auth_url = http://localhost:5000/v2.0
-auth_url = <%= KeystoneHelper.versioned_service_URL(@keystone_settings["protocol"],
-                                                    @keystone_settings["internal_url_host"],
-                                                    @keystone_settings["service_port"],
-                                                    "2.0") %>
```
rhafer: It seems all the keystone setting vanished from the file. Is that correct,. does the agent really not need those anymore? If yes, you can also cleanup the template resource in the recipe.

This also seems to apply to ```auth_region```

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
```diff
@@ -1,99 +1,184 @@
...
+# The format for an instance UUID that is passed with the log message. (string
+# value)
+#instance_uuid_format = "[instance: %(uuid)s] "
+
+# Enables or disables fatal status of deprecations. (boolean value)
+#fatal_deprecations = false
 
-# Missing in upstream template
-# Ensure that metadata server listening on UNIX socket is not expecting SSL
-use_ssl = False
```
rhafer: Are you sure we can remove this? It was added to fix an issue with the metadata proxy when enabling ssl in neutron. See: https://github.com/crowbar/crowbar-openstack/commit/faa775fce5155501a59432e3c2189caa949f2af2

chef/cookbooks/neutron/templates/default/neutron.conf.erb
sayalilunkad: Done.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
sayalilunkad: Done.

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
```diff
@@ -1,137 +1,247 @@
+[DEFAULT]
+
+#
+# From oslo.log
+#
```
rhafer: I am really unsure what all the options from oslo.log are here for. This file is for configuring the ml2_plugin so it's loaded into the neutron server. Seems quite confusing.

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
```diff
@@ -1,76 +1,210 @@
...
+
+
+[securitygroup]
+
+#
+# From neutron.ml2.linuxbridge.agent
+#
+
+# Driver for security groups firewall in the L2 agent (string value)
+#firewall_driver = <None>
```
rhafer: Aren't we loosing the firewall_driver setting here? We set this to ```neutron.agent.linux.iptables_firewall.IptablesFirewallDriver``` previously.

chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
```diff
@@ -1,212 +1,299 @@
...
+# should be false when using no security groups or using the nova security
+# group API. (boolean value)
+#enable_security_group = true
+
+# Use ipset to speed-up the iptables based security groups. Enabling ipset
+# support requires that ipset is installed on L2 agent node. (boolean value)
+#enable_ipset = true
 
 #-----------------------------------------------------------------------------
 # Sample Configurations.
```
rhafer: It seems the sample section was removed upstream

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
sayalilunkad: I think it depends on the machine because I used our packages to generate this config and I just retried and it is still 4 for me.

chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
sayalilunkad: Yes it was but I thought that was something we injected in the template for reference. If not I can remove it.

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
sayalilunkad: Done.

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
sayalilunkad: Yes these are removed. Updated the resources in the recipe.

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
sayalilunkad: Yes thank, added it back.

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
sayalilunkad: This was a recent addition since they made the configs automatically generated. This didn't seem to be the case in liberty. See this https://github.com/openstack/neutron/commit/71190773e14260fab96e78e65a290356cdc08581

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
sayalilunkad: @rhafer @toabctl If it is really not required I can try to fix it upstream as well. Again was introduced by https://github.com/openstack/neutron/commit/71190773e14260fab96e78e65a290356cdc08581

chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
rhafer: I am pretty sure it's not from us.

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
rhafer: Yeah, I know. I was just wondering if it really makes sense to have that section in ml2_conf.ini. It's just duplicating stuff. AFAIKs all thing that include ml2_conf.ini will also include neutron.conf.
This is quite confusing IMO. (Might it be some upstream bug?)

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
sayalilunkad: Created a patch upstream for fixing this: https://review.openstack.org/#/c/361232/ Will add it to our package and update the template.

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
sayalilunkad: It is removed in the stable/mitaka branch upstream but I think it didn't get into the tarball which we use.

nicolasbock:
+1

rhafer:
BTW, it seems you missed to rebase lbaas_agent.ini.erb. Or doesn't that require any changes?

sayalilunkad:
@rhafer thanks, added it.

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -81,28 +81,15 @@
...
-  nova_api_protocol = nova[:nova][:ssl][:enabled] ? "https" : "http"
-  nova_insecure = keystone_settings["insecure"] || (nova[:nova][:ssl][:enabled] && nova[:nova][:ssl][:insecure])
-
-  nova_notify = {
-    nova_url: "#{nova_api_protocol}://#{nova_api_host}:#{nova[:nova][:ports][:api]}/v2",
-    nova_insecure: nova_insecure,
-    nova_admin_username: nova[:nova][:service_user],
-    nova_admin_password: nova[:nova][:service_password]
-  }
+  nova_insecure = false
```
rhafer: nitpick: I think  ```keystone_settings["insecure"]``` is a bit more correct here.

chef/data_bags/crowbar/migrate/neutron/101_add_max_pool_size.rb
```diff
@@ -0,0 +1,11 @@
...
+  a["sql"]["max_pool_size"] = ta["sql"]["max_pool_size"] unless a["sql"].key?("max_pool_size")
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["sql"].delete("max_pool_size")
+
+  return a, d
+end
```
rhafer: You also need to update the ```schema-revision``` to 101 in ```template-neutron.json```

chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
rhafer: This really seems to be relict from the old (non-generated) upstream config. Please remove it.

chef/data_bags/crowbar/migrate/neutron/101_add_max_pool_size.rb
sayalilunkad: Done. Thanks!

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -1,850 +1,1011 @@
...
+# Verify HTTPS connections. (boolean value)
+#insecure = false
+<% if @keystone_settings['insecure'] -%>
+insecure = <%= @keystone_settings['insecure'] %>
+<% end -%>
 
-# Complete admin Identity API endpoint. This should specify
-# the unversioned root endpoint e.g. https://localhost:35357/
+# The region in which the identity server can be found. (string value)
+#region_name = <None>
```
toabctl: Don't we need to set the ```region_name``` ?

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
toabctl: @sayalilunkad Hm. ```use_ssl``` is still not here.

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
toabctl: > @toabctl Do you have any opinion on this?

if we include multiple config files with the same section, I think the last mentioned config file section wins. So this could be dangerous I think

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
toabctl: For a similar case I filled https://bugs.launchpad.net/neutron/+bug/1620211 (including a patch)

toabctl: For this case I filled https://bugs.launchpad.net/neutron/+bug/1620587

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
```diff
@@ -1,76 +1,212 @@
+[DEFAULT]
+
+#
+# From oslo.log
+#
+
+# If set to true, the logging level will be set to DEBUG instead of the default
+# INFO level. (boolean value)
+#debug = false
```
aplanas: I would set this as in the rest of configuration files

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
aplanas: According to the bug report, this is not needed anymore. Did I understand it correctly?

sayalilunkad: @aplanas yes it isn't. @toabctl  It was only a temporary fix that we did in our config earlier but is not required after  the upstream patch was merged which fixes the problem in the appropriate place rather than the config file.

chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
sayalilunkad: Done.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
rhafer: Yes, we do.

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
rhafer: Meanwhile that change has landed in our packages as well it seems. Could you re-sync this?

aplanas:
After the last changes LGTM +1

The only comment pending from the is the `debug` in the linuxbridge_agent file.

chef/cookbooks/neutron/templates/default/linuxbridge_agent.ini.erb
sayalilunkad: @aplanas See this discussion about all the options. Just waiting for some conclusion on if we should keep it or not.

chef/cookbooks/neutron/templates/default/ml2_conf.ini.erb
rhafer: @toabctl Don't we need the same fix for all the agents as well? AFAICS all of them are also now having the "oslo.log" options added (while still also loading neutron.conf)

rhafer:
Apart from all the oslo.log duplication everywhere I think this is looking good now. I think the oslo.log thing should be handled upstream and then we get that sorted out with newton.

So +1 from me.

toabctl:
+1 (but let's wait for gating)

PR #509, opened 2016-08-18
Added user and admin domain to keystone settings

rhafer:
For now (as we still use v2 to create users) this will always be "Default". But
by defining it here we can avoid having the hardcoded value spread out through
all the other cookbooks.

### DIFF
--- a/chef/cookbooks/keystone/libraries/helpers.rb
+++ b/chef/cookbooks/keystone/libraries/helpers.rb
@@ -48,6 +48,8 @@ module KeystoneHelper
                                                 node[:keystone][:api][:version])
 
       has_default_user = node["keystone"]["default"]["create_user"]
+      default_domain = "Default"
+      default_domain_id = "default"
 
       @keystone_settings ||= Hash.new
       @keystone_settings[cookbook_name] = {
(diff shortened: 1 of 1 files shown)


sayalilunkad:
+1

aplanas:
+1

Itxaka:
can we have default_domain_name AND default_domain_id just in case?

rhafer:
> can we have default_domain_name AND default_domain_id just in case?

Sure. Added.

toabctl:
+1

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -71,10 +73,14 @@ def self.keystone_settings(current_node, cookbook_name)
         "admin_tenant" => node["keystone"]["admin"]["tenant"],
         "admin_tenant_id" => node["keystone"]["admin"]["tenant_id"],
         "admin_user" => node["keystone"]["admin"]["username"],
+        "admin_domain" => default_domain,
+        "admin_domain_id" => "default",
```
Itxaka: why not default_domain_id?

rhafer: *ouch*

rhafer:
> why not default_domain_id?

@Itxaka should be fixed now

aplanas:
Still +1

toabctl:
+1 (and I think we need to fix all ```[keystone_authtoken]``` sections and add the new values)

Itxaka:
+1

PR #511, opened 2016-08-23
Added plumbing for HA to Magnum barclamp.

jgrassler:
This pull request adds HA for magnum-api and magnum-conductor to the Magnum barclamp.

### DIFF
--- a/chef/cookbooks/magnum/attributes/default.rb
+++ b/chef/cookbooks/magnum/attributes/default.rb
@@ -22,6 +22,14 @@ default[:magnum][:max_header_line] = 16_384
 
 default[:magnum][:api][:protocol] = "http"
 
+# HA
 default[:magnum][:ha][:enabled] = false
+# When HAproxy listens on the API port, make service listen elsewhere
+default[:magnum][:ha][:ports][:api] = 5611
+# pacemaker definitions
+default[:magnum][:ha][:api][:op][:monitor][:interval] = "10s"
+default[:magnum][:ha][:api][:agent] = "lsb:openstack-magnum-api"
+default[:magnum][:ha][:conductor][:op][:monitor][:interval] = "10s"
+default[:magnum][:ha][:conductor][:agent] = "lsb:openstack-magnum-conductor"
 
 default[:magnum][:cert][:cert_manager_type] = "local"
--- a/chef/cookbooks/magnum/libraries/helpers.rb
+++ b/chef/cookbooks/magnum/libraries/helpers.rb
@@ -1,0 +1,31 @@
+module MagnumHelper
+  class << self
+    def network_settings(node)
+      @ip ||= Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+      @cluster_admin_ip ||= nil
+
+      if node[:magnum][:ha][:enabled] && !@cluster_admin_ip
+        @cluster_admin_ip = CrowbarPacemakerHelper.cluster_vip(node, "admin")
+      end
+
+      if node[:magnum][:ha][:enabled]
+        bind_host = @ip
+        bind_port = node[:magnum][:ha][:ports][:api].to_i
+      else
+        bind_host = node[:magnum][:api][:bind_open_address] ? "0.0.0.0" : @ip
+        bind_port = node[:magnum][:api][:bind_port].to_i
+      end
+
+      @network_settings ||= {
+        ip: @ip,
+
+        api: {
+          bind_host: bind_host,
+          bind_port: bind_port,
+          ha_bind_host: node[:magnum][:api][:bind_open_address] ? "0.0.0.0" : @cluster_admin_ip,
+          ha_bind_port: node[:magnum][:api][:bind_port].to_i
+        },
+      }
+    end
+  end
+end
--- a/chef/cookbooks/magnum/recipes/api.rb
+++ b/chef/cookbooks/magnum/recipes/api.rb
@@ -18,8 +18,9 @@
 #
 
 keystone_settings = KeystoneHelper.keystone_settings(node, :magnum)
+network_settings = MagnumHelper.network_settings(node)
 
-magnum_port = node[:magnum][:api][:bind_port]
+magnum_port = network_settings[:api][:bind_port]
 magnum_protocol = node[:magnum][:api][:protocol]
 
 ha_enabled = node[:magnum][:ha][:enabled]
--- a/chef/cookbooks/magnum/recipes/ha.rb
+++ b/chef/cookbooks/magnum/recipes/ha.rb
@@ -1,0 +1,94 @@
+# Copyright 2016 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+unless node[:magnum][:ha][:enabled]
+  log "HA support for magnum is not enabled"
+  return
+end
+
+log "Setting up magnum HA support"
+
+network_settings = MagnumHelper.network_settings(node)
+
+ssl_enabled = (node[:magnum][:api][:protocol] == "https")
+
+haproxy_loadbalancer "magnum-api" do
+  address network_settings[:api][:ha_bind_host]
+  port network_settings[:api][:ha_bind_port]
+  use_ssl ssl_enabled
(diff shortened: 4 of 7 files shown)


jgrassler:
Ok, looks like this one is finally running:

```
root@d52-54-77-77-77-01:~ # crm resource list | grep -A 1 magnum
 Clone Set: cl-g-magnum [g-magnum]
     Started: [ d52-54-77-77-77-01 d52-54-77-77-77-02 ]
```

The code is working, but I spent some quality time scratching my head why https://github.com/jgrassler/crowbar-openstack/blob/magnum-ha/crowbar_framework/app/models/magnum_service.rb#L114 didn't do its thing (I got a lot of `INFO: HA support for magnum is not enabled`). Turns out it doesn't get refreshed by `barclamp_install.rb` but takes a Crowbar restart on top of that.

Steps for testing manually:

* Create a regular 4 node HA cloud (I hope this has gotten a bit easier by now. But you can always snatch a finished cloud from a recently finished https://ci.suse.de/view/Cloud/view/Cloud7/job/cloud-mkcloud7-job-ha-linuxbridge-x86_64/ job)
* Apply this pull request to `/opt/dell` (the regular gating run won't do for this since it doesn't build an HA cloud)
* Run
** `barclamp_install.rb /opt/dell/crowbar_framework/barclamps`
** `knife cookbook upload -o /opt/dell/chef/cookbooks magnum`
* **Restart crowbar:** `systemctl restart crowbar`
* Assign the `data` cluster to the `magnum-server` role.

It will have to be tested manually since even a `mkcloud` job with `hacloud=1` won't test it without https://github.com/SUSE-Cloud/automation/pull/1182 enabled.

chef/cookbooks/magnum/attributes/default.rb
```diff
@@ -23,5 +23,14 @@
 default[:magnum][:api][:protocol] = "http"
 
 default[:magnum][:ha][:enabled] = false
+# HA
+default[:magnum][:ha][:enabled] = false
```
vuntz: You already have this attribute two lines earlier ;-)

chef/cookbooks/magnum/attributes/default.rb
```diff
@@ -23,5 +23,14 @@
...
+# When HAproxy listens on the API port, make service listen elsewhere
+default[:magnum][:ha][:ports][:api] = 9611
```
vuntz: So far, we've been trying to use ports in the 5xxx range, iirc. Any reason to not do this here? (I'm afraid that 96xx is sometimes already used by some other openstack service)

chef/cookbooks/magnum/recipes/role_magnum_server.rb
```diff
@@ -17,6 +17,7 @@
 if CrowbarRoleRecipe.node_state_valid_for_role?(node, "magnum", "magnum-server")
   include_recipe "magnum::setup"
   include_recipe "magnum::common"
+  include_recipe "magnum::ha"
```
vuntz: Just a guess, but I would have made it run after api/conductor as I assume that they configure stuff that is needed when starting the services.

chef/cookbooks/magnum/recipes/sql.rb
```diff
@@ -58,6 +58,8 @@
   only_if { !ha_enabled || is_cluster_founder }
 end
 
+crowbar_pacemaker_sync_mark "create-magnum_database"
```
vuntz: When you have this sync mark, you usually need the "wait-magnum_database" somewhere earlier.

chef/cookbooks/magnum/libraries/helpers.rb
```diff
@@ -0,0 +1,30 @@
+module MagnumHelper
+  class << self
+    def network_settings(node)
```
vuntz: I think you need to use this new helper to put the right port and address in the config file, not just for the ha config.

chef/cookbooks/magnum/recipes/ha.rb
```diff
@@ -0,0 +1,94 @@
...
+
+pacemaker_transaction "magnum server" do
+  cib_objects transaction_objects
+  # note that this will also automatically start the resources
+  action :commit_new
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+crowbar_pacemaker_order_only_existing "o-#{clone_name}" do
+  ordering ["postgresql", "rabbitmq", "cl-keystone", clone_name]
```
vuntz: Just double-checking: don't we need to depend on some other services. Like, say, heat? Or nova?

chef/cookbooks/magnum/attributes/default.rb
jgrassler: Oops :-)

Fixed.

chef/cookbooks/magnum/attributes/default.rb
jgrassler: I basically added `100` to the regular API port, a choice I'm not overly attached to. I used `5611` now. Is that ok? It doesn't appear anywhere else so far, whereas the more straightforward choice of `5511` is already taken by `glance-registry`.

chef/cookbooks/magnum/recipes/role_magnum_server.rb
jgrassler: Somehow it worked out, but that does sound a good deal saner. Fixed.

chef/cookbooks/magnum/recipes/sql.rb
jgrassler: It's there, a few lines further up:

https://github.com/jgrassler/crowbar-openstack/blob/63d657afb8e9ca863ffe7e33e937435818be283e/chef/cookbooks/magnum/recipes/sql.rb#L27

That being said, there are two  `create-magnum_database` sync marks, though:

* https://github.com/jgrassler/crowbar-openstack/blob/63d657afb8e9ca863ffe7e33e937435818be283e/chef/cookbooks/magnum/recipes/sql.rb#L61
* https://github.com/jgrassler/crowbar-openstack/blob/63d657afb8e9ca863ffe7e33e937435818be283e/chef/cookbooks/magnum/recipes/sql.rb#L88

I removed the one on line 61.

chef/cookbooks/magnum/libraries/helpers.rb
jgrassler: The port, yes (I think the address is taken care of by CrowbarHelper.get_host_for_admin_url()). I modified `api.rb` to use `MagnumHelper for the port (address is still taken care of by `CrowbarHelper`, which appears to be HA aware).

chef/cookbooks/magnum/recipes/ha.rb
jgrassler: Oh, I missed that one. Yes, we depend on Heat. I added that (if dependencies are transitive that should be sufficient, otherwise we'll have to add most OpenStack services as well...).

chef/cookbooks/magnum/recipes/api.rb
```diff
@@ -18,8 +18,9 @@
 #
 
 keystone_settings = KeystoneHelper.keystone_settings(node, :magnum)
+network_settings = MagnumHelper.network_settings(node)
 
-magnum_port = node[:magnum][:api][:bind_port]
+magnum_port network_settings[:api][:bind_port]
```
vuntz: Missing `=`

vuntz:
You're missing at least an update here: https://github.com/jgrassler/crowbar-openstack/blob/acfe5fd177cdbfa86783560c6507e2f644a3fad3/chef/cookbooks/magnum/recipes/common.rb#L36

chef/cookbooks/magnum/recipes/api.rb
jgrassler: Fixed.

jgrassler:
> You're missing at least an update here: https://github.com/jgrassler/crowbar-openstack/blob/acfe5fd177cdbfa86783560c6507e2f644a3fad3/chef/cookbooks/magnum/recipes/common.rb#L36

Indeed. Turns out all of that logic is in MagnumHelper already. Fixed.

chef/cookbooks/magnum/libraries/helpers.rb
```diff
@@ -0,0 +1,28 @@
...
+      @ip ||= Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+      @cluster_admin_ip ||= nil
+
+      if node[:magnum][:ha][:enabled] && !@cluster_admin_ip
+        @cluster_admin_ip = CrowbarPacemakerHelper.cluster_vip(node, "admin")
+      end
+
+      bind_host = node[:magnum][:api][:bind_open_address] ? "0.0.0.0" : @ip
```
vuntz: This should never be `0.0.0.0` if HA is enabled, see https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/glance/libraries/helpers.rb for an example.

jgrassler: Ok, fixed.

vuntz:
+1.

Hint for other reviewers: compare this to what is done in the glance barclamp for HA.

aplanas:
LGTM +1

PR #512, opened 2016-08-24
Adds HA support for Barbican

sayalilunkad:


chef/cookbooks/barbican/recipes/ha.rb
```diff
@@ -0,0 +1,93 @@
...
+# resources
+crowbar_pacemaker_sync_mark "sync-barbican_before_ha"
+
+# Avoid races when creating pacemaker resources
+crowbar_pacemaker_sync_mark "wait-barbican_ha_resources"
+
+primitives = []
+transaction_objects = []
+
+["worker", "keystone_listener", "api"].each do |service|
```
vuntz: Isn't keystone-listener only running if some attribute is set? (btw, is ist keystone_listener or keystone-listener?)

sayalilunkad: It should be keystone-listener. And yes there is an option enable_keystone_listener. Then I guess we don't do this here?

vuntz: Well, you should do the keystone-listener bit, but based on the value of the option.

chef/cookbooks/barbican/recipes/ha.rb
```diff
@@ -0,0 +1,93 @@
...
+  pacemaker_primitive primitive_name do
+    agent node[:barbican][:ha][service.to_sym][:agent]
+    op node[:barbican][:ha][service.to_sym][:op]
+    action :update
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+  primitives << primitive_name
+  transaction_objects << "pacemaker_primitive[#{primitive_name}]"
+end
+
```
vuntz: For the record: we need to have keystone-listener, but in a condition, see https://github.com/crowbar/crowbar-openstack/pull/512#discussion_r78153442

sayalilunkad: Updated it to add the condition.

jgrassler:
[HA mkcloud](https://ci.suse.de/job/openstack-mkcloud/34644/) passed , so all this needs now is reviewing.

chef/cookbooks/barbican/attributes/default.rb
```diff
@@ -26,3 +26,18 @@
 default[:barbican][:api][:ssl] = false
 
 default[:barbican][:logfile] = "/var/log/barbican/barbican-api.log"
+
+# HA attributes
+default[:barbican][:ha][:enabled] = false
+
+# Ports to bind to when haproxy is used for the real ports
+default[:barbican][:ha][:ports][:api] = 9411
```
vuntz: Similar comment I made to @jgrassler for magnum: I think we try to use some port in the 5xxx range. Not critical, but maybe something to do here too?

chef/cookbooks/barbican/attributes/default.rb
```diff
@@ -26,3 +26,18 @@
...
+
+# pacemaker definitions
+default[:barbican][:ha][:api][:op][:monitor][:interval] = "10s"
+default[:barbican][:ha][:retry][:op][:monitor][:interval] = "10s"
```
vuntz: That line can be deleted.

chef/cookbooks/barbican/recipes/api.rb
```diff
@@ -34,8 +34,9 @@
 end
 
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+network_settings = BarbicanHelper.network_settings(node)
 
-bind_port = node[:barbican][:api][:bind_port]
+bind_port = network_settings[:api][:bind_port]
```
vuntz: This is not fully correct: this variable is used both for apache (correct usage) and for defining the endpoint (incorrect usage).

chef/cookbooks/barbican/recipes/common.rb
```diff
@@ -76,7 +78,7 @@
     database_connection: database_connection,
     kek: node[:barbican][:kek],
     keystone_listener: node[:barbican][:enable_keystone_listener],
-    host_href: "#{barbican_protocol}://#{public_host}:#{node[:barbican][:api][:bind_port]}",
+    host_href: "#{barbican_protocol}://#{public_host}:#{bind_port}",
```
vuntz: Similarly here, I'm not sure if the bind_port usage is correct.

chef/cookbooks/barbican/recipes/common.rb
```diff
@@ -76,7 +78,7 @@
...
     rabbit_settings: fetch_rabbitmq_settings,
     keystone_settings: KeystoneHelper.keystone_settings(node, @cookbook_name),
```
vuntz: I'm surprised that we have this `bind_open_address` attribute but it seems to be completely unused for this template? Does this attribute actually do something right now?

chef/cookbooks/barbican/recipes/api.rb
sayalilunkad: As i see the network_settings[:api][:bind_port] also points to node[:barbican][:api][:bind_port] so it should be ok to use it in both places.

chef/cookbooks/barbican/attributes/default.rb
sayalilunkad: Done

chef/cookbooks/barbican/attributes/default.rb
sayalilunkad: Done

chef/cookbooks/barbican/recipes/common.rb
sayalilunkad: It is just used in the helper function as far as I see.

vuntz: Well, then we should possibly kill this attribute because it's misleading, as barbican will not respect it.

chef/cookbooks/barbican/recipes/api.rb
vuntz: No, it's not true when HA is enabled.

sayalilunkad: Makes sense, changed it.

chef/cookbooks/barbican/libraries/helpers.rb
```diff
@@ -0,0 +1,30 @@
...
+  class << self
+    def network_settings(node)
+      @ip ||= Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+      @cluster_admin_ip ||= nil
+
+      if node[:barbican][:ha][:enabled] && !@cluster_admin_ip
+        @cluster_admin_ip = CrowbarPacemakerHelper.cluster_vip(node, "admin")
+      end
+
+      bind_host = @ip
```
vuntz: Shouldn't that be "0.0.0.0"? Although it doesn't matter much as `bind_host` is not used at all (and could be dropped).

chef/cookbooks/barbican/recipes/api.rb
```diff
@@ -84,9 +86,9 @@
...
   service_type "key-manager"
   endpoint_region keystone_settings["endpoint_region"]
-  endpoint_publicURL "http://#{public_host}:#{bind_port}"
-  endpoint_adminURL "http://#{admin_host}:#{bind_port}"
-  endpoint_internalURL "http://#{admin_host}:#{bind_port}"
+  endpoint_publicURL "http://#{public_host}:#{barbican_port}"
+  endpoint_adminURL "http://#{admin_host}:#{barbican_port}"
+  endpoint_internalURL "http://#{admin_host}:#{barbican_port}"
   action :add_endpoint_template
 end
```
vuntz: Commenting here as I can't comment below, but I'm nearly sure that bind_host/bind_port are not correct for barbican-api.conf.

chef/cookbooks/barbican/recipes/common.rb
```diff
@@ -21,9 +21,10 @@
 package "openstack-barbican"
 
 db_settings = fetch_database_settings
-
 db_conn_scheme = db_settings[:url_scheme]
 
+network_settings = BarbicanHelper.network_settings(node)
```
vuntz: This isn't used, so I guess we don't need to change this file at all.

chef/cookbooks/barbican/libraries/helpers.rb
sayalilunkad: Removed it.

chef/cookbooks/barbican/attributes/default.rb
```diff
@@ -26,3 +26,17 @@
 default[:barbican][:api][:ssl] = false
 
 default[:barbican][:logfile] = "/var/log/barbican/barbican-api.log"
+
+# HA attributes
+default[:barbican][:ha][:enabled] = false
+
+# Ports to bind to when haproxy is used for the real ports
+default[:barbican][:ha][:ports][:api] = 5411
```
sayalilunkad: Changed it anyways.

chef/cookbooks/barbican/recipes/common.rb
sayalilunkad: Removed.

toabctl:
nit: Can you follow the commit msg style and use "barbican: Add HA support" or something similar please?

vuntz:
As discussed yesterday, we should call `barbican-manage` in there.

chef/cookbooks/barbican/recipes/common.rb
```diff
@@ -67,6 +66,36 @@
   action :grant
```
vuntz: Sorry, hadn't realized... But usually, the three steps above for database creation have `only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }` and are inside a sync mark.

Btw, I think `ha_enabled` isn't defined here (although you already use it).

chef/cookbooks/barbican/recipes/common.rb
```diff
@@ -65,8 +70,39 @@
   privileges db_settings[:privs]
   provider db_settings[:user_provider]
   action :grant
+  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+crowbar_pacemaker_sync_mark "wait-barbican_db_sync"
```
vuntz: I'd remove that line and...

chef/cookbooks/barbican/recipes/common.rb
```diff
@@ -65,8 +70,39 @@
...
+ruby_block "mark node for barbican db_sync" do
+  block do
+    node.set[:barbican][:db_synced] = true
+    node.save
+  end
+  action :nothing
+  subscribes :create, "execute[barbican-manage db_sync]", :immediately
+end
+
+crowbar_pacemaker_sync_mark "create-barbican_db_sync"
```
vuntz: ... rename that one to "create-barbican_database"

If you disagree, then we're at least missing the "create-barbican_database" bit ;-)

toabctl:
retriggered gating.

toabctl:
So after the automation fix got merged (https://github.com/SUSE-Cloud/automation/commit/fc37d3b510a202aee649db7d528fe3c56cf0d518) the HA deployment of barbican does not work. Error in the proposal is:
```
Failed to apply the proposal: exception before calling chef (undefined local variable or method `controller_nodes' for #)
```

chef/cookbooks/barbican/recipes/ha.rb
```diff
@@ -0,0 +1,100 @@
...
+log "Setting up barbican HA support"
+
+network_settings = BarbicanHelper.network_settings(node)
+
+haproxy_loadbalancer "barbican-api" do
+  address network_settings[:api][:ha_bind_host]
+  port network_settings[:api][:ha_bind_port]
+  use_ssl ssl_enabled
```
vuntz: `ssl_enabled` is not defined...

chef/cookbooks/barbican/recipes/api.rb
aplanas: ahh maybe you can remove lines 37 and 39?

sayalilunkad: @aplanas why? we need this here.

aplanas: Uau, sorry. You are right : )

PR #515, opened 2016-08-30
Add public IP allocation

jgrassler:
Allocate a public IP in the Trove barclamp so it will also work if it is
deployed on a standalone machine without other barclamps. Without this code it
won't work if it is deployed to a machine that doesn't have at least one other
barclamp assigning a public IP.

Note: since this commit adds a apply_role_pre_chef_call() function to the Trove
barclamp's model it also contains commented HA code to make enabling HA for
Trove easier later on.

### DIFF
--- a/crowbar_framework/app/models/trove_service.rb
+++ b/crowbar_framework/app/models/trove_service.rb
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-class TroveService < ServiceObject
+class TroveService < PacemakerServiceObject
   def initialize(thelogger)
     super(thelogger)
     @bc_name = "trove"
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/trove_service.rb
```diff
@@ -86,4 +86,29 @@ def validate_proposal_after_save(proposal)
     end
     super
   end
+
+  def apply_role_pre_chef_call(old_role, role, all_new_nodes)
+    @logger.debug("Trove apply_role_pre_chef_call: entering #{all_new_nodes.inspect}")
+
+    server_elements, server_nodes, ha_enabled = role_expand_elements(role, "trove-server")
```
jgrassler: I'd rather merge this one as it is despite Hound's objections. That way somebody implementing HA later on only needs to uncomment the HA related code (rather than modifying variable names on top of that). Thoughts?

toabctl:
@jgrassler please update the commit msg to ```trove: Add public IP allocation```

crowbar_framework/app/models/trove_service.rb
```diff
@@ -86,4 +86,31 @@ def validate_proposal_after_save(proposal)
...
+    #                                          vip_networks)
+    # role.save if dirty
+
+    net_svc = NetworkService.new @logger
+    # All nodes must have a public IP, even if part of a cluster; otherwise
+    # the VIP can't be moved to the nodes
+    server_nodes.each do |node|
+      net_svc.allocate_ip "default", "public", "host", node
+
+      # allocate_virtual_ips_for_any_cluster_in_networks_and_sync_dns(server_elements, vip_networks)
```
vuntz: This line should not be in the loop.

jgrassler: Indeed. Fixed.

vuntz:
+1

toabctl:
@jgrassler lgtm but can you update the commit msg and add a ```trove:``` prefix please?

jgrassler:
Ok, done. I rebased while I was at it.

toabctl:
+1

vuntz:
Still +1

toabctl:
@jgrassler not sure if it's related but gating failed with: ```Error: Committing the crowbar 'default' proposal for 'trove' failed (1).```

jgrassler:
Yes, this looks related indeed:

`{"error":"Failed to apply the proposal: exception before calling chef (undefined method `role_expand_elements' for #\u003cTroveService:0x0000000993b328\u003e)"}`

`role_expand_elements()` is a bit of HA plumbing we seem to be missing. Looking into it...

jgrassler:
This (inheriting from `PacemakerServiceObject` rather than `ServiceObject`) should give us the missing method.

toabctl:
+1

toabctl:
@vuntz can you merge it please?

vuntz:
+1

PR #516, opened 2016-08-31
Merge barbican roles

sayalilunkad:


toabctl:
+1

crowbar_framework/app/models/barbican_service.rb
```diff
@@ -85,18 +58,14 @@ def create_proposal
     @logger.debug("Barbican create_proposal: entering")
     base = super
 
-    node_roles = NodeObject.find("roles:barbican-server") +
-      NodeObject.find("roles:barbican-worker") +
-      NodeObject.find("roles:barbican-retry")
+    node_roles = NodeObject.find("roles:barbican-controller")
```
vuntz: Unrelated to this PR, so don't block on this, but I don't see the point of this variable. I see it's used below, but I don't understand the intent...

chef/data_bags/crowbar/migrate/barbican/100_update_barbican_roles.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  roles = ["barbican-server", "barbican-retry", "barbican-worker", "barbican-keystone_listener"]
+  roles.each do |role|
+    nodes = NodeObject.find("roles:#{role}")
+    nodes.each do |node|
+      node.add_to_run_list("barbican-controller",
+                           td["element_run_list_order"][@role],
+                           td["element_states"][@role])
+      node.delete_from_run_list(@role)
+      node.save
```
vuntz: I don't think this is fully correct, as you might end up with 4 barbican-controller nodes (if all services were on different nodes before). But it's no big deal: this is a new barclamp, so the migration won't really be used in real life.

chef/cookbooks/barbican/recipes/retry.rb
```diff
@@ -20,7 +20,3 @@
 include_recipe "#{@cookbook_name}::common"
 
 package "openstack-barbican-retry"
-
-service "openstack-barbican-retry" do
-  action [:enable, :start]
-end
```
vuntz: Maybe we should simply kill the whole recipe? I don't see the point of installing the package if we don't use it ;-)

vuntz:
I'm +0.9 here; the last comment is the only one that I believe should really be considered.

chef/data_bags/crowbar/migrate/barbican/100_update_barbican_roles.rb
sayalilunkad: Yes I am not entirely sure either.

crowbar_framework/app/models/barbican_service.rb
```diff
@@ -85,19 +58,13 @@ def create_proposal
...
     server_nodes = nodes.select { |n| n.intended_role == "controller" }
     server_nodes = [nodes.first] if server_nodes.empty?
 
     base["deployment"][@bc_name]["elements"] = {
-      "barbican-server" => [server_nodes.first.name],
-      "barbican-worker" => [server_nodes.first.name],
-      "barbican-retry" => [server_nodes.first.name]
-    } unless node_roles.nil? || server_nodes.nil?
+      "barbican-controller" => [server_nodes.first.name]
+    } unless NodeObject.find("roles:barbican-controller").nil? || server_nodes.nil?
```
vuntz: Heh, you misunderstood me: I still don't see the point of the `NodeObject.find("roles:barbican-controller").nil?` test here. I'd just remove it. But as said, can be done in followup PR as it's not new..

vuntz:
+1, but please track my last comment somewhere so it gets fixed.

crowbar_framework/app/models/barbican_service.rb
sayalilunkad: Ok, but I guess that check is to see if the role already exists, so why would you want to remove that? And would you want to change that for all the services? Because I think all have this check.

vuntz: But we don't do that for other barclamps, do we?

sayalilunkad: I think we do for a few of them. Ceilometer for example.

toabctl:
+1

PR #517, opened 2016-09-01
Magnum: Partially revert commit 326bd034d81135f40f5b7dfa887d73e973af52ab

mjura:
This logic will return nodes for allocation and is not checking if barbican_server exist:
```
    if select_nodes_for_role(nodes, "barbican-server", "controller").empty?
      base["attributes"][@bc_name][:cert][:cert_manager_type] = "local"
    else
      base["attributes"][@bc_name][:cert][:cert_manager_type] = "barbican"
    end
```

### DIFF
--- a/crowbar_framework/app/models/magnum_service.rb
+++ b/crowbar_framework/app/models/magnum_service.rb
@@ -87,12 +87,6 @@ class MagnumService < PacemakerServiceObject
     base["attributes"][@bc_name][:db][:password] = random_password
     base["attributes"][@bc_name][:trustee][:domain_admin_password] = random_password
 
-    if select_nodes_for_role(nodes, "barbican-controller", "controller").empty?
-      base["attributes"][@bc_name][:cert][:cert_manager_type] = "local"
-    else
-      base["attributes"][@bc_name][:cert][:cert_manager_type] = "barbican"
-    end
-
     @logger.debug("Magnum create_proposal: exiting")
     base
   end


toabctl:
@mjura can you add a short description why we need to revert this?

mjura:
@toabctl : This is because there is wrong usage of `select_nodes_for_role(nodes, "barbican-server", "controller")` in the code. This function will return nodes which can be allocated for `barbican_server` but in barclamp it is used for checking if `barbican_server` was deployed or not.

toabctl:
@mjura can you add that to the commit msg please?

mjura:
@toabctl : It was added

sayalilunkad:
I don't think this is the right way to fix this problem. I am working on a patch to find_applied_proposals so we can have a proper check for this condition. Maybe we can leave it as it is till then if it does not break anything.

mjura:
@sayalilunkad : I've opened this PR only to not forget about this issue, when your patch will be ready, we can close it

mjura:
We still would like to merge this change

mjura:
It was rebased and conflicts were solved, we still need this change

crowbar_framework/app/models/magnum_service.rb
```diff
@@ -87,12 +87,6 @@ def create_proposal
     base["attributes"][@bc_name][:db][:password] = random_password
     base["attributes"][@bc_name][:trustee][:domain_admin_password] = random_password
 
-    if select_nodes_for_role(nodes, "barbican-controller", "controller").empty?
-      base["attributes"][@bc_name][:cert][:cert_manager_type] = "local"
```
aplanas: Makes sense to keep only this line here? Or is the default value?

mjura: This is default value for it. We can remove all

PR #518, opened 2016-09-01
keystone: Fix openstack command line options

rhafer:
Adding missing --os-project-name option to get a scoped token. Use
--os-auth-url instead of --os-url. --os-url is only relevant when
authenticating with an already existing token). This broke when we switched
away from using the admin token for authentication.


sayalilunkad:
+1

aspiers:
+1

PR #519, opened 2016-09-01
Heat: Change clients endpoints on publicURL

mjura:
We would like to change endpoints on publicURL for Heat
and Magnum clients. This will be used by OS::Heat::WaitCondition
resource and wc_notify.

### DIFF
--- a/chef/cookbooks/heat/templates/default/heat.conf.erb
+++ b/chef/cookbooks/heat/templates/default/heat.conf.erb
@@ -667,7 +667,7 @@ insecure = <%= @keystone_settings['insecure'] %>
 # Type of endpoint in Identity service catalog to use for communication with
 # the OpenStack service. (string value)
 #endpoint_type = <None>
-endpoint_type=internalURL
+endpoint_type=publicURL
 
 # Optional CA cert file to use in SSL connections. (string value)
 #ca_file = <None>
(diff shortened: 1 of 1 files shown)


jgrassler:
+1

toabctl:
just asking - do the other clients also need this change?

mjura:
@toabctl As @jgrassler said changing endpoints for other services can have some bad consequences.  Better would be to not change it for other services, I think.

jgrassler:
I think I only mentioned that on IRC, so I'll provide some context just to be on the safe side:

I used to use a Heat template to create Nova flavors for quickly setting up Magnum bays in clouds without our Tempest flavors available. That ceased working a while ago because it ran afoul of the following line in Nova's `policy.json`:

https://github.com/openstack/nova/blob/stable/mitaka/etc/nova/policy.json#L145

Presumably that happened in the course of template refactoring. So in a nutshell we'll need to carefully review this kind of change for each services that base their policy decisions on the API type used. For Magnum/Heat it's fine because they do not make this distinction (Heat) or base it on authentication context rather than URL type (Magnum).

mjura:
@jgrassler : Thank you a lot for explanations :)

nkrinner:
+1

PR #521, opened 2016-09-05
Infoblox

jgrassler:
This pull request adds support for the [new style infoblox Neutron plugin](https://github.com/openstack/networking-infoblox/).

### DIFF
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -88,6 +88,7 @@ when "suse"
     lbaas_haproxy_group: "haproxy",
     f5_agent_pkg: "",
     f5_agent_name: "f5-openstack-agent",
+    infoblox_agent_name: "openstack-neutron-infoblox-ipam-agent",
     metadata_agent_name: "openstack-neutron-metadata-agent",
     metadata_agent_pkg: "openstack-neutron-metadata-agent",
     metering_agent_pkg: "openstack-neutron-metering-agent",
@@ -108,6 +109,9 @@ when "suse"
     cisco_opflex_pkgs: ["agent-ovs",
                         "lldpd",
                         "openstack-neutron-opflex-agent"],
+    infoblox_pkgs: ["python-infoblox-client",
+                    "openstack-neutron-infoblox",
+                    "openstack-neutron-infoblox-ipam-agent"],
     user: "neutron",
     group: "neutron",
   }
@@ -130,6 +134,7 @@ when "rhel"
     lbaas_haproxy_group: "nogroup",
     f5_agent_pkg: "",
     f5_agent_name: "f5-openstack-agent",
+    infoblox_agent_name: "",
     metadata_agent_name: "neutron-metadata-agent",
     metadata_agent_pkg: "openstack-neutron",
     metering_agent_pkg: "openstack-neutron-metering-agent",
@@ -149,6 +154,7 @@ when "rhel"
     cisco_opflex_pkgs: ["agent-ovs",
                         "lldpd",
                         "neutron-opflex-agent"],
+    infoblox_pkgs: [],
     user: "neutron",
     group: "neutron",
   }
@@ -172,6 +178,7 @@ else
     lbaas_haproxy_group: "nogroup",
     f5_agent_pkg: "",
     f5_agent_name: "f5-oslbaasv2-agent",
+    infoblox_agent_name: "",
     metadata_agent_name: "neutron-metadata-agent",
     metadata_agent_pkg: "neutron-metadata-agent",
     metering_agent_pkg: "neutron-plugin-metering-agent",
@@ -187,6 +194,7 @@ else
     cisco_apic_pkgs: [""],
     cisco_apic_gbp_pkgs: [""],
     cisco_opflex_pkgs: [""],
+    infoblox_pkgs: [],
     user: "neutron",
     group: "neutron",
   }
@@ -220,5 +228,9 @@ default[:neutron][:ha][:neutron_ha_tool][:op][:start][:timeout] = "120s"
 default[:neutron][:ha][:server][:enabled] = false
 default[:neutron][:ha][:server][:server_ra] = "service:#{node[:neutron][:platform][:service_name]}"
 default[:neutron][:ha][:server][:op][:monitor][:interval] = "10s"
+default[:neutron][:ha][:infoblox][:enabled] = false
+default[:neutron][:ha][:infoblox][:infoblox_ra] =
+  "lsb:#{node[:neutron][:platform][:infoblox_agent_name]}"
+default[:neutron][:ha][:infoblox][:op][:monitor][:interval] = "10s"
 # Ports to bind to when haproxy is used for the real ports
 default[:neutron][:ha][:ports][:server] = 5530
--- a/chef/cookbooks/neutron/recipes/common_config.rb
+++ b/chef/cookbooks/neutron/recipes/common_config.rb
@@ -111,6 +111,14 @@ end
 os_sdn_net = Barclamp::Inventory.get_network_definition(node, "os_sdn")
 mtu_value = os_sdn_net.nil? ? 1500 : os_sdn_net["mtu"].to_i
 
+ipam_driver = nil
+infoblox_settings = nil
+
+if neutron[:neutron][:use_infoblox]
+  ipam_driver = "infoblox"
+  infoblox_settings = neutron[:neutron][:infoblox]
+end
+
 template "/etc/neutron/neutron.conf" do
     cookbook "neutron"
     source "neutron.conf.erb"
@@ -143,7 +151,9 @@ template "/etc/neutron/neutron.conf" do
       dvr_enabled: neutron[:neutron][:use_dvr],
       network_nodes_count: network_nodes_count,
       dns_domain: neutron[:neutron][:dhcp_domain],
-      mtu_value: mtu_value
+      mtu_value: mtu_value,
+      infoblox: infoblox_settings,
+      ipam_driver: ipam_driver
     )
 end
 
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -20,6 +20,8 @@ use_zvm = node[:neutron][:networking_plugin] == "ml2" && !zvm_compute_node.empty
 
 pkgs = node[:neutron][:platform][:pkgs] + node[:neutron][:platform][:pkgs_fwaas]
 pkgs += node[:neutron][:platform][:pkgs_lbaas] if node[:neutron][:use_lbaas]
+pkgs += node[:neutron][:platform][:infoblox_pkgs] if node[:neutron][:use_infoblox]
+
 if use_hyperv
   pkgs << node[:neutron][:platform][:hyperv_pkg]
 end
@@ -373,6 +375,15 @@ service node[:neutron][:platform][:service_name] do
   subscribes :restart, resources("template[#{plugin_cfg_path}]")
   subscribes :restart, resources("template[/etc/neutron/neutron.conf]")
   provider Chef::Provider::CrowbarPacemakerService if ha_enabled
+end
+
+if node[:neutron][:use_infoblox]
+  service node[:neutron][:platform][:infoblox_agent_name] do
+    supports status: true, restart: true
+    action [:enable, :start]
+    subscribes :restart, resources("template[/etc/neutron/neutron.conf]")
+    provider Chef::Provider::CrowbarPacemakerService if ha_enabled
+  end
 end
 
 include_recipe "neutron::api_register"
--- a/chef/cookbooks/neutron/recipes/server_ha.rb
+++ b/chef/cookbooks/neutron/recipes/server_ha.rb
@@ -29,21 +29,43 @@ crowbar_pacemaker_sync_mark "sync-neutron_before_ha"
 # Avoid races when creating pacemaker resources
 crowbar_pacemaker_sync_mark "wait-neutron_ha_resources"
 
+primitives = []
 transaction_objects = []
 
-primitive_name = "neutron-server"
-pacemaker_primitive primitive_name do
+server_primitive_name = "neutron-server"
+pacemaker_primitive server_primitive_name do
   agent node[:neutron][:ha][:server][:server_ra]
   op node[:neutron][:ha][:server][:op]
   action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
-transaction_objects << "pacemaker_primitive[#{primitive_name}]"
+transaction_objects << "pacemaker_primitive[#{server_primitive_name}]"
+primitives << server_primitive_name
 
-clone_name = "cl-#{primitive_name}"
+if node[:neutron][:use_infoblox]
+  infoblox_primitive_name = "infoblox-agent"
+  pacemaker_primitive infoblox_primitive_name do
+    agent node[:neutron][:ha][:infoblox][:infoblox_ra]
+    op node[:neutron][:ha][:infoblox][:op]
+    action :update
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+  transaction_objects << "pacemaker_primitive[#{infoblox_primitive_name}]"
+  primitives << infoblox_primitive_name
+end
+
+group_name = "g-neutron"
+pacemaker_group group_name do
+  members primitives
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects << "pacemaker_group[#{group_name}]"
+
+clone_name = "cl-#{group_name}"
 pacemaker_clone clone_name do
-  rsc primitive_name
-  meta ({ "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node) })
+  rsc group_name
+  meta({"clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node)})
   action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
--- a/chef/cookbooks/neutron/templates/default/neutron.conf.erb
+++ b/chef/cookbooks/neutron/templates/default/neutron.conf.erb
@@ -163,6 +163,9 @@ allow_overlapping_ips = False
 # (default behavior), no IPAM driver is used. In order to use the reference
 # implementation of Neutron IPAM driver, use 'internal'. (string value)
 #ipam_driver = <None>
+<% unless @ipam_driver.nil? -%>
+ipam_driver = <%= @ipam_driver %>
+<% end -%>
 
 # If True, then allow plugins that support it to create VLAN transparent
 # networks. (boolean value)
@@ -1596,3 +1599,74 @@ key_file = <%= @ssl_key_file %>
 # Sets the list of available ciphers. value should be a string in the OpenSSL
 # cipher list format. (string value)
 #ciphers = <None>
+
+<% if @infoblox -%>
+
+[infoblox]
+
+# ID used for selecting a particular grid from one or  more grids to serve
+# networks in Infoblox backend.
+
+# cloud_data_center_id = 0
+cloud_data_center_id = <%= @infoblox[:cloud_data_center_id] %>
+
+# Number of Infoblox IPAM agent workers to run
+# ipam_agent_workers = 1
+ipam_agent_workers = <%= @infoblox[:ipam_agent_workers] %>
+
+<% @infoblox[:grids].each_index do |grid| -%>
+# Per-grid configuration for infoblox grid <%= grid %>
+[infoblox-dc:<%= grid %>]
+
+# Admin user name to access the grid master or cloud platform appliance.
+# admin_user_name =
+admin_user_name = <%= @infoblox[:grids][grid][:admin_user_name] %>
+
+# Admin user password to access the grid master or cloud platform appliance.
+# admin_password =
+admin_password = <%= @infoblox[:grids][grid][:admin_password] %>
+
+# The name of the Grid Master (GM) This has to be the exact GM name registered
+# in the Infoblox grid.
+grid_master_name = <%= @infoblox[:grids][grid][:grid_master_name] %>
+
+# Host IP or name of the grid master.
+# grid_master_host =
+grid_master_host = <%= @infoblox[:grids][grid][:grid_master_host] %>
+
+# The name of data center to identify.
+# data_center_name =
+data_center_name = <%= @infoblox[:grids][grid][:data_center_name] %>
+
+# The WAPI version to use. Version 2.2.2 or later is recommended, if your grid
+# supports it (WAPI version 2.3 is supported in NIOS 7.3)
+# wapi_version =
+wapi_version = <%= @infoblox[:grids][grid][:wapi_version] %>
+
+# Maximum number of objects to be returned. If set to a negative number the
+# appliance will return an error when the number of returned objects would
+# exceed the setting. If this is set to a positive number, the results will be
+# truncated when necessary.
+# wapi_max_results = -1000
+
+wapi_max_results = <%= @infoblox[:grids][grid][:wapi_max_results] %>
+
+# Ensure whether WAPI requests sent over HTTPS require SSL verification.
+# ssl_verify = False
+ssl_verify = <%= @infoblox[:grids][grid][:ssl_verify] %>
+
+# HTTP pool connections.
+# http_pool_connections = 100
+http_pool_connections = <%= @infoblox[:grids][grid][:http_pool_connections] %>
+
+# HTTP pool max size.
+# http_pool_maxsize = 100
+http_pool_maxsize = <%= @infoblox[:grids][grid][:http_pool_maxsize] %>
+
+# HTTP request timeout.
+# http_request_timeout = 120
+http_request_timeout = <%= @infoblox[:grids][grid][:http_request_timeout] %>
+
+<% end -%>
+
+<% end -%>
--- a/chef/data_bags/crowbar/migrate/neutron/102_infoblox_update.rb
+++ b/chef/data_bags/crowbar/migrate/neutron/102_infoblox_update.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+def upgrade(ta, td, a, d)
+  unless a.key? "use_infoblox"
+    a["use_infoblox"] = ta["use_infoblox"]
+  end
+  # Since we will backport this to Cloud 6, we use 'grid_defaults`, one of the
+  # newly introduced keys from this migration to determine whether we are
+  # dealing with a database state from its twin in Cloud 6. In that case we
+  # won't touch it. If, on the other hand, we are dealing with the Cloud 5
+  # version from October 2015 we'll just overwrite it since the infoblox plugin
+  # changed a great deal since then.
+  unless (a.key? "infoblox") && (a[:infoblox].key? "grid_defaults")
+    a["infoblox"] = ta["infoblox"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless ta.key? "use_infoblox"
+    a.delete("use_infoblox")
+  end
+  unless ta.key? "infoblox"
+    a.delete("infoblox")
+  end
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-neutron.json
+++ b/chef/data_bags/crowbar/template-neutron.json
@@ -79,42 +79,26 @@
       },
       "use_infoblox": false,
       "infoblox": {
-        "ib_wapi": "",
-        "ib_username": "admin",
-        "ib_password": "",
-        "ib_members": "",
-        "ib_dnsserver": [],
-        "use_host_records_for_ip_allocation": true,
-        "bind_dns_records_to_fixed_address": "record:a,record:ptr",
-        "unbind_dns_records_from_fixed_address": "record:a,record:ptr",
-        "delete_dns_records_associated_with_fixed_address": "record:a,record:ptr,record:txt",
-        "use_dhcp_for_ip_allocation_record": true,
-        "enable_dhcp_relay": false,
-        "dhcp_relay_interface": "br-fixed",
-        "dhcp_relay_management_network": "",
-        "dhcp_relay_management_network_view": "",
-        "ib_conditionals": [
+        "cloud_data_center_id": 0,
+        "ipam_agent_workers": 1,
+        "grids": [
+            {
+            "admin_user_name": "admin",
+            "admin_password": "password",
+            "grid_master_host": "gridmaster1.example.net",
+            "grid_master_name": "gridmaster1",
+            "data_center_name": "dc1"
+            }
+          ],
+        "grid_defaults":
           {
-            "condition": "tenant",
-            "is_external": false,
-            "network_view": "default",
-            "dns_view": "default",
-            "dhcp_members": "\"<next-available-member>\"",
-            "dns_members": "\"<next-available-member>\"",
-            "domain_suffix_pattern": "{subnet_name}.cloud-test.com",
-            "hostname_pattern": "host-{ip_address}"
-          },
-          {
-            "condition": "global",
-            "is_external": true,
-            "network_view": "default",
-            "dns_view": "default",
-            "dhcp_members": "\"<next-available-member>\"",
-            "dns_members": "\"<next-available-member>\"",
-            "domain_suffix_pattern": "{subnet_name}.cloud-test.com",
-            "hostname_pattern": "host-{ip_address}"
+          "http_pool_connections": 100,
+          "http_pool_maxsize": 100,
+          "http_request_timeout": 120,
+          "ssl_verify": false,
+          "wapi_version": "2.2.2",
+          "wapi_max_results": -1000
           }
-        ]
       }
     }
   },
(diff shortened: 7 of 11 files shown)


jgrassler:
This is now ready for review. I tested it manually up to the point of provoking the following exception in `/var/log/neutron/neutron-server.log`

```
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource Traceback (most recent call last):
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/api/v2/resource.py", line 84, in resource
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     result = method(request=request, **args)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/api/v2/base.py", line 410, in create
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     return self._create(request, body, **kwargs)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/oslo_db/api.py", line 148, in wrapper
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     ectxt.value = e.inner_exc
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/oslo_utils/excutils.py", line 220, in __exit__
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     self.force_reraise()
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/oslo_utils/excutils.py", line 196, in force_reraise
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     six.reraise(self.type_, self.value, self.tb)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/oslo_db/api.py", line 138, in wrapper
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     return f(*args, **kwargs)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/api/v2/base.py", line 521, in _create
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     obj = do_create(body)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/api/v2/base.py", line 503, in do_create
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     request.context, reservation.reservation_id)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/oslo_utils/excutils.py", line 220, in __exit__
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     self.force_reraise()
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/oslo_utils/excutils.py", line 196, in force_reraise
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     six.reraise(self.type_, self.value, self.tb)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/api/v2/base.py", line 496, in do_create
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     return obj_creator(request.context, **kwargs)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/plugins/ml2/plugin.py", line 892, in create_subnet
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     result, mech_context = self._create_subnet_db(context, subnet)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/plugins/ml2/plugin.py", line 881, in _create_subnet_db
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     result = super(Ml2Plugin, self).create_subnet(context, subnet)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/db/db_base_plugin_v2.py", line 703, in create_subnet
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     return self._create_subnet(context, subnet, subnetpool_id)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/db/db_base_plugin_v2.py", line 594, in _create_subnet
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     subnetpool_id)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/db/ipam_pluggable_backend.py", line 434, in allocate_subnet
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     ipam_driver = driver.Pool.get_instance(subnetpool, context)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/neutron/ipam/driver.py", line 54, in get_instance
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     return driver_class(subnet_pool, context)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/networking_infoblox/ipam/driver.py", line 51, in func
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     return f(*args, **kwargs)
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/networking_infoblox/ipam/driver.py", line 91, in __init__
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     self._grid_manager.get_config()
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/networking_infoblox/neutron/common/grid.py", line 85, in get_config
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     self.grid_config.sync()
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/networking_infoblox/neutron/common/grid.py", line 244, in sync
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     discovered_config = self._discover_config(self._get_gm_member())
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource   File "/usr/lib/python2.7/site-packages/networking_infoblox/neutron/common/grid.py", line 305, in _get_gm_member
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource     raise exc.InfobloxCannotFindMember(member="GM")
2016-09-12 11:43:57.868 18138 ERROR neutron.api.v2.resource InfobloxCannotFindMember: Cannot find the member: 'GM'
```

This means the infoblox-ipam-agent is running and configured in `neutron.conf` but cannot connect to the bogus Infoblox grid master I configured. This is about as far as we can test things without an actual Infoblox API to play with. Testing this is a bit tricky right now since packages are only in my home project and/or Cloud:OpenStack:Master (some dependencies) so far. Steps for testing:

1. Build a cloud using `mkcloud instonly`
2. Apply this commit's changes
3. Apply barclamps up to the Neutron barclamp
4. Add repositories:
  * zypper ar http://download.opensuse.org/repositories/home:/jgrassler:/branches:/Cloud:/OpenStack:/Master/SLE_12_SP1/ jgr-master
  * zypper ar http://download.opensuse.org/repositories/Cloud:/OpenStack:/Master/SLE_12_SP1/ com
5. Set `use_infoblox` to `true` in Neutron barclamp and deploy Neutron barclamp
6. Deploy Nova barclamp

Note: the Nova barclamp needed a little change as well because `infoblox-ipam-agent` needs Nova to use the `messagingv2` driver for RabbitMQ (see https://github.com/openstack/networking-infoblox/blob/master/doc/source/installation.rst#nova).

In the (unlikely) case that there are no further comments: please do not merge this right away, I'd like to rebase this mess of micro commits before this happens, but I'd like to retain the history just in case I need to revert/revisit stuff.

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -83,6 +83,7 @@
     zvm_agent_name: "openstack-neutron-zvm-agent",
     lbaas_agent_pkg: "openstack-neutron-lbaas-agent",
     lbaas_agent_name: "openstack-neutron-lbaas-agent",
+    infoblox_agent_name: "openstack-neutron-infoblox-ipam-agent",
```
vuntz: Can you define this for the other distros too?

chef/cookbooks/neutron/libraries/helpers.rb
```diff
@@ -67,4 +67,44 @@ def self.get_network_node_from_neutron_attributes(node)
       Chef::Node.load(network_node_name)
     end
   end
+
+  # Validates infoblox configuration and returns a configuration hash ready for
+  # passing to the neutron.conf template.
+  def self.get_infoblox_config(node)
+    return { use_infoblox: false } unless node[:neutron][:use_infoblox]
```
vuntz: I would just call the attribute `enabled` since we know the hash we'll get is about Infoblox.

chef/cookbooks/neutron/libraries/helpers.rb
```diff
@@ -67,4 +67,44 @@ def self.get_network_node_from_neutron_attributes(node)
...
+    node[:neutron][:infoblox][:use_infoblox] = node[:neutron][:use_infoblox]
```
vuntz: Doesnt't reall matter, but I'd put `= true` directly here.

chef/cookbooks/neutron/libraries/helpers.rb
```diff
@@ -67,4 +67,44 @@ def self.get_network_node_from_neutron_attributes(node)
...
+      raise "No grid configurations in node[:neutron][:infoblox][:grids]. " \
+            "Please specify at least one."
+    end
+
+    if node[:neutron][:infoblox][:cloud_data_center_id].to_i >=
+        node[:neutron][:infoblox][:grids].length
+      raise "Invalid cloud_data_center_id: " \
+            "`#{node[:neutron][:infoblox][:cloud_data_center_id]}` is not " \
+            "a valid index into node[:neutron][:infoblox][:grids]."
+    end
```
vuntz: Can we also have these checks in the validation of `neutron_service.rb`?

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -72,6 +72,8 @@
 
 keystone_settings = KeystoneHelper.keystone_settings(neutron, @cookbook_name)
```
vuntz: Personal preference: I would not have the empty line here.

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -116,6 +118,9 @@
 os_sdn_net = Barclamp::Inventory.get_network_definition(node, "os_sdn")
 mtu_value = os_sdn_net.nil? ? 1500 : os_sdn_net["mtu"].to_i
 
+ipam_driver = nil
+ipam_driver = "infoblox" if infoblox_settings[:use_infoblox]
```
vuntz: Why not `ipam_driver = infoblox_settings[:use_infoblox] ? "infoblox" : nil`?

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -31,23 +31,24 @@
...
-pacemaker_primitive primitive_name do
-  agent node[:neutron][:ha][:server][:server_ra]
-  op node[:neutron][:ha][:server][:op]
-  action :update
-  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
-end
-transaction_objects << "pacemaker_primitive[#{primitive_name}]"
+["neutron-server", "infoblox"].each do |primitive_name|
+  pacemaker_primitive primitive_name do
+    agent node[:neutron][:ha][primitive_name.to_sym]["#{primitive_name}_ra".to_sym]
```
vuntz: You're breaking stuff for neutron-server here as the attribute is `:server`, not `:neutron-server`.

vuntz: Also, I would probably name the primitive for the infoblox bit `neutron-infoblox-ipam-agent`, for consistency with the other names we have.

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -370,6 +372,15 @@
   provider Chef::Provider::CrowbarPacemakerService if ha_enabled
 end
 
+if node[:neutron][:use_infoblox]
+  service node[:neutron][:platform][:infoblox_agent_name] do
```
vuntz: Just to confirm: this agent should be running on the same node as neutron-server, not on the same node as other network agents. Is that correct?

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -31,23 +31,24 @@
...
+  end
+  transaction_objects << "pacemaker_primitive[#{primitive_name}]"
 
-clone_name = "cl-#{primitive_name}"
-pacemaker_clone clone_name do
-  rsc primitive_name
-  meta ({ "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node) })
-  action :update
-  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  clone_name = "cl-#{primitive_name}"
```
vuntz: Are we sure the infoblox ipam agent can run active/active?

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -163,6 +163,9 @@ allow_overlapping_ips = False
 # (default behavior), no IPAM driver is used. In order to use the reference
 # implementation of Neutron IPAM driver, use 'internal'. (string value)
 #ipam_driver = <None>
+<% unless @ipam_driver.nil? -%>
+ipam_driver=<%= @ipam_driver %>
```
vuntz: Spaces around `=`?

chef/cookbooks/neutron/libraries/helpers.rb
```diff
@@ -67,4 +67,44 @@ def self.get_network_node_from_neutron_attributes(node)
...
+                   :ssl_verify,
+                   :wapi_max_results,
+                   :wapi_version]
+
+    # Fill in defaults for individual grid configurations ([infoblox-dc:<id>])
+
+    node[:neutron][:infoblox][:grids].each do |grid|
+      defaultable.each do |d|
+        unless grid.key?(d)
+          grid[d] = node[:neutron][:infoblox][:grid_defaults][d]
```
vuntz: Hrm. Not a big fan of this, for a couple of reasons:

- this sets an attribute on the node, so you should either use `node.set` or `node.default` or something like this
- generally speaking, we should avoid setting attributes in the namespace of a barclamp on a node: that is, we should avoid setting stuff in `node[:nova]`, `node[:keystone]`, etc. That's because the attributes in this namespace are actually note on the node itself, but inherited from a role the node has, and if we start adding them on the node, it will possibly be confusing, cause bugs (because of unclear priority between role attribute and node attribute), and so on.

I would probably just kill that code and instead, in the template, do something like: `option = grid_option.nil? grid_default : grid_option`

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -1594,3 +1597,73 @@ key_file = <%= @ssl_key_file %>
...
+
+# HTTP pool max size.
+# http_pool_maxsize = 100
+http_pool_maxsize = <%= @infoblox[:grids][grid][:http_pool_maxsize] %>
+
+# HTTP request timeout.
+# http_request_timeout = 120
+http_request_timeout = <%= @infoblox[:grids][grid][:http_request_timeout] %>
+<% end -%>
+
```
vuntz: Empty line should be before the `<% end -%>` bit, to have an empty line between two sections.

crowbar_framework/app/models/nova_service.rb
```diff
@@ -176,6 +176,13 @@ def create_proposal
     base["attributes"]["nova"]["db"]["password"] = random_password
     base["attributes"]["nova"]["neutron_metadata_proxy_shared_secret"] = random_password
 
+    neutron = Proposal.find_by(barclamp: "neutron",
+                               name: base["attributes"][@bc_name]["neutron_instance"])
+
+    # Use messagingv2 if the infoblox agent is enabled for the Neutron barclamp
+    use_infoblox = neutron["attributes"]["neutron"]["use_infoblox"]
+    base[:attributes][:nova][:notification_driver] = "messagingv2" if use_infoblox
```
vuntz: Can't we just always switch to messagingv2 instead of adding an attribute?

chef/data_bags/crowbar/migrate/neutron/102_infoblox_update.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+def upgrade(ta, td, a, d)
+  unless a.key? "use_infoblox"
+    a["use_infoblox"] = ta["use_infoblox"]
+  end
+  # Since we will backport this to Cloud 6, we use, 'grid_defaults`, one of the
```
vuntz: "we use," => "we use"?

chef/cookbooks/neutron/attributes/default.rb
jgrassler: I can, but not sure if it's a good idea: right now there are no packages as far as I can see, so that name might change (e.g. the RHEL package might end up getting named `openstack-neutron-infoblox-agent` rather than `openstack-neutron-infoblox-ipam-agent`.

vuntz: Sure, just make the thing empty. We do that for some other agents already.

chef/cookbooks/neutron/recipes/common_config.rb
jgrassler: I don't like the `? :` operator very much - it is easy to overlook and hence almost invariably slows at least me down when reading the code. I.e. this is mostly a matter of personal preference :-)

If you'd rather have a one liner here I can change it, though.

chef/cookbooks/neutron/recipes/server.rb
jgrassler: Correct, yes. Its task is providing `neutron-server` with IP addresses and other network configuration it retrieves from the Infoblox appliance.

chef/cookbooks/neutron/recipes/server_ha.rb
jgrassler: Reasonably sure... the vendor claims that is the case: https://github.com/openstack/networking-infoblox/blob/master/doc/source/installation.rst#installing-the-driver

vuntz: Thanks for the confirmation :-)

chef/cookbooks/neutron/recipes/common_config.rb
vuntz: Nah, I'm fine. It's a matter of taste, so doesn't matter.

chef/cookbooks/neutron/libraries/helpers.rb
jgrassler: Hmm. Sounds to me like we could improve matters greatly by moving all code modifying stuff to `crowbar_framework/app/models/neutron_server.rb`. This would make the whole thing more transparent too. Once that's done I can get rid of get_infoblox_config entirely (by that time there won't be much left anyway :-)).

vuntz: That makes sense!

crowbar_framework/app/models/nova_service.rb
jgrassler: We could (we do that kind of thing in the tempest barclamp for instance to match Heat's `heat_stack_owner_roles setting`), but I went for an attribute for transparency's sake: that way the user can easily see that Nova switched to `messagingv2`.

chef/cookbooks/neutron/libraries/helpers.rb
jgrassler: Ok, done :-)

While I was at it I shortened it a bit, too since it doesn't really make sense to have a list of defaults in the code when it's already in the data bag. Also, I'm not quite sure what I'm doing in `neutron_server.rb` makes sense: are the contents of the data bag already available in `base[:attributes]` when `create_proposal()` runs?

This is in a separate commit as well so I can revisit it for amending.

chef/data_bags/crowbar/migrate/neutron/102_infoblox_update.rb
jgrassler: Fixed.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
jgrassler: Done.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
jgrassler: Fixed.

chef/cookbooks/neutron/recipes/server_ha.rb
jgrassler: Yeah. I was toying with the idea of `not` making this a loop. I guess i should have stuck with that idea :-)

I gave it a bit of an overhaul now: I modeled it on the way Magnum does HA (i.e. I created a group/clone group). I put this into a separate commit in case it needs revisiting.

chef/cookbooks/neutron/recipes/common_config.rb
jgrassler: No longer relevant anyway, `get_infoblox_settings()` being a thing of the past now.

chef/cookbooks/neutron/libraries/helpers.rb
jgrassler: Sure. In fact, I moved them over completely since they don't serve much of a purpose in the helper once they're in `neutron_service.rb`.

chef/cookbooks/neutron/libraries/helpers.rb
jgrassler: Agreed, that makes the code a bit easier on the eyes. Done.

chef/cookbooks/neutron/libraries/helpers.rb
jgrassler: Ok, done.

chef/cookbooks/neutron/attributes/default.rb
jgrassler: Ok, done.

jgrassler:
Ok, the last 3 commits hopefully fix all bugs I created while addressing the comments. So this is ready for another round of reviews now.

crowbar_framework/app/models/nova_service.rb
dirkmueller: I think that should be transparent to the user (read: he shouldn't care). iirc messagingv2 is the recommended choice meanwhile, we should just always use it.

crowbar_framework/config/locales/neutron/en.yml
```diff
@@ -105,6 +105,8 @@ en:
         no_valid_ip: 'Multicast group for VXLAN broadcast emulation %{mcast_group} is not a valid IP address'
         no_valid_multicast_ip: 'Multicast group for VXLAN broadcast emulation %{mcast_group} is not a valid multicast IP address.'
         network: 'Network "%{ext_net}" cannot be used as an additional external network'
+        infoblox_grids: "No grid configurations in infoblox[:grids]. Please specify at least one."
+        infoblox_dc_id: "Invalid cloud_data_center_id: `%{dc_id}` is not a valid index into infoblox[:grids]."
```
dirkmueller: '%{dc_id}' needs to be a index in the range (1-x) into infoblox[:grids]

crowbar_framework/config/locales/neutron/en.yml
```diff
@@ -105,6 +105,8 @@ en:
         no_valid_ip: 'Multicast group for VXLAN broadcast emulation %{mcast_group} is not a valid IP address'
         no_valid_multicast_ip: 'Multicast group for VXLAN broadcast emulation %{mcast_group} is not a valid multicast IP address.'
         network: 'Network "%{ext_net}" cannot be used as an additional external network'
+        infoblox_grids: "No grid configurations in infoblox[:grids]. Please specify at least one."
```
dirkmueller: Please specificy at least one grid configuration in infoblox[:grids]

crowbar_framework/app/models/nova_service.rb
jgrassler: Ok, I made `messagingv2` the default now.

crowbar_framework/config/locales/neutron/en.yml
jgrassler: Done.

crowbar_framework/config/locales/neutron/en.yml
jgrassler: Done.

jgrassler:
Heads-up: rebased this on current master (schema revision for Nova needed bumping).

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -29,21 +29,44 @@
...
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects << "pacemaker_group[#{group_name}]"
+
+clone_name = "cl-#{group_name}"
 pacemaker_clone clone_name do
-  rsc primitive_name
-  meta ({ "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node) })
+  rsc group_name
+  meta("clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node))
```
vuntz: Are you sure we can remove the `{}` bits here? If it's because of hound, I think it's a bug in hound.

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -99,6 +99,16 @@ def create_proposal
...
 
+    # Fill in missing defaults for infoblox grid configurations
+    base[:attributes][:neutron][:infoblox][:grids].each do |grid|
+      defaults = base["attributes"]["neutron"]["infoblox"]["grid_defaults"]
+      defaults.each_key.each do |d|
+        unless grid.key?(d)
+          grid[d] = defaults[d]
+        end
+      end
+    end
```
vuntz: Hrm, I somehow doubt it should be in `create_proposal`, as this won't be done when the user modifies the proposal and adds a grid. Shouldn't it be in some `save` method instead?

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -286,6 +310,7 @@ def validate_proposal_after_save(proposal)
 
     validate_ml2(proposal) if plugin == "ml2"
     validate_dvr(proposal)
+    validate_infoblox(proposal)
```
vuntz: Shouldn't we validate only if `use_infoblox` is `true`?

chef/data_bags/crowbar/template-nova.json
```diff
@@ -10,6 +10,7 @@
       "glance_instance": "none",
       "cinder_instance": "none",
       "neutron_instance": "none",
+      "notification_driver": "messagingv2",
```
vuntz: I'm still not sold on the attribute here, as opposed to always setting this in the conf file. I just don't see the benefit of having it configurable -- is there one?

chef/cookbooks/neutron/recipes/server_ha.rb
jgrassler: It is because of hound. And it does actually work as it's supposed to (Ruby syntax allows you to omit the braces if the only parameter is a hash). That being said I don't really like it either. It is the sort of thing that causes every reader to stop and wonder (and write a snippet of sample code to test if this actually works in my case), all for the dubious value of slightly terser code. So I'm all for teaching hound not to enforce this :-)

chef/data_bags/crowbar/template-nova.json
jgrassler: I originally used that so I could modify the Nova barclamp's messaging driver transparently based on the value of the Neutron barclamp's `use_infoblox` setting. So if you'd rather have this hardwired I won't argue about it. I only left it exposed in case a hypothetical somebody needs that for a hypothetical reason some time in the future, which is not much of a case for leaving it exposed :-)

vuntz: I'd go for hardcoding it, yes.

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -78,50 +78,34 @@
       },
       "use_infoblox": false,
```
vuntz: Feel free to ignore, this just came to my mind during the monthly demo: we could clean up attributes and have `[:infoblox][:enabled]` instead of `[:use_infoblox]` and `[:infoblox]`.

crowbar_framework/app/models/neutron_service.rb
jgrassler: That sounds better, yes. I'm having some trouble finding a suitable `save` method to use, though, though: I tried overriding and then calling both `ServiceObject's` `save_proposal!()` or  `proposal_edit()` method both looked like likely candidates, but overriding and then calling them with the modified proposal didn't have any effect in either case. Is there another save method that's flying in under the radar somewhere?

jgrassler: Never mind. I PEBKACed on that one by only applying the changes to my git checkout but not `/opt/dell` :-)

Works now.

crowbar_framework/app/models/neutron_service.rb
jgrassler: Indeed. Done.

chef/data_bags/crowbar/template-nova.json
jgrassler: Ok, done.

chef/data_bags/crowbar/template-neutron.json
jgrassler: Yes, I'll ignore this for the time being. Thanks for the offer :-)

I'd rather do this cleanup in a follow-up pull request.

jgrassler:
All comments addressed. I'm leaving the commits as they are for now. I'll rebase before merge.

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -29,21 +29,44 @@
...
+    op node[:neutron][:ha][:infoblox][:op]
+    action :update
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+  transaction_objects << "pacemaker_primitive[#{infoblox_primitive_name}]"
+  primitives << infoblox_primitive_name
+end
+
+group_name = "g-neutron"
+
```
vuntz: For cleanup: you can kill this empty line.

chef/cookbooks/neutron/recipes/server_ha.rb
vuntz: I'd really add the `{}` back, tbh, especially as we have it like this everywhere else. Can be in a cleanup PR.

jgrassler:
Ok,  I removed the empty line and restored the braces (no way I'm going to wait until a cleanup PR if you're giving me carte blanche to spite Hound _now_). Let's see how that squash feature works out...

chef/cookbooks/neutron/recipes/server_ha.rb
jgrassler: Done.

aplanas:
Can you use the pattern: `component: ...` for the commits that are not going to be squashed? Also the hound complain is easy to fix.

But LGTM basically (not tested)

vuntz:
> Also the hound complain is easy to fix.

As discussed, we actually want to keep the code as it is.

jgrassler:
| Can you use the pattern: component: ... for the commits that are not going to be squashed?

I am using it already: All commits after "neutron/nova: re-added infoblox support" will be squashed since I only left them in there temporarily (in case I needed to revisit them).

rsalevsky:
Needs a rebase.

jgrassler:
Rebase done: Squashed commit messages and rebased on current master (there was a small conflict in default.rb: some line numbers were off due to the addition of the lbaas stuff).

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -29,21 +29,43 @@
...
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects << "pacemaker_group[#{group_name}]"
+
+clone_name = "cl-#{group_name}"
 pacemaker_clone clone_name do
-  rsc primitive_name
-  meta ({ "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node) })
+  rsc group_name
+  meta({"clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node)})
```
rsalevsky: `meta{ "clone-max" =>CrowbarPacemakerHelper.num_corosync_nodes(node) }` should fix the hound issue

jgrassler: It does (I had that exact fix in place at first), but we decided on overruling Hound on this because it makes the code look ambiguous needlessly. Unfortunately Github hid the discussion in the course of a force push, but here it is: https://github.com/crowbar/crowbar-openstack/pull/521#discussion_r81309533

PR #522, opened 2016-09-05
Skip installation of qemu-block-rbd on s390x

dirkmueller:
This only exists on platforms for which ceph exists, which is
aarch64 and x86_64 right now.


aplanas:
+1

vuntz:
+1

PR #525, opened 2016-09-14
Only start barbican-keystone-listener when enabled

dirkmueller:
Otherwise starting aborts and the barclamp apply fails


rhafer:
+1

jgrassler:
+1

PR #406, opened 2016-06-14
OpenDayLight Integration with Neutron

mmnelemane:
The integration allows configuration of OpenDayLight Controller as the manager for the openvswitch switches on the compute and controller nodes. The UI is not fully complete but allows for selecting opendaylight as a ML2 Mechanism driver and set the ODL parameters in the raw window to apply the changes. Requires installation of crowbar/crowbar-opendaylight to setup opendaylight controller on the controller node.

PR #526, opened 2016-09-14
Opendaylight Neutron integration

mmnelemane:
The integration allows configuration of OpenDayLight Controller as the manager for the openvswitch switches on the compute and controller nodes. The UI is not fully complete but allows for selecting opendaylight as a ML2 Mechanism driver and set the ODL parameters in the raw window to apply the changes. Requires installation of crowbar/crowbar-opendaylight to setup opendaylight controller on the controller node.


mmnelemane:
Duplicate PR in https://github.com/crowbar/crowbar-openstack/pull/406

PR #527, opened 2016-09-15
swift: install keystonemiddleware package in case of keystone auth

aplanas:
switf-proxy can be configured with keystone auth. Paste will try
to load the authtoken filter_factory, that resides in the package
python-keystonemiddleware.

Without this package, the service proxy-swift will fail to load
with an error:

ImportError: No module named keystonemiddleware.auth_token

### DIFF
--- a/chef/cookbooks/swift/recipes/dispersion.rb
+++ b/chef/cookbooks/swift/recipes/dispersion.rb
@@ -105,13 +105,14 @@ template "/etc/swift/dispersion.conf" do
   #notifies :run, "execute[populate-dispersion]", :immediately
 end
 
-# NOTE(toabctl): swift-dispersion-populate process hangs so add a timeout as workaround
+# NOTE(aplanas): swift-dispersion-populate process can be slow, and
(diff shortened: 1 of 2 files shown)


rsalevsky:
Can you explain in the commit message why this is required?

aplanas:
* Add explanation in the commit message

rsalevsky:
+1

aplanas:
* Add a commit to fix bsc#989683

nicolasbock:
+1

dirkmueller:
Ignoring LGTM..

aspiers:
I'm not sure why the "swift: remove swift-dispersion-populate timeout" commit was piggy-backed onto this PR - to me it looks like an independent topic. Unfortunately it has caused quite a lot of breakage of the gate. Also in general I think any execution of a non-trivial command which redirects both STDOUT and STDERR to `/dev/null` is a really bad idea.  As a result, in this case the gate is breaking and we have no idea why because the error is being thrown away.  So in the future please don't do that.

aplanas:
As commented, the fail is in the only_if, not in command.

PR #530, opened 2016-09-22
nova: Allow to define a custom network for xCAT access (bsc#976778)

rhafer:


rhafer:
> LGTM, but migration file is mising

Argh. I simply forgot to ```git add``` it. Should be there now. Thanks.

PR #531, opened 2016-09-26
aarch64/s390x always support KVM

dirkmueller:
KVM support was implemented in s370 Second Edition, so anything
s390 or newer supports it.

aarch64 also has no feature flag to disable KVM


tboerger:
The bot seriously still pings people not being involved in the project anymore??? @aspiers

PR #532, opened 2016-09-28
neutron: Simplify code to get nova_fixed interface

rhafer:
We're using the Barclamp::Inventory.Network class now.

Note: Barclamp::Inventory.Network.interface() returns the interface name
including the vlan tag if use_vlan is true for the network while
Barclamp::Inventory.Network.lookup_interface_info didn't include it.
This is not a real problem for "nova_fixed". As "use_vlan" is always set to
"false" when the "nova_fixed" network is enabled on any node (see
NeutronService.rb).

### DIFF
--- a/chef/cookbooks/neutron/libraries/helpers.rb
+++ b/chef/cookbooks/neutron/libraries/helpers.rb
@@ -30,9 +30,7 @@ module NeutronHelper
     fixed_physnet = ""
     if node[:crowbar_wall][:network][:nets][:nova_fixed]
       nova_fixed_net = BarclampLibrary::Barclamp::Inventory.get_network_by_type(node, "nova_fixed")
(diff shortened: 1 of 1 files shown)


rhafer:
This needs to be merged before: https://github.com/crowbar/crowbar-core/pull/681

rhafer:
Gating failed because of a stupid typo in my patch. Should be fixed now.

rhafer:
Now gating failed with a timeout in " execute[populate-dispersion] (swift::dispersion)". I guess this is unrelated. Restarting.

aspiers:
@rhafer Regarding the gate failure, see [swift populate-dispersion timeout (Trello)](https://trello.com/c/ZepcSmJ4/223-swift-populate-dispersion-timeout) and today's IRC discussion with @aplanas / @toabctl.

PR #534, opened 2016-09-28
magnum: Use correct port for endpoint

jgrassler:
This commit sets the port used for Magnum's service catalog entries to
`node[:magnum][:bind_port]`, as opposed to the port retrieved from `MagnumHelper`
that was used before this commit. The port retrieved from `MagnumHelper` is the
service's actual listening port, which is incorrect in an HA setting. There,
the endpoint catalog entries must use the port where the haproxy instance
forwarding to the actual service port listens.

### DIFF
--- a/chef/cookbooks/magnum/recipes/api.rb
+++ b/chef/cookbooks/magnum/recipes/api.rb
@@ -20,7 +20,7 @@
 keystone_settings = KeystoneHelper.keystone_settings(node, :magnum)
 network_settings = MagnumHelper.network_settings(node)
 
-magnum_port = network_settings[:api][:bind_port]
+magnum_port = node[:magnum][:api][:bind_port]
 magnum_protocol = node[:magnum][:api][:protocol]
 
 ha_enabled = node[:magnum][:ha][:enabled]


chef/cookbooks/magnum/recipes/api.rb
```diff
@@ -21,6 +21,7 @@
 network_settings = MagnumHelper.network_settings(node)
 
 magnum_port = network_settings[:api][:bind_port]
```
toabctl: ```magnum_port``` is no longer used. Can be removed or just use it instead of ```endpoint_port```

jgrassler: Ah, I overlooked that, thanks. Fixed.

aplanas:
gating broker because swif/ceilometer issue. Rebuilding.

chef/cookbooks/magnum/recipes/api.rb
```diff
@@ -20,7 +20,7 @@
 keystone_settings = KeystoneHelper.keystone_settings(node, :magnum)
 network_settings = MagnumHelper.network_settings(node)
```
vuntz: You can actually remove that line, can't you?

PR #539, opened 2016-10-03
nova: Remove ec2 from enabled_ssl_apis in nova.conf

vuntz:
EC2 support has been moved to another project, out of nova.


sayalilunkad:
+1

PR #540, opened 2016-10-04
swift-proxy-server: increase number of workers (workaround bsc#989683)

aplanas:


chef/cookbooks/swift/templates/default/proxy-server.conf.erb
```diff
@@ -29,7 +29,10 @@ user = <%= @user %>
 # cores in the system.  It's worth noting that individual workers will
 # use many eventlet co-routines to service multiple concurrent requests.
 # workers = auto
-workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
+# Workaround for bsc#989683. Sometimes swift-proxy service get blocked
+# in the evenlet loop, and the worker become useless.
+# workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
+workers = 6
```
nicolasbock: I don't understand the problem: Why does raising the number of workers work around this bug? As it is right now, we use `workers >= 2`, i.e. there are always at least 2 workers. So if one blocks there should be another one.

aspiers:
Please could you provide a proper explanation in the commit message in the future?  Thanks.

PR #541, opened 2016-10-05
swift: Fix variable replacement in error message

toabctl:
When deploying swift and not enough storage nodes are available, the error
message was:

Need at least as many swift-storage nodes as zones; only %{swift-storage}
     swift-storage node was set for 2 zones.


rsalevsky:
Feel free to merge this PR once the gating is green.

aplanas:
Ceilometer / swift again in the gating : (

PR #542, opened 2016-10-06
Add configuration option for multipath (bsc#997607)

s-t-e-v-e-n-k:
Currently, mulitpath support is set by default in the cinder and
nova cookbooks, and both cookbooks force it to be true, but the
built package overrides nova's setting to false.

Remove that default support, and add a new option to cinder which allows
operators to enable or disable multipath support for both cinder
and nova.

### DIFF
--- a/chef/cookbooks/nova/recipes/config.rb
+++ b/chef/cookbooks/nova/recipes/config.rb
@@ -133,10 +133,13 @@ keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
 rbd_enabled = false
 
+use_multipath = false
+
 cinder_servers = search_env_filtered(:node, "roles:cinder-controller") || []
 if cinder_servers.length > 0
   cinder_server = cinder_servers[0]
   cinder_insecure = cinder_server[:cinder][:api][:protocol] == "https" && cinder_server[:cinder][:ssl][:insecure]
+  use_multipath = cinder_server[:cinder][:use_multipath]
 
   if node.roles.include? "nova-compute-kvm"
     cinder_server[:cinder][:volumes].each do |volume|
@@ -310,7 +313,6 @@ template "/etc/nova/nova.conf" do
             ec2_dmz_host: public_api_host,
             libvirt_migration: node[:nova]["use_migration"],
             migration_host: migration_host,
-            libvirt_enable_multipath: node[:nova][:libvirt_use_multipath],
             shared_instances: node[:nova]["use_shared_instance_storage"],
             force_config_drive: node[:nova]["force_config_drive"],
             glance_server_protocol: glance_server_protocol,
@@ -335,6 +337,7 @@ template "/etc/nova/nova.conf" do
             neutron_has_tunnel: neutron_has_tunnel,
             keystone_settings: keystone_settings,
             cinder_insecure: cinder_insecure || keystone_settings["insecure"],
+            use_multipath: use_multipath,
             ceph_user: ceph_user,
             ceph_uuid: ceph_uuid,
             ssl_enabled: api[:nova][:ssl][:enabled],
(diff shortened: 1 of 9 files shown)


chef/cookbooks/nova/recipes/config.rb
```diff
@@ -151,6 +154,14 @@
   include_recipe "nova::ceph"
 end
 
+if use_multipath
+  package "multipath-tools"
+
+  service "multipathd" do
+    action [:enable, :start]
+  end
+end
```
rhafer: Hm, it seems wrong to move this to the config recipe. Isn't that part only needed on the compute nodes? AFAICS config.rb is also run on the controller.

chef/data_bags/crowbar/migrate/cinder/102_use_multipath.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["use_multipath"] = false
```
rhafer: Usually we do use the value from the template here. So:

```a["use_multipath"] = ta["use_multipath"]```

File: chef/cookbooks/nova/recipes/config.rb (hunk 1 above)
Comment (s-t-e-v-e-n-k) [inline chef/cookbooks/nova/recipes/config.rb]: I was trying to avoid duplicating the cinder servers bit, but I didn't consider that the cookbooks are run on the wrong hosts, I'll switch it around.

File: chef/cookbooks/nova/recipes/config.rb [hunk 3]
```diff
@@ -133,10 +133,13 @@
 
 rbd_enabled = false
 
+use_multipath = false
+
 cinder_servers = search_env_filtered(:node, "roles:cinder-controller") || []
 if cinder_servers.length > 0
   cinder_server = cinder_servers[0]
   cinder_insecure = cinder_server[:cinder][:api][:protocol] == "https" && cinder_server[:cinder][:ssl][:insecure]
+  use_multipath = cinder_server[:cinder][:use_multipath]
```
Itxaka: `use_multipath = cinder_server[:cinder][:use_multipath] ? true : false` to save a var :P

toabctl:
@s-t-e-v-e-n-k please prefix the commit msg with ```cinder:```

chef/cookbooks/nova/recipes/config.rb
toabctl: the variable is used in line 340. I like it the way it is currently

chef/data_bags/crowbar/migrate/cinder/102_use_multipath.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["use_multipath"] = ta["use_multipath"]
```
toabctl: Do we want to backport this to Cloud6? If yes, you need to check if the key is already avaialble (for the upgrade from C6 to C7)

s-t-e-v-e-n-k: Yes, that was the plan, that's a good point, I'll make that change.

s-t-e-v-e-n-k:
I am going to quickly re-test this tomorrow dealing with review comments, and the hound violation.

chef/data_bags/crowbar/migrate/cinder/102_use_multipath.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "use_multipath"
+    a["use_multipath"] = ta["use_multipath"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  del a["use_multipath"]
```
dirkmueller: unless ta.key? "use_multipath"

chef/data_bags/crowbar/migrate/cinder/102_use_multipath.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "use_multipath"
+    a["use_multipath"] = ta["use_multipath"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless a.key? "use_multipath"
```
dirkmueller: unless ta.key..

PR #543, opened 2016-10-06
Fix XEN Kernel recognition

rsalevsky:
Since SP2 there is no separate XEN Kernel anymore.

### DIFF
--- a/chef/cookbooks/nova/libraries/boot_kernel.rb
+++ b/chef/cookbooks/nova/libraries/boot_kernel.rb
@@ -25,7 +25,10 @@ module NovaBootKernel
 
     # trigger reboot through reboot_handler, if kernel-$flavor is not yet
     # running
-    node.run_state[:reboot] = true unless `uname -r`.include?(flavor)
+    unless Dir.exist?("/proc/xen") && flavor == "xen" ||
+        !Dir.exist?("/proc/xen") && flavor == "default"
+      node.run_state[:reboot] = true
+    end
   end
 
   protected
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -143,7 +143,7 @@ case node[:nova][:libvirt_type]
             action :nothing
             supports status: true, start: true, stop: true, restart: true
             # restart xend only when xen kernel is already present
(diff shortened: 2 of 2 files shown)


aplanas:
is this fixing bsc#1001993?

rsalevsky:
Looks like the same issue. I will update the commit message once gating is green.

chef/cookbooks/nova/libraries/boot_kernel.rb
```diff
@@ -25,7 +25,7 @@ def self.set_boot_kernel_and_trigger_reboot(node, flavor = "default")
 
     # trigger reboot through reboot_handler, if kernel-$flavor is not yet
     # running
-    node.run_state[:reboot] = true unless `uname -r`.include?(flavor)
+    node.run_state[:reboot] = true unless Dir.exist?("/proc/xen")
```
bmwiedemann: I think, this would not work for the case where you want to switch back to kernel-default without xen.
There you would need to use something like
```
is_xen xor flavor = "xen"
```

chef/cookbooks/nova/libraries/boot_kernel.rb
```diff
@@ -25,7 +25,9 @@ def self.set_boot_kernel_and_trigger_reboot(node, flavor = "default")
 
     # trigger reboot through reboot_handler, if kernel-$flavor is not yet
     # running
-    node.run_state[:reboot] = true unless `uname -r`.include?(flavor)
```
bmwiedemann: I thought, the uname -r is always returning the same, because xen uses kernel-default now.

rsalevsky: Yes, but do we have only xen and default als flavor?

rsalevsky:
The fix works but we hit now another issue while rebooting. See https://trello.com/c/1zhuA44m/46-debug-memballoon-xen

chef/cookbooks/nova/libraries/boot_kernel.rb
dguitarbite: +1, I think this check is necessary. Xen will not work otherwise (Type I).

rsalevsky: This part of the code is out of date. ;)

PR #544, opened 2016-10-06
dispersion: restart swift-proxy service to avoid bsc#1003366

aplanas:
extreme workaround to bsc#1003366. Python 2.7.9 and greenlet 0.4.5
produce GIL deadlock in this service. A restart usually resolves the
issue in local testing. This needs to be removed once the bug is
fixed.


aplanas:
@rsalevsky lets wait gating to see if help.

Also there is an old card created in the keep-jenkins-green board. The bug is new, because is a better description of the issue.

PR #545, opened 2016-10-07
Added a location constraint to ensure services are started on upgraded nodes only

jsuchome:
Requires 
- https://github.com/crowbar/crowbar-ha/pull/146
- https://github.com/crowbar/crowbar-ha/pull/149

I've added this everywhere where we use the location requirement for controller nodes.

### DIFF
--- a/chef/cookbooks/aodh/recipes/aodh_ha.rb
+++ b/chef/cookbooks/aodh/recipes/aodh_ha.rb
@@ -83,6 +83,11 @@ end
 location_name = openstack_pacemaker_controller_only_location_for clone_name
 transaction_objects << "pacemaker_location[#{location_name}]"
 
+if CrowbarPacemakerHelper.being_upgraded?(node)
+  upgrade_location_name = upgraded_only_location_for clone_name
+  transaction_objects << "pacemaker_location[#{upgrade_location_name}]"
+end
+
 pacemaker_transaction "aodh" do
   cib_objects transaction_objects
   # note that this will also automatically start the resources
--- a/chef/cookbooks/ceilometer/recipes/central_ha.rb
+++ b/chef/cookbooks/ceilometer/recipes/central_ha.rb
@@ -35,6 +35,11 @@ transaction_objects << "pacemaker_primitive[#{service_name}]"
 location_name = openstack_pacemaker_controller_only_location_for service_name
 transaction_objects << "pacemaker_location[#{location_name}]"
 
+if CrowbarPacemakerHelper.being_upgraded?(node)
+  upgrade_location_name = upgraded_only_location_for service_name
+  transaction_objects << "pacemaker_location[#{upgrade_location_name}]"
+end
+
 pacemaker_transaction "ceilometer central" do
   cib_objects transaction_objects
   # note that this will also automatically start the resources
--- a/chef/cookbooks/ceilometer/recipes/mongodb.rb
+++ b/chef/cookbooks/ceilometer/recipes/mongodb.rb
@@ -69,6 +69,11 @@ if ha_enabled
   location_name = openstack_pacemaker_controller_only_location_for clone_name
   transaction_objects << "pacemaker_location[#{location_name}]"
 
+  if CrowbarPacemakerHelper.being_upgraded?(node)
+    upgrade_location_name = upgraded_only_location_for clone_name
+    transaction_objects << "pacemaker_location[#{upgrade_location_name}]"
+  end
+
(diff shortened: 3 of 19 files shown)


chef/cookbooks/aodh/recipes/aodh_ha.rb
```diff
@@ -83,6 +83,11 @@
 location_name = openstack_pacemaker_controller_only_location_for clone_name
 transaction_objects << "pacemaker_location[#{location_name}]"
 
+if CrowbarPacemakerHelper.being_upgraded?(node)
```
jsuchome: Here I'm not sure if I have to check for `being_upgraded?`, because we're also checking for this condition inside of `upgraded_only_location_for`.

However, leaving it only inside `upgraded_only_location_for` would mean that the location is still created, but not used, right? Or what would be put into `transaction_objects` in case we do not do this in a block with `if CrowbarPacemakerHelper.being_upgraded?(node)`?

aspiers: > Here I'm not sure if I have to check for `being_upgraded?`, because we're also checking for this condition inside of `upgraded_only_location_for`.

We should definitely avoid *all* this repetition, not just the `being_upgraded?` check.  It should be possible to add a single line in 29 places, not 4 almost identical lines in 29 places.

> However, leaving it only inside `upgraded_only_location_for` would mean that the location is still created, but not used, right?

No - why do you think that?  If the `only_if` block evaluates to false then the resource's action will not be executed.

> Or what would be put into `transaction_objects` in case we do not do this in a block with `if CrowbarPacemakerHelper.being_upgraded?(node)`?

Yes, that is a real problem.  I don't have a good solution yet, as mentioned in https://github.com/crowbar/crowbar-ha/pull/149/files#r82425013

At worst we can avoid the repetition via a library helper, I guess.

jsuchome:
[(13) Orchestrate upgrade of controller nodes in HA setup (upgrade step 12 part 2)](https://trello.com/c/2PMGauJx/118-13-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2)

chef/cookbooks/aodh/recipes/aodh_ha.rb
jsuchome: Yes, different helper will help here

aspiers: Sorry - I was half wrong when I wrote:
> > However, leaving it only inside upgraded_only_location_for would mean that the location is still created, but not used, right?
>
> No - why do you think that? If the only_if block evaluates to false then the resource's action will not be executed.

Whilst it's true that the location resource's action (i.e. `:update`) will not be executed unless the node is being upgraded, if the resource is contained in `transaction_objects`, then the `:commit_new` action will create the constraint, and we don't want that to happen unless we're within the upgrade workflow.

> I guess we can pass transaction_objects to the helper, and the helper then modifies it.

I think you meant "definition" rather than "helper" here; the definition calls the helper, and the helper merely constructs a Pacemaker configuration string which `crmsh` can understand.

But I wonder if that approach of passing something to modify is a tiny bit ugly.  The helper might have to ensure that the constraint gets added in the right place, after the resource it refers to, if Pacemaker is fussy about ordering of new objects within transactions.  Well, maybe it's not too bad.

The problem stems from the fact that `only_if` gets evaluated during Chef's execution phase, whereas this `if CrowbarPacemakerHelper.being_upgraded?(node)` gets evaluated during Chef's compile phase.  So we can't rely on the result of `only_if` to determine whether or not to add the constraint to `transaction_objects`.  But that's OK, because definitions are also processed during Chef's compile phase, and we can safely rely on the `being_upgraded?` check during the compile phase.  (This is still duplicating the same check across two phases, but the duplication only exists once within the definition rather than in every barclamp, so IMHO it's an acceptably small DRY violation.)  So I think we can simply move `if CrowbarPacemakerHelper.being_upgraded?(node)` inside the definition, and either

a) return `nil` if `being_upgraded?` is false, in which case the caller doesn't add anything to `transaction_objects`, or
b) follow @vuntz's suggestion of having the definition insert the new constraint into `transaction_objects` if `being_upgraded?` is true.

Either way, this will require changes to https://github.com/crowbar/crowbar-ha/pull/149.

aspiers:
@jsuchome IIUC this is now obsolete  as we are doing it a different way - is that right?

jsuchome:
Oh, true

PR #546, opened 2016-10-07
keystone: Use apache frontend

toabctl:
The native frontend (eventlet based) is no longer available and upstream
recommends to use a "real" webserver so use Apache.

### DIFF
--- a/chef/data_bags/crowbar/template-keystone.json
+++ b/chef/data_bags/crowbar/template-keystone.json
@@ -4,7 +4,7 @@
   "attributes": {
     "keystone": {
       "debug": false,
-      "frontend": "native",
+      "frontend": "apache",
       "verbose": true,
       "use_syslog": false,
       "policy_file": "policy.json",
@@ -118,7 +118,7 @@
     "keystone": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 101,
+      "schema-revision": 102,
       "element_states": {
         "keystone-server": [ "readying", "ready", "applying" ]
       },
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -134,6 +134,8 @@ elsif node[:keystone][:frontend] == "apache"
     service_name node[:keystone][:service_name]
     supports status: true, restart: true
     action [:disable, :stop]
+    # allow to fail here because there may not be a service "keystone"
+    ignore_failure true
   end
 
   include_recipe "apache2"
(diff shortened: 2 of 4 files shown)


chef/data_bags/crowbar/template-keystone.json
```diff
@@ -4,7 +4,7 @@
   "attributes": {
     "keystone": {
       "debug": false,
-      "frontend": "native",
+      "frontend": "apache",
```
dirkmueller: you need to add a migration for this.

aplanas:
Note: keystone/recipes/server.rb still try to start the service

chef/data_bags/crowbar/template-keystone.json
toabctl: @dirkmueller fixed

toabctl:
> Note: keystone/recipes/server.rb still try to start the service

Where? It tries to stop and I think we should allow to fail there (if the init script no longer exists after an upgrade)

aplanas:
@toabctl but this will work in a non-upgrade scenario? There is not keystone service, but there is a `:disable, :stop` action for the frontend `apache` branch.

toabctl:
@aplanas hm. true. we should allow to fail the stop action. Or remove it completely?

aplanas:
@toabctl not sure, I think that during the upgrade process all the services are down, so maybe is ok to remove it. @jsuchome can provide more info.

toabctl:
@aplanas better now?

aplanas:
something is missing in the `auth_token` configuration: apache wakeup properly but the keystone_register wakeup service fails. The logs shows 

    File "/usr/lib/python2.7/site-packages/keystonemiddleware/auth_token/__init__.py", line 542, in _fetch_token raise NotImplemented()
        TypeError: 'NotImplementedType' object is not callable

This method is implemented in `AuthProtocol` class. Keystone-paste looks OK

aplanas:
I found the problem in the keystonemiddleware package. The current version is 4.0.0, but newton requires >=4.5.0 <=4.9.0

toabctl:
> I found the problem in the keystonemiddleware package. The current version is 4.0.0, but newton requires >=4.5.0 <=4.9.0

Yeah. Sorry. I thought I fixed that already but the package build fails in D:C:7:Newton.

toabctl:
@dirkmueller are the changes ok? Can we merge this?

toabctl:
> Only one last suggestion here: update the wsgi from https://github.com/openstack/keystone/blob/stable/newton/httpd/wsgi-keystone.conf

@aplanas I think @rhafer already synced that a while ago with the config upstream provides.

PR #547, opened 2016-10-07
ceilometer: Use auth_type to password (bsc#1001273)

aplanas:
Ceilometer 7.0 change the section service_credentials.  If auth_type
is not set, some queries to keytone will use the legacy v2.0 API,
and this cause problems (POST can be blocked in SSL connections).

If we set this value to `password` V3 will be used.

Update `username` parameter to `user-name`, to avoid deprecation
warning.

### DIFF
--- a/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
+++ b/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
@@ -1506,6 +1506,7 @@ interface = internalURL
 # Authentication type to load (string value)
 # Deprecated group/name - [DEFAULT]/auth_plugin
 #auth_type = <None>
+auth_type = password
 
 # Config Section from which to load plugin specific options (string value)
 #auth_section = <None>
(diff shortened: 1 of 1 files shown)


rsalevsky:
Can you add the bsc number to the commit message header? Then the build service will track this issue.

aplanas:
@rsalevsky not sure what you mean here. The bsc is referenced in the commit message and in the comment.

toabctl:
> @rsalevsky not sure what you mean here. The bsc is referenced in the commit message and in the comment.

@aplanas I think @rsalevsky wants to have the bug number in the first line of the commit message. Like ```ceilometer: Use auth_type to password (bsc #1001273)```

aplanas:
@toabctl I understand now, thanks!

@rsalevsky done!

AFAIK OBS detect the bsc reference code in the commit description too. But clearly is better to put it in the commit title for the changelog.

nicolasbock:
Gating passed without ssl. I started a rebuild with ssl.

nicolasbock:
`manila_tempest_tests.tests.api.admin.test_share_networks.ShareNetworkAdminTest` and `manila_tempest_tests.tests.api.test_share_networks.ShareNetworksTest' are failing now. This doesn't seem to be connected to ceilometer.

PR #548, opened 2016-10-11
[stable/3.0] Backport #549

mmnelemane:
@mfriesenegger found issue in reading configuration file for opflex agent. This commit fixes that.

### DIFF
--- a/chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
+++ b/chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
@@ -14,7 +14,7 @@
       "ca-store": "/etc/ssl/certs/"
     },
     "inspector": {
-    "enabled": true,
+      "enabled": true,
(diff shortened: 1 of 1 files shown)


chef/cookbooks/neutron/templates/default/10-opflex-agent-ovs.conf.erb
```diff
@@ -39,8 +39,8 @@
       "ovs-bridge-name": "br-int",
       "encap": {
         "vxlan" : {
-          "encap-iface": <%= @opflex_vxlan_encap_iface %>,
-          "uplink-iface": <%= @opflex_vxlan_uplink_iface %>,
+          "encap-iface": "<%= @opflex_vxlan_encap_iface %>",
+          "uplink-iface": "<%= @opflex_vxlan_uplink_iface %>",
```
aplanas: Good catch. I am worried about that there are some boolean values with quotes too below. Also I guess that vlan and port (below) are numeric values?

PR #549, opened 2016-10-11
Fix for Fatal error in reading opflex config file

mmnelemane:
@mfriesenegger identified a fatal error in reading opflex config file. This commit is to fix that and correct few other values.


rsalevsky:
Was there a bug report?

mmnelemane:
@rsalevsky : there wasn't a bug report. It came up when Mike was testing on Cisco hardware and we had a short discussion and mail exchanges on this.

rsalevsky:
Ok, can you add the bsc number to the commit message? So `Fix for Fatal error in reading opflex config file (bsc#1004056)` would be perfect.

PR #550, opened 2016-10-11
Raise availability-zone timeout consistently

dirkmueller:
There were two places that set availability zones. Set both to the same,
higher timeout to accomodate slowness on s390x


dguitarbite:
+1

PR #551, opened 2016-10-11
Configure sahara plugins

toabctl:


toabctl:
@rsalevsky This is needed for the full tempest run. We need to configure the ```fake``` plugin (I'll created another automation PR for that) so the sahara tempest tests can pass.

toabctl:
With this, ```# ostestr  -r '^sahara.tests' ``` all theses tests pass for me.

Itxaka:
+1

Will we need to backport this to 3.0?

PR #552, opened 2016-10-11
tempest: Disable personality tests

toabctl:
The tests are currently failing when running a full tempest run.
The tests for personality can not pass because they need an enabled
"inject_partition" in nova.conf .

### DIFF
--- a/chef/cookbooks/tempest/templates/default/tempest.conf.erb
+++ b/chef/cookbooks/tempest/templates/default/tempest.conf.erb
@@ -514,6 +514,8 @@ interface_attach = <%= @use_interface_attach %>
 
 # Does the test environment support server personality (boolean value)
 #personality = true
+# NOTE(toabctl): to use this, we need to enable inject_partition in nova.conf
+personality = false
 
 # Does the test environment support attaching an encrypted volume to a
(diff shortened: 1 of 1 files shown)


toabctl:
@rsalevsky another PR which "fixes" 2 tempest tests from the full testrun. I guess the real fix would be to enable the inspect mode via libguestfs but for the next milestone this should help.
Failing tests are:
```
tempest.api.compute.servers.test_server_personality.ServerPersonalityTestJSON.test_can_create_server_with_max_number_personality_files [18.898299s] ... FAILED
tempest.api.compute.servers.test_server_personality.ServerPersonalityTestJSON.test_create_server_with_personality
```
from https://ci.suse.de/job/openstack-mkcloud/36866/consoleFull

toabctl:
@vuntz maybe enabling the libguestfs support to inject files into the guest is something we want?

toabctl:
> @vuntz maybe enabling the libguestfs support to inject files into the guest is something we want?

I had a quick look howto enable it but failed: https://bugzilla.suse.com/show_bug.cgi?id=1004185

sayalilunkad:
+1

PR #553, opened 2016-10-12
manila: Fix keystone default domain settings

toabctl:
The settings for the different clients (nova, cinder, neutron) must be
set. Otherwise the manila-share service complains with:

Expecting to find domain in project - the server could not comply with the \
          request since it is either malformed or otherwise incorrect. The \
          client is assumed to be in error. (HTTP 400)


toabctl:
@rsalevsky This helps with plenty of failing manila tempest tests during the full test run (see https://ci.suse.de/job/openstack-mkcloud/36960/console)

toabctl:
All manila api tests (```ostestr -r '^manila_tempest_tests.tests.api'```) are passing with this.

sayalilunkad:
+1

PR #557, opened 2016-10-13
Tempest: Fixes typo in config

sayalilunkad:


chef/cookbooks/tempest/templates/default/tempest.conf.erb
```diff
@@ -1201,7 +1201,7 @@ dhcp_client =
 
 # Whether or not cinder is expected to be available (boolean value)
 #cinder = true
-cinder = <%= @use_docker ? false : true %>
+cinder = <%= @use_cinder ? false : true %>
```
vuntz: Hrm, I think that part was correct, as I don't believe we have a `use_cinder` variable.

sayalilunkad: Oh ok. So it does depend on docker but in case of docker it should always be false?

toabctl: you need to have a look if the test itself respects the ```cinder``` flag from tempest I think

PR #561, opened 2016-10-18
Tuning rabbitmq (bsc #1004284)

mmnelemane:
Adding tuning parameters for rabbitmq as suggested from http://www.rabbitmq.com/networking.html.


chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -21,6 +21,30 @@
...
+                               "net.ipv4.tcp_keepalive_intvl = 3")
+    rc.insert_line_if_no_match(/^net.ipv4.tcp_keepalive_probes = 6/,
+                               "net.ipv4.tcp_keepalive_probes = 6")
+    rc.insert_line_if_no_match(/^net.core.wmem_max = 16777216/, "net.core.wmem_max = 16777216")
+    rc.insert_line_if_no_match(/^net.core.rmem_max = 16777216/, "net.core.rmem_max = 16777216")
+    rc.insert_line_if_no_match(/^net.ipv4.tcp_fin_timeout = 20/, "net.ipv4.tcp_fin_timeout = 20")
+    rc.insert_line_if_no_match(/^net.ipv4.tcp_tw_reuse = 1/, "net.ipv4.tcp_tw_reuse = 1")
+    rc.insert_line_if_no_match(/^net.core.somaxconn = 4096/, "net.core.somaxconn = 4096")
+    rc.insert_line_if_no_match(/^net.ipv4.tcp_syncookies = 0/, "net.ipv4.tcp_syncookies = 0")
+    rc.insert_line_if_no_match(/^net.ipv4.tcp_max_syn_backlog = 8192/,
```
dirkmueller: these parameters are already set elsewhere in crowbar.

mmnelemane: @dirkmueller: Among the parameters set here, I see that only rmem_max and wmem_max are set in the network barclamp only for 1G/10G interfaces. Do you suggest to move these settings to some place in network barclamp ?

vuntz: We should not change /etc/sysctl.conf, but add a new file in /etc/sysctl.d

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -20,11 +20,12 @@
 
 package "rabbitmq-server"
 package "rabbitmq-server-plugins" if node[:platform_family] == "suse"
+tuning_file = "/etc/sysctl.d/60-sysctl-rabbitmq.conf"
 
 directory "/etc/rabbitmq/" do
   owner "root"
   group "root"
-  mode 0755
+  mode "0755"
```
toabctl: unrelated

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -20,11 +20,12 @@
```
toabctl: nit: commit msg should start with ```rabbitmq: ...```

vuntz:
This needs more research and discussion, as the settings the way they are may not make sense. Closing for now.

PR #562, opened 2016-10-18
Add another constraint to the controller only locati

jsuchome:
Make sure that the services supposed to run on controller nodes
do not get started when the node is in pre-upgrade state.

See also https://bugzilla.suse.com/show_bug.cgi?id=1005346

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/ha_helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/ha_helpers.rb
@@ -32,7 +32,7 @@
 module OpenStackHAHelper
   def self.controller_only_location(location, service)
     "location #{location} #{service} resource-discovery=exclusive " \
-      "rule 0: OpenStack-role eq controller"
+      "rule 0: OpenStack-role eq controller and pre-upgrade eq false"
   end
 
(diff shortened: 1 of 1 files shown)


jsuchome:
Alternatively, we could pass an argument to `openstack_pacemaker_controller_only_location_for` definition (https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/crowbar-openstack/definitions/openstack_pacemaker_controller_only_location_for.rb) to create `pacemaker_location` with one or two constraints

aspiers:
Although if you have 5 spare seconds perhaps you could fix the commit message :)

aspiers:
@rhafer You were right to be suspicious about the reversed logic - it totally broke HA.  Hopefully fixed by #568.

aspiers:
BTW I am sorry for not spotting this in my review. But I think we can all blame the lack of CI anyway :-)  This is exactly why I submitted [#stableCI multiple configs for Crowbar repo PR gates (Trello)](https://trello.com/c/IcWmgI8E/30-8-stableci-multiple-configs-for-crowbar-repo-pr-gates) which will catch this kind of thing in the future.

PR #564, opened 2016-10-19
Use "service" resource agent where "lsb" is used

toabctl:
The "service" resource agent intelligently figures out which system service
is used. It tries first LSB and then systemd.
This is useful while we transition to systemd services in our packages.


chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -203,22 +203,22 @@
 default[:neutron][:lbaas][:f5][:parent_ssl_profile] = "clientssl"
 
 default[:neutron][:ha][:network][:enabled] = false
-default[:neutron][:ha][:network][:l3_ra] = "lsb:#{node[:neutron][:platform][:l3_agent_name]}"
-default[:neutron][:ha][:network][:lbaas_ra] = "lsb:#{node[:neutron][:platform][:lbaas_agent_name]}"
-default[:neutron][:ha][:network][:lbaasv2_ra] = "lsb:#{node[:neutron][:platform][:lbaasv2_agent_name]}"
+default[:neutron][:ha][:network][:l3_ra] = "service:#{node[:neutron][:platform][:l3_agent_name]}"
+default[:neutron][:ha][:network][:lbaas_ra] = "service:#{node[:neutron][:platform][:lbaas_agent_name]}"
+default[:neutron][:ha][:network][:lbaasv2_ra] = "service:#{node[:neutron][:platform][:lbaasv2_agent_name]}"
 default[:neutron][:ha][:network][:f5_ra] = "systemd:#{node[:neutron][:platform][:f5_agent_name]}"
```
aplanas: we want to move this too to `service:`?

toabctl: fixed

toabctl:
> Can you submit a similar PR for crowbar-ha (there's one use of lsb: there)?

@vuntz It's already in a ```if SLE11``` so that doesn't matter. And where we already have ```systemd:``` I wouldn't change it to ```service:```. But I could remove the ```if SLE11``` code I think.

vuntz:
Failure of HA gating is in the ceph testuite, and related to the move to SES 4, not to this PR.

aspiers:
Just realised that this PR rules out one of the approaches suggested for resolving [bsc#980341 (systemd thinks it owns services defined by lsb primitives)](https://bugzilla.suse.com/show_bug.cgi?id=980341), since `service:` prefers LSB to systemd.

aspiers:
#597 partially reverts this.

PR #566, opened 2016-10-19
Fix typo and make heat tested in tempest

vmoravec:


dirkmueller:
(waiting for gating..)

PR #568, opened 2016-10-21
don't break whole cluster when not upgrading

aspiers:
https://github.com/crowbar/crowbar-openstack/pull/562 totally broke HA cluster setup, because when the pre-upgrade node attribute is not present, the location constraints generated by this helper prevent anything useful from running.

So instead we reverse the

    pre-upgrade eq false

logic to

    pre-upgrade ne true

so that the constraints will *only* prevent resources from running when pre-upgrade is set to true, and not when it isn't set at all.

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/ha_helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/ha_helpers.rb
@@ -32,7 +32,7 @@
 module OpenStackHAHelper
   def self.controller_only_location(location, service)
     "location #{location} #{service} resource-discovery=exclusive " \
-      "rule 0: OpenStack-role eq controller and pre-upgrade eq false"
+      "rule 0: OpenStack-role eq controller and pre-upgrade ne true"
   end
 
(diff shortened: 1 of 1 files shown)


aspiers:
**DO NOT MERGE UNTIL THIS IS TESTED WITH HA ENABLED**

aspiers:
Just triggered a build with `hacloud=1`.

aspiers:
[P2: cloud-mkcloud7-job-ha-compute-x86_64](https://trello.com/c/ZKRtrPzC/227-p2-cloud-mkcloud7-job-ha-compute-x86-64)

aspiers:
[Error executing action `create` on resource 'ruby_block[wait for fs-postgresql to be started]'](https://trello.com/c/tHUXTxDt/287-error-executing-action-create-on-resource-ruby-block-wait-for-fs-postgresql-to-be-started)

aspiers:
HA gate has already got to trove which proves it's fixed the issue.

PR #575, opened 2016-10-27
manila: Add NetApp driver for managed mode

toabctl:
Manila has for some drivers different modes. DHSS=True means
in the NetApp case, that the vserver creation in handled by the Manila
driver. Support this mode in addition to the DHSS=False case.


toabctl:
I don't plan to work on this. Closing.

PR #576, opened 2016-10-27
Deploy keystone wsgi mod with SSL

nicolasbock:
When using the wsgi apache module we need to enable SSL for it not in
the `keystone.conf` but in apache.

---

This change is [Reviewable](https://reviewable.io/reviews/crowbar/crowbar-openstack/576)

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -157,6 +157,8 @@ elsif node[:keystone][:frontend] == "apache"
     "${APACHE_LOG_DIR}"
   end
 
+  apache_module "ssl" if node[:keystone][:api][:protocol] == "https"
+
   template "#{node[:apache][:dir]}/sites-available/keystone.conf" do
     path "#{node[:apache][:dir]}/vhosts.d/keystone.conf" if node[:platform_family] == "suse"
     source "apache_keystone.conf.erb"
@@ -166,6 +168,10 @@ elsif node[:keystone][:frontend] == "apache"
       bind_admin_host: bind_admin_host,
       bind_service_port: bind_service_port, # public port
       bind_service_host: bind_service_host,
+      ssl_enable: node[:keystone][:api][:protocol] == "https",
+      ssl_certfile: node[:keystone][:ssl][:certfile],
+      ssl_keyfile: node[:keystone][:ssl][:keyfile],
+      ssl_ca_certs: node[:keystone][:ssl][:ca_certs],
       processes: 3,
       threads: 10
     )
@@ -245,7 +251,6 @@ template "/etc/keystone/keystone.conf" do
       ssl_enable: (node[:keystone][:frontend] == "native" && node[:keystone][:api][:protocol] == "https"),
       ssl_certfile: node[:keystone][:ssl][:certfile],
       ssl_keyfile: node[:keystone][:ssl][:keyfile],
-      ssl_cert_required: node[:keystone][:ssl][:cert_required],
       ssl_ca_certs: node[:keystone][:ssl][:ca_certs],
       rabbit_settings: fetch_rabbitmq_settings
     )
@@ -403,7 +408,6 @@ if node[:keystone][:api][:protocol] == "https"
     keyfile node[:keystone][:ssl][:keyfile]
     group node[:keystone][:group]
     fqdn node[:fqdn]
-    cert_required node[:keystone][:ssl][:cert_required]
     ca_certs node[:keystone][:ssl][:ca_certs]
   end
 end
--- a/chef/cookbooks/keystone/templates/default/apache_keystone.conf.erb
+++ b/chef/cookbooks/keystone/templates/default/apache_keystone.conf.erb
@@ -7,6 +7,13 @@ Listen <%= @bind_service_host %>:<%= @bind_service_port %>
     WSGIPassAuthorization On
     LimitRequestBody 114688
 
+<% if @ssl_enable %>
+    SSLEngine on
+    SSLCertificateFile <%= @ssl_certfile %>
+    SSLCertificateKeyFile <%= @ssl_keyfile %>
+    SSLCACertificateFile <%= @ssl_ca_certs %>
+<% end %>
+
     <IfVersion >= 2.4>
       ErrorLogFormat "%M"
     </IfVersion>
@@ -36,6 +43,13 @@ Listen <%= @bind_admin_host %>:<%= @bind_admin_port %>
     WSGIApplicationGroup %{GLOBAL}
     WSGIPassAuthorization On
     LimitRequestBody 114688
+
+<% if @ssl_enable %>
+    SSLEngine on
+    SSLCertificateFile <%= @ssl_certfile %>
+    SSLCertificateKeyFile <%= @ssl_keyfile %>
+    SSLCACertificateFile <%= @ssl_ca_certs %>
+<% end %>
 
     <IfVersion >= 2.4>
       ErrorLogFormat "%M"
--- a/chef/data_bags/crowbar/migrate/keystone/103_remove_cert_required.rb
+++ b/chef/data_bags/crowbar/migrate/keystone/103_remove_cert_required.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["ssl"].delete("cert_required")
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["ssl"]["cert_required"] = ta["ssl"]["cert_required"]
+  return a, d
+end
--- a/chef/cookbooks/keystone/attributes/default.rb
+++ b/chef/cookbooks/keystone/attributes/default.rb
@@ -58,7 +58,6 @@ default[:keystone][:ssl][:certfile] = "/etc/keystone/ssl/certs/signing_cert.pem"
 default[:keystone][:ssl][:keyfile] = "/etc/keystone/ssl/private/signing_key.pem"
 default[:keystone][:ssl][:generate_certs] = false
 default[:keystone][:ssl][:insecure] = false
-default[:keystone][:ssl][:cert_required] = false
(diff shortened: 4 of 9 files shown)


nicolasbock:
Once non-SSL gating is through, I'll re-trigger with SSL.

nicolasbock:
Keystone is deploying now with SSL. The gating failure is something else I think. I am looking into it right now.

nicolasbock:
The openstack client is failing on `openstack network list`.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/crowbar/crowbar-openstack/576#-:-KVBBGg7w4oZT4l4f-ck:b-5bb0lq)*

rsalevsky:
[cloud-mkcloud7-job-ssl-x86_64](https://trello.com/c/Ex9ZJ8bd/246-cloud-mkcloud7-job-ssl-x86-64)

toabctl:
@nicolasbock Isn't there a ```a2enmod ssl``` missing to be sure ssl is enabled? Also (nitpick), please prefix the commit message with ```keystone: ...```

chef/cookbooks/keystone/templates/default/apache_keystone.conf.erb
```diff
@@ -7,6 +7,14 @@ Listen <%= @bind_service_host %>:<%= @bind_service_port %>
     WSGIPassAuthorization On
     LimitRequestBody 114688
 
+<% if @ssl_enable %>
+    SSLEngine on
+    SSLVerifyDepth 1
```
toabctl: 1 is the default.

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -166,6 +168,11 @@
       bind_admin_host: bind_admin_host,
       bind_service_port: bind_service_port, # public port
       bind_service_host: bind_service_host,
+      ssl_enable: node[:keystone][:api][:protocol] == "https",
+      ssl_certfile: node[:keystone][:ssl][:certfile],
+      ssl_keyfile: node[:keystone][:ssl][:keyfile],
+      ssl_cert_required: node[:keystone][:ssl][:cert_required],
```
toabctl: you are not using this anywhere. What is the intention of this config parameter?

chef/cookbooks/keystone/templates/default/apache_keystone.conf.erb
```diff
@@ -37,6 +45,14 @@ Listen <%= @bind_admin_host %>:<%= @bind_admin_port %>
...
```
toabctl: same here. 1 is the default

chef/cookbooks/keystone/recipes/server.rb
nicolasbock: Hmm, a copy and paste oversight :)

nicolasbock:
@toabctl I have updated the PR

chef/cookbooks/keystone/recipes/server.rb
toabctl: If it's no longer used, please remove ```cert_required``` from the schema and from the proposal via a migration.

nicolasbock: We are still using it in `keystone.conf`: https://github.com/nicolasbock/crowbar-openstack/blob/5d92c39b2a4d00bff02bebbe9f564140bb5d6304/chef/cookbooks/keystone/recipes/server.rb#L254

nicolasbock:
@toabctl Could you have another look?

toabctl:
@nicolasbock the ```cert_required``` option is still there. I know it's used in our template but we can remove it there because this is only used in the case where eventlet is used (and because our template is still Mitaka based). And that's no longer possible. So please remove ```cert_required``` with a migration and all places where it's used.
The change from upstream where all this got removed is https://review.openstack.org/#/c/249486/

toabctl:
Alternative is to remove the ```cert_required``` option together with all other eventlet related options when we update the templates. @nicolasbock you can choose :)

nicolasbock:
@toabctl I removed `cert_required`. I am not 100% sure I got the migration right. Could you double check it? Thanks!

nicolasbock:
@toabctl done.

nicolasbock:
@toabctl Could you have another look please?

chef/data_bags/crowbar/migrate/keystone/103_remove_cert_required.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  a["ssl"].delete("cert_required") \
+    unless ta["ssl"].key?("cert_required")
```
toabctl: why not just deleting it?

nicolasbock:
@toabctl I patched a fresh 'instonly' deploy and now the keystone barclamp webform is broken. The SSL option is grayed out and I can't select `https` anymore. I suspect that removing https://github.com/crowbar/crowbar-openstack/pull/576/files#diff-fca7695ab0df9a673d75e40a60e495e9L40 is the problem, but I am not sure how to fix it at this point. More reading rails/haml I guess :)

toabctl:
> @toabctl I patched a fresh 'instonly' deploy and now the keystone barclamp webform is broken. The SSL option is grayed out and I can't select https anymore. I suspect that removing https://github.com/crowbar/crowbar-openstack/pull/576/files#diff-fca7695ab0df9a673d75e40a60e495e9L40 is the problem, but I am not sure how to fix it at this point. More reading rails/haml I guess :)

I think you just did not recompile the assets. your patch works for me

chef/cookbooks/keystone/templates/default/apache_keystone.conf.erb
```diff
@@ -7,6 +7,13 @@ Listen <%= @bind_service_host %>:<%= @bind_service_port %>
     WSGIPassAuthorization On
     LimitRequestBody 114688
 
+<% if @ssl_enable %>
+    SSLEngine on
+    SSLCertificateFile  <%= @ssl_certfile %>
```
aplanas: nit: double space

chef/cookbooks/keystone/templates/default/apache_keystone.conf.erb
```diff
@@ -37,6 +44,13 @@ Listen <%= @bind_admin_host %>:<%= @bind_admin_port %>
...
```
aplanas: nit: double space

nicolasbock:
@toabctl I'll try that again.

aplanas:
Update is not working. If I move from keystone w/out ssl to ssl using this patch, the system try to connect to the http endpoint:

```bash
export OS_AUTH_URL='https://d52-54-77-77-77-01.vo7.cloud.suse.de:5000/v3/'
openstack --debug endpoint list --insecure
```

This produce this:

```
...
REQ: curl -g -i --insecure -X GET https://d52-54-77-77-77-01.vo7.cloud.suse.de:5000/v3/ -H "Accept: application/json" -H "User-Agent: osc-lib keystoneauth1/2.12.2 python-requests/2.11.1 CPython/2.7.9"
...
...
"POST /v3/auth/tokens HTTP/1.1" 201 14549
... "catalog": [{"endpoints": [{"url": "http://10.165.198.2:5000/v3/", "interface": "public", "region": "CustomRegion", "region_id": "CustomRegion", "id": "ef6ecb9179784f569e07a45bf92adb88"}, {"url": "http://d52-54-77-77-77-01.vo7.cloud.suse.de:5000/v3/", "interface": "internal", "region": "CustomRegion", "region_id": "CustomRegion", "id": "0fde6e969aad4c83b14a9b8f6e45f4e3"}, {"url": "http://d52-54-77-77-77-01.vo7.cloud.suse.de:35357/v3/", "interface": "admin", "region": "CustomRegion", "region_id": "CustomRegion", "id": "7977bf0440e64994ade7fbde2bc8bccf"}], "type": "identity", "id": "725a2732f9ca4312bbd35ff3420cbea9", "name": "keystone"},
...
...
Making authentication request to https://d52-54-77-77-77-01.vo7.cloud.suse.de:5000/v3/auth/tokens
...
REQ: curl -g -i --insecure -X GET http://d52-54-77-77-77-01.vo7.cloud.suse.de:35357/v3/ -H "Accept: application/json" -H "User-Agent: osc-lib keystoneauth1/2.12.2 python-requests/2.11.1 CPython/2.7.9"
...
404
```

So everything goes well and use the HTTPS, that generate a good response. But the catalog contain the HTTP URL, so fail after the authentication, when ask for the service.

Is possible to update the catalog before, or the migration from HTTP to HTTPS is not supported / expected?

Also I need to create manually `/etc/keystone/credential-keys` directory

nicolasbock:
@aplanas Thanks for testing! I'll have to check how to update the endpoints during a change. I'll also have to investigate that directory. I can't find any reference to it in the cookbooks.

nicolasbock:
@aplanas I looked a bit into the endpoint challenge. Right now we are bootstrapping keystone the first time we deploy a keystone proposal. If we change the ssl option for keystone, then we will have to add new endpoints and remote the old ones. I tried to simply add a `keystone_register` block to the keystone recipe but that didn't quite work as expected. It adds the new endpoints but it fails at deleting the old ones. After a change the endpoints contain both ssl and non-ssl versions. I think this will require a bit more careful work.

toabctl:
@aplanas , @nicolasbock I think the problem with switching from non-ssl to ssl is unrelated to this PR. I'm pretty sure this was broken before. I would ignore it (create an extra bug report for it) and check the case if a fresh deployment with ssl works or not.

nicolasbock:
@toabctl Yes I agree. I don't think that it's working for a fresh SSL deployment either though. Neutron is not starting up. I am looking into it.

toabctl:
The neutron thing is another problem. Can we merge this, @aplanas , @nicolasbock ? it is solving the mentioned bug...

toabctl:
I filled bsc#1011199 for the followup problem

PR #578, opened 2016-10-28
Use systemd resource for aodh

dirkmueller:


toabctl:
Retriggering gate with ```hacloud=1```

toabctl:
I don't think we need this.The ```service`` RA really just checks for a file in ```/etc/init.d``` and used that if available. Otherwise it uses the systemd service. So -1 for now

dirkmueller:
not needed anymore, globally switched to systemd instead

PR #579, opened 2016-10-31
nova: Install openstack-nova-consoleauth for suse/rhel

toabctl:
The service is always started some lines later for suse/rhel
so install the packages before using the service.


chef/cookbooks/nova/recipes/vncproxy.rb
```diff
@@ -27,11 +27,12 @@
   end
 end
 
+package "openstack-nova-consoleauth" if %w(rhel suse).include?(node[:platform_family])
```
toabctl: It's consistent with the rest of the code so I'll ignore hound here.

PR #581, opened 2016-11-01
Adjust 100_add_tunnel_csum migration to allow backport

rhafer:
Don't touch the tunnel_csum attribute if it already exists before
migration


rhafer:
> please prefix the commit msg with neutron:

fixed

PR #582, opened 2016-11-01
[stable/3.0] Add tunnel_csum attribute to neutron barclamp

rhafer:
(cherry picked from commit eca6eb435189d1c63c431e7df4ae2af97d527d57)

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -223,6 +223,7 @@ if neutron[:neutron][:networking_plugin] == "ml2"
         use_l2pop: neutron[:neutron][:use_l2pop] &&
             (ml2_type_drivers.include?("gre") || ml2_type_drivers.include?("vxlan")),
         dvr_enabled: neutron[:neutron][:use_dvr],
+        tunnel_csum: neutron[:neutron][:ovs][:tunnel_csum],
         bridge_mappings: bridge_mappings
       )
     end
--- a/chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
@@ -186,6 +186,9 @@ enable_distributed_routing = True
 # carrying GRE/VXLAN tunnel. The default value is False.
 #
 # tunnel_csum = False
+<% if @tunnel_csum -%>
+tunnel_csum = True
+<% end -%>
 
 # (StrOpt) agent_type to report.
 # This config entry allows configuration of the neutron agent type reported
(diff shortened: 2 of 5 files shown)


chef/data_bags/crowbar/migrate/neutron/49_add_tunnel_csum.rb
```diff
@@ -0,0 +1,14 @@
+def upgrade(ta, td, a, d)
+  a["ovs"] ||= {}
+  a["ovs"]["tunnel_csum"] = ta["ovs"]["tunnel_csum"]
+
+  return a, d
+end
+
```
nicolasbock: Do you want to fold https://github.com/crowbar/crowbar-openstack/pull/581 into this PR?

rhafer: It's not really needed here. Do you think I should?

nicolasbock: Well, here is what I was thinking: The migrations are run every time the barclamp is run. If the user sets `a['ovs']['tunnel_csum']` to something other than what's in the default attributes it will be overwritten. But it's very possible that I am misunderstanding when migrations are applied, and it's not necessary to add extra checks here :smile:.

rhafer: Ok. So actually the migrations on run when a barclamp is installed (by ```barclamp_install```). And specific migration is only ever executed if the ```schema_revision``` of the barclamp to install is greater than the ```schema_revision``` of the already installed barclamp.

So changes done in https://github.com/crowbar/crowbar-openstack/pull/581 are only needed because we backport the migration to stable/3.0. That means it might be that the ```a["ovs"]["tunnel_csum"]``` attribute already exists when the migration in master is executed.

We would need the same check here if we planned to backport it to even older releases. Which I strongly doubt, because the tunnel_csum option doesn't have an effect on the older openvswitch versions we're using there.

nicolasbock: Thanks for the explanation. It's smarter than I thought :smile:. Please ignore my previous comment then. We shouldn't make the migration code more complicated than is necessary.

rhafer:
I added the do not merge label here for now as this change requires an updated openvswitch to really have an effect. (it doesn't break anything though when the old openvswitch is still installed.

dirkmueller:
well, then we can still merge it if it doesn't break anything, right?

rhafer:
> Should we merge this if it doesn't affect anything so we dont have this PR lingering around?

Yeah, true. Sorry for the late reply I somehow overlooked this.

Rebased the PR and remove the label.

PR #584, opened 2016-11-03
nova: install libosinfo to provide hardware properties when driver is libvirt

sidthe:
For the libvirt driver, by default hardware properties will be retrieved from the Glance image and if such haven’t been provided, it will use a libosinfo database to get those values.

### DIFF
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -59,6 +59,9 @@ case node[:nova][:libvirt_type]
         action :nothing
       end.run_action(:install)
 
+      # install libosinfo to provide hardware properties when driver is libvirt
+      package "typelib-1_0-Libosinfo-1_0"
+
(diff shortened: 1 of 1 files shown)


dguitarbite:
+1

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -59,6 +59,9 @@
         action :nothing
       end.run_action(:install)
 
+      # install libosinfo to provide hardware properties when driver is libvirt
+      package "typelib-1_0-Libosinfo-1_0"
```
toabctl: I think on Ubuntu/Debian, the package name is different (```gir1.2-libosinfo-1.0```. Please add a ```if``` or set ```package_name```

toabctl: I missed the big ```if``` which is already there for suse

sidthe: yes, thx! the if there is a deal breaker

PR #586, opened 2016-11-07
glance: remove conversion task api support from glance

sidthe:
The conversion task API for glance is being deprecated. The PR removes the attributes and tasks related to this service, effectively reverting this PR https://github.com/crowbar/crowbar-openstack/commit/2123499b53c7820c8cb8850a99dfe9d171dfc3a1 as well as migrates the schema

### DIFF
--- a/chef/cookbooks/glance/templates/default/glance-api.conf.erb
+++ b/chef/cookbooks/glance/templates/default/glance-api.conf.erb
@@ -1861,8 +1861,7 @@ flavor = keystone
 # estimations and you should do them based on the worst case scenario
 # and be prepared to act in case they were wrong. (string value)
 #work_dir = <None>
-work_dir = <%= node[:glance][:conversion][:work_dir] %>
-
+work_dir = /var/lib/glance/taskflow
 
 [taskflow_executor]
 
@@ -1885,6 +1884,3 @@ work_dir = <%= node[:glance][:conversion][:work_dir] %>
 # using the RBD backend, this should be set to 'raw' (string value)
 # Allowed values: qcow2, raw, vmdk
 #conversion_format = <None>
-<% if node[:glance][:conversion][:enabled] -%>
-conversion_format = <%= node[:glance][:conversion][:conversion_format] %>
-<% end -%>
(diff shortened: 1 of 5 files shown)


chef/cookbooks/glance/templates/default/glance-api.conf.erb
```diff
@@ -1861,7 +1861,6 @@ flavor = keystone
 # estimations and you should do them based on the worst case scenario
 # and be prepared to act in case they were wrong. (string value)
 #work_dir = <None>
-work_dir = <%= node[:glance][:conversion][:work_dir] %>
```
aplanas: I am not sure that it is OK to remove this, according to the description. This is related to any asynchronous operation, that can be used outside the conversion API

sidthe: agreed offline with Alberto that for now, it will make sense to hardcode the work_dir path instead of creating another attribute group

chef/cookbooks/glance/templates/default/glance-api.conf.erb
```diff
@@ -1861,8 +1861,7 @@ flavor = keystone
...
-
+work_dir = /var/lib/glance/taskflow
```
sidthe: @aplanas is that allright?

chef/cookbooks/glance/templates/default/glance-api.conf.erb
```diff
@@ -1885,6 +1884,3 @@ work_dir = <%= node[:glance][:conversion][:work_dir] %>
 # using the RBD backend, this should be set to 'raw' (string value)
 # Allowed values: qcow2, raw, vmdk
 #conversion_format = <None>
-<% if node[:glance][:conversion][:enabled] -%>
-conversion_format = <%= node[:glance][:conversion][:conversion_format] %>
```
toabctl: Hm. I remember that I also read something about the deprecation but now looking into the code ( https://git.openstack.org/cgit/openstack/glance/tree/glance/async/flows/convert.py?h=master#n35 ) it seems that it's not deprecated. Can you clarify please?

vuntz: @toabctl the feature is deprecated upstream since Mitaka, and scheduled for removal. So we don't want to have people start using it now...

toabctl: Ok. Found it now on http://docs.openstack.org/releasenotes/glance/mitaka.html#upgrade-notes

PR #587, opened 2016-11-08
api: replace (tenant_id) with (project_id) in service endpoint registration urls

sidthe:
Keystone supports $(project_id)s in the catalog. It works the same as $(tenant_id)s. Use of $(tenant_id)s is deprecated and catalog endpoints should be updated to use $(project_id)s.


rhafer:
@sidthe: I'd really appreciate if you put this sentence from the pull request summary into the commit message:

> Keystone supports $(project_id)s in the catalog. It works the same as $(tenant_id)s. Use of $(tenant_id)s is deprecated and catalog endpoints should be updated to use $(project_id)s.

That way it ends up in the git log output. Otherwise this looks good to me.

sidthe:
@rhafer thanks! is that what you expected?

PR #588, opened 2016-11-08
neutron/nova: re-added infoblox support

jgrassler:
This is a backport of https://github.com/crowbar/crowbar-openstack/pull/521 (master). Please do not merge this until https://github.com/crowbar/crowbar-openstack/pull/521 is merged (I don't think there are going to be any more changes to the pull request for `master`, but let's wait for it to land to be on the safe side).

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -1488,7 +1488,7 @@ executor_thread_pool_size = 256
 # The Drivers(s) to handle sending notifications. Possible values are
 # messaging, messagingv2, routing, log, test, noop (multi valued)
 #notification_driver =
-notification_driver = messaging
+notification_driver = messagingv2
 
 # AMQP topic used for OpenStack notifications. (list value)
 # Deprecated group/name - [rpc_notifier2]/topics
--- a/crowbar_framework/app/models/neutron_service.rb
+++ b/crowbar_framework/app/models/neutron_service.rb
@@ -74,6 +74,22 @@ class NeutronService < PacemakerServiceObject
     answer
   end
 
+  def save_proposal!(prop, options = {})
+    # Fill in missing defaults for infoblox grid configurations
+    if prop.raw_data[:attributes][:neutron][:use_infoblox]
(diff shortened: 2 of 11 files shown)


chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -1488,7 +1488,7 @@ executor_thread_pool_size = 256
 # The Drivers(s) to handle sending notifications. Possible values are
 # messaging, messagingv2, routing, log, test, noop (multi valued)
 #notification_driver =
-notification_driver = messaging
+notification_driver = messagingv2
```
toabctl: Why that? IIRC we had a spike about moving to v2 and there were some problems. Does this work now?

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -74,6 +74,22 @@ def proposal_dependencies(role)
     answer
   end
 
+  def save_proposal!(prop, options = {})
+    # Fill in missing defaults for infoblox grid configurations
```
toabctl: Not sure, but can't we do that during the migration?

chef/cookbooks/nova/templates/default/nova.conf.erb
jgrassler: The infoblox plugin needs Nova to speak `messagingv2` in order to understand Nova's notifications. I initially made this configurable with the Neutron barclamp's infoblox part  setting it on an as-required basis, but we ended up settling for hardwiring this in order to avoid extra complexity and code that manipulates other barclamps' proposals. As far as I am aware it hasn't caused major trouble (this has been active in `master` and thus Cloud 7 for quite a while now).

crowbar_framework/app/models/neutron_service.rb
jgrassler: No, because there can be multiple grid configurations (we only have one in the list right now, because support for that is [not supported in the infoblox plugin, yet](https://github.com/openstack/networking-infoblox/blob/stable/liberty/doc/source/installation.rst#neutron), but that may change in the future).

PR #589, opened 2016-11-08
Revert "neutron: Fix device_driver for Neutron Lbaas agent"

toabctl:
This reverts commit 208ec23e473ac64a6603f50192254aa327c55c80.
It does not work with Newton. The error (neutron-lbaas.log) on Newton is:

Could not load neutron_lbaas.services.loadbalancer.drivers.haproxy.\
      namespace_driver.HaproxyNSDriver


jgrassler:
Some context on this: `neutron_lbaas.services.loadbalancer.drivers.haproxy. namespace_driver.HaproxyNSDriver` does exist in `stable/mitaka` but no longer in `stable/newton`.

PR #590, opened 2016-11-09
nova: Allocate IPs on compute nodes from the ceph_client network

toabctl:
crowbar-ceph got recently CephFS support. With that, the instances
running on compute nodes need access to the Ceph cluster to be able
to mount shares.
For that, the default network for ceph client traffic changed to
the public network (set via the "ceph_client" attribute).

See https://github.com/crowbar/crowbar-ceph/commit/4f893f6bbf6c3 for
details about the new ceph_client attribute.

### DIFF
--- a/crowbar_framework/app/models/glance_service.rb
+++ b/crowbar_framework/app/models/glance_service.rb
@@ -132,6 +132,25 @@ class GlanceService < PacemakerServiceObject
       end
     end
 
+    # allocate a IP from the ceph_client network if Ceph is used
+    ceph_proposal = Proposal.find_by(barclamp: "ceph")
+    network_proposal = Proposal.find_by(barclamp: "network")
+
+    if ceph_proposal
+      ceph_client = ceph_proposal["attributes"]["ceph"]["client_network"]
+      # is the ceph_client network really available?
+      if network_proposal["attributes"]["network"]["networks"][ceph_client].nil?
+        raise I18n.t(
+          "barclamp.#{@bc_name}.validation.ceph_client_network_not_available",
+          ceph_client: ceph_client
+        )
+      end
+      server_nodes.each do |n|
+        @logger.info("Allocating an IP from the Ceph client network '#{ceph_client}' for node #{n}")
+        net_svc.allocate_ip "default", ceph_client, "host", n
+      end
+    end
+
     # Setup virtual IPs for the clusters
     allocate_virtual_ips_for_any_cluster_in_networks_and_sync_dns(server_elements, vip_networks)
 
--- a/crowbar_framework/app/models/nova_service.rb
+++ b/crowbar_framework/app/models/nova_service.rb
@@ -306,6 +306,9 @@ class NovaService < PacemakerServiceObject
 
     neutron_service = NeutronService.new @logger
 
+    ceph_proposal = Proposal.find_by(barclamp: "ceph")
+    network_proposal = Proposal.find_by(barclamp: "network")
+
     compute_nodes_for_network.each do |n|
       neutron_service.enable_neutron_networks(neutron["attributes"]["neutron"],
                                               n, net_svc,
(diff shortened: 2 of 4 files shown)


crowbar_framework/app/models/nova_service.rb
```diff
@@ -314,6 +316,13 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
         net_svc.allocate_ip("default", role.default_attributes["nova"]["migration"]["network"],
                             "host", n)
       end
+
+      # allocate a IP from the ceph_client network if Ceph is used
+      if ceph_proposal
```
toabctl: Hm. I think that recommendation is stupid is that case. I'll ignore it.

jgrassler: Agreed.

crowbar_framework/app/models/glance_service.rb
```diff
@@ -132,6 +132,16 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
       end
     end
 
+    # allocate a IP from the ceph_client network if Ceph is used
+    ceph_proposal = Proposal.find_by(barclamp: "ceph")
+    if ceph_proposal
+      ceph_client = ceph_proposal["attributes"]["ceph"]["client_network"]
```
rhafer: Do we really know that the network referenced here really exists in the network barclamp? Otherwise I'd suggest to check the proposal for it an error out if it doesn't exist. Should be as simple as:

```
if network_proposal["attributes"]["network"]["networks"][ceph_client].nil?
  raise I18n.t(errormsg_here)
end
```

crowbar_framework/app/models/nova_service.rb
```diff
@@ -314,6 +316,13 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+        ceph_client = ceph_proposal["attributes"]["ceph"]["client_network"]
```
rhafer: The same as for the glance barlcamp applies here.

crowbar_framework/app/models/glance_service.rb
toabctl: Done

crowbar_framework/app/models/nova_service.rb
toabctl: Done

crowbar_framework/app/models/glance_service.rb
```diff
@@ -132,6 +132,24 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
 
+    # allocate a IP from the ceph_client network if Ceph is used
+    ceph_proposal = Proposal.find_by(barclamp: "ceph")
+    network_proposal = Proposal.find_by(barclamp: "network")
+
+    if ceph_proposal
+      ceph_client = ceph_proposal["attributes"]["ceph"]["client_network"]
+      # is the ceph_client network really available?
+      if network_proposal["attributes"]["network"]["networks"][ceph_client].nil?
+        err = "ceph_client network '#{ceph_client}' not available in network proposal"
```
rhafer: Hm, shouldn't the error message be translateable? I.e. use I18n.t() here?

toabctl: @rhafer ah. sorry. done.

crowbar_framework/app/models/glance_service.rb
```diff
@@ -132,6 +132,25 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+    # allocate a IP from the ceph_client network if Ceph is used
+    ceph_proposal = Proposal.find_by(barclamp: "ceph")
+    network_proposal = Proposal.find_by(barclamp: "network")
+
+    if ceph_proposal
+      ceph_client = ceph_proposal["attributes"]["ceph"]["client_network"]
+      # is the ceph_client network really available?
+      if network_proposal["attributes"]["network"]["networks"][ceph_client].nil?
+        raise I18n.t(
+          "barclamp.#{@bc_name}.validation.ceph_client_network_not_available",
```
rhafer: Seems you forgot to add the string to ```crowbar_framework/config/locales/glanc/en.yml``` :)

toabctl: @rhafer done (hopefully)

PR #593, opened 2016-11-09
nova: Use keystone v2.0 for compute HA resources

vuntz:
fence_compute doesn't work with v3 yet, unfortunately.


sayalilunkad:
+1 after hound fix.

vuntz:
> +1 after hound fix.

Fixed.

dirkmueller:
This needs to be backported to Cloud6 then.. I really wonder how we ever tested that successfully.. (well we didn't I guess thats why)

PR #595, opened 2016-11-10
Rename cookbook READMEs correctly

s-t-e-v-e-n-k:
There were a bunch of README files with .me extensions, move them to
have the correct .md extension.


aplanas:
I want to note that this was merged with the gating red, and the gating was showing the issue. Maybe a `tail /var/log/crowbar/component_install.log` can helps is to diagnose the issue.

But in any case is not OK to merge with the gating red.

PR #596, opened 2016-11-10
cinder: register cinder v3 service and endpoint

sidthe:
To aid support for API microversions, the Cinder team has decided to up the endpoint of cinder from v2 to v3. The v3 endpoint is similar in functionality to v2, the difference being mainly the introduction of API microversioning support.

### DIFF
--- a/chef/cookbooks/cinder/recipes/api.rb
+++ b/chef/cookbooks/cinder/recipes/api.rb
@@ -130,6 +130,35 @@ keystone_register "register cinder endpoint v2" do
   action :add_endpoint_template
 end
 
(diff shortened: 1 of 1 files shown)


sidthe:
@aplanas actually this adds the v3 as well. I think yes, we want to keep all for compatibility reasons

toabctl:
@sidthe can you adjust the commit msg?now you have to read th  e whole msg. see https://wiki.openstack.org/wiki/GitCommitMessages

sidthe:
thanks for the hint @toabctl goot to know

PR #597, opened 2016-11-10
[Newton+] glance/aodh/ceilometer: Use systemd resource class

toabctl:
Instead of using the service resource class, use directly systemd.
There may be some problems with the service resource class.


toabctl:
Retriggered gate with HA and newtoncloud7

aplanas:
Gating moves from neutron, but fails in horizon

toabctl:
> cinder is missing, nova, magnum, sahara, kerysone and manila too

its not missing. the packages are still using sysv init scripts

aplanas:
Ah OK, so service: for lsb: is working

toabctl:
> Could you address the hound issues? Otherwise looks great!

@nicolasbock done

aspiers:
Would have been helpful to mention that this partially reverts https://github.com/crowbar/crowbar-openstack/pull/564.  Also "There may be some problems with the service resource class" is very vague; can you please give more details? (including a bugzilla link if appropriate)

aspiers:
@toabctl Still waiting for clarification on this. Maybe it was already given via another medium (e.g. email) but I can't find it - really it needs to be clarified here for future reference.

PR #598, opened 2016-11-10
nova: set cache.enabled in nova.conf

sidthe:
Option memcached_servers is deprecated in Mitaka. Operators should use oslo.cache configuration instead. Specifically enabled option under [cache] section should be set to True and the url(s) for the memcached servers should be in [cache]/memcache_servers option.

memcached was already enabled. This PR sets the cache to enabled

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -2609,6 +2609,7 @@ os_region_name=<%= @keystone_settings['endpoint_region'] %>
 # production deployments.  Small workloads (single process) like devstack can
 # use the dogpile.cache.memory backend. (string value)
 #backend = dogpile.cache.null
+backend = dogpile.cache.memcached
 
 # Arguments supplied to the backend module. Specify this option once per
 # argument to be passed to the dogpile.cache backend. Example format:
@@ -2622,6 +2623,7 @@ os_region_name=<%= @keystone_settings['endpoint_region'] %>
 
 # Global toggle for caching. (boolean value)
 #enabled = false
+enabled = true
 
 # Extra debugging from the cache backend (cache keys, get/set/delete/etc
 # calls). This is only really useful if you need to see the specific cache-


sidthe:
@toabctl done

chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -2620,6 +2620,7 @@ os_region_name=<%= @keystone_settings['endpoint_region'] %>
 
 # Global toggle for caching. (boolean value)
 #enabled = false
+enabled = true
```
toabctl: is that enough?what about the option ```#backend = dogpile.cache.null``` ?is that also needed to enable caching?

sidthe: At least according to the current nova.conf guidelines and AFAIK thr scope of this card this was just about that specific parameter.

sidthe:
@dirkmueller is that allright?

chef/cookbooks/nova/templates/default/nova.conf.erb
sidthe: @toabctl regarding #backend = dogpile.cache.null this is for non-production/dev environments only. Otherwise it's either oslo - which is our case here or redis.

vuntz:
> can you remove the setting of memcached_servers then as well ? otherwise this looks incomplete to me

So I don't know about Newton, but in Mitaka, the option is actually still required :/ Without this (and even with the cache.enabled option set), we get back to the "no working console in horizon when HA is deployed" issue.

sidthe:
@dirkmueller taking into account the comment from @vuntz I will be returning back the config option for memcahced_servers

toabctl:
retriggered gating...

chef/cookbooks/nova/templates/default/nova.conf.erb
dirkmueller: from my reading of the code backend needs to be set to ```dogpile.cache.memcached```

sidthe: @dirkmueller thanks! ok, I just looked at it again and you guys are right. Before I change anything - I wonder if instead of hardcoding this to the nova.conf.erb, maybe it makes sense to actually create a new attribute group instead and set a placeholder for both parameters?

dirkmueller: I am not sure this makes sense, we don't want to support any other backend other than memcached any time soon.. this attributification imho only makes sense if we want to make that changeable. but we're always deploying with memcache

sidthe: @dirkmueller thanks! That's fine. Just wanted to know what you think of it.

sidthe:
@dirkmueller changed the parameter and shortened the commit message

PR #599, opened 2016-11-10
manila: Allow to use a HA cluster for manila-share in active/passive

sayalilunkad:


rsalevsky:
Need a rebase.

dirkmueller:
[P3: Investigate manila-share active/passive (SAP)](https://trello.com/c/Nzv9W8Sh/120-p3-investigate-manila-share-active-passive-sap)

chef/cookbooks/manila/recipes/share.rb
```diff
@@ -38,4 +38,37 @@
...
+  use_pacemaker_provider ha_enabled
+end
+
+if ha_enabled
+  log "HA support for manila share is enabled"
+
+  # Create manila-share HA specific config file
+  service_host = CrowbarPacemakerHelper.cluster_vhostname(node)
+
+  template "/etc/manila/manila-share.conf" do
```
vuntz: I guess this needs to be updated due to the change we did in config file locations, see https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/cinder/recipes/volume.rb#L303 and https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/cinder/attributes/default.rb#L20

chef/cookbooks/manila/recipes/role_manila_share.rb
```diff
@@ -16,4 +16,5 @@
 
 if CrowbarRoleRecipe.node_state_valid_for_role?(node, "manila", "manila-share")
   include_recipe "manila::share"
+  include_recipe "manila::share_ha"
```
vuntz: This shouldn't be here, instead it should be in the `if ha_enabled` code. See https://github.com/crowbar/crowbar-openstack/commit/0c87c97f8ed9dec479102ceb6e965c7208a5ae4f#diff-e44ebd5ea17511e644658be6d5ac87bfR314

toabctl:
Closing. Please reopen if you still want to work on this.

PR #602, opened 2016-11-15
fix undefined method `key?' for nil:NilClass

MaximilianMeister:
when running the bin/rake crowbar:schema_migrate_prod


toabctl:
@MaximilianMeister please prefix the commit msg with ```neutron:```

MaximilianMeister:
> please prefix the commit msg with neutron:

@toabctl done

PR #604, opened 2016-11-15
Do not execute dispersion recipe in readying state.

jsuchome:
Restart of swift-proxy from dispersion recipe during readying
(introduced by commit 77ec11da5b7db42bedfdd3829234957af1c28011)
could break pacemaker cluster setup.

### DIFF
--- a/chef/data_bags/crowbar/template-swift.json
+++ b/chef/data_bags/crowbar/template-swift.json
@@ -100,9 +100,9 @@
     "swift": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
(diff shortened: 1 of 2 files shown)


chef/data_bags/crowbar/template-swift.json
```diff
@@ -100,9 +100,9 @@
     "swift": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 22,
+      "schema-revision": 23,
```
vuntz: Bump to 100 please :-)

PR #605, opened 2016-11-16
aodh/ceilometer: put behind apache2

dirkmueller:
The builtin wsgi server in Newton does not read the port setting
from the config file anymore, so any attempt at setting that (which
is needed to be customized for HA) is failing.

### DIFF
--- a/chef/cookbooks/aodh/recipes/aodh.rb
+++ b/chef/cookbooks/aodh/recipes/aodh.rb
@@ -207,9 +207,44 @@ crowbar_pacemaker_sync_mark "create-aodh_db_sync"
 service "aodh-api" do
   service_name node[:aodh][:api][:service_name]
   supports status: true, restart: true, start: true, stop: true
-  action [:enable, :start]
-  subscribes :restart, resources("template[/etc/aodh/aodh.conf]")
-  provider Chef::Provider::CrowbarPacemakerService if ha_enabled
+  action [:disable, :stop]
+  ignore_failure true
+end
+
+include_recipe "apache2"
+if node[:platform_family] == "rhel"
+  package "mod_wsgi"
+else
+  include_recipe "apache2::mod_wsgi"
+end
+apache_module "version"
+
+apache_site "000-default" do
+  enable false
+end
+
+apache_log_dir = if node[:platform_family] == "suse"
+  "/var/log/apache2"
+else
+  "${APACHE_LOG_DIR}"
+end
+
+template "#{node[:apache][:dir]}/sites-available/aodh-api.conf" do
+  path "#{node[:apache][:dir]}/vhosts.d/aodh-api.conf" if node[:platform_family] == "suse"
+  source "apache_aodh.conf.erb"
+  variables(
+    apache_log_dir: apache_log_dir,
+    bind_host: bind_host,
+    bind_port: bind_port,
+    # There is tough science behind those numbers..
+    processes: 3,
+    threads: 3
+  )
+  notifies :restart, resources(service: "apache2"), :immediately
+end
+
+apache_site "aodh.conf" do
+  enable true
 end
 
 service "aodh-evaluator" do
--- a/chef/cookbooks/ceilometer/recipes/server.rb
+++ b/chef/cookbooks/ceilometer/recipes/server.rb
@@ -187,10 +187,53 @@ end
 service "ceilometer-api" do
   service_name node[:ceilometer][:api][:service_name]
   supports status: true, restart: true, start: true, stop: true
-  action [:enable, :start]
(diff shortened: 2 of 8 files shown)


chef/cookbooks/aodh/recipes/aodh.rb
```diff
@@ -207,9 +207,45 @@
...
+  ignore_failure true
+end
+
+include_recipe "apache2"
+if %w(rhel).include?(node[:platform_family])
+  package "mod_wsgi"
+else
+  include_recipe "apache2::mod_wsgi"
+end
+include_recipe "apache2::mod_rewrite"
```
aplanas: Why do we need this?

chef/cookbooks/aodh/recipes/aodh.rb
```diff
@@ -207,9 +207,44 @@
 service "aodh-api" do
```
toabctl: maybe add a comment here that this can be removed for SOC8

chef/cookbooks/aodh/recipes/aodh.rb
```diff
@@ -207,9 +207,44 @@
...
+  enable false
+end
+
+apache_log_dir = if node[:platform_family] == "suse"
+  "/var/log/apache2"
+else
+  "${APACHE_LOG_DIR}"
+end
+
+template "#{node[:apache][:dir]}/sites-available/aodh.conf" do
```
toabctl: maybe ```aodh-api.conf``` ?that's what we use for barbican.conf (but we use also keystone.conf, openstack-dashboard.conf so it's already inconsistent)

chef/cookbooks/aodh/recipes/aodh.rb
```diff
@@ -207,9 +207,44 @@
 service "aodh-api" do
   service_name node[:aodh][:api][:service_name]
   supports status: true, restart: true, start: true, stop: true
-  action [:enable, :start]
-  subscribes :restart, resources("template[/etc/aodh/aodh.conf]")
-  provider Chef::Provider::CrowbarPacemakerService if ha_enabled
+  action [:disable, :stop]
```
toabctl: maybe add a comment here that this can be removed for SOC8

chef/cookbooks/ceilometer/recipes/server.rb
```diff
@@ -187,10 +187,44 @@
...
+  enable false
+end
+
+apache_log_dir = if node[:platform_family] == "suse"
+  "/var/log/apache2"
+else
+  "${APACHE_LOG_DIR}"
+end
+
+template "#{node[:apache][:dir]}/sites-available/ceilometer.conf" do
```
toabctl: same here - ```ceilometer-api.conf``` ?

toabctl:
open questions (but not blocking this PR):
- do we want to remove the systemd service files from the packages?
- howto handle the -config-dir thing with apache?

chef/cookbooks/aodh/recipes/aodh.rb
dirkmueller: we won't use this code anymore for soc 8 :)

chef/cookbooks/aodh/recipes/aodh.rb
dirkmueller: I like it, changed

dirkmueller:
still waiting for packaging to be updated..

chef/cookbooks/aodh/recipes/aodh.rb
toabctl: lets see...

jgrassler:
> open questions (but not blocking this PR):
 > do we want to remove the systemd service files from the packages?

I'd rather err on the side of keeping at least these services that can run standalone running in a standalone fashion. That way people have a fighting chance to get small-scale deployments with standalone API services running on OpenSUSE - otherwise they will have Apache as a hard requirement.

toabctl:
Let's see if HA gating passes

chef/cookbooks/aodh/recipes/aodh.rb
```diff
@@ -207,9 +207,44 @@
...
+end
+
+template "#{node[:apache][:dir]}/sites-available/aodh-api.conf" do
+  path "#{node[:apache][:dir]}/vhosts.d/aodh-api.conf" if node[:platform_family] == "suse"
+  source "apache_aodh.conf.erb"
+  variables(
+    apache_log_dir: apache_log_dir,
+    bind_host: bind_host,
+    bind_port: bind_port,
+    # There is tough science behind those numbers..
```
aplanas: Is an elaborate alias for 'magic number', I like it : ))

PR #609, opened 2016-11-17
tempest: Don't assume that heat is deployed on tempest node (bsc#991873)

vuntz:
https://bugzilla.suse.com/show_bug.cgi?id=991873


aspiers:
#toomanyreviewers

toabctl:
> #toomanyreviewers

we need github-alytics to judtify all these reviewers ;)

PR #610, opened 2016-11-17
magnum: Rename service_type from "container" to "container-infra" bsc#1010673

mjura:
OpenStack Magnum changed service_type name from "container" to "container-infra".
It was done in upstream bug https://bugs.launchpad.net/magnum/+bug/1584251

### DIFF
--- a/chef/cookbooks/magnum/recipes/api.rb
+++ b/chef/cookbooks/magnum/recipes/api.rb
@@ -75,8 +75,8 @@ keystone_register "register magnum service" do
   port keystone_settings["admin_port"]
   auth register_auth_hash
   service_name "magnum"
-  service_type "container"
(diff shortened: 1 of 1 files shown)


chef/cookbooks/magnum/recipes/api.rb
```diff
@@ -75,7 +75,7 @@
   port keystone_settings["admin_port"]
   auth register_auth_hash
   service_name "magnum"
-  service_type "container"
+  service_type "container-infra"
   service_description "Openstack Magnum - Containers as a Service"
```
dirkmueller: the description is "Container Infrastructure Management Service" these days. can you update that as well?

mjura: Thank you for pointing me this, it is done

PR #611, opened 2016-11-17
barbican: Fix typos in HA setup

dirkmueller:
This proposal never applies due to essential typos


toabctl:
Interessting that this was not detected while testing.

dirkmueller:
@toabctl with HA will fail due to the missing aodh fixes in the other PR.. but the fixes in this one are safe to merge and tested

PR #612, opened 2016-11-18
glance: add uid and gid check

matelakat:
Make the recipe fail if the node is not compatible with hardwired
uid/gid/username/groupname constraints.

### DIFF
--- a/chef/cookbooks/glance/recipes/common.rb
+++ b/chef/cookbooks/glance/recipes/common.rb
@@ -22,7 +22,28 @@ package "glance" do
   package_name "openstack-glance" if %w(rhel suse).include?(node[:platform_family])
 end
 
-GlanceHelper.verify_user_and_group_ids(node)
+ruby_block "Adjust glance uid/gid" do
+  block do
+    Chef::Log.info("Verifying uid/gid for glance")
+
+    cmdline = "groupmod -g #{node[:glance][:gid]} #{node[:glance][:group]}"
+    Chef::Log.info("running #{cmdline}")
+    cmd = Chef::ShellOut.new(cmdline).run_command
+    if cmd.exitstatus.nonzero?
+      Chef::Log.error("Failed to update glance uid/gid")
+    end
+
+    cmdline = "usermod -u #{node[:glance][:uid]} -g #{node[:glance][:gid]} #{node[:glance][:user]}"
+    Chef::Log.info("running #{cmdline}")
+    cmd = Chef::ShellOut.new(cmdline).run_command
+    if cmd.exitstatus.nonzero?
+      Chef::Log.error("Failed to update glance uid/gid")
+    end
+
+  end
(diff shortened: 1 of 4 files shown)


nicolasbock:
@matelakat, @vuntz: I tried it out and it works as advertised by @matelakat. However, when a user re-installs a node and that node has a mismatching uid/gid the node ends up in 'problem' state without indication on the crowbar UI what went wrong. The traceback with the error message goes to `/var/log/crowbar/crowbar_join/chef.log` on that node (which is picked up by supportconfig) and is not visible directly on the crowbar node.

nicolasbock:
The CI failure is related to nova and not to this PR.

nicolasbock:
Fixed hound complaints.

nicolasbock:
@vuntz Could you have a look at the last commit? This commit would address your comment regarding what to do with previously installed nodes. For this to work it is necessary to stop all glance service before running `usermod`. So far I haven't quite figured out how to do that.

dirkmueller:
[P2: bsc#1004514: L3: glance uid: changed after reinstallation of a cluster node](https://trello.com/c/45uLzu6M/172-p2-bsc-1004514-l3-glance-uid-changed-after-reinstallation-of-a-cluster-node)

Itxaka:
Shouldnt we update this to do it in a more chef-i way by setting user/group resources like @nicolasbock did in the rabbitmq cookbook?

chef/cookbooks/glance/recipes/common.rb
```diff
@@ -22,6 +22,55 @@
...
+      Chef::Log.info("error from ajdusting glance uid/gid")
+      cmd.stderr.split("\n").each do |l|
+        Chef::Log.info(l)
+      end
+    end
+    if not cmd.stdout.empty?
+      Chef::Log.info("output from ajdusting glance uid/gid")
+      cmd.stdout.split("\n").each do |l|
+        Chef::Log.info(l)
+      end
```
vuntz: Do we really need to put the output in the chef logs?

nicolasbock:
>Does this really do what is intended? I mean, `notifies` is not 
>supposed to work from here.

Good point. I moved the notifies statement outside the `block do` 
block.

>Do we really need to put the output in the chef logs?

I removed those logging statements.

chef/cookbooks/glance/recipes/common.rb
```diff
@@ -22,6 +22,29 @@
   package_name "openstack-glance" if %w(rhel suse).include?(node[:platform_family])
 end
 
+ruby_block "Adjust glance uid/gid" do
```
matelakat: Do we really want to change gids and uids? What's going to happen with files that were already created, having the old gid/uid?

matelakat:
@nicolasbock It might be a good idea to re-open this PR with your name?

nicolasbock:
@matelakat I agree, I'll do that.

nicolasbock:
Closing in favor of #800.

PR #615, opened 2016-11-23
Fix cinder openstack migrations

rsalevsky:


rsalevsky:
[P1: Make frequent chef-client runs configuarble](https://trello.com/c/M4JbV95Y/192-p1-make-frequent-chef-client-runs-configuarble)

PR #616, opened 2016-11-23
keystone: Do not bootstrap keystone after upgrade

jsuchome:
This may be wrong. Also, we need to take care for non-HA case as well, I assume

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -447,7 +447,9 @@ execute "keystone-manage bootstrap" do
   action :run
   only_if do
     !node[:keystone][:bootstrap] &&
-      (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node))
+      (!ha_enabled || (CrowbarPacemakerHelper.is_cluster_founder?(node) &&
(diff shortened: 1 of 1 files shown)


chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -447,7 +447,9 @@
   action :run
   only_if do
     !node[:keystone][:bootstrap] &&
-      (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node))
+      (!ha_enabled || (CrowbarPacemakerHelper.is_cluster_founder?(node) &&
+        !CrowbarPacemakerHelper.being_upgraded?(node))
+      )
```
rhafer: @jsuchome Hm, I somehow thing it would be cleaner to just set node[:keystone][:bootstrap] on the node(s) as part of the upgrade. Similar do how we reset the db_synced flag during the 5->6 upgrade. (Which reminds me, are we already doing that for the 6->7 upgrade somewhere?)

jsuchome: Sounds like a better approach, indeed

jsuchome:
Closing, some new could would be part of upgrade related methods in crowbar-core

PR #619, opened 2016-11-25
upgrade: Do not start neutron-ha-tool for node in the upgrade state

jsuchome:
During the initial post-upgrade chef-client run, network routers
are still present on another node.
We do not want neutron-ha-tool to create new ones automatically, so
we prevent the creation of neutron-ha-tool primitive and hope to handle the situation
manually later.

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -169,6 +169,11 @@ pacemaker_transaction "neutron agents" do
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
+if CrowbarPacemakerHelper.being_upgraded?(node)
+  log "Skipping neutron-ha-tool resource creation during the upgrade"
+  use_l3_agent = false
+end
+
 if use_l3_agent
   # FIXME: neutron-ha-tool can't do keystone v3 currently
   os_auth_url_v2 = KeystoneHelper.versioned_service_URL(keystone_settings["protocol"],


jsuchome:
So... @abel-navarro @rhafer , do I need to do the same for `cl-g-neutron-agents`? Possibly for anything else?

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -189,7 +189,7 @@
       "os_insecure"    => keystone_settings["insecure"] || node[:neutron][:ssl][:insecure]
     })
     op node[:neutron][:ha][:neutron_ha_tool][:op]
-    action :update
+    action CrowbarPacemakerHelper.being_upgraded?(node) ? :create : :update
```
vuntz: Is it useful to create it, though? Shouldn't we simply skip the whole block if `CrowbarPacemakerHelper.being_upgraded?(node)`? (we already have the block inside a `if use_l3_agent`, so should be easy to do).

jsuchome: My idea is to:

1. upgrade node1, let the network traffic go through node2
2. run chef-client on node1
3. kill the routers on node2 and create them on node1. I would assume that when the pacemaker resources are already created at node1, the easy command would be just to call 'crm resource start XY' without any need for creation. 

What would be the other option when we wouldn't have resources reated? Run chef client on node1 again?

jsuchome: There's at least one problem here, though - i'm skipping the transaction

rhafer: > I thought to have this work in an optimized way, we wouldn't want to let neutron-ha-tool do everything on its own, but we would "manually" migrate routers one by one. Which means we don't want the pacemaker resource to do this for us.

Right.

rhafer: > > Ah, so you mean we would even ignore those pacemaker resources at all. (And assume they'd get created on some later regular chef-client run)

> Exactly. But my understanding is partial, so I don't know if it's fine.

Well. We will need to make sure to have them created at some point. But during the upgrade (or maybe at least until e.g. the 2nd network node is upgraded) I'd not create those resources.

jsuchome: So just skip whole `network_agents_ha` recipe? Or is there anything from it that is still needed?

rhafer: Only the part the creates the ha-tool resource and the ordering primitive for it. But isn't your patch already doing that?

Well it creates the ha_tool_transaction_objects but then doesn't commit it. But that effectively is not creating anything, right?

I would probably first try with skipping the entire ```if use_l3_agent``` block starting at line 172

rhafer: @jsuchome BTW, ```CrowbarPacemakerHelper.being_upgraded?(node)``` basically uses ```node["crowbar_wall"]["crowbar_upgrade_step"] == "prepare-os-upgrade"``` as an indication for whether the node is being upgraded. I wonder if this is really the right condition here. At which point is ```node["crowbar_wall"]["crowbar_upgrade_step"]``` deleted (or set back to ```"none"```)? I assume that's happening at some point after crowbar_join completed, though I haven't found the code for that yet? But can we be sure that the manual router migration procedure did run already at that point? Because as soon as chef-client runs again on the node after ```node["crowbar_wall"]["crowbar_upgrade_step"]```` was reset the ha-tool resources will be created.

jsuchome:
[(5) S51P4 Solve network traffic transition during upgrade of controller nodes](https://trello.com/c/QhV7YPbo/185-5-s51p4-solve-network-traffic-transition-during-upgrade-of-controller-nodes)

jsuchome:
I know that 1st commit is not needed with 2nd one, just keeping it now for history...

PR #621, opened 2016-11-30
heat: Minify heat.conf template

dirkmueller:
remove all comments and whitespace in prepration of Newton port


nicolasbock:
This comment is not meant as a criticism, I am just curious: how does this minification help in the Newton port?

toabctl:
nit: commit msg should start with ```heat:```

dirkmueller:
@nicolasbock its part of the grant master plan to only set those settings that we need to be set.. I thought it is easier to get the comment-deletion part in before we actually look at the templates.

jgrassler:
@aplanas: Yes, that's the plan eventually. But this is just a preparatory first step to make sorting the rest out later easier (no useless comments to sift through).

aplanas:
@jgrassler sorry, I do not get it. I was thinking that this exercise with Heat was to test the new layout in a isolated way. Part of the test was moving the new clean config file to the new house. Did I missunderstand this?

jgrassler:
@aplanas:
| remove all comments and whitespace in prepration of Newton port

I understand this as "remove irrelevant noise" so the actual work of testing the new layout requires less scrolling.

nicolasbock:
@dirkmueller Thanks. It definitely is a lot more readable without the comments :)

dirkmueller:
@aplanas yes, but I wanted to do that in a separate pull request (to isolate signal from noise)

dirkmueller:
@aplanas added the move.. will likely not work right now as the packaging changes for that are not yet in the tree.

toabctl:
retriggered gating

PR #623, opened 2016-11-30
magnum: add x509keycert store option

sidthe:
Introduce a new certificate store called x509keypair to magnum.

### DIFF
--- a/crowbar_framework/app/helpers/barclamp/magnum_helper.rb
+++ b/crowbar_framework/app/helpers/barclamp/magnum_helper.rb
@@ -21,7 +21,8 @@ module Barclamp
       options_for_select(
         [
           ["Local", "local"],
-          ["Barbican", "barbican"]
+          ["Barbican", "barbican"],
+          ["Magnum Database (x509keypair)", "x509keypair"]
         ],
         selected.to_s
       )


crowbar_framework/app/helpers/barclamp/magnum_helper.rb
```diff
@@ -21,7 +21,8 @@ def cert_manager_types(selected)
       options_for_select(
         [
           ["Local", "local"],
-          ["Barbican", "barbican"]
+          ["Barbican", "barbican"],
+          ["X509keypair", "x509keypair"]
```
vuntz: Asking out loud: is it possible to have a string that is slightly more understandable than X509keypair in the webui? :-)

sidthe:
@vuntz fair. Let's label it inline with the documentation and have it as Magnum Database. Is that allright?

vuntz:
> @vuntz fair. Let's label it inline with the documentation and have it as Magnum Database. Is that allright?

Actually, I was trying to find what label would be best and your proposal is quite good. Maybe "Magnum Database (x509keypair)" so that it's also explicit what this is for experts?

sidthe:
> Actually, I was trying to find what label would be best and your proposal is quite good. Maybe "Magnum Database (x509keypair)" so that it's also explicit what this is for experts?

I don't have a preference. Googling "magnum database certificate" leads to this result http://docs.openstack.org/developer/magnum/userguide.html which specifically tells what magnum database option is.

shall I append the x509keypair to the name?

vuntz:
> shall I append the x509keypair to the name?

I'd do it, I guess, yes. But I won't fight over it.

mjura:
I would prefer to keep it simple, my proposal is to change it to just "Database" which is most accurate.

sidthe:
@vuntz changed
@mjura simply Database would also work for me. @vuntz has the final word

PR #624, opened 2016-12-01
horizon:  add unversioned keystone endpoint url

sidthe:
It is no longer necessary to include the version suffix into OPENSTACK_KEYSTONE_URL setting. Thanks to a recent update of django-openstack-auth library as of 2.3.0 release, Horizon will append the proper version suffix to the URL based on the value stored inside OPENSTACK_API_VERSIONS[‘identity’] setting.

### DIFF
--- a/chef/cookbooks/keystone/libraries/helpers.rb
+++ b/chef/cookbooks/keystone/libraries/helpers.rb
@@ -46,6 +46,9 @@ module KeystoneHelper
                                                 node[:fqdn],
                                                 node[:keystone][:api][:service_port],
                                                 node[:keystone][:api][:version])
+      unversioned_internal_auth_url = service_URL(node[:keystone][:api][:protocol],
+                                                  node[:fqdn],
+                                                  node[:keystone][:api][:service_port])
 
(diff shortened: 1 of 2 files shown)


chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -40,12 +40,10 @@ def self.keystone_settings(current_node, cookbook_name)
                                    node[:keystone][:api][:admin_port])
       public_auth_url = versioned_service_URL(node[:keystone][:api][:protocol],
```
s-t-e-v-e-n-k: I question the use of versioned_service_URL() without passing a version through.

sidthe:
@s-t-e-v-e-n-k correct. I took it from the wrong angle. But now I fixed it.

toabctl:
@sidthe Can yopu explain in the commit msg why we want todo that? http://docs.openstack.org/developer/horizon/topics/settings.html#openstack-keystone-url  has still a versioned url as default.

sidthe:
@toabctl fixed. Also rebased.

sidthe:
@toabctl I tested it, it works properly

PR #625, opened 2016-12-01
horizon: Do not require other openstack proposals for deployment

sidthe:
Horizon no longer requires Nova (or Glance) to function; it will run as long as keystone is present (for instance, swift-only deployments).

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -298,8 +298,13 @@ else
   neutron_use_vpnaas = false
 end
 
-nova = get_instance("roles:nova-controller")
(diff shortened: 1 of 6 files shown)


sidthe:
@vuntz just to clarify, by removing from horizon proposal, you mean removing from the horizon schema via a schema migration?

vuntz:
> @vuntz just to clarify, by removing from horizon proposal, you mean removing from the horizon schema via a schema migration?

yes

sidthe:
@vuntz is that allright?

vuntz:
> @vuntz is that allright?

Nearly: you need to remove nova_instance from the json file too :-)

sidthe:
@vuntz done

sidthe:
@vuntz that's a little strange. I just rebased and I don't see why this might have happened. I bumped it again, let's see.

PR #628, opened 2016-12-05
nova: Use systemd resource class

toabctl:
Instead of using the service resource class, use directly systemd.
There may be some problems with the service resource class.


toabctl:
Do not merge before https://build.opensuse.org/request/show/443782 was accepted and the packages landed in C:O:Newton

PR #629, opened 2016-12-05
nova/cinder: set shared lock path when collocated with cinder

sidthe:


sidthe:
@vuntz I can add the comment but I would probably prefer to move it in crowbar-openstack cookbook

sidthe:
@vuntz I had to revert back to the original solution, as group and directory resource usage from the library ended up being problematic. I tested the current solution - it performs as expected. I also added the comments.

sidthe:
[cinder/nova: Shared lock_path](https://trello.com/c/mLv0tInF/359-cinder-nova-shared-lock-path)

chef/cookbooks/cinder/recipes/common.rb
```diff
@@ -104,6 +104,22 @@
   bind_port = node[:cinder][:api][:bind_port]
 end
 
+# lock path prevents race conditions for cinder-volume and nova-compute on same
+# node. Keep code in sync between cinder and nova recipe
+need_shared_lock_path = node.roles.include?("cinder-volume") && \
+  node.roles.any? { |role| /^nova-compute-/ =~ role }
+if need_shared_lock_path
+  group "openstack" do
+    members ["nova", "cinder"]
```
vuntz: Ah, that's slightly annoying: gating fails because the nova user doesn't exist yet (nova packages are installed a bit later).

I'd suggest simply adding the cinder user here (and similarly, only the nova user in the nova recipe). That would solve it.

sidthe:
@vuntz thanks, I saw it failing yesterday, but it was too late to fix it. Let's see if this time it goes through properly

chef/cookbooks/cinder/recipes/common.rb
```diff
@@ -104,6 +104,23 @@
   bind_port = node[:cinder][:api][:bind_port]
 end
 
+# lock path prevents race conditions for cinder-volume and nova-compute on same
+# node. Keep code in sync between cinder and nova recipes. For reference check
+# http://docs.openstack.org/releasenotes/nova/newton.html
```
vuntz: Are you sure it's the newton release notes? I thought it was mitaka...

sidthe: > Are you sure it's the newton release notes? I thought it was mitaka...
that's what google returned to me when copy pasting the note.

sidthe: @vuntz look at version 14.0.0 in the release notes of newton

vuntz: Indeed, my bad.

PR #631, opened 2016-12-05
Systemd cinder

aplanas:


toabctl:
cinder packages are already in D:C:7:S so so is ready for merging...

PR #634, opened 2016-12-05
A better Cisco ACI switch ports example

mfriesenegger:
An example of multiple switch ports that will help a customer with a multi switch configurations.


crowbar_framework/app/models/neutron_service.rb
```diff
@@ -33,11 +33,11 @@ def self.networking_plugins_valid
   end
 
   def self.networking_ml2_type_drivers_valid
-    ["vlan", "gre", "vxlan"]
+    ["vlan", "gre", "vxlan", "opflex"]
   end
 
   def self.networking_ml2_mechanism_drivers_valid
-    ["linuxbridge", "openvswitch", "cisco_nexus"]
+    ["linuxbridge", "openvswitch", "cisco_nexus", "cisco_apic_ml2", "apic_gbp"]
```
vuntz: I may be wrong, but I believe that adding these options here mean that we should update the javascript so that the UI does the right thing.

mfriesenegger: Are you asking about the UI so a user can enable support for ACI from Crowbar UI?

vuntz: @mmnelemane we can also get in touch with the design team. But at the very minimum, when these options are picked we should display something that says "you will need to edit the raw attributes before applying". Right now, this will be too confusing.

PR #635, opened 2016-12-06
Do not require crowbar-pacemaker cookbook to run when we configure HA in openstack

vuntz:
We make it possible to not run the crowbar-pacemaker cookbook when we run some OpenStack cookbook that use HA and require some specific setup (for drbd or haproxy, for instance).

This depends on and needs to be merged at the same time as: https://github.com/crowbar/crowbar-ha/pull/164

### DIFF
--- a/chef/cookbooks/aodh/recipes/aodh_ha.rb
+++ b/chef/cookbooks/aodh/recipes/aodh_ha.rb
@@ -13,6 +13,8 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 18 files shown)


toabctl:
@vuntz travis needs to be fixed.

vuntz:
> @vuntz travis needs to be fixed.

Rebased, that should fix it.

vuntz:
Marking as "do not merge" as this will break HA without the other PR.

vuntz:
Dependency got merged, merging.

PR #642, opened 2016-12-07
nova-manage api_db sync needs to run before nova-manage db sync

jsuchome:
@jgrassler I think you can provide some better reasoning for the commit message


jgrassler:
How about:

"Sync API DB before main DB

Some migrations move things from the main Nova database to the API database. Since the API database needs to have an up-to-date schema for that to work, it needs to be created/updated first."

PR #643, opened 2016-12-07
neutron: Replace /etc/sysconfig/neutron usage with links

toabctl:
/etc/sysconfig/neutron is no longer supported. Instead use symlinks
to enable/disable specific configs for the neutron-server.

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -46,32 +46,43 @@ end
 
 include_recipe "neutron::common_config"
 
+# set core plugin for neutron-server
 if node[:neutron][:networking_plugin] == "vmware"
-  plugin_cfg_path = "/etc/neutron/plugins/vmware/nsx.ini"
+  core_link = "/etc/neutron/plugins/vmware/nsx.ini"
 else
-  plugin_cfg_path = "/etc/neutron/plugins/ml2/ml2_conf.ini"
+  core_link = "/etc/neutron/plugins/ml2/ml2_conf.ini"
 end
 
-template "/etc/sysconfig/neutron" do
-  source "sysconfig.neutron.erb"
-  owner "root"
-  group "root"
-  mode 0640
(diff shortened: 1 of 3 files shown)


chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -242,17 +251,12 @@
...
-             for i in $NEUTRON_PLUGIN_CONF; do \
-               CONF_ARGS="$CONF_ARGS --config-file $i"; \
-             done; \
-             neutron-db-manage --config-file /etc/neutron/neutron.conf $CONF_ARGS upgrade head'
   when "debian"
     command 'source /etc/default/neutron-server; \
              neutron-db-manage --config-file /etc/neutron/neutron.conf --config-file $NEUTRON_PLUGIN_CONFIG upgrade head'
   else
-    command "neutron-db-manage --config-file /etc/neutron/neutron.conf upgrade head"
+    # FIXME(toabctl): Does neutron-db-manage need the ml2 plugin configs???
```
dirkmueller: I think it does.. its the one that initializes the core plugin specific database, so it needs to know about the core plugin configuration.

toabctl: Then I wonder how we want to handle this in the package case. There we have in the service file ```ExecStartPre=-/usr/bin/neutron-db-manage upgrade head```

aplanas: @toabctl For me makes sense

toabctl: @aplanas what makes sense? How would you fix it in the service file?

aplanas: @toabctl uhm I was understanding that your suggestion was to add `ExecStartPre` in each plugin service, and not only in openstack-neutron.service one. Did I understand wrong?

toabctl: @dirkmueller done

PR #644, opened 2016-12-08
cinder/nova: Allow setting a fixed-key for encrypted volumes

toabctl:
With this key set in the cinder proposal, volumes casn be encrypted.
Be careful: For *all* volumes the *same* key is used!


toabctl:
Tested manually and works. This fixes:
```
# ostestr -r '^tempest.scenario.test_encrypted_cinder_volumes.TestEncryptedCinderVolumes.test_encrypted_cinder_volumes_cryptsetup'
```

MaximilianMeister:
test was green, just change in commit message

PR #646, opened 2016-12-08
nova: during upgrade, we have to execute online migrations

jsuchome:


chef/cookbooks/nova/recipes/database.rb
```diff
@@ -79,6 +79,7 @@
   user node[:nova][:user]
   group node[:nova][:group]
   command "nova-manage db sync"
+  ignore_failure true
```
jgrassler: This should be the result of  `CrowbarPacemakerHelper.being_upgraded?(node)` since we should only ignore failures in a situation where we upgrade from Cloud 6 to Cloud 7. Also, this should probably have a comment explaining what the issue is.

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -88,6 +89,46 @@
...
+  command "nova-manage db online_data_migrations"
+  ignore_failure CrowbarPacemakerHelper.being_upgraded?(node)
+  action :run
+  only_if do
+    !node[:nova][:db_synced] && CrowbarPacemakerHelper.being_upgraded?(node) &&
+      (!node[:nova][:ha][:enabled] || CrowbarPacemakerHelper.is_cluster_founder?(node))
+  end
+end
+
+# Update Nova DB to revision 334 (most recent Newton migration)
```
jgrassler: This `db sync` and the `online_data_migrations` following it should only run if `CrowbarPacemakerHelper.being_upgraded?(node)` is true.

jsuchome: @jgrassler see `only_if` part

vuntz:
I'm not a big fan of the `ignore_failure` bits. Can't we tell to do a `db sync` up until a specific version, then do our magic (up to a version), then `db sync` again, etc.?

jgrassler:
Hmm. That might work indeed. I'll try it and comment here.

Itxaka:
Agreed with @vuntz here, from nova-manage version >14 (at least) you can specify up to which version you want to upgrade by using `nova-manage db sync --version VERSION` so it would be good to have those versions in there and let it fail if any other issues happen (database down?)

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -88,6 +89,46 @@
...
+execute "nova-manage db sync (continue)" do
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage db sync"
+  ignore_failure true
```
jgrassler: We definitely won't need an `ignore_failure` here. Only the first `db_sync` should fail, and if the second fails we'll want to know about it.

jgrassler:
I played with this a bit and it looks like this sequence should work:

```
nova-manage api_db sync
nova-manage db sync --version 329 # this is the only change
nova-manage db online_data_migrations  # prints error messages (for online migrations that will simply not run until the corresponding offline migrations are in place - harmless) but exits 0
nova-manage db_sync
nova-manage db online_data_migrations
```

"should work" because the CI cloud I tested this on got rebuilt just as I was about to hit Enter on the second `db sync :-(

I'm fairly sure there won't be any trouble, though since the net effect of the first `db sync` is still the same. The only change is that it will exit 0 rather than 1 because it can execute all planned migrations with the `--version` option in place.

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -79,6 +79,7 @@
   user node[:nova][:user]
   group node[:nova][:group]
   command "nova-manage db sync"
```
jgrassler: Add a `--version 329` here..

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -79,6 +79,7 @@
...
+  ignore_failure CrowbarPacemakerHelper.being_upgraded?(node)
```
jgrassler: ..and remove this.

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -88,6 +89,46 @@
...
 end
 
+# Perform online migrations up to revision 329 (the ones for later revisions
+# will fail. These errors can probably be ignored, but will double-check to
+# make sure)
+execute "nova-manage db online_data_migrations" do
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage db online_data_migrations"
+  ignore_failure true
```
jgrassler: This is a bit debatable: Despite the error messages it spews, the `online_data_migrations` command will exit 0. That being said, the `ignore_failure` thing might be a prudent thing, still, in case somebody upstream feels like fixing the exit status. I don't think this is likely, though. Nova is very conservative about touching migration related stuff during a release, so they'll probably leave this part alone.

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -78,7 +78,7 @@
 execute "nova-manage db sync" do
```
jgrassler: One more thing (sorry, I forgot about that earlier): The first `db sync` (this one) and...

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -88,6 +88,45 @@
   end
 end
 
+# Perform online migrations up to revision 329 (the ones for later revisions
+# will fail. These errors can probably be ignored, but will double-check to
+# make sure)
+execute "nova-manage db online_data_migrations" do
```
jgrassler: ...the first `online_data_migrations` should only be run in the upgrade case, i.e. when being_upgraded?(node) is `true`.

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -88,6 +88,45 @@
...
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage db online_data_migrations"
+  ignore_failure true
+  action :run
+  only_if do
+    !node[:nova][:db_synced] && CrowbarPacemakerHelper.being_upgraded?(node) &&
```
jgrassler: And of course the `being_upgraded` part of this condition and...

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -88,6 +88,45 @@
...
+end
+
+# Run online migration again to cover the ones that failed in the first pass.
+execute "nova-manage db online_data_migrations (continue)" do
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage db online_data_migrations"
+  action :run
+  only_if do
+    !node[:nova][:db_synced] && CrowbarPacemakerHelper.being_upgraded?(node) &&
```
jgrassler: ...the `being_upgraded` part of this condition need to go if the previous `db sync` and `db online_data_migrations` cycle is conditional on `being_upgraded?(node) == true` already.

jsuchome:
Hm, during local testing the upgrade case, first invocation fails:

```
---- Begin output of nova-manage db sync --version 329 ----
STDOUT: WARNING: cell0 mapping not found - not syncing cell0.
error: Function 'downgrade' is not defined in this script
STDERR: Option "verbose" from group "DEFAULT" is deprecated for removal.  Its value may be silently ignored in the future.
Option "rpc_backend" from group "DEFAULT" is deprecated for removal.  Its value may be silently ignored in the future.
Option "rpc_conn_pool_size" from group "DEFAULT" is deprecated. Use option "rpc_conn_pool_size" from group "oslo_messaging_rabbit".
``

jsuchome:
[(5) S51P3 (S50P7) DB migrations when skipping Mitaka release](https://trello.com/c/lt3wdHEk/313-5-s51p3-s50p7-db-migrations-when-skipping-mitaka-release)

PR #647, opened 2016-12-11
Cache keystone settings per instance not per barclamp

dirkmueller:
per barclamp is actually unnecessarily conservative, per keystone
instance is enough. even in the dumb nova-compute node case this
reduces cache misses by factor two.

### DIFF
--- a/chef/cookbooks/keystone/libraries/helpers.rb
+++ b/chef/cookbooks/keystone/libraries/helpers.rb
@@ -11,6 +11,8 @@ module KeystoneHelper
   end
 
   def self.keystone_settings(current_node, cookbook_name)
+    instance = current_node[cookbook_name][:keystone_instance] || "default"
+
     # Cache the result for each cookbook in an instance variable hash. This
     # cache needs to be invalidated for each chef-client run from chef-client
     # daemon (which are all in the same process); so use the ohai time as a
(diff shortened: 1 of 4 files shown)


chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -90,18 +92,14 @@ def self.keystone_settings(current_node, cookbook_name)
...
+    @keystone_settings[instance].merge(
+      service_user: current_node[cookbook_name][:service_user],
+      service_password: current_node[cookbook_name][:service_password])
   end
 
-  private
-  def self.search_for_keystone(node, cookbook_name)
-    instance = node[cookbook_name][:keystone_instance] || "default"
-
+  private_class_method def self.search_for_keystone(node, instance)
```
aspiers: I don't get this change but I'll take it on trust it's right :-)

dirkmueller: Thats what houndci told me. apparently ```private:``` does not apply to class methods. some ruby crap..

dirkmueller:
so this doesn#t work because now we're caching so good that the values that are cached are incorrect (because on first run they're not yet set). requires some more cleanup..

chef/cookbooks/keystone/libraries/helpers.rb
rsalevsky: http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Lint/IneffectiveAccessModifier

PR #648, opened 2016-12-12
nova: reduce excessive node searches on compute role nodes

dirkmueller:
This was only a marginal security feature to begin with, as
you could just copy the other instance ssh key with one hop
more. Created a followup card to make this more secure.

### DIFF
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -300,42 +300,30 @@ elsif !node[:nova]["use_shared_instance_storage"]
   end
 end
 
-# Create and distribute ssh keys for nova user on all compute nodes
-
-# if for some reason, we only have one of the two keys, we recreate the keys
-# (no need to check the case where public key doesn't exist, as the execute
-# resource deals with that)
-if ::File.exist?("#{node[:nova][:home_dir]}/.ssh/id_rsa.pub") && !::File.exist?("#{node[:nova][:home_dir]}/.ssh/id_rsa")
-  file "#{node[:nova][:home_dir]}/.ssh/id_rsa.pub" do
-    action :delete
-  end
-end
-
-execute "Create Nova SSH key" do
-  command "su #{node[:nova][:user]} -c \"ssh-keygen -q -t rsa  -P '' -f '#{node[:nova][:home_dir]}/.ssh/id_rsa'\""
-  creates "#{node[:nova][:home_dir]}/.ssh/id_rsa.pub"
-  only_if { ::File.exist?(node[:nova][:home_dir]) }
-end
-
-ruby_block "nova_read_ssh_public_key" do
-  block do
-    service_ssh_key = File.read("#{node[:nova][:home_dir]}/.ssh/id_rsa.pub")
-    if node[:nova][:service_ssh_key] != service_ssh_key
-      node.set[:nova][:service_ssh_key] = service_ssh_key
-      node.save
-    end
-  end
-  only_if { ::File.exist?("#{node[:nova][:home_dir]}/.ssh/id_rsa.pub") }
+directory "#{node[:nova][:home_dir]}/.ssh" do
+  mode 0o700
+  owner node[:nova][:user]
+  action :create
+  recursive true
 end
 
 ssh_auth_keys = ""
-compute_nodes = node_search_with_cache("roles:nova-compute-#{node[:nova][:libvirt_type]}")
-compute_nodes.each do |n|
-  ssh_auth_keys += n[:nova][:service_ssh_key]
-end
-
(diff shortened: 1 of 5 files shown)


chef/data_bags/crowbar/template-nova.json
```diff
@@ -138,7 +139,7 @@
     "nova": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 108,
+      "schema-revision": 109,
```
toabctl: Please rebase and update the schema-revision

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -284,47 +284,30 @@
...
-  end
-  only_if { ::File.exist?("#{node[:nova][:home_dir]}/.ssh/id_rsa.pub") }
+ssh_auth_keys = ""
+if node["roles"].include?("nova-compute-zvm")
+  ssh_auth_keys += node[:nova][:zvm][:zvm_xcat_ssh_key]
 end
 
-ssh_auth_keys = ""
-%w(kvm xen docker qemu zvm).each do |hypervisor|
-  search_env_filtered(:node, "roles:nova-compute-#{hypervisor}") do |n|
```
vuntz: I'm somehow uncomfortable sharing the same private key for all nodes, but maybe it's not that much of an issue?

That being said, an easy win here would be to search for `nova-compute-*` and then filter the results locally to only keep the hypervisors we care about. That's already a 80% win, and less disruptive.

vuntz: https://github.com/crowbar/crowbar-openstack/pull/663 implements the easy win.

dirkmueller: @vuntz this didn't fix anything for me. it seems the result set is big (lots of nodes with lots of attributes) and thats what is killing it. 

the proper way to make this more secure is by restricting ssh login for that ssh key, which I filed a card for: https://trello.com/c/iT5iff2j - 

even in this secure setup the remote ssh private key is just one "scp nova@foo:.ssh/id_rsa ." away.

rsalevsky:
needs a rebase but looks good to me +1

chef/data_bags/crowbar/migrate/nova/204_add_remotefs_sshkey.rb
```diff
@@ -0,0 +1,20 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "compute_remotefs_sshkey"
+    a["compute_remotefs_sshkey"] = %x[
+      t=$(mktemp)
+      rm -f $t
```
Itxaka: If something in this block fails, we would get the wrong values in here. It should not happen but maybe there is a way of doing this more safely?

nicolasbock:
The ed25519 curve is one of my favorites! :wink:

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -300,46 +300,30 @@
   end
 end
```
matelakat: Commit message is broken.

chef/data_bags/crowbar/migrate/nova/206_add_remotefs_sshkey.rb
```diff
@@ -0,0 +1,20 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "compute_remotefs_sshkey"
+    a["compute_remotefs_sshkey"] = %x[
+      t=$(mktemp)
```
matelakat: We could use `mktemp -d` here as well

crowbar_framework/app/models/nova_service.rb
```diff
@@ -194,6 +194,13 @@ def create_proposal
     base["attributes"]["nova"]["neutron_metadata_proxy_shared_secret"] = random_password
 
     base["attributes"]["nova"]["ec2-api"]["db"]["password"] = random_password
+    base["attributes"]["nova"]["compute_remotefs_sshkey"] = %x[
+      t=$(mktemp)
+      rm -f $t
+      ssh-keygen -q -t ed25519 -N "" -f $t
+      cat $t
+      rm -f $t ${t}.pub
```
matelakat: this shellout seems to be code duplication for generating an ssh key, also used in the migration script.

matelakat:
[backport avoid excessive searches](https://trello.com/c/MJpY601t/146-backport-avoid-excessive-searches)

matelakat:
checking why the package build failed..

matelakat:
> checking why the package build failed..

on my local env, build succeeded, re-running the gate.

Itxaka:
build was aborted, retriggering

matelakat:
This has been aborted again, triggering another re-build

PR #649, opened 2016-12-12
nova: Change name for docker compute_driver

sayalilunkad:
Since nova does not support out-of-tree drivers anymore we introduced
this workaround https://review.openstack.org/#/c/408148/ in our package
so the config change is required.


sayalilunkad:
needs https://build.opensuse.org/request/show/445466 to be in the package first.

rsalevsky:
Is the nova boot error related?

sayalilunkad:
@rsalevsky should not be.

sayalilunkad:
@rsalevsky on second thought they are probably related. Will rebuild and check.

sayalilunkad:
Closing since https://github.com/crowbar/crowbar-openstack/pull/683

PR #651, opened 2016-12-13
neutron: Rework workaround for populating database

vuntz:
The workaround stopped working with systemd because it doesn't wait for
the service to be really up, and so stops the service before the
population has been done.

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -265,9 +265,43 @@ if ha_enabled && CrowbarPacemakerHelper.is_cluster_founder?(node) && !node[:neut
   # quickly starting and stopping the service.
   # https://bugs.launchpad.net/neutron/+bug/1326634
   # https://bugzilla.novell.com/show_bug.cgi?id=889325
-  service "workaround for races in initial db population" do
+  service "workaround for races in initial db population (starting)" do
     service_name node[:neutron][:platform][:service_name]
-    action [:start, :stop]
+    action :start
+  end
+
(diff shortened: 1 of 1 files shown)


vuntz:
While this may unbreak the CI for HA, I don't think that's the correct fix: the fact that systemd doesn't wait for the service to be up will have other impacts, most notably on neutron-ha-tool. I'm nearly sure that we added the wait for the service to be up in the init script for that a few months ago.

@aspiers, do I remember correctly?

@rsalevsky cc'ing you as you may care about that for the milestone

aplanas:
Will try to move the logic here in the package. I was under the impression that @jgrassler and @toabctl had already a proposal here.

toabctl:
@vuntz I think we should fix it upstream. See https://bugs.launchpad.net/neutron/+bug/1649234 . And then use ```type=notify``` for the neutron-server .

toabctl:
Looks like neutron-server already supports the SD_NOTIFY interface. So having ```type=simple``` I get:
```
# systemctl restart openstack-neutron ; curl -X GET http://localhost:9696
curl: (7) Failed to connect to localhost port 9696: Connection refused
```

Having ``type=notify``` I get:
```
# systemctl restart openstack-neutron ; curl -X GET http://localhost:9696
{"versions": [{"status": "CURRENT", "id": "v2.0", "links": [{"href": "http://localhost:9696/v2.0/", "rel": "self"}]}]}
```

toabctl:
I changed the type in the package. Let's see if that helps.

aplanas:
IMHO this can be closed

vuntz:
Thanks, let's see how it goes with the updated package.

PR #652, opened 2016-12-13
Cross az attach

toabctl:


chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -3388,7 +3388,8 @@ insecure = <%= @cinder_insecure ? 'True' : 'False' %>
 # Cinder (or allow_availability_zone_fallback=False in cinder.conf), the volume
 # create request will fail and the instance will fail the build request.
 # (boolean value)
-#cross_az_attach = true
+#cross_az_attach >= true
```
aplanas: > ?

chef/data_bags/crowbar/template-nova.json
```diff
@@ -138,7 +139,7 @@
     "nova": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 108,
+      "schema-revision": 110,
```
aplanas: the number match with the migration, is ok to jump from 108 to 110?

chef/cookbooks/nova/templates/default/nova.conf.erb
toabctl: fixed

jsuchome:
Don't we want to have an UI element for that?

toabctl:
> Don't we want to have an UI element for that?

I don't think so.

PR #655, opened 2016-12-14
nova: Make "cross_az_attach" configurable

toabctl:
(cherry picked from commit 6fc4f5ae069ccc9e311aa8cc0098ac8aca9ec9b6)

### DIFF
--- a/chef/data_bags/crowbar/template-nova.json
+++ b/chef/data_bags/crowbar/template-nova.json
@@ -20,6 +20,7 @@
       "setup_shared_instance_storage": false,
       "use_shared_instance_storage": false,
       "use_migration": false,
+      "cross_az_attach": true,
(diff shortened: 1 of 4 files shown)


chef/data_bags/crowbar/template-nova.json
```diff
@@ -96,7 +97,7 @@
     "nova": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 45,
+      "schema-revision": 46,
```
aplanas: This is ok, but you need to rename the migration file

toabctl: ah. thanks. Fixed

toabctl:
@dirkmueller fixed.

PR #656, opened 2016-12-14
keystone: minify the keystone configuration file

Itxaka:


Itxaka:
[(5) Update templates for config files for Newton - Keystone](https://trello.com/c/t0rgQ51S/375-5-update-templates-for-config-files-for-newton-keystone)

chef/cookbooks/keystone/attributes/default.rb
```diff
@@ -29,6 +29,7 @@
 default[:keystone][:verbose] = true
 default[:keystone][:domain_specific_drivers] = false
 default[:keystone][:domain_config_dir] = "/etc/keystone/domains"
+default[:keystone][:config_file] = "/etc/keystone/keystone.conf.d/100-keystone.conf"
```
nicolasbock: Don't we need a migration?

Itxaka: This is only specific for chef attributes/cookbooks so it should not need a migration. IIRC the migrations are only for crowbar <-> chef attributes, so anything that can be configured from crowbar UI/comes from the user as a json for the proposal needs it.

toabctl: Hm. Do we need that attribute? For heat we just adjusted the template path...

Itxaka: If in the future we need to change it, its easier to have it on one place, especially on the attributes which is where this stuff is supposed to go. That way we wont have to go over the code to search for those occurrences :)

sidthe: I think Itxaka is right here. It makes way more sense to do it this way

toabctl: The idea was to have multiple config files (one for logging, one for authtoken, one for messaging, ...). How would you handle that?

toabctl: but I'm ok with keeping it as suggested.

Itxaka: I did not know about that, but still I would do it through attributes. Not sure how this change will affect to using multiple config files per section, as we are gonna have to redo and probably move it to the attributes anyway (Much easier!)


Attributes:
```
default[:keystone][:config_dir] = "/etc/keystone/keystone.conf.d/"
default[:keystone][:config_files] = ["100-general.conf", "101-authoken.conf", "102-logging.conf"]
```

Recipe:
```
node[:keystone][:config_files].each do |config|
  template "#{node[:keystone][:config_dir]}#{config}" do
    source "#{config}.erb"
    owner "root"
    group node[:keystone][:group]
    mode 0640
    ...
  end
end
```

PR #663, opened 2016-12-16
Decrease amount of searches done in cookbooks for compute nodes

vuntz:


chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -317,9 +317,9 @@
 end
 
 ssh_auth_keys = ""
-%w(kvm xen docker qemu zvm).each do |hypervisor|
-  search_env_filtered(:node, "roles:nova-compute-#{hypervisor}") do |n|
-    ssh_auth_keys += n[:nova][:service_ssh_key]
+search_env_filtered(:node, "roles:nova-compute-*") do |n|
+  ["kvm", "xen", "docker", "qemu", "zvm"].each do |hypervisor|
```
dirkmueller: it really should only inject ssh keys for hypervisor hosts of the same type. there is e.g. no point in collecting ssh keys of xen hosts on the kvm host. 

anyway, this doesn't fix anything in my tests since its the size of the query that is the problem not the number of queries.

vuntz: > it really should only inject ssh keys for hypervisor hosts of the same type. there is e.g. no point in collecting ssh keys of xen hosts on the kvm host.

And you're obviously correct, done.

vuntz:
> I would still prefer to pick my compute-node-ssh-solution :-)

Right, I think your PR there is better long-term (assuming it's okay security-wise). Just doing some simple work that I know can be backported to Cloud 6 for now, hence this commit.

PR #665, opened 2016-12-16
sahara: minify sahara.conf template

sidthe:


sidthe:
[(5) Update templates for config files for Newton - Sahara](https://trello.com/c/6TN6KIuJ/380-5-update-templates-for-config-files-for-newton-sahara)

PR #666, opened 2016-12-16
cookbooks: Use new node_search_with_cache helper

vuntz:
~~Requires https://github.com/crowbar/crowbar-core/pull/906 (merged)~~

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/common.rb
+++ b/chef/cookbooks/ceilometer/recipes/common.rb
@@ -6,10 +6,9 @@ if node[:ceilometer][:use_mongodb]
   db_connection = nil
 
   if node[:ceilometer][:ha][:server][:enabled]
-    db_hosts = search(:node,
-                      "ceilometer_ha_mongodb_replica_set_member:true AND roles:ceilometer-server AND "\
-                      "ceilometer_config_environment:#{node[:ceilometer][:config][:environment]}"
-                     )
+    db_hosts = node_search_with_cache("roles:ceilometer-server").select do |n|
+      n[:ceilometer][:ha][:mongodb][:replica_set][:member] rescue false
+    end
     unless db_hosts.empty?
       mongodb_servers = db_hosts.map { |s| "#{Chef::Recipe::Barclamp::Inventory.get_network_by_type(s, "admin").address}:#{s[:ceilometer][:mongodb][:port]}" }
       db_connection = "mongodb://#{mongodb_servers.sort.join(",")}/ceilometer?replicaSet=#{node[:ceilometer][:ha][:mongodb][:replica_set][:name]}"
@@ -20,7 +19,7 @@ if node[:ceilometer][:use_mongodb]
   # been set on any node (yet), we just fallback to using the first
   # ceilometer-server node
   if db_connection.nil?
-    db_hosts = search_env_filtered(:node, "roles:ceilometer-server")
+    db_hosts = node_search_with_cache("roles:ceilometer-server")
     db_host = db_hosts.first || node
     mongodb_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(db_host, "admin").address
     db_connection = "mongodb://#{mongodb_ip}:#{db_host[:ceilometer][:mongodb][:port]}/ceilometer"
@@ -39,7 +38,7 @@ else
     db_password = node[:ceilometer][:db][:password]
   else
     # pickup password to database from ceilometer-server node
-    node_controllers = search(:node, "roles:ceilometer-server") || []
+    node_controllers = node_search_with_cache("roles:ceilometer-server")
     if node_controllers.length > 0
       db_password = node_controllers[0][:ceilometer][:db][:password]
     end
--- a/chef/cookbooks/ceilometer/recipes/common.rb
+++ b/chef/cookbooks/ceilometer/recipes/common.rb
@@ -3,27 +3,7 @@ include_recipe "apache2"
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
 if node[:ceilometer][:use_mongodb]
-  db_connection = nil
-
-  if node[:ceilometer][:ha][:server][:enabled]
-    db_hosts = node_search_with_cache("roles:ceilometer-server").select do |n|
-      n[:ceilometer][:ha][:mongodb][:replica_set][:member] rescue false
-    end
-    unless db_hosts.empty?
-      mongodb_servers = db_hosts.map { |s| "#{Chef::Recipe::Barclamp::Inventory.get_network_by_type(s, "admin").address}:#{s[:ceilometer][:mongodb][:port]}" }
-      db_connection = "mongodb://#{mongodb_servers.sort.join(",")}/ceilometer?replicaSet=#{node[:ceilometer][:ha][:mongodb][:replica_set][:name]}"
-    end
-  end
-
-  # if this is a cluster, but the replica set member attribute hasn't
-  # been set on any node (yet), we just fallback to using the first
-  # ceilometer-server node
-  if db_connection.nil?
-    db_hosts = node_search_with_cache("roles:ceilometer-server")
-    db_host = db_hosts.first || node
-    mongodb_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(db_host, "admin").address
-    db_connection = "mongodb://#{mongodb_ip}:#{db_host[:ceilometer][:mongodb][:port]}/ceilometer"
-  end
+  db_connection = CeilometerHelper.mongodb_connection_string(node, cookbook_name)
 else
   db_settings = fetch_database_settings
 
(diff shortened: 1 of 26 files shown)


chef/cookbooks/nova/recipes/compute_ha.rb
```diff
@@ -35,7 +35,7 @@
 internal_auth_url_v2 = \
   "#{keystone_settings["protocol"]}://" \
   "#{keystone_settings["internal_url_host"]}:#{keystone_settings["service_port"]}/v2.0/"
-neutrons = search(:node, "roles:neutron-server AND roles:neutron-config-#{nova[:nova][:neutron_instance]}")
+neutrons = node_search_with_cache("roles:neutron-server")
```
vuntz: So here it doesn't matter because we have `nova[:nova][:neutron_instance]} == node[:nova][:neutron_instance]}` by definition (node and nova are both part of the same nova proposal, so with the same attribute), so simply using the "magic" behavior will work.

chef/cookbooks/swift/recipes/proxy.rb
```diff
@@ -277,8 +277,7 @@
 end
 
 ## make sure to fetch ring files from the ring compute node
-env_filter = " AND swift_config_environment:#{node[:swift][:config][:environment]}"
-compute_nodes = search(:node, "roles:swift-ring-compute#{env_filter}")
+compute_nodes = node_search_with_cache("roles:swift-ring-compute")
```
vuntz: Happily ignoring :-)

chef/cookbooks/swift/recipes/ring-compute.rb
```diff
@@ -48,7 +46,7 @@
 
 ####
 # collect the current contents of the ring files.
-nodes = search(:node, "roles:swift-storage#{env_filter}")
+nodes = node_search_with_cache("roles:swift-storage")
```
vuntz: Ignoring, as per https://github.com/crowbar/crowbar-openstack/pull/666/files#r94943772

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -16,8 +16,7 @@
 
 neutron = nil
 if node.attribute?(:cookbook) && node[:cookbook] == "nova"
-  neutrons = search(:node, "roles:neutron-server AND \
-                    roles:neutron-config-#{node[:nova][:neutron_instance]}")
+  neutrons = node_search_with_cache("roles:neutron-server", node[:nova][:neutron_instance])
```
vuntz: As discussed on jangouts, yes, they are :-)

vuntz:
Rebased

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -20,8 +20,11 @@
 if node.attribute?(:cookbook) and node[:cookbook] == "nova"
   neutrons = node_search_with_cache("roles:neutron-server", node[:nova][:neutron_instance])
   neutron = neutrons.first || raise("Neutron instance '#{node[:nova][:neutron_instance]}' for nova not found")
+  nova = node
```
matelakat: At this point we know that `node[:cookbook] == "nova"`, and based on that information, we say that this is the nova controller. Is that always true?

chef/cookbooks/ceilometer/recipes/mongodb.rb
```diff
@@ -82,9 +82,9 @@
     # replicaset
     package("ruby#{node["languages"]["ruby"]["version"].to_f}-rubygem-mongo").run_action(:install)
 
-    members = search(:node,
-                     "ceilometer_ha_mongodb_replica_set_member:true AND "\
-                     "ceilometer_config_environment:#{node[:ceilometer][:config][:environment]}")
+    members = node_search_with_cache("roles:ceilometer-server").select do |n|
+      n[:ceilometer][:ha][:mongodb][:replica_set][:member] rescue false
```
matelakat: Seems like it's the same query as above. Maybe a candidate for extracting?

chef/cookbooks/ceilometer/recipes/server.rb
```diff
@@ -23,10 +23,9 @@
   mongodb_nodes = nil
 
   if ha_enabled
-    mongodb_nodes = search(:node,
-                           "ceilometer_ha_mongodb_replica_set_member:true AND "\
-                           "ceilometer_config_environment:#{node[:ceilometer][:config][:environment]}"
-      )
+    mongodb_nodes = node_search_with_cache("roles:ceilometer-server").select do |n|
+      n[:ceilometer][:ha][:mongodb][:replica_set][:member] rescue false
```
matelakat: Same here

chef/cookbooks/neutron/recipes/common_config.rb
vuntz: No, it's not a nova controller. It's a nova node, but nova nodes have the attributes that we need too.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
aspiers: @vuntz For the benefit of those of us who missed the jangouts discussion, presumably this is because `neutron_config_environment:foo` is Chef magic for searching for nodes which have `node[:neutron][:config][:environment]` set to `foo`.  And if a node has a neutron barclamp proposal `foo` applied to it then it will have both this and also the role `neutron-config-foo`, which is where that attribute comes from.  Right?

vuntz: @aspiers yes

chef/cookbooks/neutron/recipes/common_config.rb
aspiers: I found this pretty confusing (not least because apparently this is using a very old Dell hack to provide the `include_recipe` caller via `node[:cookbook]`), and the commit message gives very little context.

So please can you expand the commit message and rename the variables to make this clearer?  It's not immediately obvious what variables called `neutron` and `nova` actually refer to.  I suggest `s/neutrons/neutron_servers/`, `s/neutron/neutron_server/`, and `s/nova/nova_node/` (not so sure about the last one; can you think of anything a bit more informative?)

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -5,10 +5,9 @@
   db_connection = nil
 
   if node[:ceilometer][:ha][:server][:enabled]
-    db_hosts = search(:node,
-                      "ceilometer_ha_mongodb_replica_set_member:true AND roles:ceilometer-server AND "\
-                      "ceilometer_config_environment:#{node[:ceilometer][:config][:environment]}"
-                     )
+    db_hosts = node_search_with_cache("roles:ceilometer-server").select do |n|
+      n[:ceilometer][:ha][:mongodb][:replica_set][:member] rescue false
```
aspiers: By now you must be familiar with the reasons for Hound complaining about this. So are you deliberately disagreeing with them, or just old habits dying hard?

vuntz: I simply disagree about avoiding `rescue` when we access an attribute with a deep path that may not exist.

chef/cookbooks/neutron/recipes/common_config.rb
vuntz: > It's not immediately obvious what variables called neutron and nova actually refer to. I suggest s/neutrons/neutron_servers/, s/neutron/neutron_server/, and s/nova/nova_node/ (not so sure about the last one; can you think of anything a bit more informative?)

That would definitely not be part of this PR according to the way you split commits ;-)

chef/cookbooks/ceilometer/recipes/mongodb.rb
vuntz: @matelakat I added a commit for this.

chef/cookbooks/neutron/recipes/common_config.rb
vuntz: > So please can you expand the commit message and rename the variables to make this clearer?

Done.

vuntz: > It's not immediately obvious what variables called neutron and nova actually refer to. I suggest s/neutrons/neutron_servers/, s/neutron/neutron_server/, and s/nova/nova_node/ (not so sure about the last one; can you think of anything a bit more informative?)

@aspiers I started doing that, but I feel this is actually way too prone to errors to be part of this PR. I can create a PR after this one just for this, though.

aspiers: > That would definitely not be part of this commit according to the way you split commits ;-)

Correct - if you look closely, you'll see that I wasn't asking the renaming to be part of this commit :-)  Ideally commits refactoring for clarity should happen first, so that when it comes to the non-refactoring commits, they already benefit from the added clarity.

But I'm OK with this being done in a separate PR.  And thanks for the much improved commit message!

vuntz:
@aspiers is this fine with you with the latest changes?

vuntz:
Rebased

vuntz:
HA job found a bug, fixed.

vuntz:
@dirkmueller @rsalevsky @matelakat are your +1 still valid?

vuntz:
Rebased: aodh changes are gone now that the aodh barclamp has been split (and doesn't use mongodb anymore).

aplanas:
Retrying HA job

aspiers:
@dirkmueller I think this one is good to merge now.

(and #toomanyreviewers)

PR #667, opened 2016-12-16
ha: Re-order rabbitmq UID/GID and pacemaker resource creation

nicolasbock:
We reset the UID/GID of rabbitmq to `91` to make it consistent across
the cluster. This change has to happen before we create the fs-rabbitmq
pacemaker resource, otherwise rabbitmq can not be started due to
potentially wrong ownership values on its home directory. This patch
re-orders the two relevant Chef resources so that the UID/GID is changed
before pacemaker creates the new resource.

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -18,6 +18,23 @@
 # limitations under the License.
 #
 
+# Create rabbitmq user with fixed UID/GID values to allow for HA.
+group "rabbitmq" do
+  gid node[:rabbitmq][:gid]
+  action :create
+  not_if "grep -q rabbitmq:x /etc/group"
+end
+
+user "rabbitmq" do
+  comment "user for RabbitMQ messaging server"
+  uid node[:rabbitmq][:uid]
+  gid node[:rabbitmq][:gid]
+  home "/var/lib/rabbitmq"
+  shell "/sbin/nologin"
+  action :create
+  not_if "grep -q rabbitmq:x /etc/passwd"
+end
+
 package "rabbitmq-server"
 package "rabbitmq-server-plugins" if node[:platform_family] == "suse"
 
@@ -26,6 +43,19 @@ directory "/etc/rabbitmq/" do
   group "root"
   mode 0755
   action :create
+  recursive true
+end
+
+# Update ownerships
+["/var/lib/rabbitmq", "/var/run/rabbitmq", "/var/log/rabbitmq",
+ "/var/log/rabbitmq"].each do |d|
+  directory d do
+    owner "rabbitmq"
+    group "rabbitmq"
+    action :create
+    recursive true
+    notifies :restart, "service[rabbitmq-server]"
+  end
 end
 
 template "/etc/rabbitmq/rabbitmq-env.conf" do
@@ -68,5 +98,3 @@ service "rabbitmq-server" do
   action [:enable, :start]
   provider Chef::Provider::CrowbarPacemakerService if node[:rabbitmq][:ha][:enabled]
 end
-
-
--- a/chef/cookbooks/rabbitmq/recipes/ha.rb
+++ b/chef/cookbooks/rabbitmq/recipes/ha.rb
@@ -137,37 +137,6 @@ end
 
 crowbar_pacemaker_sync_mark "create-rabbitmq_ha_storage"
 
-# Ensure that uid/gid are the same on all nodes
-#
-# This is really hacky!!!
-#
-# We need to have the uid/gid be the same accross all nodes, but rabbitmq
-# packages don't use a static uid/gid. So we pick one (91) which doesn't seem
-# to be in use:
-#  - see base-passwd package for Debian: http://sources.debian.net/src/base-passwd/3.5.32/passwd.master
-#  - see wiki page for Fedora: https://fedoraproject.org/wiki/PackageUserRegistry
-# We could use the chef user/group LWRP to do the change, but we also need to
-# run other commands (to change ownership of existing files), so we might as
-# well do a big script.
-# We also stop rabbitmq in case it was running before, so we can change the uid
-# without any impact. Pacemaker will start the process on one other node later
-# on anyway.
-static_uid = 91
-static_gid = 91
-bash "assign static uid to rabbitmq" do
- code <<EOC
- service rabbitmq-server stop > /dev/null;
- groupmod -g #{static_gid} rabbitmq;
- usermod -u #{static_uid} -g #{static_gid} rabbitmq;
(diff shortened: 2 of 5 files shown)


Itxaka:
Maybe Im not seeing but why not, instead of using this hacky method, just create the user+group before installing the rabbitmq package? That way it doesn't matter if its HA or not or you move into an HA from a non-HA, the user/group will be always the same one on all installs (Assuming the package install wont overwrite the user)

And we can remove that whole bash block and use a normal ruby block for it (Which I also think is a no-op) with a link to the docs mentioning the same UID/GID for rabbitmq (https://www.rabbitmq.com/pacemaker.html)

nicolasbock:
@Itxaka Yes, you might be right that this strategy would allow us to remove the bash block. I'll try that. Thanks!

nicolasbock:
@Itxaka Still hasn't gone through gating but what do you think?

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -21,6 +21,18 @@
 package "rabbitmq-server"
 package "rabbitmq-server-plugins" if node[:platform_family] == "suse"
 
+# Create rabbitmq user with fixed UID/GID values to allow for HA.
+group "rabbitmq"
```
vuntz: Missing `do`.

nicolasbock:
@vuntz Ok, that's a good point. I was only considering the mount point. Let me redo this.

nicolasbock:
@vuntz I have moved this a bit around. The user is now create _before_ the package is installed to ensure that the UID/GID are set correctly. This should also guarantee that the other files and directories owned by `rabbitmq` are created with the correct UID/GID.

nicolasbock:
I have tested this on vp5 and all directories changed in the bash resource have the correct owner, i.e. 91/91.

nicolasbock:
Triggered HA gating.

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -18,6 +18,22 @@
...
+# Create rabbitmq user with fixed UID/GID values to allow for HA.
+group "rabbitmq" do
+  gid "91"
+  :create
+end
+
+user "rabbitmq" do
+  comment "user for RabbitMQ messaging server"
+  uid "91"
+  gid "91"
```
dirkmueller: Should be without string literal quoting - these are numbers (and maybe should be chef attributes instead of being hardcodef)

vuntz:
> The user is now create before the package is installed to ensure that the UID/GID are set correctly.

Sorry, but you can't assume that: the package may have been installed before (manually, for instance -- in case someone preallocated nodes with all required packages preinstalled).

chef/cookbooks/rabbitmq/recipes/default.rb
Itxaka: +1 for attributes. attributes are life. embrace life.

Itxaka:
@vuntz I believe that when a user id is modified, automatically all of its files under its home dir will be changed to reflect that.

for /var/log and /var/run, we can just create them after the user/group creation so they are created (or modified if they already exists) with the proper ID (var/run should not be an issue, even removing it altogether doesn't affect rabbit)

The only issue with this is that we would need to stop rabbitmq-server if its running and  if we need to change the user/group id, which could be a problem becuase I dont think the `:before` timer for `notify` is available on chef 11? If it was we could do:


```
service 'rabbitmq' do
  action :nothing
end

group 'rabbitmq' do
  action :create
  gid 91
  notifies :stop, 'service[rabbitmq]', :before
  only_if { node[:etc][:group][:rabbitmq][:uid] != 91 }
end

user 'rabbitmq' do
  action :create
  uid 91
  gid 91
  notifies :stop, 'service[rabbitmq]', :before
  only_if { node[:etc][:passwd][:rabbitmq][:uid] != 91 }
end

directory '/var/log/rabbitmq' do
  owner 'rabbitmq'
  group 'rabbitmq'
  action :create
end

service 'rabbitmq' do
  action :start
end
```

So this would only stop rabbitmq if is installed and the user/group ID needs to be changed. And the other resource are a no-op I think so it should not impact that much the chef run.

Thougths?

chef/cookbooks/rabbitmq/recipes/default.rb
nicolasbock: I agree, I'll move those to attributes.

nicolasbock:
I opened an alternative (#670) in which I move the `bash` resource further up. I don't see a reason why we can't adjust ownerships already as early as in #670.

nicolasbock:
@Itxaka Thanks for the suggestion! I have added `directory` resources and attributes.

nicolasbock:
Gating is presumably failing due to #668.

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -18,6 +18,22 @@
 # limitations under the License.
 #
 
+# Create rabbitmq user with fixed UID/GID values to allow for HA.
+group "rabbitmq" do
+  gid node[:rabbitmq][:gid]
+  :create
+end
+
+user "rabbitmq" do
```
Itxaka: I dont think this will work if rabbitmq is already running in the system as you cant change a user ID which has any process running.

nicolasbock: @Itxaka I have added a guard to the `user` and `group` resources.

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -18,6 +18,23 @@
 # limitations under the License.
 #
 
+# Create rabbitmq user with fixed UID/GID values to allow for HA.
+group "rabbitmq" do
+  gid node[:rabbitmq][:gid]
+  action :create
+  not_if "grep -q rabbitmq:x /etc/group"
```
Itxaka: This should not be necessary, if it exists it will just try to modify it to update it to the provided params

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -18,6 +18,23 @@
...
+end
+
+user "rabbitmq" do
+  comment "user for RabbitMQ messaging server"
+  uid node[:rabbitmq][:uid]
+  gid node[:rabbitmq][:gid]
+  home "/var/lib/rabbitmq"
+  shell "/sbin/nologin"
+  action :create
+  not_if "grep -q rabbitmq:x /etc/passwd"
```
Itxaka: same as above

vuntz:
For the record, I'm more comfortable with https://github.com/crowbar/crowbar-openstack/pull/670

nicolasbock:
Closing since #670 was merged.

PR #671, opened 2016-12-18
Fix unversioned internal auth url for the HA case

dirkmueller:
Since the attribute was never set, it fell back to the non-HA case
and pointed to the non-VIP endpoint, which tentantivel might not work.

### DIFF
--- a/chef/cookbooks/keystone/libraries/helpers.rb
+++ b/chef/cookbooks/keystone/libraries/helpers.rb
@@ -10,6 +10,28 @@ module KeystoneHelper
     service_URL(protocol, host, port) + "/" + version + "/"
   end
 
+  def self.admin_auth_url(node, admin_host)
+    service_URL(node[:keystone][:api][:protocol], admin_host, node[:keystone][:api][:admin_port])
+  end
+
+  def self.public_auth_url(node, public_host)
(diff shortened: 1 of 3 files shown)


chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -63,17 +47,29 @@ def self.keystone_settings(current_node, cookbook_name)
         # version to be a "v3.0" for the v3 API instead of the "v3" or "3" that
         # is used everywhere else.
         "api_version_for_middleware" => "v%.1f" % node[:keystone][:api][:version],
-        "admin_auth_url" => node[:keystone][:api][:admin_URL] || admin_auth_url,
-        "public_auth_url" => node[:keystone][:api][:versioned_public_URL] || public_auth_url,
-        "internal_auth_url" => node[:keystone][:api][:versioned_internal_URL] || internal_auth_url,
-        "unversioned_internal_auth_url" => node[:keystone][:api][:unversioned_internal_URL] || \
-          unversioned_internal_auth_url,
+        # keep in sync with recipes/server.rb
```
toabctl: in sync with what exactly? I don't get that.

dirkmueller: @toabctl done

PR #672, opened 2016-12-19
manila: minify config template and add common attribute for it

Itxaka:


Itxaka:
[(5) Update templates for config files for Newton - Manila](https://trello.com/c/dTscHrhq/377-5-update-templates-for-config-files-for-newton-manila)

Itxaka:
Thanks @toabctl !

PR #674, opened 2016-12-19
Decrease amount of node saves

vuntz:
Saving is not cheap, and most of the time, it's not needed (as no attribute changed, or because we will already save later).


vuntz:
Was done in other PR referenced here.

PR #675, opened 2016-12-19
Cinder configuration minification and relocation

aplanas:


aplanas:
[(5) Update templates for config files for Newton - Cinder](https://trello.com/c/QLNt5bu6/372-5-update-templates-for-config-files-for-newton-cinder)

aplanas:
@Itxaka indeed! done!

PR #676, opened 2016-12-19
neutron: add port_security to ml2 extension_drivers

bmwiedemann:
as per
http://specs.openstack.org/openstack/neutron-specs/specs/kilo/ml2-ovs-portsecurity.html
https://wiki.openstack.org/wiki/Neutron/ML2PortSecurityExtensionDriver

this should allow us to run our own routers and DHCP-servers within the cloud
that would allow us to test SOC within OpenStack

using
neutron net-update ournet --port-security-enabled=False
or neutron port-update
to override the defaults (default is to filter as before)


toabctl:
nit: commit msg should start with ```neutron:```

PR #677, opened 2016-12-19
cinder: looplvm: avoid one level of indirection

bmwiedemann:
to simplify code and reduce sources of errors


toabctl:
nit. commit msg should start with ```cinder:..```

PR #678, opened 2016-12-19
horizon: minify config templates and point the config to the new dir

Itxaka:


Itxaka:
[(5) Update templates for config files for Newton - Horizon](https://trello.com/c/k4QN23kq/374-5-update-templates-for-config-files-for-newton-horizon)

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -352,7 +352,8 @@
 
 crowbar_pacemaker_sync_mark "wait-horizon_config"
 
-local_settings = "#{dashboard_path}/openstack_dashboard/local/local_settings.py"
+local_settings = "#{dashboard_path}/openstack_dashboard/local/"\
```
nicolasbock: Maybe space before `\`?

Itxaka: good point, ammended

Itxaka:
Fails with:

```
django.core.exceptions.ImproperlyConfigured: settings.DATABASES is improperly configured. Please supply the ENGINE value. Check settings documentation for more details.
```

But also has this above
```
ERROR:root:Can not exec settings snippet 100-local_settings.py
```


Which may be related to this[0]:
```
 The file must start with an underscore (_) because Python can not load files starting with a number.
```

Sending a new commit to fix this.

[0] http://docs.openstack.org/developer/horizon/topics/settings.html#openstack-settings-partial

Itxaka:
now Im lost, still failing with the same error. Re-running the barclamp makes it work so... race condition? upstream problem with the secret key creation/lock? ¬_¬

Itxaka:
May have something to do with the secret key (which executes code) running twice as it appears on both files, and hitting the file lock on teh key store.

nicolasbock:
@Itxaka it's passing now, it looks like your last commit did it.

PR #679, opened 2016-12-19
ceilometer: update conf template to newton

sidthe:
remove or fix parameters against newton config for ceilometer


sidthe:
[(5) Update templates for config files for Newton - Ceilometer](https://trello.com/c/mE4o4pAj/371-5-update-templates-for-config-files-for-newton-ceilometer)

PR #680, opened 2016-12-19
Nova configuration and new location

aplanas:


aplanas:
[(5) Update templates for config files for Newton - Nova](https://trello.com/c/qbXXTDIE/379-5-update-templates-for-config-files-for-newton-nova)

toabctl:
@aplanas nova-api does not start:
```
2016-12-19 16:53:14.708 1839 ERROR nova MissingRequiredOptions: Auth plugin requires parameters which were not given: auth_url

```

aplanas:
@toabctl actually the fail is because of some crudini bits in /etc/nova/nova.conf that comes from the package. I guess that when the changes in the package are merged, and the crudini info is not there anymore, this will pass (I tested)

aplanas:
* Remove some blank lines and expand some `if`. Still generate the same config file

rsalevsky:
Needs a rebase

chef/cookbooks/nova/attributes/default.rb
```diff
@@ -20,6 +20,7 @@
 
 default[:nova][:debug] = false
 default[:nova][:max_header_line] = 16384
+default[:nova][:config_file] = "/etc/nova/nova.conf.d/100-nova.conf"
```
jsuchome: Is this correct? My installed file is `010-nova.conf`

toabctl:
Taking over from @aplanas with https://github.com/crowbar/crowbar-openstack/pull/697

PR #682, opened 2016-12-19
barbican: update template to newton, remove not found configs, remove…

Itxaka:
… configs already on default values

### DIFF
--- a/chef/cookbooks/barbican/templates/default/barbican.conf.erb
+++ b/chef/cookbooks/barbican/templates/default/barbican.conf.erb
@@ -2,13 +2,7 @@
 
 debug=<%= node[:barbican][:debug] %>
 host_href = <%= @host_href %>
-backlog = 4096
-max_allowed_secret_in_bytes = 10000
-max_allowed_request_size_in_bytes = 1000000
 sql_connection = <%= @database_connection %>
-sql_idle_timeout = 3600
-default_limit_paging = 10
-max_limit_paging = 100
 
 [oslo_messaging_rabbit]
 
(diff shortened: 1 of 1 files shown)


Itxaka:
[(5) Update templates for config files for Newton - Barbican](https://trello.com/c/6ozqOvAG/370-5-update-templates-for-config-files-for-newton-barbican)

Itxaka:
rebuild with want_barbican=1 triggered

Itxaka:
This works *but* barbican isnt working on the master branch anyway. So Im not sure this patch should go in until we can have barbican working on the master branch as to find if there is any issues with this patch

Itxaka:
Depends on https://github.com/crowbar/crowbar-openstack/pull/686

toabctl:
> Depends on #686

It is the same repo. Why are you not just adding the commits needed here to this PR? This would be less confusing imo.

Itxaka:
@toabctl PR separation, plus I rather check if barbican is really working *before* this patch and *after* it instead of killing myself trying to see why the template change isnt working just to see that it has nothing to do with the template.

Itxaka:
FYI: not using transport_url in barbican due to iot being broken: https://github.com/openstack/barbican/blob/stable/newton/etc/barbican/barbican.conf#L133

Itxaka:
https://github.com/crowbar/crowbar-openstack/pull/686 was merged so tihs should be good to go after the reviews!

chef/cookbooks/barbican/templates/default/barbican.conf.erb
```diff
@@ -2,13 +2,7 @@
```
toabctl: What about moving the config to the conf.d/ dir? Is that another commit?

Itxaka: wow, I thougth it was in there already o_O , will update the PR

Itxaka: ahh, that was already in: https://github.com/crowbar/crowbar-openstack/commit/ffc5168cbf672a514635ab22e19912b6a5399aba

PR #683, opened 2016-12-19
Remove nova-docker support

dirkmueller:
This is abandoned upstream in Newton so there is no point in
keeping it available.


chef/cookbooks/tempest/templates/default/tempest.conf.erb
```diff
@@ -1362,7 +1359,7 @@ dhcp_client =
 
 # Whether or not cinder is expected to be available (boolean value)
 #cinder = true
-cinder = <%= @use_docker ? false : true %>
+cinder = true
```
toabctl: Is the default, can be removed

rsalevsky:
Just for the record: We also need to remove the docker CI job and the trello card.

dirkmueller:
@vuntz done

PR #684, opened 2016-12-19
Revert "Revert "neutron: Disable l2pop mechanism unless DVR is enabled""

dirkmueller:
l2pop with HA/vxlan just doesn't seem to work at all. Only enable
what upstream does in devstack: when DVR is used.

This reverts commit 8ed94d918e0ba241e70a16d37d4d63177c2804cc.

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -223,7 +223,7 @@ if neutron[:neutron][:networking_plugin] == "ml2"
       variables(
         ml2_type_drivers: ml2_type_drivers,
         tunnel_types: ml2_type_drivers.select { |t| ["vxlan", "gre"].include?(t) },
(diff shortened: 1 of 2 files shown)


dirkmueller:
already cherry picked into the packages in order to unblock gating

vuntz:
We might want to reintroduce the l2pop attribute then: there are some deployments out there who want l2pop without DVR.

vuntz:
0a2d8f853704838ee596dca5bca0f7332f2d8c46 is the commit I was mentioning. I guess we also need to expose that in the UI now (since DVR is exposed there).

@dirkmueller want me to work on that?

dirkmueller:
@vuntz yes please, might not get to it in time

vuntz:
> @vuntz yes please, might not get to it in time

Done in https://github.com/crowbar/crowbar-openstack/pull/751 -- should supersede this PR.

dirkmueller:
Not needed anymore

PR #686, opened 2016-12-20
barbican: Set missing port and proper var for host in case of non-HA

Itxaka:
The missing/wrong vars were causing to make barbican bind to `localhost:80` so it barbican was working but not listening

### DIFF
--- a/chef/cookbooks/barbican/recipes/api.rb
+++ b/chef/cookbooks/barbican/recipes/api.rb
@@ -120,8 +120,8 @@ if node[:barbican][:ha][:enabled]
   bind_host = admin_address
   bind_port = node[:barbican][:ha][:ports][:api]
 else
-  bind_host = node[:barbican][:api][:host]
-  bind_port = node[:barbican][:api][:port]
+  bind_host = node[:barbican][:api][:bind_host]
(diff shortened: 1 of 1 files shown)


chef/cookbooks/barbican/attributes/default.rb
```diff
@@ -22,7 +22,8 @@
 override[:barbican][:user] = "barbican"
 
 default[:barbican][:debug] = false
-default[:barbican][:api][:bind_host] = "*"
```
toabctl: Removing this does not work for the HA case. The attribute is used in:
```
libraries/helpers.rb
22:          ha_bind_host: node[:barbican][:api][:bind_host] == "*" ? "0.0.0.0" : @cluster_admin_ip,
```

Itxaka:
thanks @toabctl I changed it to fix it properly instead of changing stuff around. Now it will use the proper binf_host/bind_port vars that are used in the rest of the code

Itxaka:
[(5) Update templates for config files for Newton - Barbican](https://trello.com/c/6ozqOvAG/370-5-update-templates-for-config-files-for-newton-barbican)

PR #687, opened 2016-12-20
crowbar: remove slash after port in rabbit settings

sidthe:
the rabbit[:rabbitmq][:vhost] attribute already has a slash in its value for the vhost. The extra slash in the helpers library generates a string that contains two slashes as a result, such as ip_add:port//vhost. This PR removes the extra slash

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -131,7 +131,7 @@ class CrowbarOpenStackHelper
           vhost: rabbit[:rabbitmq][:vhost],
           url: "rabbit://#{rabbit[:rabbitmq][:user]}:" \
             "#{rabbit[:rabbitmq][:password]}@" \
-            "#{rabbit[:rabbitmq][:address]}:#{rabbit[:rabbitmq][:port]}/" \
+            "#{rabbit[:rabbitmq][:address]}:#{rabbit[:rabbitmq][:port]}" \
             "#{rabbit[:rabbitmq][:vhost]}"
         }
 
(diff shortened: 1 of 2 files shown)


toabctl:
@sidthe can you add your explanation to the commit msg please?

sidthe:
@toabctl done

nicolasbock:
@dirkmueller, @sidthe, I added a schema validation pattern.

sidthe:
@nicolasbock thanks! was on my todo list. However, this being Ruby, I think the right pattern would be `"/^\//"` but not 100% sure

sidthe:
[(1) Remove trailing slash in rabbit settings](https://trello.com/c/5QsmtWP3/413-1-remove-trailing-slash-in-rabbit-settings)

nicolasbock:
@sidthe Good point. I wasn't sure how that pattern was supposed to look like and tested it only in an online schema validation tool. I was going to wait for gating to tell me how right or wrong I was :)

sidthe:
@nicolasbock this one is pretty nice for the matter http://rubular.com/

Itxaka:
*super* nitpick. I dont think that we should mention "trailing" as its not trailing anywhere, its in the middle of the full address or at the start of the vhost. It confused me a bit at the start after reading trailing to find the regex matching at the start of the line

sidthe:
@Itxaka ok - once the build passes I will squash all commits into something that removes trailing

nicolasbock:
@sidthe thanks for the link!

nicolasbock:
I triggered a rebuild because the previous build was recycled already.

sidthe:
After a little testing and talking with Michal who already had experience with the transport url as part of his work, I'll be dropping that, as the trailing slash in this case is not treated as a separator

PR #688, opened 2016-12-20
keystone: drop the use of 'use_syslog' as its deprecated in newton

Itxaka:
http://docs.openstack.org/newton/config-reference/tables/conf-changes/keystone.html


Itxaka:
[(5) Update templates for config files for Newton - Keystone](https://trello.com/c/t0rgQ51S/375-5-update-templates-for-config-files-for-newton-keystone)

PR #695, opened 2016-12-21
aodh: upgrade config template to newton

sidthe:


sidthe:
[(5) Update templates for config files for Newton - aodh](https://trello.com/c/3uuoP2bG/369-5-update-templates-for-config-files-for-newton-aodh)

chef/cookbooks/aodh/templates/default/aodh.conf.erb
```diff
@@ -22,12 +23,7 @@ identity_uri = <%= @keystone_settings['admin_auth_url'] %>
 admin_user = <%= @keystone_settings['service_user'] %>
```
toabctl: These are also deprecated.

PR #696, opened 2016-12-21
ceilometer: replace deprecated parameters in template

sidthe:


sidthe:
[(5) Update templates for config files for Newton - Ceilometer](https://trello.com/c/mE4o4pAj/371-5-update-templates-for-config-files-for-newton-ceilometer)

PR #698, opened 2016-12-22
[3.0] cookbooks: Use new node_search_with_cache helper

vuntz:
~~Requires https://github.com/crowbar/crowbar-core/pull/931 (merged)~~

~~Contains https://github.com/crowbar/crowbar-openstack/pull/681 (merged)~~

~~Contains https://github.com/crowbar/crowbar-openstack/pull/805 (merged)~~

Backport of https://github.com/crowbar/crowbar-openstack/pull/666

### DIFF
--- a/chef/cookbooks/ceilometer/libraries/mongodb.rb
+++ b/chef/cookbooks/ceilometer/libraries/mongodb.rb
@@ -27,6 +27,48 @@ include Chef::Mixin::ShellOut
 
 module CeilometerHelper
   class << self
+    def replica_set_members(node)
+      CrowbarUtilsSearch.node_search_with_cache(node,
+                                                "roles:ceilometer-server",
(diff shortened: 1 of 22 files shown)


vuntz:
> That PR contained another change, and although I might be missing something, but I haven't found that in the history:
> d848e5f
> Looking at the source code we might need that one as well?

I was initially thinking it's only for master, but upon digging, it's actually also valid for stable/3.0. Do you want me to backport that as a separate PR, or here?

matelakat:
@vuntz Yes, I think it would make sense to have the backport of #391 separately, so this PR will only be dealing with the evil change.

vuntz:
@matelakat I rebased on top of https://github.com/crowbar/crowbar-openstack/pull/805 (so it appears here also).

vuntz:
We won't need that backport anymore.

PR #699, opened 2016-12-23
keystone: update ldap config parameters in template

sidthe:


chef/cookbooks/keystone/attributes/default.rb
```diff
@@ -92,19 +92,26 @@
...
 default[:keystone][:ldap][:group_member_attribute] = "member"
-default[:keystone][:ldap][:group_attribute_ignore] = ""
-default[:keystone][:ldap][:group_allow_create] = true
-default[:keystone][:ldap][:group_allow_update] = true
-default[:keystone][:ldap][:group_allow_delete] = true
-default[:keystone][:ldap][:use_pool] = true
 default[:keystone][:ldap][:group_members_are_ids] = false
+default[:keystone][:ldap][:group_additional_attribute_mapping] = ""
+default[:keystone][:ldap][:group_ad_nesting] = false
+default[:keystone][:ldap][:use_pool] = true
```
Itxaka: This is the default value already provided in the default config, if we are not gonna change it, doesn't it make sense to just remove it?

After all, we want to only add keys that we know we are overriding while letting the default values stay.

chef/cookbooks/keystone/attributes/default.rb
```diff
@@ -92,19 +92,26 @@
...
+default[:keystone][:ldap][:pool_size] = 10
+default[:keystone][:ldap][:pool_retry_max] = 3
```
Itxaka: from here downwards, all keys are already in their default values as commented above. It migth make more sense to just remove them unless we are gonna override them.

chef/cookbooks/keystone/attributes/default.rb
sidthe: @Itxaka for LDAP the rules is to expose all parameters to the user, this is what the card is all about

sidthe:
[(2) Ensure new Keystone LDAP options are in schema](https://trello.com/c/5WlLyj70/396-2-ensure-new-keystone-ldap-options-are-in-schema)

chef/cookbooks/keystone/attributes/default.rb
sidthe: @Itxaka as written above. For LDAP we expose everything to the user.

chef/cookbooks/keystone/attributes/default.rb
Itxaka: Looks good then :D

PR #702, opened 2016-12-27
keystone: remove PKI format option for users

sidthe:
PKI tokens in Keystone are known to not be fully secure, we should not offer this option anymore


sidthe:

crowbar_framework/app/helpers/barclamp/keystone_helper.rb
```diff
@@ -31,7 +31,6 @@ def server_frontends_for_keystone(selected)
     def token_formats_for_keystone(selected)
       options_for_select(
         [
-          ["PKI", "PKI"],
```
Itxaka: Do we still need this if there is only one option to choose? Should be removed IMHO

sidthe: @Itxaka we still need this, as there is a plan to have fernet as an option as well

mjura: If there is only one option, there is no need to select, choose option can be removed from Keystone proposal

mjura:
In general looks ok, but as I said there is no need to keep select option in Keystone proposal only with one choice

sidthe:
@mjura see above the same question from Itxaka and my reply to it. The plan is to have fernet introduced as an option very soon, so no need to drop the select.

mjura:
@sidthe: ok please fix then comments in https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/resources/sync_mark.rb#L32

toabctl:
@sidthe what happens if somebody uses PKI and upgrades? Isn't there a migration needed to UUID?

sidthe:
@toabctl we've looked at it with Itxaka, and we couldn't find anything that would hint to a possibility of such a migration. You have any ideas?

PR #706, opened 2016-12-29
neutron: run multiple api workers

Itxaka:


Itxaka:
[(2) neutron: Run multiple api_workers](https://trello.com/c/gX8NK4yn/402-2-neutron-run-multiple-api-workers)

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -204,7 +204,7 @@ use_ssl = <%= @ssl_enabled ? "True" : "False" %>
 # default is equal to the number of CPUs available for best performance.
 # (integer value)
 #api_workers = <None>
-api_workers = 0
+api_workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
```
Itxaka: Im still not sure the reasoning behind this.

Anything lower than 4 cpus will use 2 workers.
Anything over 6 cpus will use 4 workers always.

It does not seem to increase workers linearly with cpu, which is something that I would expect. 

Maybe it needs a more complex helper to calculate the amount of workers properly?

toabctl: @Itxaka why not leaving the default (None)?
```
# Number of separate API worker processes for service. If not specified, the
# default is equal to the number of CPUs available for best performance.
# (integer value)
```

Itxaka: Yeah, I guess thats the correct way. I thougth that the method used across all cookbooks for calculating the api_workers was due to not have workers on all cpus, but it doesnt look like it. Im not sure whats the real policy here regarding workers == cpus or we want to have less workers or whatnot. But I guess thats out of scope for this :D

dirkmueller: @Itxaka we definitely don't want to have workers==cpu as that is excessive on big machines (think 32 cores and more). also at least in the past neutron used to have terrible races with large number of workers.. better be conservative and use the same code like the others, even if its bad.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -204,7 +204,6 @@ use_ssl = <%= @ssl_enabled ? "True" : "False" %>
 # default is equal to the number of CPUs available for best performance.
 # (integer value)
 #api_workers = <None>
-api_workers = 0
```
dirkmueller: I suggest to not just use the default as that is excessive on large core-machines (think 32 cores and more). also at least in the past neutron used to have terrible races with api_workers !=0. we might want to be conservative here and use exactly the same code like everywhere else (```<%= [node["cpu"]["total"], 2, 4].sort[1] %>```)

chef/cookbooks/neutron/templates/default/neutron.conf.erb
toabctl: @Itxaka then let's go back to your initial proposal.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
Itxaka: Updated!

toabctl:
Tempest (linuxbridge) passed. It failed because of radosgw. merging...

PR #709, opened 2017-01-02
trove: Handle radosgw being on a different node properly

dirkmueller:
The mechanical code changes just for the sake of satisfying a
stupid robot are actually done by humans, and humans tend to
make mistakes.


dirkmueller:
@rhafer ah I see. updated commit message.

PR #711, opened 2017-01-02
keystone: migrate value of token_format

sidthe:
As part of https://github.com/crowbar/crowbar-openstack/pull/702 while the behavior of 702 is correct, it leaves the value for the token_format incorrect in the attribute list, which is a bug. This PR migrates the value of the attribute since in the new release the PKI is not supported.


sidthe:

PR #714, opened 2017-01-03
tempest: minify config template

dirkmueller:


dirkmueller:
[(5) Update templates for config files for Newton - Tempest](https://trello.com/c/x99NxFiG/415-5-update-templates-for-config-files-for-newton-tempest)

PR #720, opened 2017-01-04
crowbar: remove dependency on ca.pem availability

sidthe:
not all services have a ca.pem file. This PR makes the entry optional to the availability of the users set path


sidthe:
[(5) Split aodh out of ceilometer barclamp](https://trello.com/c/ZtGoKQwG/403-5-split-aodh-out-of-ceilometer-barclamp)

PR #721, opened 2017-01-04
upgrade: fake the nodes_db_dump step for now until it is implemented

MaximilianMeister:
see https://github.com/crowbar/crowbar-core/pull/946#issuecomment-270151292


jsuchome:
Ah, it's part of crowbar-openstack? Strange, is the only upgrade related API in this module? If so, should not we rather move it to upgrade controller of crowbar-core?

MaximilianMeister:
handeled in https://github.com/crowbar/crowbar-core/pull/946 now as it's the same problem as with the `adminbackup` API. we need to make a distinction between using the upgrade status library or not. the later implementation might happen in this repo though

PR #724, opened 2017-01-05
keystone: Remove unused options

nicolasbock:
We are not using '.*dumb_member' in the template anymore so remove it
from the conf file. Also sorted the ldap keys in the template and schema
files to make comparisons between the two easier.

### DIFF
--- a/chef/cookbooks/keystone/templates/default/keystone.conf.erb
+++ b/chef/cookbooks/keystone/templates/default/keystone.conf.erb
@@ -61,8 +61,6 @@ url = <%= node[:keystone][:ldap][:url] %>
 user = <%= node[:keystone][:ldap][:user] %>
 password = <%= node[:keystone][:ldap][:password] %>
 suffix = <%= node[:keystone][:ldap][:suffix] %>
-use_dumb_member = <%= node[:keystone][:ldap][:use_dumb_member] %>
-dumb_member = <%= node[:keystone][:ldap][:dumb_member] %>
 allow_subtree_delete = <%= node[:keystone][:ldap][:allow_subtree_delete] %>
 query_scope = <%= node[:keystone][:ldap][:query_scope] %>
 page_size = <%= node[:keystone][:ldap][:page_size] %>


chef/data_bags/crowbar/template-keystone.json
```diff
@@ -63,58 +63,58 @@
         "ca_certs": "/etc/keystone/ssl/certs/ca.pem"
       },
       "ldap" : {
-        "url": "ldap://localhost",
-        "user": "dc=Manager,dc=example,dc=com",
-        "password": "",
-        "suffix": "cn=example,cn=com",
+        "alias_dereferencing": "default",
         "allow_subtree_delete": false,
+        "auth_pool_connection_lifetime": 60
```
Itxaka: missing comma

chef/data_bags/crowbar/template-keystone.schema
```diff
@@ -68,58 +68,58 @@
                       "ca_certs": { "type": "str" }
                     }},
                     "ldap" : { "type" : "map", "mapping": {
-                      "url": { "type": "str" },
-                      "user": { "type": "str" },
-                      "password": { "type": "str" },
-                      "suffix": { "type": "str" },
+                      "alias_dereferencing": { "type": "str" },
                       "allow_subtree_delete": { "type": "bool" },
+                      "auth_pool_connection_lifetime": { "type": "int" }
```
Itxaka: missing comma

nicolasbock:
@Itxaka Fixed.

nicolasbock:
@sidthe Why did we remove the `.*dumb` settings? Are they deprecated from keystone? Or did we decide that we shouldn't expose those through the barclamp anymore?

nicolasbock:
@sidthe thanks.

sidthe:
@nicolasbock why did you rearrange stuff this way? Before this change, the way the parameters were set was reflecting the order in which they appear in the actual keystone.conf file which at least for me in some cases was helpful during migration

nicolasbock:
@sidthe I can re-arrange it back. Now they are alphabetically sorted in the conf / json / schema files.

nicolasbock:
@sidthe I agree though that we should definitely keep the order synched.

sidthe:
@nicolasbock I agree that the way you arranged them is easier on the eye, but it does not strike me as helpful during future cross checks... Currently, they are also synced in many cases to the order in the attributes file

PR #725, opened 2017-01-05
nova: Add missing ec2-api schema

toabctl:
When deploying ec2-api, a password was never generated
and the schema for the database parameters was missing in the proposal.


toabctl:
> I'm somewhat sceptical about the new toplevel attribute tree. I would really really prefer to move that under the nova tree, because this would be the first barclamp where we make this different.

@dirkmueller fixed but ec2api is stil broken. Needs more commits. Adding wip

toabctl:
Note: This was tested with the ```aws``` CLI:

1. zypper in python-pip
2. pip install aws
3. openstack ec2 credentials list
4. Add the credentials to .aws/config . Looks like:
```
# cat .aws/config 
[default]
aws_access_key_id = 8df8e1bf61da47b9bb084812941b4216 
aws_secret_access_key = 892f0edaee5a417e80011e34c804656e 
region = CustomRegion
```
5. test with: aws --endpoint-url http://localhost:8788 ec2 describe-instances

toabctl:
rebased

PR #733, opened 2017-01-06
trove: Fix config file path

toabctl:
Commit dfb322935fca7 switched to the new config file paths but also
changed the common config file /etc/trove/trove.conf to
/etc/trove/trove-api.conf.d/100-trove-api.conf which seems not to work.

### DIFF
--- a/chef/cookbooks/trove/attributes/default.rb
+++ b/chef/cookbooks/trove/attributes/default.rb
@@ -22,7 +22,7 @@ default[:trove][:debug] = false
 
 default[:trove][:user] = "trove"
 default[:trove][:group] = "trove"
-default[:trove][:api][:config_file] = "/etc/trove/trove-api.conf.d/100-trove-api.conf"
+default[:trove][:config_file] = "/etc/trove/trove.conf.d/100-trove.conf"
 default[:trove][:conductor][:config_file] = \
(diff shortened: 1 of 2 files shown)


toabctl:
UNTESTED!

Itxaka:
I dont really think this is the real issue.

Deploying a clean machine with no trove and then deploying trove leaves it in a wrong state due the error mentioned on irc: `20046 ERROR root NoSuchTableError: instances`

This can be fixed by running `trove-manage --config-file /etc/trove/trove-api.conf.d/100-trove-api.conf db_sync`

After that, trove-api/conductor/taskmanager start with no issues. 

Furthermore, I re-tested that the trove binaries pick the override config file and they work, so this is pointing to a different issue, maybe during the chef-run?

Itxaka:
oh my gooood, another dumb mistake on my part:


from the original commit:
```
execute "trove-manage db sync" do
  command "trove-manage --config-file #{node[:trove][:api][:config_file]} db_sync"
  user node[:trove][:user]
  group node[:trove][:group]
  action :nothing
  only_if { !node[:trove][:db_synced] }
end
```

Action :nothing is WRONG! Thats the real issue, this could get closed.

dirkmueller:
hopefully we don't need this anymore

PR #735, opened 2017-01-06
trove: Fix wrong db_sync resource action

Itxaka:
action :nothing is a leftover of a more complicated setup to be triggered that I experimented on while doing https://github.com/crowbar/crowbar-openstack/commit/dfb322935fca75f6e2b6887fad77c27f90d9c380 and blocks the database for trove to be created+synced, breaking the tempest run for all jobs in the CI


Itxaka:
This is not necessarily required urgently as the real fix is done on the package, which also calls db_sync[0]

[0] https://build.opensuse.org/request/show/449014

vuntz:
Merging to not break the CI over the week-end.

PR #739, opened 2017-01-07
swift: Switch to 620x port range

dirkmueller:
Swift upstream changed from 600x to 620x port range some time ago
in order to avoid port collissions.


dirkmueller:
[(5) S53P4 switch swift to Newton version](https://trello.com/c/IywUtcYv/430-5-s53p4-switch-swift-to-newton-version)

PR #742, opened 2017-01-09
cinder: Avoid self-dependency

dirkmueller:
The generated systemd file had a dependency "Before=" pointing to
itself, causing systemd warnings every chef run


dirkmueller:
[(5) P3: bsc#1004620 - shutdown takes 90s from faulty cinder-looplvm.service](https://trello.com/c/W0PuSXn4/2183-5-p3-bsc-1004620-shutdown-takes-90s-from-faulty-cinder-looplvm-service)

Itxaka:
retriggered jenkins

PR #743, opened 2017-01-09
keystone: revert back certificate attributes setup

sidthe:
Due to the fact that Ceph depends on Keystone signing certificates for communication with Keystone we need to have the right code to set up the attributes and related content.

relates to this https://github.com/crowbar/crowbar-ceph/pull/83

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -286,6 +286,79 @@ end
 
 crowbar_pacemaker_sync_mark "create-keystone_db_sync"
 
+ruby_block "synchronize signing keys for founder and remember them for non-HA case" do
+  only_if { (!ha_enabled || (ha_enabled && CrowbarPacemakerHelper.is_cluster_founder?(node))) &&
+            (node[:platform_family] == "suse") }
+  block do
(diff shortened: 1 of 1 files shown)


sidthe:
needed for https://github.com/crowbar/crowbar-ceph/pull/83

sidthe:
I don't see a way to optimize the block size and usage of blocks across multi lines, so I will be ignoring that. This was supposed to be a partial code revert.

sidthe:

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -286,6 +286,84 @@
 
 crowbar_pacemaker_sync_mark "create-keystone_db_sync"
```
vuntz: You're missing this bit here:
```ruby
# Make sure the certificates code is run on the founder first
crowbar_pacemaker_sync_mark "wait-keystone_certificates" do
  fatal true
end
```

sidthe: indeed, thx!

aplanas:
Actually I agree with @vuntz to do a proper git revert here. Usually helps.

sidthe:
@aplanas if it is not critical to you I would not do it, as I already did some changes here and there, so as long as this is critical I'd just live it the way it is

aplanas:
@sidthe I do not think that is critical, sure.

PR #744, opened 2017-01-10
neutron-ha-tool: increase start timeout (bsc#1011655)

matelakat:
The resource agent for neutron-ha-tool triggers a run of router
migration when start is received. Given that neutron-ha-tool will
migrate routers synchronously, its run will take significantly more
time than the original 120s. This would make the resource fail,
whereas in reality it just hasn't finished yet.

Our measurements showed that the time needed to create a router takes
around 12 seconds, router removal should be much quicker. In total we
are accounting 20 seconds for a router migration.

In case of a failed L3 agent, the start operation will be called that
triggers the router migration. Having a 30 minute timeout for it
leaves room for migrating 90 routers. If the operation times out, the
neutron-ha-tool will be immediately moved to the next node, where it
can move 90 more before timing out, and being moved to the next
node. This sets a (network node number) * 90 limit for the routers
that could be migrated.

https://bugzilla.suse.com/show_bug.cgi?id=1011655


matelakat:
[P2.5: bsc#1011655 - neutron-ha-tool 120s start timeout too low for large clouds](https://trello.com/c/hAWQkAM8/31-p2-5-bsc-1011655-neutron-ha-tool-120s-start-timeout-too-low-for-large-clouds)

aplanas:
nit: if comes from a bug, we need to put the bsc# number in the commit message (same for the backport)

matelakat:
@aplanas @aspiers change amended, please review

matelakat:
@dirkmueller I thought we wait until gate passes with the merges? That was the reason for me not pushing the merge button.

dirkmueller:
@matelakat: what could possibly go wrong with increasing timeout ;-) ha-tool is not tested by mkcloud

PR #749, opened 2017-01-12
glance: fix registry config whitespace

Itxaka:
seems like the existance of the whitespace was
causing issue when parsing the config


Itxaka:
Will wait for ssl ci job to pass before merging, just in case :o

dirkmueller:
seems to work, thanks!

PR #751, opened 2017-01-12
Re-add use_l2pop attribute

vuntz:
Quoting @dirkmueller:

> l2pop with HA/vxlan just doesn't seem to work at all. Only enable what upstream does in devstack: when DVR is used.

This should replace https://github.com/crowbar/crowbar-openstack/pull/684

### DIFF
--- a/chef/data_bags/crowbar/migrate/neutron/105_add_back_use_l2pop.rb
+++ b/chef/data_bags/crowbar/migrate/neutron/105_add_back_use_l2pop.rb
@@ -1,0 +1,16 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "use_l2pop"
+    # Only enable it if DVR is used
+    a["use_l2pop"] = a["use_dvr"]
+  end
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless ta.key? "use_l2pop"
+    a.delete("use_l2pop")
+  end
+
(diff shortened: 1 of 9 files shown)


chef/data_bags/crowbar/migrate/neutron/105_add_back_use_l2pop.rb
```diff
@@ -0,0 +1,16 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "use_l2pop"
+    # Only enable it if DVR is used
+    a["use_l2pop"] = a["use_dvr"]
```
aplanas: why is not better to add this logic when we export the variable `use_l2pop` in the templates? if dvr are enabled or disabled after the barclamp, use_l2pop will not change with this approach

vuntz: Because some users want to use l2pop without DVR.

aplanas: Reading the quote from dirk looks like that this option was is not OK outside DVR.

vuntz: It is okay, but it comes with some tradeoffs so people might not always want it.

PR #753, opened 2017-01-12
Improve LBaaS UI in crowbar

vuntz:


rsalevsky:
Do you have a screenshot from the improved UI? This will make the review easier.

vuntz:
@rsalevsky:

capture du 2017-01-12 16-59-34

dirkmueller:
Looks much better, thanks. do we want to remove the use_lbaas option from UI ? I don't think we need to expose this, since there is no obvious downside of enabling it and its enabled anyway by default.

vuntz:
> do we want to remove the use_lbaas option from UI ? I don't think we need to expose this, since there is no obvious downside of enabling it and its enabled anyway by default.

Good point, and I agree. Done.

PR #754, opened 2017-01-12
neutron: Fix F5 attributes and provide reasonable defaults

vuntz:


chef/data_bags/crowbar/template-neutron.json
```diff
@@ -48,12 +48,12 @@
...
       "f5": {
         "icontrol_hostname": "",
-        "icontrol_username": "",
-        "icontrol_password": "",
-        "parent_ssl_profile": "",
-        "external_physical_mappings": "",
-        "vtep_folder": "",
-        "vtep_selfip_name": "",
+        "icontrol_username": "admin",
+        "icontrol_password": "admin",
```
dirkmueller: I"m not a fan of hardcoded passwords (since you can't read it in the ui anyway). can we just leave it empty?

vuntz: @dirkmueller done

vuntz:
@dirkmueller rebased

vuntz:
> Personally I wouldn't worry about a migration for this since it's just recently been introduced and not yet released but it lgtm

I was thinking the same, but then realized that beta participants might upgrade from one version to the other with a zypper up. I'd like to avoid being the one breaking that ;-)

dirkmueller:
I'm almost sure that the f5 support wasn't in M6 yet... Anyway there is mire important matter

PR #755, opened 2017-01-12
swift: disable splice (bsc#1019668)

aplanas:


aplanas:
Revert: https://github.com/crowbar/crowbar-openstack/pull/759

PR #758, opened 2017-01-13
[stable/3.0] Provide .openrc for nodes with neutron-network role

jsuchome:
So we can execute neutron-ha-tool or nova commands if they
are in the different cluster than keystone.

OK, I'm not sure if this is correct, so it's rather initial proposal...

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -36,6 +36,19 @@ if use_l3_agent
     #sensitive true
     action :create
   end
+
+  # We need .openrc present at network node so the node can use neutron-ha-tool even
+  # when located in separate cluster
+  template "/root/.openrc" do
+    source "openrc.erb"
+    cookbook "keystone"
+    owner "root"
+    group "root"
+    mode 0600
+    variables(
+      keystone_settings: keystone_settings
+    )
+  end
 end
 
 # Wait for all "neutron-network" nodes to reach this point so we know that they will


jsuchome:
@dirkmueller this proposal is meant for Cloud6...

jsuchome:
[(8) S54P6 (S49P16) Orchestrate upgrade of controller nodes in HA setup (upgrade step 12 part 2 -- HA/multiple clusters)](https://trello.com/c/MquyXrTn/252-8-s54p6-s49p16-orchestrate-upgrade-of-controller-nodes-in-ha-setup-upgrade-step-12-part-2-ha-multiple-clusters)

chef/cookbooks/keystone/recipes/openrc.rb
```diff
@@ -0,0 +1,27 @@
...
+# useful with .openrc
+package "python-openstackclient"
+
+template "/root/.openrc" do
+  source "openrc.erb"
+  owner "root"
+  group "root"
+  mode 0600
+  variables(
+    keystone_settings: KeystoneHelper.keystone_settings(node, @cookbook_name)
```
vuntz: I'm afraid that the `@cookbook_name` bit here will not result in what you expect when included from neutron/nova recipes on nodes that are not keystone nodes...

jsuchome: Uhm... so you think I should rather use the template directly from `network_agents.rb` with `keystone_settings` retrieved from the neutron cookbook?

vuntz: Or you make this a LWRP where the attributes would be the node and a cookbook name?

jsuchome: Actually, are you sure there's something broken with this? The only part where `keystone_settings` uses `cookbook_name` is for setting `service_user`/`service_password` values and those are not needed by `openrc.erb`

vuntz: Isn't it also used to find the keystone server with the search query?

jsuchome: The search is there, but `cookbook` seems irrelevant https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/keystone/libraries/helpers.rb#L104

vuntz: @jsuchome it's used there https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/keystone/libraries/helpers.rb#L36 and this has an impact on the search query

jsuchome:
@vuntz do you agree with this version of PR?

PR #763, opened 2017-01-15
ec2-api: Limit ec2 metadata workers

dirkmueller:
the default is number of cpus which is just eating too much memory for no real gain


aplanas:
rechecking for HA

PR #767, opened 2017-01-16
magnum: add SSL support

sidthe:
This PR adds ssl support for Magnum API


sidthe:
[(3) S53P5 S46P16: magnum: Add SSL support](https://trello.com/c/M3vczJEm/197-3-s53p5-s46p16-magnum-add-ssl-support)

mjura:
Hmm it looks that Magnum doesn't support `ca_files`

sayalilunkad:
I am not sure about this myself but I wonder if it is common to have the same data stored both as attribute and in data bags?

nicolasbock:
>I am not sure about this myself but I wonder if it is common to 
>have the same data stored both as attribute and in data bags?

I looked through our other recipes and it is common to do that. 
Whether we need to is another question though :smile:.

sidthe:
@sayalilunkad can you give an example? not sure what you refer to? what we store in data bags is actually barclamp related, and attributes file is independent of that

mjura:
This can be merged after changes to `openstack-magnum` package will be approved.
https://build.opensuse.org/request/show/453568

dirkmueller:
@mjura this branch needs rebasing first. Any chance that you could answer tje review comments in the previous package submission? Also it would be good to have this submitted to cloud:openstack:master as well/first

dirkmueller:
@sidthe could you rebase, please?

sidthe:
@dirkmueller squashed/rebased

chef/cookbooks/magnum/templates/default/magnum.conf.erb
```diff
@@ -13,6 +13,13 @@ control_exchange = magnum
 [api]
 port = <%= @bind_port %>
 host = <%= @bind_host %>
+<% if node[:magnum][:api][:protocol] == "https" %>
+ssl_cert_file = <%= node[:magnum][:ssl][:certfile] %>
+ssl_key_file = <%= node[:magnum][:ssl][:keyfile] %>
+enabled_ssl = false
+<% else -%>
+enabled_ssl = false
+<% end %>
```
vuntz: Hrm, do we need this if listening behind apache?

chef/data_bags/crowbar/migrate/magnum/101_add_ssl_attributes.rb
```diff
@@ -0,0 +1,15 @@
+def upgrade(ta, td, a, d)
+  a["ssl"] = {}
+  a["ssl"]["certfile"]       = ta["ssl"]["certfile"]
+  a["ssl"]["keyfile"]        = ta["ssl"]["keyfile"]
+  a["ssl"]["generate_certs"] = ta["ssl"]["generate_certs"]
+  a["ssl"]["insecure"]       = ta["ssl"]["insecure"]
+  a["ssl"]["cert_required"]  = ta["ssl"]["cert_required"]
+  a["ssl"]["ca_certs"]       = ta["ssl"]["ca_certs"]
```
vuntz: Just do `a["ssl"] = ta["ssl"]` ;-)

crowbar_framework/config/locales/magnum/en.yml
```diff
@@ -31,4 +31,15 @@ en:
         cert_header: 'Certificate Manager'
         cert:
           cert_manager_type: 'Plugin'
-
+        api:
+          protocol: 'Protocol'
+          bind_open_address: 'Open Address'
+          bind_port: 'Port'
```
vuntz: I don't think these two lines are used, I'd drop them.

chef/cookbooks/magnum/templates/default/magnum.conf.erb
dirkmueller: yep, doesn't look correct.

chef/cookbooks/magnum/recipes/api.rb
```diff
@@ -99,6 +117,27 @@
...
+  if node[:magnum][:ssl][:cert_required]
+    ssl_cacert node[:magnum][:ssl][:ca_certs]
+  end
+end
+
+apache_site "magnum-api.conf" do
+  enable true
+end
+
 magnum_service "api" do
```
dirkmueller: that needs to be removed now

chef/cookbooks/magnum/templates/default/magnum.conf.erb
sidthe: yes, leftover from the first implementation

chef/cookbooks/magnum/recipes/api.rb
sidthe: not sure I understand why?

Also, any hint regarding setup.rb adjustment? I can't seem to figure it out?

chef/data_bags/crowbar/migrate/magnum/101_add_ssl_attributes.rb
sidthe: ok

crowbar_framework/config/locales/magnum/en.yml
sidthe: right

chef/cookbooks/magnum/recipes/api.rb
dirkmueller: --insecure needs to be used when magnum is behind insecure ssl

sidthe:
@vuntz  @dirkmueller hope that's ok

chef/cookbooks/magnum/templates/default/magnum.conf.erb
```diff
@@ -13,6 +13,10 @@ control_exchange = magnum
 [api]
 port = <%= @bind_port %>
 host = <%= @bind_host %>
+<% if node[:magnum][:api][:protocol] == "https" %>
+ssl_cert_file = <%= node[:magnum][:ssl][:certfile] %>
+ssl_key_file = <%= node[:magnum][:ssl][:keyfile] %>
+<% end %>
```
vuntz: No need to add this as really, the magnum wsgi app knows nothing about SSL. Also, port / host can be dropped too for the same reason.

sidthe: dropped

chef/cookbooks/magnum/recipes/api.rb
```diff
@@ -16,19 +16,37 @@
 # Cookbook Name:: magnum
 # Recipe:: api
 #
+include_recipe "apache2"
+include_recipe "apache2::mod_wsgi"
+include_recipe "apache2::mod_rewrite"
```
vuntz: Stupid question: why do we need mod_rewrite?

chef/cookbooks/magnum/templates/default/magnum.conf.erb
```diff
@@ -10,10 +10,6 @@ executor_thread_pool_size = 256
 transport_url = <%= @rabbit_settings[:url] %>
 control_exchange = magnum
 
-[api]
-port = <%= @bind_port %>
-host = <%= @bind_host %>
```
vuntz: I guess next time I'll be more explicit ;-) Removing this means you can also remove bits from common.rb.

chef/cookbooks/magnum/recipes/api.rb
sidthe: gods of copy paste, removed

chef/cookbooks/magnum/templates/default/magnum.conf.erb
sidthe: ok, fixed

dirkmueller:
hmm, PR gating with want_magnum=1 fails.. :-/

sidthe:
@dirkmueller :-/ I haven't done any manual tests on that today, last time I think it was building and at least the manual patching was doing the job (endpoint reconfiguration, conf files), but as you remember the actual communication via ssl to magnum api was not working, so I was a little surprised you re-opened this today. Were there any fixes that would make that work?

dirkmueller:
@sidthe normal pr testing does not test magnum because the test automation is incomplete and conflicts with everything else. So it might not necessarily be a new problem.

sidthe:
@mjura can you please have a look at the magnum job? any idea why testing fails there. Does not look like it is SSL related.

mjura:
@sidthe : I've submitted fix for Magnum jenkins job https://github.com/SUSE-Cloud/automation/pull/1650

sidthe:
@mjura looks like @dirkmueller did there something somewhere according to the comment he left in your PR. I just rebased and retriggered the gating with want magnum for both jobs, lets see

sidthe:
same issue as before here. Dirk, weren't you saying that there was a fix you submitted somewhere?

dirkmueller:
@sidthe yes, and the jenkins run without your pull request is green: https://ci.suse.de/user/cloud/my-views/view/Cloud7/job/cloud-mkcloud7-job-magnum-x86_64/ 

That does might point to an issue in your PR, although I have not debugged it.

chef/cookbooks/magnum/recipes/api.rb
```diff
@@ -99,6 +116,23 @@
...
 
-magnum_service "api" do
-  use_pacemaker_provider ha_enabled
+crowbar_openstack_wsgi "WSGI entry for magnum-api" do
+  bind_host bind_host
+  bind_port bind_port
+  daemon_process "magnum-api"
+  user node[:magnum][:user]
+  group node[:magnum][:group]
+  script_alias "/srv/www/magnum-api/app.wsgi"
```
dirkmueller: thats the default I think..

chef/cookbooks/magnum/recipes/api.rb
```diff
@@ -99,6 +116,23 @@
 
 crowbar_pacemaker_sync_mark "create-magnum_register"
 
-magnum_service "api" do
-  use_pacemaker_provider ha_enabled
+crowbar_openstack_wsgi "WSGI entry for magnum-api" do
+  bind_host bind_host
```
dirkmueller: node.normal[:apache][:listen_ports_crowbar] needs to be adjusted for this switch to the WSGI service.

chef/cookbooks/magnum/recipes/api.rb
sidthe: right, but this is not set as part of the package, which was one of the things that I believe was broken in the first place, and currently is not set either. I guess it needs either an upstream patch or some work in our own packages? My understanding is that without this, it won't work

dirkmueller: my understanding is that the LWRP defaults to a script_alias that you set explictly when the daemon_process parameter is set properly. see https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/crowbar-openstack/providers/wsgi.rb#L144

sidthe: I am probably missing the point. You are concerned that I explicitly list the default and this is unnecessary?

sidthe:
@dirkmueller looks like it finally works. The magnum recipe was not installing the openstack-magnum-api package where you had the wsgi fix, but now it does and manual patching of my mkcloud gets me results

sidthe:
@Itxaka looks like adding flavors breaks the gating for me

chef/cookbooks/magnum/recipes/api.rb
dirkmueller: Yes, violation of dry

sidthe: :) that's over neat picking, since I guess this is what we do everywhere else. Anyway - I will remove it. Maybe you have an idea why flavors are breaking the build?

Itxaka:
@sidthe rebase pls!

sidthe:
@Itxaka it was rebased yesterday night, at least 10 hours after your code was merged. Has anything changed?

Itxaka:
@sidthe there was a patch for the flavors merged 8 hours ago: https://github.com/crowbar/crowbar-openstack/pull/835 and the last rebase was I believe 13 hours ago?

sidthe:
@vuntz are you sure there is anything to be done? When I look at the Magnum code for HA - the ssl looks properly enabled there. It was there before I even touched the ssl implementation.

sidthe:
@mjura since those tests are not even running with SSL on, what could be the reason the cluster does not get deployed? In my manual tests, this patch works properly

vuntz:
> @vuntz are you sure there is anything to be done? When I look at the Magnum code for HA - the ssl looks properly enabled there. It was there before I even touched the ssl implementation.

Ah, indeed.

mjura:
@sidthe: I will try rebuild it and check it on existing env

chef/cookbooks/magnum/recipes/api.rb
```diff
@@ -99,6 +118,25 @@
 
 crowbar_pacemaker_sync_mark "create-magnum_register"
 
-magnum_service "api" do
```
dirkmueller: just dropping this is not good enough, we want to support upgrade. so you need to stop and disable the systemd service / pacemaker resources.

sidthe: is that better @dirkmueller ?

sayalilunkad:
While testing this it fails with
>FATAL: NameError: Cannot find a resource for magnum_api_resource on suse version 12.2

And it makes sense because magnum_api_resource is not defined before this

sidthe:
thanks Sayali. The gods of copy-paste were not good to me. I have another problem with SSL again though, which seems unrelated. Will need to figure it out.

chef/cookbooks/magnum/recipes/api.rb
```diff
@@ -118,6 +118,22 @@
...
+  ignore_failure true
+end
+
+magnum_api_primitive_name = "magnum-api"
+
+pacemaker_primitive magnum_api_resource do
+  agent node[:magnum][:ha][:api][:agent]
+  action [:stop, :delete]
+  only_if "crm configure show #{magnum_api_primitive_name}"
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
toabctl: are you sure that you can use ```only_if``` twice?

sidthe:
@dirkmueller we looked at it with Tom, and it works now. Could you please reconsider the change request status?

sidthe:
@dirkmueller done

dirkmueller:
Can we remove the donotmerge label? Has this been successfully tested ?

sidthe:
@dirkmueller it has been manually tested. Last time we looked at it jointly with Tom and it was doing what it is supposed to do.

rsalevsky:
Needs a rebase.

sayalilunkad:
what is the status here?

vuntz:
Obsoleted by https://github.com/crowbar/crowbar-openstack/pull/1608

PR #769, opened 2017-01-16
Swift: minify template

sayalilunkad:


sayalilunkad:
[(5) S53P7 Update templates for config files for Newton - Swift](https://trello.com/c/1EdmK6Gb/428-5-s53p7-update-templates-for-config-files-for-newton-swift)

chef/cookbooks/swift/templates/default/swift.conf.erb
```diff
@@ -1,7 +1,185 @@
 [swift-hash]
+
+# swift_hash_path_suffix and swift_hash_path_prefix are used as part of the
+# hashing algorithm when determining data placement in the cluster.
+# These values should remain secret and MUST NOT change
+# once a cluster has been deployed.
+# Use only printable chars (python -c "import string; print(string.printable)")
+
```
nicolasbock: Why are you adding back the comments after 'minifying' the template first? Is that what we are doing now?

sayalilunkad: yes that's how I understand it, so first we minimized config templates and then update the comments that are generated from the new release(here newton).

aplanas:
@sayalilunkad actually is a bit different AFAIK:

1.- The package generate a new config file that is put in place in /etc/swift/
2.- The package create a new directory something like /etc/swift/swift.conf.d
3.- We remove all comments from the old .conf and move to the new directory (we can this mimify), we need to update crowbar to save it in the new place and double check that the package create this this dir
4.- We undated the mimified version, with no comments, that is win the new place

You can check nova, neutron and others as example

chef/cookbooks/swift/recipes/default.rb
```diff
@@ -25,7 +25,7 @@
   package "swift"
 end
 
-template "/etc/swift/swift.conf" do
+template "/etc/swift/swift.conf.d/100-swift.conf" do
```
aplanas: Can you check the recipes from Nova and other as examples? We are using some metadata in node[] to store this and reuse.

chef/cookbooks/swift/recipes/dispersion.rb
```diff
@@ -93,7 +93,7 @@
   swift_cmd="swift"
 end
 
-template "/etc/swift/dispersion.conf" do
+template "/etc/swift/swift.conf.d/100-dispersion.conf" do
```
aplanas: same here (and in the rest)

chef/cookbooks/swift/recipes/default.rb
Itxaka: Agreed, would be better to store it into the chef cookbook attributes :)

chef/cookbooks/swift/templates/default/swift.conf.erb
```diff
@@ -1,120 +1,7 @@
...
-# in metadata however the ring file name will always be 'object.ring.gz' for
-# backwards compatibility.  If no policies are defined a policy with index 0
-# will be automatically created for backwards compatibility and given the name
-# Policy-0.  A default policy is used when creating new containers when no
-# policy is specified in the request.  If no other policies are defined the
-# policy with index 0 will be declared the default.  If multiple policies are
-# defined you must define a policy with index 0 and you must specify a
-# default.  It is recommended you always define a section for
-# storage-policy:0.
 [storage-policy:0]
```
Itxaka: I think this is already provided in the file /etc/swift/swift.conf by default which is provided by the package so we may not need it in the override files. Its gonna be there anyway.

chef/cookbooks/swift/recipes/default.rb
sayalilunkad: done!

chef/cookbooks/swift/templates/default/swift.conf.erb
sayalilunkad: ok so it would be alright to remove this section right?

sayalilunkad: done!

aplanas:
There are more files in "/etc/swift", like "proxy-server.conf" and other. I wonder is we want to update them too

Itxaka:
@aplanas I think that those other files (proxy, etc..) should be ignored as long as we dont have anything to override on those, which doesnt seem like we have...yet :D

sayalilunkad:
@aplanas @Itxaka as the other config files are maintained upstream I have minified those as well.

chef/cookbooks/swift/templates/default/account-server.conf.erb
```diff
@@ -1,197 +1,9 @@
...
-# db_preallocation = off
-#
-# eventlet_debug = false
-#
-# You can set fallocate_reserve to the number of bytes you'd like fallocate to
-# reserve, whether there is space for the given file size or not.
-# fallocate_reserve = 0
-
-[pipeline:main]
-pipeline = healthcheck recon account-server
```
dirkmueller: You're deleting the pipeline here

chef/cookbooks/swift/templates/default/account-server.conf.erb
```diff
@@ -1,197 +1,9 @@
...
-# replication_server = false
-
-[filter:healthcheck]
-use = egg:swift#healthcheck
-# An optional filesystem path, which if present, will cause the healthcheck
-# URL to return "503 Service Unavailable" with a body of "DISABLED BY FILE"
-# disable_path =
-
-[filter:recon]
-use = egg:swift#recon
```
dirkmueller: This needs to be remain set

chef/cookbooks/swift/templates/default/account-server.conf.erb
```diff
@@ -1,197 +1,9 @@
...
-# You can search logs for this message if space is not being reclaimed
-# after you delete account(s).
-# Default is 2592000 seconds (30 days). This is in addition to any time
-# requested by delay_reaping.
-# reap_warn_after = 2592000
-
-# Note: Put it at the beginning of the pipeline to profile all middleware. But
-# it is safer to put this after healthcheck.
-[filter:xprofile]
-use = egg:swift#xprofile
```
dirkmueller: Same

chef/cookbooks/swift/templates/default/container-server.conf.erb
```diff
@@ -1,210 +1,11 @@
...
-# db_preallocation = off
-#
-# eventlet_debug = false
-#
-# You can set fallocate_reserve to the number of bytes you'd like fallocate to
-# reserve, whether there is space for the given file size or not.
-# fallocate_reserve = 0
-
-[pipeline:main]
-pipeline = healthcheck recon container-server
```
dirkmueller: Pipeline is needed

chef/cookbooks/swift/templates/default/container-server.conf.erb
```diff
@@ -1,210 +1,11 @@
...
-# Will sync each container at most once per interval
-# interval = 300
-#
-# Maximum amount of time to spend syncing each container per pass
-# container_time = 60
-
-# Note: Put it at the beginning of the pipeline to profile all middleware. But
-# it is safer to put this after healthcheck.
-[filter:xprofile]
-use = egg:swift#xprofile
```
dirkmueller: Is needed

chef/cookbooks/swift/templates/default/object-expirer.conf.erb
```diff
@@ -1,113 +1,6 @@
...
-#  processes with process set to 0, 1, and 2
-# process = 0
-# The expirer will re-attempt expiring if the source object is not available
-# up to reclaim_age seconds before it gives up and deletes the entry in the
-# queue.
-# reclaim_age = 604800
-# recon_cache_path = /var/cache/swift
-
-[pipeline:main]
-pipeline = catch_errors proxy-logging cache proxy-server
```
dirkmueller: Is needed

chef/cookbooks/swift/templates/default/swift.conf.erb
```diff
@@ -1,120 +1,4 @@
...
-# in metadata however the ring file name will always be 'object.ring.gz' for
-# backwards compatibility.  If no policies are defined a policy with index 0
-# will be automatically created for backwards compatibility and given the name
-# Policy-0.  A default policy is used when creating new containers when no
-# policy is specified in the request.  If no other policies are defined the
-# policy with index 0 will be declared the default.  If multiple policies are
-# defined you must define a policy with index 0 and you must specify a
-# default.  It is recommended you always define a section for
-# storage-policy:0.
-[storage-policy:0]
```
dirkmueller: Is needed

sayalilunkad:
@dirkmueller I thought these are same as the default so we don't need to have them in the minified templates. But will add them again.

sayalilunkad:
@aplanas I was also under the same impression. And yes need to fix the CI problem.

aplanas:
That explain the fail in gating : ((

sayalilunkad:
@toabctl @aplanas yikes then I will move it back to /etc/swift/, thanks!

aplanas:
I am still +1

sayalilunkad:
@dirkmueller can you please merge this?

PR #770, opened 2017-01-17
swift: Fail to create swift proposal if there is no keystone

rsalevsky:
This change was not wanted by upstream in 2013, since then this patch is in our package but a lot changed since then so let's try it again.


rsalevsky:
@vuntz @dirkmueller This patch is already in our package this is the reason why I submitted it. If it is really incomplete then we should follow up here and create a regular sprint card.

dirkmueller:
@rsalevsky ok, you're right, lets get rid of the custom patch in the package and clean this up in a new pr.

PR #771, opened 2017-01-17
Cinder no v1

dirkmueller:


vuntz:
@dirkmueller not sure if you still want that?

PR #705, opened 2016-12-28
aodh: create aodh barclamp

sidthe:
this PR creates a separate barclamp for aodh

PR #773, opened 2017-01-18
nova: Add SSL support for EC2API

sayalilunkad:


chef/data_bags/crowbar/migrate/nova/113_add_ec2_ssl.rb
```diff
@@ -0,0 +1,10 @@
+def upgrade(ta, td, a, d)
+ a["ec2-api"]["ssl"]["enabled"]       = ta["ec2-api"]["ssl"]["enabled"]
+ a["ec2-api"]["ssl"]["certfile"]       = ta["ec2-api"]["ssl"]["certfile"]
+ a["ec2-api"]["ssl"]["keyfile"]        = ta["ec2-api"]["ssl"]["keyfile"]
+ a["ec2-api"]["ssl"]["generate_certs"] = ta["ec2-api"]["ssl"]["generate_certs"]
+ a["ec2-api"]["ssl"]["insecure"]       = ta["ec2-api"]["ssl"]["insecure"]
+ a["ec2-api"]["ssl"]["cert_required"]  = ta["ec2-api"]["ssl"]["cert_required"]
+ a["ec2-api"]["ssl"]["ca_certs"]       = ta["ec2-api"]["ssl"]["ca_certs"]
+ return a, d
+end
```
nicolasbock: No `downgrade`?

nicolasbock:
The gating failure is what we are currently seeing in our [Jenkins SSL test](https://ci.suse.de/view/Cloud/view/Cloud7/job/cloud-mkcloud7-job-ssl-x86_64/). I am looking into that right now, don't have a solution so far.

sayalilunkad:
@nicolasbock yes I see that, thanks for looking into it. I have saved the current build and enabled SSL for ec2-api in the nova barclamp since that need to be done manually. So if you would like to take a look at the barclamp you can find it at http://crowbar.vk4.cloud.suse.de/

nicolasbock:
@sayalilunkad Thanks! I'll have a look there.

chef/data_bags/crowbar/migrate/nova/113_add_ec2_ssl.rb
```diff
@@ -0,0 +1,21 @@
+def upgrade(ta, td, a, d)
+  a["ec2-api"]["ssl"]["enabled"] = ta["ec2-api"]["ssl"]["enabled"]
```
dirkmueller: this doesn't work, and you can make it much simpler via

```ruby

a["ec2-api"]["ssl"] = ta["ec2-api"]["ssl"]
```

crowbar_framework/config/locales/nova/en.yml
```diff
@@ -82,6 +82,17 @@ en:
             keyfile: 'SSL (Private) Key File'
         logging_header: 'Logging'
         verbose: 'Verbose Logging'
+        ec2-api_header: 'EC2API Configuration'
```
dirkmueller: whats the point of duplicating that ? nova already has a ssl configuration, we don't need to ask twice for the same info.

chef/data_bags/crowbar/migrate/nova/113_add_ec2_ssl.rb
sayalilunkad: I think this was not working as expected because of a typo in the attributes. Fixed that. Also it would make it uniform with the other setting if I kept the format ["ec2-api"]["ssl"]["enabled"]

crowbar_framework/config/locales/nova/en.yml
sayalilunkad: ok, removing the common configuration like insecure, generate_crets, cert_required..the others are required I guess since we need different certificates..

sidthe:
at a glance, a couple of questions
1) isn't that supposed to go behind apache?
2) in ha, I see ssl stays as not enabled, I guess it is supposed to be set?
3) In the newton ec2 conf file (absent from the minification we have) there is `ec2api_use_ssl` parameter that I guess might need to be set to true if ssl is enabled?

sayalilunkad:
@sidthe 
>1. isn't that supposed to go behind apache?

what do you mean exactly?
>2. in ha, I see ssl stays as not enabled, I guess it is supposed to be set?

even in non-ha it is not getting set. Will fix it, it should be set.
>3. In the newton ec2 conf file (absent from the minification we have) there is ec2api_use_ssl parameter that I guess might need to be set to true if ssl is enabled?

possibly, need to check. thanks!

sidthe:
@sayalilunkad 

> isn't that supposed to go behind apache?
>what do you mean exactly?

So you might want to check with other guys, but my understanding was there is a preference to use apache as the ssl gateway, going forward, for all services to avoid broken ssl implementations between various openstack components. 

An example is here
https://github.com/crowbar/crowbar-openstack/pull/767/files#diff-db3e16d4f0315ce5c90981613106ea47R120

or here

https://github.com/crowbar/crowbar-openstack/pull/705/files#diff-9113c9b97278a251a6dfcb8b9f8ba203R201

sayalilunkad:
@sidthe https://github.com/SUSE-Cloud/automation/pull/1634 should fix the problem to set https by default.

sidthe:
@sayalilunkad looks like there is some conflict there, can you rebase it?

chef/cookbooks/ec2-api/recipes/ec2api.rb
```diff
@@ -167,6 +167,17 @@
   action :add_endpoint_template
 end
 
+# ec2-api ssl
+if node[:nova]["ec2-api"][:ssl][:enabled]
+  ssl_setup "setting up ssl for ec2-api" do
+    generate_cert true
```
vuntz: Hard-coding to `true` seems awfully wrong.

sayalilunkad: no, I should change that. It was only for testing :)

sayalilunkad:
@sidthe rebased, and will take a look at the HA case, thanks!

sidthe:
@sayalilunkad I retriggered the build with want ssl set to 1, cause yesterday it was failing on an unrelated dns issue - and I see that so far it looks ok

chef/cookbooks/ec2-api/recipes/ec2api_ha.rb
```diff
@@ -35,7 +35,7 @@
 haproxy_loadbalancer "ec2-metadata" do
   address "0.0.0.0"
   port node[:nova][:ports][:ec2_metadata]
-  use_ssl false
+  use_ssl node[:nova]["ec2-api"][:ssl][:enabled]
   servers CrowbarPacemakerHelper.haproxy_servers_for_service(
     node, "nova", "ec2-api", "ec2_metadata"
   )
```
s-t-e-v-e-n-k: Please rebase, there's a missing ec2-s3 service here now.

sayalilunkad: done

sidthe:
so, the failure on ha side seems unrelated

sayalilunkad:
[(5) S51P11 (S50P12 S49P14) Re-add Nova EC2 API that got removed in Mitaka and added as new project](https://trello.com/c/x51hX6aZ/269-5-s51p11-s50p12-s49p14-re-add-nova-ec2-api-that-got-removed-in-mitaka-and-added-as-new-project)

chef/data_bags/crowbar/migrate/nova/113_add_ec2_ssl.rb
dirkmueller: Right, but you don't need to copy each individual entry in the hash, just assigning the hash is enough.

chef/cookbooks/ec2-api/recipes/ec2api.rb
```diff
@@ -179,6 +179,17 @@
   action :add_endpoint_template
 end
 
+# ec2-api ssl
+if node[:nova]["ec2-api"][:ssl][:enabled]
```
s-t-e-v-e-n-k: I'm little unclear how this will behave with the HA+SSL case? Is this something we want running on different nodes with different certificates generated, or only on the cluster founder?

toabctl:
> I'm still not convinced that we want separate SSL settings for ec2-api. wouldn't it be easier to just reuse the nova ssl config variables. it doesn't look nice in the barclamp view to have twice different ssl settings listed.
> 
> I do understand that ec2 and nova can be deployed on different nodes but that would still not matter..

IMO we would like to have an extra barclamp for ec2. There is no reason to deploy it with nova together. From that POV different ssl settings might make sense.

toabctl:
@sayalilunkad needs a rebase.

dirkmueller:
@toabctl  I agree  that this should have been a separate barclamp, but the way it is implemented right now it is not and i don't see a reason for uglifying the user experience just because we at some point intend to do it properly

sayalilunkad:
@dirkmueller but even if you look at novnc it has it's own certificates at least. Do you not want anything at all in the UI for ec2?

toabctl:
@dirkmueller I'm also fine with using the same SSL settings.

Itxaka:
If we re-use the nova ssl settings then we need to modify the automation repo to take that into account:
https://github.com/SUSE-Cloud/automation/blob/master/scripts/qa_crowbarsetup.sh#L2205

sayalilunkad:
Updated to use all setting from nova as discussed with @dirkmueller

sayalilunkad:
@Itxaka dirk reverted that commit.

Itxaka:
@s-t-e-v-e-n-k @dirkmueller @vuntz re-review?

s-t-e-v-e-n-k:
@dirkmueller Re-review this, please?

s-t-e-v-e-n-k:
I've just been testing this with HA and SSL enabled, and it does not work, and that is due to file permissions:

From /var/log/ec2-api/ec2-api.log:

2017-03-08 05:27:14.421 19391 ERROR oslo_service.service   File "/usr/lib64/python2.7/ssl.py", line 574, in __init__
2017-03-08 05:27:14.421 19391 ERROR oslo_service.service     self._context.load_cert_chain(certfile, keyfile)
2017-03-08 05:27:14.421 19391 ERROR oslo_service.service IOError: [Errno 13] Permission denied

The certificate is fine, everyone can read it.

root@d52-54-77-77-77-01:~ # ls -lh /etc/nova/ssl/certs/
total 4.0K
-rw-r--r-- 1 root root 1.9K Mar  7 05:28 signing_cert.pem

The private key file is not:

root@d52-54-77-77-77-01:~ # ls -lh /etc/nova/ssl/private/
total 4.0K
-rw-r----- 1 root nova 3.2K Mar  7 05:28 signing_key.pem

Only users in group nova can read it, which ec2-api isn't. The quick and dirty solution is to add the ec2-api UNIX user to the nova group, the proper solution would be to separate the SSL configs and set up ec2 with its own SSL certificates and keys.

vuntz:
@s-t-e-v-e-n-k hmm, so I'm of the opinion that using keys specific to ec2-api would be better. I'd prefer to go that way (and that'd make it easier to move ec2-api to its own barclamp eventually).

FWIW, I'm not a fan of using the nova group, but we could use the openstack group (which is right now only used for cinder and nova, you can search for that in the cookbooks). So if people want to go in that direction, that'd work for me.

sayalilunkad:
@vuntz @s-t-e-v-e-n-k In the first few iterations of this PR I had used the method suggested first that is letting ec2-api have its own SSL certificates and keys. At that time @dirkmueller suggested the existing patch as a better option since it has minimum change and that would have been better if we were to include it in the release. But since it did not make it to the release I am happy to revert the change to keep the ec2-api stuff separate from  nova and for the time being have the setting in the nova barclamp(which was also how it was previously).

s-t-e-v-e-n-k:
@sayalilunkad Yes, please revert to your earlier commit with the separate keys, and I'll test it in a full HA/SSL environment.

sayalilunkad:
and @s-t-e-v-e-n-k thanks for spotting the problem. I have reverted to the earlier version will test it soon too. Another thing to note is that @dirkmueller said that  he would prefer a separate setting for ec2-api, but with optional way of overwriting certfile/keyfile like it is done for novnc but I think that will lead us to the same permission problem again since we will be using the nova certificates by default in that case so I am not keeping it optional.

s-t-e-v-e-n-k:
The resurrected changes look good to me, except it's missing at least generate self-signed certificates, insecure and maybe the CA option like nova's SSL configuration has -- we should dig and see which of the last two are represented in ec2api.conf. NoVNC can rely on falling back to nova's SSL config since they run as the same user, we don't have that luxury for ec2-api.

vuntz:
> The resurrected changes look good to me, except it's missing at least generate self-signed certificates, insecure and maybe the CA option like nova's SSL configuration has -- we should dig and see which of the last two are represented in ec2api.conf.

We definitely want insecure.

sayalilunkad:
@s-t-e-v-e-n-k @vuntz I have removed the options like generate_cert that we shared with nova and replaced it with attributes specific to ec2-api. I have also added the ca file and insecure option but limiting what goes to the UI to avoid making the nova barclamp bigger than it already is.

sayalilunkad:
also opened this https://github.com/SUSE-Cloud/automation/pull/1744

s-t-e-v-e-n-k:
[2017-03-10T06:52:54+00:00] FATAL: NoMethodError: undefined method `ca_cert' for Chef::Resource::SslSetup 
controller1 (d52-54-77-77-77-01.vp4.cloud.suse.de): 
188: certfile node[:nova]["ec2-api"][:ssl][:certfile] 
189: keyfile node[:nova]["ec2-api"][:ssl][:keyfile] 
190: group node[:nova]["ec2-api"][:group] 
191: fqdn node[:fqdn] 
192: cert_required node[:nova]["ec2-api"][:ssl][:cert_required] 
193>> ca_cert node[:nova]["ec2-api"][:ssl][:ca_cert] 
194: end 
195: end 

It's defined as ca_certs in the template and schema, and referenced in the cookbook as ca_cert.

Furthermore, I don't really think adding them to the template to show up in the custom view would require that much more extra space, and I don't think barclamp configuration vertical space is at that much of a premium, so it would certainly be less awkward if everything was in the template.

sayalilunkad:
@s-t-e-v-e-n-k it should be ca_certs. I have fixed that bit.

sayalilunkad:
Added UI changes, but not tested yet.

sayalilunkad:
Seems to work for me. Only problem I repeatedly see is I need to restart crowbar to see the nova custom view but don't see anything suspicious in the logs so wonder where that comes from.

s-t-e-v-e-n-k:
This now requires a rebase, and shifting the migration, 115 has been claimed.

sayalilunkad:
@s-t-e-v-e-n-k done, thanks

chef/cookbooks/ec2-api/templates/default/ec2api.conf.erb
```diff
@@ -13,7 +13,16 @@ s3_listen = <%= @bind_host %>
...
 transport_url = <%= @rabbit_settings[:url] %>
 full_vpc_support = false
-
+<% if node[:nova]["ec2-api"][:ssl][:enabled] %>
+ec2api_use_ssl = true
+metadata_use_ssl = true
+ssl_cert_file = <%= node[:nova]["ec2-api"][:ssl][:certfile] %>
+ssl_key_file = <%= node[:nova]["ec2-api"][:ssl][:keyfile] %>
+<% if node[:nova]["ec2-api"][:ssl][:cert_required] %>
+  ssl_ca_file = <%= node[:nova]["ec2-api"][:ssl][:ca_certs] %>
```
dirkmueller: this doesn't work, configuration keys need to be at the beginning of a line

crowbar_framework/config/locales/nova/en.yml
dirkmueller: hmm, its possible to specify different certificates but they wouldn't be needed.. anyway, its consistent, so I don't object

sayalilunkad:
@dirkmueller just created a fix for HA failure https://github.com/crowbar/crowbar-openstack/pull/868

s-t-e-v-e-n-k:
This has been replaced with https://github.com/crowbar/crowbar-openstack/pull/872 .

PR #775, opened 2017-01-18
[nova] Assure that libvirtd is restarted before nova-compute (bsc#101…

rhafer:
…6302)

nova-compute is currently suffering from race conditions when libvirtd
is being restarted during the startup of nova-compute. (For details see:
https://bugs.launchpad.net/nova/+bug/1654207

Our recipes currently are implemented in a way that makes that race
condition trigger pretty often on the initial deploymented and
especially on the crowbar_join run during the upgrade from an older
release. One reason for this is, that the libvirtd restart happens
after/during the restart of nova-compute.

This patch adds a workaround so that libvirtd is restarted before
nova-compute, by utilizing immediate notifications.

### DIFF
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -73,7 +73,7 @@ case node[:nova][:libvirt_type]
           libvirtd_listen_addr: listen_addr,
           libvirtd_auth_tcp: node[:nova]["use_migration"] ? "none" : "sasl"
         )
-        notifies :restart, "service[libvirtd]", :delayed
+        notifies :create, "ruby_block[restart_libvirtd]", :immediately
       end
 
       case node[:nova][:libvirt_type]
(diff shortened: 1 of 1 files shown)


rhafer:
@vuntz this utilizes a similar trick to what I initially tried for https://github.com/crowbar/crowbar-core/pull/868, as I remember you didn't particularly like it I am now trying to sneak it in in another place ;)

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -184,7 +184,25 @@
...
             group: libvirt_group
         )
-        notifies :restart, "service[libvirtd]"
+        notifies :create, "ruby_block[restart_libvirtd]", :immediately
+      end
+
+      # This block is here to allow to restart libvirtd as soon as possible
+      # after configuration changes (notified from the qemu.conf and libvirtd.conf
+      # templates above), while avoiding it to restart multiple times (like it would
+      # when we'd sent an :immediate restart notification directly to the service.
```
jsuchome: `)` is missing

rhafer: Fixed

PR #776, opened 2017-01-18
glance: disable glance v1 by default

dirkmueller:


dirkmueller:
[(3) glance: Do not deploy API v1 by default](https://trello.com/c/H4WXL803/2028-3-glance-do-not-deploy-api-v1-by-default)

PR #777, opened 2017-01-19
swift: Remove slogging leftovers

dirkmueller:
This has always been disabled and certainly would need adjustments
with the newer Swift


rsalevsky:
@dirkmueller I guess we have to remove it also from: https://build.suse.de/package/view_file/Devel:Cloud:7:Staging/crowbar-openstack/swift-override.rb?expand=1

PR #778, opened 2017-01-20
[nova] Assure that libvirtd is restarted before nova-compute (bsc#101…

rhafer:
…6302)

nova-compute is currently suffering from race conditions when libvirtd
is being restarted during the startup of nova-compute. (For details see:
https://bugs.launchpad.net/nova/+bug/1654207

Our recipes currently are implemented in a way that makes that race
condition trigger pretty often on the initial deploymented and
especially on the crowbar_join run during the upgrade from an older
release. One reason for this is, that the libvirtd restart happens
after/during the restart of nova-compute.

This patch adds a workaround so that libvirtd is restarted before
nova-compute, by utilizing immediate notifications.

(cherry picked from commit 8c44da82f835c81e2377e38be5e1146a456e532d)


rhafer:
This is backporting: https://github.com/crowbar/crowbar-openstack/pull/775

PR #779, opened 2017-01-20
nova: increase sync mark timeout for compute HA

aplanas:
For vitualized installations we need to increase the timeout
of the sync-nova_compute_before_ha synchronization point.


aplanas:
Note: this was tested without https://github.com/crowbar/crowbar-core/pull/747, so this can be the wrong fix for the same issue

aspiers:
Yeah, I suspect it's the wrong fix.

PR #781, opened 2017-01-20
[stable/3.0] Tell the apache service to use pacemaker provider if needed

jsuchome:
Requires https://github.com/crowbar/crowbar-core/pull/1015

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -13,6 +13,13 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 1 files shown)


jsuchome:
Gating fails as it requires https://github.com/crowbar/crowbar-core/pull/1015

jsuchome:
[apache service not configured correctly with HA (not handled by pacemaker)](https://trello.com/c/1b5iNWF7/273-apache-service-not-configured-correctly-with-ha-not-handled-by-pacemaker)

PR #784, opened 2017-01-23
nova: fix ec2api endpoint port

dirkmueller:
Apparently noone ever tried to use ec2api so far..


sayalilunkad:
@dirkmueller what command caused the problem? Because I had tested it with some of the basic ec2-api commands and it worked..

sidthe:
I have also this change done in my HA PR for EC2 - along with other fixes (or Steves HA PR) on which I am currently stuck and I don't seem to find anyone who could help :(

dirkmueller:
@sidthe i'm trying to add test coverage before looking at your pr

PR #785, opened 2017-01-24
magnum: set auth_url properly for authtoken section

dirkmueller:


chef/cookbooks/magnum/templates/default/magnum.conf.erb
```diff
@@ -16,23 +16,29 @@ host = <%= @bind_host %>
 
 [barbican_client]
 region_name = <%= @keystone_settings['endpoint_region'] %>
+insecure = <%= @keystone_settings['insecure'] %>
```
aplanas: do we need the insecure entries too?

dirkmueller: we don't need them, I think they're not used at this point in time. just included them for consistency reasons (extra settings usually don't hurt)

chef/cookbooks/magnum/templates/default/magnum.conf.erb
```diff
@@ -16,23 +16,29 @@ host = <%= @bind_host %>
...
 
 [glance_client]
 region_name = <%= @keystone_settings['endpoint_region'] %>
 insecure = <%= @keystone_settings['insecure'] %>
+endpoint_type = internalURL
 
 [heat_client]
 region_name = <%= @keystone_settings['endpoint_region'] %>
 insecure = <%= @keystone_settings['insecure'] %>
+endpoint_type = internalURL
```
mjura: This is used by Waiting Condition to call back, and it has to stay publicURL

chef/cookbooks/magnum/templates/default/magnum.conf.erb
```diff
@@ -55,14 +62,17 @@ admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
 [magnum_client]
 region_name = <%= @keystone_settings['endpoint_region'] %>
 insecure = <%= @keystone_settings['insecure'] %>
+endpoint_type = internalURL
```
mjura: This is used by Waiting Condition to call back, and it has to stay publicURL

mjura:
What is explanation for using internalURL instead of publicURL. Magnum is deploying COE cluster and some features requires to communicate back to OpenStack services to setup additional services or for monitoring. In most of this clients has to stay publicURL. This change will break Kubernetes deployment. strong -1

PR #788, opened 2017-01-24
neutron: Add ability to configure the f5_ha_type option for F5 LBaaS + one fix

vuntz:


crowbar_framework/app/helpers/barclamp/neutron_helper.rb
```diff
@@ -73,6 +73,17 @@ def networking_lbaasv2_drivers(selected)
       )
     end
 
+    def f5_ha_type(selected)
+      options_for_select(
+        [
+          [t(".f5.ha_types.standalone"), "standalone"],
+          [t(".f5.ha_types.pair"), "pair"],
+          [t(".f5.ha_types.scalen"), "scalen"]
```
matelakat: Might be a stupid question: What does `".f5.ha_types.scalen"` reference? Is it just an arbitrary string?

vuntz: This string: https://github.com/crowbar/crowbar-openstack/pull/788/files#diff-1671f88e92d191883e65b12ee99abf44R72

matelakat: thanks!

vuntz:
@matelakat @aplanas @aspiers are you still +1? I had force-pushed to add the missing migration (that was not there due to me forgetting to do "git add").

PR #792, opened 2017-01-24
Neutron ha tool servicify

matelakat:
This is an experimental patch for turning neutron-ha-tool into a service

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -217,6 +217,30 @@ if use_l3_agent
                                                          keystone_settings["service_port"],
                                                          "2.0")
 
+  neutron_l3_ha_service_options = node[:neutron][:ha][:neutron_l3_ha_service][:service_options]
+  env_vars_for_neutron_l3_ha_service = {
+      'OS_AUTH_URL' => os_auth_url_v2,
+      'OS_REGION_NAME' => keystone_settings['endpoint_region'],
+      'OS_TENANT_NAME' => keystone_settings['admin_tenant'],
+      'OS_USERNAME' => keystone_settings['admin_user'],
+      'OS_INSECURE' => keystone_settings['insecure'] || node[:neutron][:ssl][:insecure],
+      'NHAS_SECONDS_TO_SLEEP_BETWEEN_CHECKS' => neutron_l3_ha_service_options[:seconds_to_sleep_between_checks],
+      'NHAS_AGENT_CHECK_TIMEOUT' => neutron_l3_ha_service_options[:agent_check][:timeout],
+      'NHAS_AGENT_CHECK_SHUTDOWN_TIMEOUT' => neutron_l3_ha_service_options[:agent_check][:shutdown_timeout],
+      'NHAS_REPLICATE_DHCP_TIMEOUT' => neutron_l3_ha_service_options[:replicate_dhcp][:timeout],
+      'NHAS_REPLICATE_DHCP_SHUTDOWN_TIMEOUT' => neutron_l3_ha_service_options[:replicate_dhcp][:shutdown_timeout],
+      'NHAS_AGENT_MIGRATE_TIMEOUT' => neutron_l3_ha_service_options[:agent_migrate][:timeout],
+      'NHAS_AGENT_MIGRATE_SHUTDOWN_TIMEOUT' => neutron_l3_ha_service_options[:agent_migrate][:shutdown_timeout],
+      'NHAS_HA_TOOL' => "/usr/bin/neutron-ha-tool",
+  }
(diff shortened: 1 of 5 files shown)


chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -228,7 +228,7 @@
     owner "root"
     group "root"
     mode "0600"
-    content env_vars.map {|e| e.join('=')}.join('\n')
+    content env_vars.map {|e| e.join("=")}.join("\n")
```
aspiers: This can be squashed with the previous commit :)

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -228,7 +228,7 @@
     owner "root"
     group "root"
     mode "0600"
-    content env_vars.map {|e| e.join("=")}.join("\n")
+    content env_vars.map {|k, v| [k.upcase, v].join("=")}.join("\n")
```
aspiers: Squash this too.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -228,7 +228,7 @@
     owner "root"
     group "root"
     mode "0600"
-    content env_vars.map {|k, v| [k.upcase, v].join("=")}.join("\n")
+    content env_vars.map {|k, v| [["export", k.upcase].join(" "), v].join("=")}.join("\n")
```
aspiers: and this...

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -216,21 +216,46 @@
...
                                                          "2.0")
+  env_vars = {
+      "os_auth_url"    => os_auth_url_v2,
+      "os_region_name" => keystone_settings["endpoint_region"],
+      "os_tenant_name" => keystone_settings["admin_tenant"],
+      "os_username"    => keystone_settings["admin_user"],
+      "os_insecure"    => keystone_settings["insecure"] || node[:neutron][:ssl][:insecure]
+  }
+
+  neutron_ha_service_options = node[:neutron][:ha][:neutron_ha_service][:service_options]
```
aspiers: Traditionally we haven't treated recipes with quite the same respect as the rest of our code - typically it's been considered OK to keep adding and adding code without splitting out into more manageable chunks.  But it's not OK, so maybe we can change that?  If possible there should be a separate `neutron::ha_tool_service` recipe instead of further bloating `neutron::network_agents_ha`.  There are two potential problems though:

- I'm not sure of the scoping rules for `include_recipe`, so there might be a downside that you would have to do extra work to share `env_vars` between the files.
- The recipe would probably have to return a Chef resource to be included in the transaction, and I'm not sure whether `include_recipe` supports returning values either.

If these are insurmountable then too bad, Chef sucks.  But we should at least try.

Also, `neutron_ha_service` sounds slightly too vague a name to me, since there are other neutron services which are also HA.  This is specifically L3 HA.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -257,6 +282,68 @@
     action :create
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
+
+  # TEMPORARY SECTION BEGIN
+  # the responsibility of installing the service and the systemd unit file
+  # should be moved to the package openstack-neutron-ha-tool
```
aspiers: As discussed, maybe not.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -257,6 +282,68 @@
...
+
+  # install script
```
aspiers: So there are a whole load of extra resources here.  It's worth bearing in mind that this increases the amount of work which needs to be completed before the non-founder sync marks time out.  If we're already close to the timeouts in our slow CI environment then this might push it over the edge and introduce new Jenkins failures.

Ideally the work would be split into smaller synchronisation regions, but that might be difficult.

spec/neutron_l3_ha_service_spec.rb
```diff
@@ -0,0 +1,463 @@
...
+    before(:each) do
+      @neutron_service_script = NeutronServiceScript.new
+    end
+
+    it 'exits with 1' do
+      run_result = @neutron_service_script.run
+      expect(run_result.exitstatus).to eq 1
+    end
+
+    it 'error message complains about missing rc file' do
```
aspiers: s/it/specify/ ?

matelakat:
Closing this item, we decided to implement the service in ruby. The PR can be seen here: #840

PR #793, opened 2017-01-24
nova: Expose `use_serial` to UI

nicolasbock:
Right now the `use_serial` switch is hidden in the raw view but it's
useful enough to expose it in the UI.

### DIFF
--- a/crowbar_framework/app/views/barclamp/nova/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/nova/_edit_attributes.html.haml
@@ -99,5 +99,14 @@
 
     %fieldset
       %legend
+        = t(".serial_console_header")
+
+      .alert.alert-warning
+        = t(".serial_console_hint")
+
+      = boolean_field :use_serial
+
+    %fieldset
+      %legend
         = t(".logging_header")
       = boolean_field :verbose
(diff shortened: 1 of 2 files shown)


Itxaka (dismissed):

Itxaka:
May need to open a bug for the docs to be updated methinks

nicolasbock:
@Itxaka oh, right, totally forgot about that :smile:

nicolasbock:
Filed [bsc#1021701](https://bugzilla.suse.com/show_bug.cgi?id=1021701).

dirkmueller:
I'm totally not convinced that  it should be exposed because;
* It requires novnc to be disabled
* It is completely unprotected
* Only one client can access the console at any time

Was therr any request for this being exposed?

nicolasbock:
@dirkmueller Yes, someone asked me yesterday on `#cloud` if we could enable it. I figured I would just throw this out on GitHub since it was so straightforward to do :smile:. It sounds like we rather shouldn't do it though.

aplanas (dismissed):

vuntz (changes requested):
@nicolasbock can you add a hint (we have some in the dns barclamp, if you need an example) to explain to the user the limitations, especially the part about disabling novnc?

nicolasbock:
@vuntz I have added the hints @dirkmueller provided. Is there a way to format the hints more like a list? I tried inserting html, but that didn't work.

PR #794, opened 2017-01-25
keystone/swift: Hide unneeded options from proposal screen

dirkmueller:
This patch has been in our packages since more or less forever.
As it causes conflicts in the PR testing, upstreaming that patch
is easier.


aplanas:
maybe we can think about droping the uwsgi code

PR #795, opened 2017-01-26
swift: Extend ring file parsing

jsuchome:
There might be extra strings and newlines

### DIFF
--- a/chef/cookbooks/swift/providers/ringfile.rb
+++ b/chef/cookbooks/swift/providers/ringfile.rb
@@ -114,7 +114,8 @@ def parse_dev_info_line(line, ringinfo)
   # Line looks like this:
   # id region zone  ip address:port      replication ip:port  name                              weight partitions balance meta
   #  0      1    0  192.168.125.14:6200  192.168.125.14:6200  2d4dc9923ed244dc9cac8f283ca79748  99.00           0 -100.00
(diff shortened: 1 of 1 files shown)


chef/cookbooks/swift/providers/ringfile.rb
```diff
@@ -114,7 +114,8 @@ def parse_dev_info_line(line, ringinfo)
   # Line looks like this:
   # id region zone  ip address:port      replication ip:port  name                              weight partitions balance meta
   #  0      1    0  192.168.125.14:6200  192.168.125.14:6200  2d4dc9923ed244dc9cac8f283ca79748  99.00           0 -100.00
-  line =~ /^\s+(\d+)\s+(\d+)\s+(\d+)\s+(\d+\.\d+\.\d+\.\d+)[:\s+](\d+)\s+(\d+\.\d+\.\d+\.\d+)[:\s+](\d+)\s+(\S+)\s+([0-9.]+)\s+(\d+)\s*([-0-9.]+)\s*$/
+  #  0      1    0  192.168.125.11:6002  192.168.125.11:6002  f5e3892aad224b2896c1f865a799e96a   0.00       65536  999.99   DEL \n
```
aplanas: the port is not 6202?

jsuchome: HM. That's actually residue after the upgrade from Cloud6. Maybe, if we manage to delete it before upgrade (re bug https://bugzilla.suse.com/show_bug.cgi?id=1022228) this PR is not needed.

PR #786, opened 2017-01-24
nova: automatically create the default flavors on chef run

Itxaka:

PR #797, opened 2017-01-26
magnum: Adds some post-install tasks

sayalilunkad:
This will upload the magnum k8s image to glance by default and
create the cluster template.

### DIFF
--- a/chef/cookbooks/magnum/recipes/post_install.rb
+++ b/chef/cookbooks/magnum/recipes/post_install.rb
@@ -1,0 +1,60 @@
+#
+# Copyright 2017 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Cookbook Name:: magnum
+# Recipe:: post_install
+#
+
+return unless !node[:magnum][:ha][:enabled] || CrowbarPacemakerHelper.is_cluster_founder?(node)
+
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+
+service_sles_image_name = "openstack-magnum-k8s-image"
+
+openstack_args = "--os-username #{keystone_settings["service_user"]}"
+openstack_args += " --os-auth-type password --os-identity-api-version 3"
+openstack_args += " --os-password #{keystone_settings["service_password"]}"
+openstack_args += " --os-tenant-name #{keystone_settings["service_tenant"]}"
+openstack_args += " --os-auth-url #{keystone_settings["internal_auth_url"]}"
+openstack_args += " --os-endpoint internalURL"
+openstack_args += keystone_settings["insecure"] ? "--insecure" : ""
+
+openstack_cmd = "openstack #{openstack_args}"
+
+admin_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+
+execute "create_magnum_image" do
+  command "curl http://#{admin_address}:8091/files/#{service_sles_image_name}/\
+  #{service_sles_image_name}.#{node[:kernel][:machine]}.qcow2 | \
+  #{openstack_cmd} image create  --disk-format qcow2 --container-format bare \
+  --public --property os_distro=opensuse #{service_sles_image_name}"
+  not_if "#{openstack_cmd} image list -f value -c Name | grep -q #{service_sles_image_name}"
+  action :nothing
+end
+
+execute "create_magnum_flavor" do
+  command "#{openstack_cmd} flavor create --ram 1024 --disk 10 --vcpus 1 m1.magnum"
+  not_if "#{openstack_cmd} flavor list --all | grep -q m1.magnum"
+  action :nothing
+end
+
+# This is to trigger the above resource to run :delayed, so that they run at
+# the end of the chef-client run, after the magnum services have been restarted
+# (in case of a config change)
+execute "trigger-magnum-commands" do
+  command "true"
+  notifies :run, "execute[create_magnum_image]", :delayed
+  notifies :run, "execute[create_magnum_flavor]", :delayed
+end
--- a/chef/cookbooks/magnum/recipes/role_magnum_server.rb
+++ b/chef/cookbooks/magnum/recipes/role_magnum_server.rb
@@ -21,4 +21,5 @@ if CrowbarRoleRecipe.node_state_valid_for_role?(node, "magnum", "magnum-server")
   include_recipe "magnum::api"
   include_recipe "magnum::conductor"
   include_recipe "magnum::ha"
+  include_recipe "magnum::post_install"
 end


sayalilunkad:
[(3) S54P5 upload magnum image into glance store by default](https://trello.com/c/QOkOA9XT/442-3-s54p5-upload-magnum-image-into-glance-store-by-default)

sayalilunkad:
[(3) S54P5 upload magnum image into glance store by default](https://trello.com/c/QOkOA9XT/442-3-s54p5-upload-magnum-image-into-glance-store-by-default)

sayalilunkad:
[(3) S54P5 upload magnum image into glance store by default](https://trello.com/c/QOkOA9XT/442-3-s54p5-upload-magnum-image-into-glance-store-by-default)

sayalilunkad:
[(3) S54P5 upload magnum image into glance store by default](https://trello.com/c/QOkOA9XT/442-3-s54p5-upload-magnum-image-into-glance-store-by-default)

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,55 @@
...
+create_magnum_image << " --disk-format qcow2 --container-format bare --public"
+create_magnum_image << " --property os_distro=opensuse #{service_sles_image_name}"
+Mixlib::ShellOut.new(create_magnum_image).run_command
+
+# create_magnum_template = "#{magnum_command} cluster-template-create --name k8s_template_tls_lb_off"
+# creat_magnum_template << "--image-id #{service_sles_image_name} --keypair-id default"
+# create_magnum_template << "--external-network-id floating --flavor-id m1.smaller"
+# Mixlib::ShellOut.new(create_magnum_template).run_command
+
+node.save
```
vuntz: I don't think this is needed.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,55 @@
...
+# magnum_command = "magnum --os-username #{keystone_settings["admin_user"]}"
+# magnum_command << " --os-auth-type password --os-identity-api-version 3"
+# magnum_command << " --os-password #{keystone_settings["admin_password"]}"
+# magnum_command << " --os-tenant-name #{keystone_settings["admin_tenant"]}"
+# magnum_command << " --os-auth-url #{auth_url} #{insecure}"
+
+create_magnum_image = "#{openstack_command} image create --file #{service_sles_image_path}"
+create_magnum_image << " --disk-format qcow2 --container-format bare --public"
+create_magnum_image << " --property os_distro=opensuse #{service_sles_image_name}"
+Mixlib::ShellOut.new(create_magnum_image).run_command
```
vuntz: Can you do it in a way similar to what we do for post-install of glance and neutron? That is, we define resources that only run at the end of the chef run through a notification. This allows avoiding early chef failures that could be caused by these post install bits.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,55 @@
...
+
+insecure = keystone_settings["insecure"] ? "--insecure" : ""
+auth_url = "#{keystone_settings["protocol"]}://"\
+           "#{keystone_settings["internal_url_host"]}:"\
+           "#{keystone_settings["service_port"]}/v3"
+
+openstack_command = "openstack --os-username #{keystone_settings["admin_user"]}"
+openstack_command << " --os-auth-type password --os-identity-api-version 3"
+openstack_command << " --os-password #{keystone_settings["admin_password"]}"
+openstack_command << " --os-tenant-name #{keystone_settings["admin_tenant"]}"
```
vuntz: Would it make sense to upload it in the service tenant? That's what we do for neutron, iirc.

sayalilunkad:
@vuntz thanks for the suggestion, I have made the changes accordingly.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,106 @@
+#
+# Copyright 2016 SUSE Linux GmbH
```
rsalevsky: 2017

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,108 @@
...
+  action :nothing
+end
+
+execute "create_floating_network" do
+  command "#{neutron_cmd} net-create floating --router:external #{floating_network_type}"
+  not_if "out=$(#{neutron_cmd} net-list); [ $? != 0 ] || echo ${out} | grep -q ' floating '"
+  retries 5
+  retry_delay 10
+  action :nothing
+end
```
vuntz: Isn't the floating network already created by the neutron cookbook?

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,108 @@
...
+
+execute "create_flavor" do
+  command "#{openstack_cmd} flavor create --ram 512 --disk 8 --vcpus 1 m1.smaller"
```
vuntz: Does it make sense to use a name that suggests it's useful for magnum?

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,108 @@
...
+service_sles_image_path = Dir.glob(
+  "/srv/tftpboot/files/#{service_sles_image_name}/#{service_sles_image_name}*qcow2"
+).first
+floating_network_type = ""
+if node[:neutron][:networking_plugin] == "ml2"
+  floating_network_type = "--provider:network_type flat "
+end
+
+# Install package for magnum k8s image
+package "#{service_sles_image_name}-#{node[:kernel][:machine]}"
```
mjura: You have to put it before line #23. During first run `service_sles_image_path` variable will be empty because `/srv/tftpboot/files/#{service_sles_image_name}/#{service_sles_image_name}*qcow2` doesn't exist

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,108 @@
...
+execute "create_flavor" do
+  command "#{openstack_cmd} flavor create --ram 512 --disk 8 --vcpus 1 m1.smaller"
+  not_if "#{openstack_cmd} flavor list | grep m1.smaller"
+  retries 5
+  retry_delay 10
+  action :nothing
+end
+
+execute "create_keypair" do
+  command "#{openstack_cmd} keypair create --public-key /root/.ssh/id_rsa.pub default"
```
vuntz: I'm nearly sure it's not useful for most people to use a keypair with this ssh key... I don't think we can actually have a good keypair automatically -- if we want one, the user should provide the public key, I guess.

chef/cookbooks/magnum/recipes/post_install.rb
mjura: Hmmm can we change this flavor name to `m1.magnum`, please? --ram 1024 --disk 10 --vcpus 1

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: It is actually but I thought we can check it here anyways..

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,108 @@
...
+  not_if "#{openstack_cmd} keypair list | grep default"
+  retries 5
+  retry_delay 10
+  action :nothing
+end
+
+execute "create_magnum_template" do
+  command "magnum #{magnum_args} cluster-template-create --name k8s_template_tls_lb_off \
+  --image-id #{service_sles_image_name} --keypair-id default --external-network-id \
+  floating --flavor-id m1.smaller --coe kubernetes"
```
mjura: I would like to add more options `--tls-disabled` and `--floating-ip-enabled` and `--master-flavor-id m1.magnum`

chef/cookbooks/magnum/recipes/post_install.rb
mjura: I was also confused, but since floating network is used in cluster-template. I'm fine with it

sayalilunkad: I can remove it since neutron will definitely have to be deployed successfully before magnum.

chef/cookbooks/magnum/recipes/post_install.rb
mjura: This key will be used only for example template, after Magnum will be deployed and image will be uploaded user can create their own template.

dirkmueller: I don't think its useful to create a default keypair on the controller for this..

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,96 @@
...
+  --container-format bare --public --property os_distro=opensuse #{service_sles_image_name}"
+  not_if "#{openstack_cmd} image list --f value -c Name | grep -q #{service_sles_image_name}"
+  retries 5
+  retry_delay 10
+  action :nothing
+end
+
+execute "create_flavor" do
+  command "#{openstack_cmd} flavor create --ram 1024 --disk 10 --vcpus 1 m1.magnum"
+  not_if "#{openstack_cmd} flavor list | grep m1.magnum"
```
dirkmueller: grep -q

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: I don't mind removing it.. @mjura ?

mjura: I agree

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,96 @@
...
+execute "create_keypair" do
+  command "#{openstack_cmd} keypair create --public-key /root/.ssh/id_rsa.pub default"
+  not_if "#{openstack_cmd} keypair list | grep default"
+  retries 5
+  retry_delay 10
+  action :nothing
+end
+
+execute "create_magnum_template" do
+  command "magnum #{magnum_args} cluster-template-create --name k8s_template_tls_lb_off \
```
sayalilunkad: yes we are, the other services like neutron which have post install tasks are also using the service tenant so we decided to go with it.

dirkmueller: So the image is public but the cluster template is not?

dirkmueller: hmm, but its inconsistent then to upload the magnum image into public but hide the cluster template in a project if its not the default project. 

anyway, imho we should just not create one I guess.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,87 @@
...
+package "#{service_sles_image_name}-#{node[:kernel][:machine]}"
+
+service_sles_image_path = Dir.glob(
+  "/srv/tftpboot/files/#{service_sles_image_name}/#{service_sles_image_name}*qcow2"
+).first
+
+insecure = keystone_settings["insecure"] ? "--insecure" : ""
+
+openstack_args = "--os-username #{keystone_settings["service_user"]}"
+openstack_args = "#{openstack_args} --os-auth-type password"
```
dirkmueller: might be simpler to make use of the ```+=``` operator

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,87 @@
...
+end
+
+# This is to trigger the above resource to run :delayed, so that they run at
+# the end of the chef-client run, after the magnum services have been restarted
+# (in case of a config change)
+execute "trigger-magnum-commands" do
+  command "true"
+  notifies :run, "execute[create_magnum_image]", :delayed
+  notifies :run, "execute[create_flavor]", :delayed
+  notifies :run, "execute[create_keypair]", :delayed
```
dirkmueller: this needs to be removed

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,87 @@
...
+
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+
+service_sles_image_name = "openstack-magnum-k8s-image"
+
+# Install package for magnum k8s image
+package "#{service_sles_image_name}-#{node[:kernel][:machine]}"
+
+service_sles_image_path = Dir.glob(
+  "/srv/tftpboot/files/#{service_sles_image_name}/#{service_sles_image_name}*qcow2"
```
dirkmueller: why globbing? originally the file was intended to be installed on the admin node, and you could just fetch it from http://provisioner-server:8091/files/..

sayalilunkad: So we are installing the package openstack-magnum-k8s-image which installs the image in that location so we are fetching it from there. I am not aware of the file already being present on the admin node.

chef/cookbooks/magnum/recipes/post_install.rb
vuntz: Agreed, I'd just keep the image upload and the flavor.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,87 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Cookbook Name:: magnum
+# Recipe:: post_install
+#
+
```
vuntz: Can we do some magic to only run this on one node when doing HA? Otherwise, I'd expect some races.

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: ok if everyone thinks so then I can remove the template creation. I did it as it was mentioned in the card.

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: do we want to run it at all when HA is enabled? I guess in glance we don't do the post install steps in case ha is enabled.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,66 @@
...
+  --container-format bare --public --property os_distro=opensuse #{service_sles_image_name}"
+  not_if "#{openstack_cmd} image list --f value -c Name | grep -q #{service_sles_image_name}"
+  retries 5
+  retry_delay 10
+  action :nothing
+end
+
+execute "create_flavor" do
+  command "#{openstack_cmd} flavor create --ram 1024 --disk 10 --vcpus 1 m1.magnum"
+  not_if "#{openstack_cmd} flavor list | grep -q m1.magnum"
```
Itxaka: I think you can do `system "#{openstack_cmd} flavor show m1.magnum"` directly and it will return 0 or 1 based on if it was found or not, see https://github.com/crowbar/crowbar-openstack/pull/786/files#diff-a40ddaf23692b2b50e538c6c2621b30dR71

Plus you can use `out: File::NULL` to not have the output from the openstack command pollute the chef log :)

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,66 @@
...
+openstack_args += "--os-auth-url #{keystone_settings["internal_auth_url"]}"
+openstack_args += insecure
+
+openstack_cmd = "openstack #{openstack_args}"
+
+execute "create_magnum_image" do
+  command "#{openstack_cmd} image create --file #{service_sles_image_path} --disk-format qcow2 \
+  --container-format bare --public --property os_distro=opensuse #{service_sles_image_name}"
+  not_if "#{openstack_cmd} image list --f value -c Name | grep -q #{service_sles_image_name}"
+  retries 5
```
Itxaka: Why 5 retries? Are we hitting an issue here that we require the retry? In that case, maybe a comment would be helpful

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,66 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Cookbook Name:: magnum
+# Recipe:: post_install
+#
+
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+
+service_sles_image_name = "openstack-magnum-k8s-image"
```
Itxaka: This is better stored in the cookbook attributes IMHO so if it changes in the future its easier to track.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,66 @@
...
+
+# Install package for magnum k8s image
+package "#{service_sles_image_name}-#{node[:kernel][:machine]}"
+
+service_sles_image_path = Dir.glob(
+  "/srv/tftpboot/files/#{service_sles_image_name}/#{service_sles_image_name}*qcow2"
```
Itxaka: Same in here, why not use attributes for this?

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,66 @@
...
+execute "create_magnum_image" do
+  command "#{openstack_cmd} image create --file #{service_sles_image_path} --disk-format qcow2 \
+  --container-format bare --public --property os_distro=opensuse #{service_sles_image_name}"
+  not_if "#{openstack_cmd} image list --f value -c Name | grep -q #{service_sles_image_name}"
+  retries 5
+  retry_delay 10
+  action :nothing
+end
+
+execute "create_flavor" do
```
Itxaka: nit: "create_magnum_flavor"

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: I don't think these will be needed in many  other recipes so I would not create an attribute for just using it here..

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: same for this.

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: no not really, but I think generally we give 5 retries just in case or at least that's how it is in other places..

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: Done.

chef/cookbooks/magnum/recipes/post_install.rb
mjura: @dirkmueller : How do you want install this image on admin node? Using dependencies or chef recipe?

dirkmueller: That part is already taken care of, admin node always installs the image.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,67 @@
...
+execute "create_magnum_image" do
+  command "#{openstack_cmd} image create --file #{service_sles_image_path} --disk-format qcow2 \
+  --container-format bare --public --property os_distro=opensuse #{service_sles_image_name}"
+  not_if "#{openstack_cmd} image list --f value -c Name | grep -q #{service_sles_image_name}"
+  retries 5
+  retry_delay 10
+  action :nothing
+end
+
+execute "create_mangnum_flavor" do
```
dirkmueller: Typo

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,67 @@
...
+  command "#{openstack_cmd} flavor create --ram 1024 --disk 10 --vcpus 1 m1.magnum"
+  not_if "#{openstack_cmd} flavor show m1.magnum"
+  retries 5
+  retry_delay 10
```
dirkmueller: Why retries? We should not hide errors.imho.. are we really doing that in other places already?

chef/cookbooks/magnum/recipes/post_install.rb
Itxaka: :+1:

chef/cookbooks/magnum/recipes/post_install.rb
dirkmueller: This does not work at all..this is evaluated in first phase, but package isn't there until 2nd phse. You need to move it in a ruby block.

Also, some logic about only matching the x86 image would need to be added.

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/neutron/recipes/post_install_conf.rb#L155

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: I don't see this. At which point is it installed?

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: I think you are right about moving it to a ruby block, guess I was testing where the image was already installed. But the logic for x_86 is not required since the current query returns the correct image

> irb(main):001:0> Dir.glob("/srv/tftpboot/files/openstack-magnum-k8s-image/openstack-magnum-k8s-image*qcow2")
=> ["/srv/tftpboot/files/openstack-magnum-k8s-image/openstack-magnum-k8s-image.x86_64.qcow2"]

But in any case if it is present in the admin node I will switch to using that.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,67 @@
...
+end
+
+# This is to trigger the above resource to run :delayed, so that they run at
+# the end of the chef-client run, after the magnum services have been restarted
+# (in case of a config change)
+execute "trigger-magnum-commands" do
+  command "true"
+  notifies :run, "execute[create_magnum_image]", :delayed
+  notifies :run, "execute[create_magnum_flavor]", :delayed
+  only_if { !node[:magnum][:ha][:enabled] || CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
sayalilunkad: @dirkmueller here is the check for HA.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,72 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Cookbook Name:: magnum
+# Recipe:: post_install
+#
+
+return unless !node[:magnum][:ha][:enabled] || CrowbarPacemakerHelper.is_cluster_founder?(node)
```
sayalilunkad: Moved the check for HA here.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,61 @@
...
+
+service_sles_image_name = "openstack-magnum-k8s-image"
+
+insecure = keystone_settings["insecure"] ? "--insecure" : ""
+
+openstack_args = "--os-username #{keystone_settings["service_user"]}"
+openstack_args += " --os-auth-type password --os-identity-api-version 3"
+openstack_args += " --os-password #{keystone_settings["service_password"]}"
+openstack_args += " --os-tenant-name #{keystone_settings["service_tenant"]}"
+openstack_args += " --os-auth-url #{keystone_settings["internal_auth_url"]}"
```
dirkmueller: would be nice to add --os-endpoint to point to internal url

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,61 @@
...
+openstack_args += insecure
```
dirkmueller: insecure is used only once, maybe move L26 here?

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,61 @@
...
+openstack_cmd = "openstack #{openstack_args}"
+
+admin_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+
+execute "create_magnum_image" do
+  command "curl http://#{admin_address}:8091/files/#{service_sles_image_name} \
+  #{service_sles_image_name}-#{node[:kernel][:machine]}.qcow2 | \
+  #{openstack_cmd} image create  --disk-format qcow2 --container-format bare \
+  --public --property os_distro=opensuse #{service_sles_image_name}"
+  not_if "#{openstack_cmd} image list --f value -c Name | grep -q #{service_sles_image_name}"
```
dirkmueller: grep -qx

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,61 @@
...
+openstack_args += " --os-auth-url #{keystone_settings["internal_auth_url"]}"
+openstack_args += insecure
+
+openstack_cmd = "openstack #{openstack_args}"
+
+admin_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+
+execute "create_magnum_image" do
+  command "curl http://#{admin_address}:8091/files/#{service_sles_image_name} \
+  #{service_sles_image_name}-#{node[:kernel][:machine]}.qcow2 | \
```
dirkmueller: the name is openstack-magnum-k8s-image.x86_64.qcow2, so it should be 

```
#{service_sles_image_name}.#{node[:kernel][:machine]}.qcow2
```

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,61 @@
...
+openstack_args += " --os-tenant-name #{keystone_settings["service_tenant"]}"
+openstack_args += " --os-auth-url #{keystone_settings["internal_auth_url"]}"
+openstack_args += insecure
+
+openstack_cmd = "openstack #{openstack_args}"
+
+admin_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+
+execute "create_magnum_image" do
+  command "curl http://#{admin_address}:8091/files/#{service_sles_image_name} \
```
dirkmueller: missing slash at the end

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,61 @@
...
+  #{service_sles_image_name}-#{node[:kernel][:machine]}.qcow2 | \
+  #{openstack_cmd} image create  --disk-format qcow2 --container-format bare \
+  --public --property os_distro=opensuse #{service_sles_image_name}"
+  not_if "#{openstack_cmd} image list --f value -c Name | grep -q #{service_sles_image_name}"
+  action :nothing
+end
+
+execute "create_magnum_flavor" do
+  command "#{openstack_cmd} flavor create --ram 1024 --disk 10 --vcpus 1 m1.magnum"
+  not_if "#{openstack_cmd} flavor list | grep -q m1.magnum"
```
dirkmueller: grep -qx

sayalilunkad: does not work with the -x option.
>root@d52-54-77-77-77-01:~ # openstack --os-username magnum --os-auth-type password --os-identity-api-version 3 --os-password 8Xhq9HUTUuS0 --os-tenant-name service --os-auth-url http://d52-54-77-77-77-01.c14.cloud.suse.de:5000/v3/ --os-endpoint internalURL flavor list | grep -x m1.magnum
root@d52-54-77-77-77-01:~ # openstack --os-username magnum --os-auth-type password --os-identity-api-version 3 --os-password 8Xhq9HUTUuS0 --os-tenant-name service --os-auth-url http://d52-54-77-77-77-01.c14.cloud.suse.de:5000/v3/ --os-endpoint internalURL flavor list | grep m1.magnum
| 70012c38-cb93-4d8b-b877-1c39759da22c | m1.magnum | 1024 |   10 |         0 |     1 | True      |

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,60 @@
...
+openstack_args += " --os-auth-url #{keystone_settings["internal_auth_url"]}"
+openstack_args += " --os-endpoint internalURL"
+openstack_args += keystone_settings["insecure"] ? "--insecure" : ""
+
+openstack_cmd = "openstack #{openstack_args}"
+
+admin_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+
+execute "create_magnum_image" do
+  command "curl http://#{admin_address}:8091/files/#{service_sles_image_name}/\
```
dirkmueller: admin_address needs to be the address from the provisioner-server node, not from "this" node, as you run it on the magnum server.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -0,0 +1,60 @@
...
+  #{service_sles_image_name}.#{node[:kernel][:machine]}.qcow2 | \
+  #{openstack_cmd} image create  --disk-format qcow2 --container-format bare \
+  --public --property os_distro=opensuse #{service_sles_image_name}"
+  not_if "#{openstack_cmd} image list --f value -c Name | grep -q #{service_sles_image_name}"
```
toabctl: ```--f``` seems to e wrong. Should be a single dash

toabctl:
> Hrm, and I just realized, why not use the upload mechanism that already exists in the glance cookbook (see setup.rb recipe). This would mean we only need to modify the attributes of the glance proposal.

@vuntz is that mechanism anywhere used? I don't see where ```[:glance][:images]``` is set

vuntz:
> @vuntz is that mechanism anywhere used? I don't see where [:glance][:images] is set

Not yet, and it's set in the glance barclamp. Which is not super convenient here, I will admit :-)

chef/cookbooks/magnum/recipes/post_install.rb
sayalilunkad: I think it still returns the admin node address because we specify type as admin.

aplanas:
@sayalilunkad @toabctl witch one we can close? this or https://github.com/crowbar/crowbar-openstack/pull/838

sayalilunkad:
@aplanas I didn't realize @toabctl opened another PR, so I updated this one also.

sayalilunkad:
moved to https://github.com/crowbar/crowbar-openstack/pull/838

toabctl:
[(3) S54P5 upload magnum image into glance store by default](https://trello.com/c/QOkOA9XT/442-3-s54p5-upload-magnum-image-into-glance-store-by-default)

toabctl:
[(3) S54P5 upload magnum image into glance store by default](https://trello.com/c/QOkOA9XT/442-3-s54p5-upload-magnum-image-into-glance-store-by-default)

PR #798, opened 2017-01-26
neutron: Add attribute of choosing ovsdb/of interface (bsc#1022074)

rhafer:
Neutron uses "native" by default now. Allow to switch back to the old
vsctl/ovs_ofctl based implementations.

Workaround for: https://bugzilla.suse.com/show_bug.cgi?id=1022074

### DIFF
--- a/chef/data_bags/crowbar/template-neutron.json
+++ b/chef/data_bags/crowbar/template-neutron.json
@@ -32,7 +32,9 @@
         "multicast_group": "239.1.1.1"
       },
       "ovs": {
-        "tunnel_csum": false
+        "tunnel_csum": false,
+        "of_interface": "native",
+        "ovsdb_interface": "native"
       },
       "allow_overlapping_ips": true,
       "use_syslog": false,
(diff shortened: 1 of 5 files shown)


chef/data_bags/crowbar/template-neutron.json
```diff
@@ -118,7 +120,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 105,
+      "schema-revision": 106,
```
dirkmueller: needs rebase to 107

vuntz:
Is the intent to allow the user to switch back? Or to make it easy for us to switch back?

rhafer:
> Is the intent to allow the user to switch back? Or to make it easy for us to switch back?

Both, probably.

chef/data_bags/crowbar/template-neutron.json
rhafer: done

rhafer:
FYI: https://github.com/crowbar/crowbar-core/pull/1041 is supposed to be the real fix.

I think I still would like to see this one merged as well.

PR #799, opened 2017-01-26
Do not use pacemaker groups unless needed

vuntz:
The downside of groups is that this creates ordering constraints within
a group, and in nearly all cases, these ordering constraints are not
reflecting real dependencies.

This is then causing issues when restarting an individual resource,
because all later resources in the group are also restarted, possibly
leading to temporary failures that were not wanted.

On top of that, a helper was created to significantly reduce code duplication.

This depends on https://github.com/crowbar/crowbar-ha/pull/178 and contains https://github.com/crowbar/crowbar-openstack/pull/330.

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/server_ha.rb
+++ b/chef/cookbooks/ceilometer/recipes/server_ha.rb
@@ -38,12 +38,12 @@ services.each do |service|
   primitive_name = "ceilometer-#{service}"
 
   if node[:ceilometer][:use_mongodb]
-    order_only_existing = ["rabbitmq", "cl-keystone"]
+    order_only_existing = "( rabbitmq cl-keystone )"
   else
     # we don't make the db mandatory if not mongodb; this is debatable, but
     # oslo.db is supposed to deal well with reconnections; it's less clear about
     # mongodb
-    order_only_existing = ["postgresql", "rabbitmq", "cl-keystone"]
+    order_only_existing = "( postgresql rabbitmq cl-keystone )"
   end
 
   objects = openstack_pacemaker_controller_clone_for_transaction primitive_name do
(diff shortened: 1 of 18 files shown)


vuntz:
@dirkmueller marking this for GM as it'll be harder to fix properly in an update and this is actually pretty important to get right. That being said, it's not fully tested yet and I won't have lots of time to play with it today...

Maybe @aspiers can help, but I think he's busy too.

vuntz:
Generally speaking, this seems to work well on vo4. I see a couple of failures in tempest, but they are not related to this, afaict.

chef/cookbooks/cinder/recipes/controller_ha.rb
```diff
@@ -40,58 +40,25 @@
...
-api_primitive = "cinder-api"
-pacemaker_primitive api_primitive do
-  agent node[:cinder][:ha][:api_ra]
-  op node[:cinder][:ha][:op]
-  action :update
-  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
-end
-transaction_objects << "pacemaker_primitive[#{api_primitive}]"
+services.each do |service|
+  primitive_name = "cinder-#{service}"
```
vuntz: This is not true

chef/cookbooks/ceilometer/recipes/server_ha.rb
```diff
@@ -31,74 +31,48 @@
...
-
-if node[:ceilometer][:use_mongodb]
-  pacemaker_order "o-ceilometer-mongo" do
-    score "Mandatory"
-    ordering "cl-mongodb #{clone_name}"
-    action :update
-    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  transaction_objects.push(objects)
+
+  if node[:ceilometer][:use_mongodb]
```
vuntz: I prefer not to, here. This makes the code slightly less readable in this case.

vuntz:
I cleaned up the git history a bit, so it's now ready for review (& merging if approved).

PR #800, opened 2017-01-27
Glance uid/gid

nicolasbock:
I am taking over ownership of this PR from #612 so that I get all notifications and not @matelakat.


Itxaka:
[(2) S54P7 (P2): bsc#1004514: L3: glance uid: changed after reinstallation of a cluster node](https://trello.com/c/45uLzu6M/444-2-s54p7-p2-bsc-1004514-l3-glance-uid-changed-after-reinstallation-of-a-cluster-node)

nicolasbock:
Rebased.

Itxaka:
As in the rabbitmq PR Im gonna complain and say that mixing ruby code when we have perfectly nice user/group resources is a bit non-chef-y way of doing it.

The resources will take care of modifying if it exists so Im still not sure of the reasoning behind using raw ruby for this.

Maybe @vuntz can shade some light on this? I dont see any benefit into creating our own methods and dealing with ruby code when there is perfectly fine resources for the same issue, but dont want to block this.

chef/cookbooks/glance/recipes/common.rb
```diff
@@ -22,6 +22,29 @@
...
+
+    cmdline = "usermod -u #{node[:glance][:uid]} -g #{node[:glance][:gid]} #{node[:glance][:user]}"
+    Chef::Log.info("running #{cmdline}")
+    cmd = Chef::ShellOut.new(cmdline).run_command
+    if cmd.exitstatus.nonzero?
+      Chef::Log.error("Failed to update glance uid/gid")
+    end
+
+  end
+  notifies :stop, "service[#{node[:glance][:api][:service_name]}]", :before
```
Itxaka: unfortunately this does not work on chef 10, so the workflow needs to be different :(

chef/cookbooks/glance/libraries/helpers.rb
```diff
@@ -27,4 +27,10 @@ def network_settings(node)
       }
     end
   end
+
+  def self.verify_user_and_group_ids(node)
```
Itxaka: Is this used somewhere?

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -142,6 +142,50 @@ def self.rabbitmq_settings(node, barclamp)
...
 
+  # Verify uid for user.
+  def self.check_user(node, expected_username, expected_uid)
+    node["etc"]["passwd"].each do |username, attrs|
+      if username == expected_username
+        if attrs["uid"] != expected_uid
+          message = "user #{username} has uid #{attrs["uid"]} but " \
+            "expected #{expected_uid}"
+          Chef::Log.fatal(message)
+          raise message
```
Itxaka: raising here means breaking the chef run if this is used somewhere in a recipe I think.

Sounds like a boolean return is more sane, although Im not sure if this is useful, the user/group chef resource will check this for us and change or not the UID/GUID if necessary.

JanZerebecki:
I suspect this will not work if the glance package is already installed. As then it doesn't modify the on disk uid/gids created by the package.

A related problem with the kvm gid came up in https://jira.suse.de/browse/SCRD-8014 .

PR #802, opened 2017-01-28
neutron: Reenable qos and dns extensions

dirkmueller:
Setting extension_drivers causes *only* those drivers to be loaded,
instead of a builtin default, so suddenly a lot of extensions
got disabled.

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -149,7 +149,7 @@ when "ml2"
   os_sdn_net = Barclamp::Inventory.get_network_definition(node, "os_sdn")
   mtu_value = os_sdn_net.nil? ? 1500 : os_sdn_net["mtu"].to_i
 
-  ml2_extension_drivers = ["port_security"]
+  ml2_extension_drivers = ["port_security", "qos", "dns"]
   ml2_type_drivers = node[:neutron][:ml2_type_drivers]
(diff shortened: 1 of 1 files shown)


toabctl:
@dirkmueller, afaics the extensions where loaded but not supported by any plugins. When ```extension_drivers``` is an empty list, I get:
```
2017-01-31 16:47:07.861 5278 WARNING neutron.api.extensions [req-bf1dcd45-01fd-4870-bf8b-606dbe8ad190 - - - - -] Extension dns-integration not supported by any of loaded plugins
2017-01-31 16:47:07.883 5278 WARNING neutron.api.extensions [req-bf1dcd45-01fd-4870-bf8b-606dbe8ad190 - - - - -] Extension port-security not supported by any of loaded plugins
2017-01-31 16:47:07.886 5278 WARNING neutron.api.extensions [req-bf1dcd45-01fd-4870-bf8b-606dbe8ad190 - - - - -] Extension qos not supported by any of loaded plugins
```

And with this change, I get:
```
2017-01-31 13:40:45.524 26996 INFO neutron.api.extensions [req-f9199821-8fed-4245-b3bd-ece4678e4c61 - - - - -] Loaded extension: dns-integration
2017-01-31 13:40:45.545 26996 WARNING neutron.api.extensions [req-f9199821-8fed-4245-b3bd-ece4678e4c61 - - - - -] Extension qos not supported by any of loaded plugins
```

So looks to me that dns and qos where not enabled before @bmwiedemann change so I think we should revert this commit for now.

PR #806, opened 2017-01-30
neutron: Enable QoS service plugin

dirkmueller:


dirkmueller:
@rhafer the ml2_conf change landed via https://github.com/crowbar/crowbar-openstack/pull/802

I manually tested this in tempestful runs, and at least it doesn't immediately fail anymore (unlike now). I haven't tested it actually.

rhafer:
> I manually tested this in tempestful runs, and at least it doesn't immediately fail anymore (unlike now). See https://ci.suse.de/job/openstack-mkcloud/51146/consoleFull (search for test_qos)

@dirkmueller Hm, so we're doing the to be able to successfully skip the tests? :) Don't get me wrong. In the end I am fine with merging this, if it really works. 

It seems the previous failure (https://ci.suse.de/job/openstack-mkcloud/51221/consoleFull) is actually a bug in the testcase? This should be the fix: https://review.openstack.org/#/c/409574/

toabctl:
> This should be the fix: https://review.openstack.org/#/c/409574/

I'll test the patch and and it to the pkg if that works. Thanks @rhafer !

toabctl:
The patch works (the qos tests are now skipped) so I would be in favor to revert https://github.com/crowbar/crowbar-openstack/pull/802 and close this PR for the GMC

dirkmueller:
@toabctl  it was nevee tagged for GM anyway

dirkmueller:
not needed anymore

PR #807, opened 2017-01-30
nova: use `live_migration_inbound_addr`

sjamgade:
Incase when using network other than admin, the said variable should
have the fqdn/ip of the migration interface.

`live_migration_uri` is no longer used for migration.


sjamgade:
[(5) Check Live-Migration in a different interface solution for Newton in the current crowbar (related with bsc#990664)](https://trello.com/c/vfeddutc/399-5-check-live-migration-in-a-different-interface-solution-for-newton-in-the-current-crowbar-related-with-bsc-990664)

PR #808, opened 2017-01-30
cinder, nova: Use a tmpfiles.d config file for /var/run/openstack

vuntz:
Without this, we rely on chef-client for this directory to be created on
boot, which is undesired.

https://bugzilla.suse.com/show_bug.cgi?id=1021753

### DIFF
--- a/chef/cookbooks/cinder/recipes/common.rb
+++ b/chef/cookbooks/cinder/recipes/common.rb
@@ -113,11 +113,7 @@ if need_shared_lock_path
     members "cinder"
     append true
   end
-  directory "/var/run/openstack" do
-    owner "root"
-    group "openstack"
-    mode "0775"
-  end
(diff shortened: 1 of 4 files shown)


dirkmueller:
Is this something that shoukd be in packaging?

vuntz:
> Is this something that should be done in packaging?

I think @sidthe discussed that with @toabctl for the initial pull request, and apparently it was decided to have it in the cookbook. It's really only required because of the fact that we change the lock path in some case in the cookbook.

I don't mind either way, though.

dirkmueller:
[bsc#1021753 - systemd openstack-nova-api service won't start; permission denied /var/run/openstack/nova-iptables](https://trello.com/c/HvhDppk4/2230-bsc-1021753-systemd-openstack-nova-api-service-won-t-start-permission-denied-var-run-openstack-nova-iptables)

PR #810, opened 2017-01-31
Revert "Use SSL to talk to nova-metadata when SSL is enabled for nova"

dirkmueller:
The settings here specify the SSL settings used for the metadata proxy,
which causes in a ssl environment guests to require metadata to
be served via https://169.254.169.264:80/, which is highly unusual.
Go back to http.

This reverts commit 4a55e9117b65a9fabbb01f2bbe6a1d2d4a9cae46.


toabctl:
cc @vuntz

dirkmueller:
yeah, it surprised to me as well, and it worked in local testing but it doesn't work. so there is something else.. (I tried a couple of changes, so the problem is elsewhere). what I can confirm is that always disabling the SSL for nova-metadata api seems to work, but I guess we don't want that

dirkmueller:
fixed via https://review.openstack.org/#/c/427943/

PR #811, opened 2017-01-31
neutron: Reenable qos and dns extensions

bmwiedemann:
Setting extension_drivers causes *only* those drivers to be loaded,
instead of a builtin default, so suddenly a lot of extensions
got disabled.

(cherry picked from commit 4316cf735f1e3785a12511ff1dbb5c0b8c570c67)
backport of #802


bmwiedemann:
so maybe that means, the qos plugin was not active before and we should rather drop it from this list to not alter the previous behaviour?

bmwiedemann:
/usr/lib/python2.7/site-packages/neutron/plugins/ml2/config.py has
```python
    cfg.ListOpt('extension_drivers',
                default=[],
```

bmwiedemann:
closing because probably not needed for Cloud6

PR #812, opened 2017-01-31
keystone: Create dummy HA resource to allow ordering constraints

vuntz:
We want other openstack services to start after keystone. However, with
the move to a wsgi app behind apache, we didn't have any resource
specific to keystone in pacemaker, and blindly ordering things after
apache2 doesn't look like a good idea (as it could be hosting just
horizon, for instance).

So we create a dummy resource for keystone that does nothing, but that
can be used for the constraints.


vuntz:
> looks like it doesn't quite work (see jenkins failure)

Was submitted too fast, and the code it depends upon was not in the package yet. Re-running testbuild; that should fix it.

PR #814, opened 2017-01-31
Revert "neutron: Add a HA order constraint between L2 agent and DHCP agent"

dirkmueller:
Potentially this does not work (as seen in jenkins). investigating..

Reverts crowbar/crowbar-openstack#813


dirkmueller:
looks like these were caused by the missing trackupstream run

PR #815, opened 2017-02-01
ec2-api: Change listen ports if using HA

s-t-e-v-e-n-k:
Since nothing was changing the listening ports for the ec2-api service,
when deploying in a HA environment, either haproxy would grab the port,
and then the service wouldn't start; or the service would, and then
haproxy would be sad. Change the listening ports if HA is enabled.

### DIFF
--- a/chef/cookbooks/ec2-api/attributes/default.rb
+++ b/chef/cookbooks/ec2-api/attributes/default.rb
@@ -21,6 +21,5 @@ default[:nova]["ec2-api"][:db][:password] = nil # must be set by wrapper
 default[:nova]["ec2-api"][:user] = "ec2-api"
 default[:nova]["ec2-api"][:group] = "ec2-api"
 default[:nova]["ec2-api"][:ha][:enabled] = false
-default[:nova]["ec2-api"][:api][:port] = 8788
 
 default[:nova]["ec2-api"][:config_file] = "/etc/ec2api/ec2api.conf.d/100-ec2api.conf"
--- a/chef/cookbooks/ec2-api/recipes/ec2api.rb
+++ b/chef/cookbooks/ec2-api/recipes/ec2api.rb
@@ -22,12 +22,24 @@ package "openstack-ec2-api-metadata"
 package "openstack-ec2-api-s3"
 
 # NOTE: ec2 is deployed via the nova barclamp
-ha_enabled  = node[:nova][:ha][:enabled]
+ha_enabled  = node[:nova]["ec2-api"][:ha][:enabled]
+# SSL support is for the entire barclamp
 ssl_enabled = node[:nova][:ssl][:enabled]
 db_settings = fetch_database_settings "nova"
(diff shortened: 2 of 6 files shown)


chef/cookbooks/ec2-api/recipes/ec2api.rb
```diff
@@ -22,12 +22,24 @@
...
+  bind_host = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+  bind_port_ec2api = node[:nova][:ha][:ports][:ec2_api]
+  bind_port_metadata = node[:nova][:ha][:ports][:ec2_metadata]
+  bind_port_s3 = node[:nova][:ha][:ports][:ec2_s3]
+else
+  bind_host = "0.0.0.0"
+  bind_port_ec2api = node[:nova][:ports][:ec2_api]
+  bind_port_metadata = node[:nova][:ports][:ec2_metadata]
+  bind_port_s3 = node[:nova][:ports][:ec2_s3]
+end["ec2-api"]
```
Itxaka: This ` ["ec2-api"]` seems to be wrong.

crowbar_framework/app/models/nova_service.rb
```diff
@@ -272,6 +272,13 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
     dirty = prepare_role_for_ha_with_haproxy(role, ["nova", "ha", "enabled"], ha_enabled, controller_elements, vip_networks)
     role.save if dirty
 
+    ec2_controller_elements, ec2_controller_nodes, ec2_ha_enabled = role_expand_elements(role, "ec2-api")
+    reset_sync_marks_on_clusters_founders(ec2_controller_elements)
+    Openstack::HA.set_controller_role(ec2_controller_nodes) if ec2_ha_enabled
```
vuntz: We usually add an empty line after this one.

chef/cookbooks/ec2-api/attributes/default.rb
```diff
@@ -21,6 +21,5 @@
 default[:nova]["ec2-api"][:user] = "ec2-api"
 default[:nova]["ec2-api"][:group] = "ec2-api"
 default[:nova]["ec2-api"][:ha][:enabled] = false
-default[:nova]["ec2-api"][:api][:port] = 8788
```
vuntz: Strictly speaking, I'd actually move the attributes out of nova/attributes/default.rb to here instead of removing things here.

s-t-e-v-e-n-k: We can't. The pacemaker code in crowbar-ha calls name.to_sym so they have to be in a top level, and pretty much everything else is under :nova already.

crowbar_framework/app/models/nova_service.rb
dirkmueller: done

dirkmueller:
@Itxaka  / @vuntz  please re-review

PR #817, opened 2017-02-01
keystone: Don't get in the way when having multiple regions (bsc#1012108)

bhdn:
These changes allow administrators to set up manually multiple keystone endpoints for different regions. Otherwise crowbar would replace any endpoint of the same service with the one used in the current deployment, defeating the approach for a multi region deployment.

### DIFF
--- a/chef/cookbooks/keystone/providers/register.rb
+++ b/chef/cookbooks/keystone/providers/register.rb
@@ -280,7 +280,7 @@ action :add_endpoint do
     if endpoints.empty?
       _create_item(http, headers, path, body, name)
       endpoint_updated = true
-    elsif endpoint_needs_update interface, endpoints, new_resource
+    elsif _not_keystone(new_resource) && endpoint_needs_update(interface, endpoints, new_resource)
       path = "#{path}/#{endpoints[interface]["id"]}"
       endpoint_updated = _update_item(http, headers, path, body, name)
     end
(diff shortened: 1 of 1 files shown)


bhdn:
Should I do something about these indentation complaints from Hound?

dirkmueller:
@bhdn yes please, would be nice to keep this consistently indented

bhdn:
Do you have any reservations about the behavior of this patch? I would like to prepare a PTF out of it.

bhdn:
[P2: bsc#1012108: L3: add multiregion support for keystone](https://trello.com/c/v5iPUm3l/196-p2-bsc-1012108-l3-add-multiregion-support-for-keystone)

bhdn:
[P2: bsc#1012108: L3: add multiregion support for keystone](https://trello.com/c/v5iPUm3l/196-p2-bsc-1012108-l3-add-multiregion-support-for-keystone)

toabctl:
triggered CI

cmurphy:
This bug is reported for SOC6. I can't reproduce the issue on SOC7, I think because we now use keystone-manage bootstrap and only trigger that once. Should this just be proposed to stable/3.0?

bhdn:
Ok, I will create a new PR for that.

vuntz:
> This bug is reported for SOC6. I can't reproduce the issue on SOC7, I think because we now use keystone-manage bootstrap and only trigger that once. Should this just be proposed to stable/3.0?

We actually still need that for stable/4.0 & master: this is a patch that enables people to do multiregion in case they know what they are doing.

rsalevsky:
@bhdn Please rebase.

dirkmueller:
@bhdn needs rebase ..

rsalevsky:
@cmurphy Could you maybe help rebasing this PR as you worked recently on the conflicting code?

cmurphy:
@rsalevsky sure

bhdn:
Please go ahead. The code has been refactored and I won't have the time to look at it now.

cmurphy:
Done

rsalevsky:
@cmurphy The HA gating failure looks related.

cmurphy:
@vuntz do you remember the rationale behind f2e60631bdeba01bcc0be3367fe3cb7e3c267504? The address defaults to "0.0.0.0" and it is clearly not designed to handle the HA case, where my_admin_host uses a helper to get either the cluster domain name or node fqdn depending on the case.

dirkmueller:
@cmurphy I bet this is supposed to enable a 2nd crowbar installation register "its" users against the first crowbar installation's keystone (so rather than using the 2nds crowbar internal keystone have it update stuff on the primary keystone that is actually managed by teh first crowbar). 

It does look like the patch is wrong / incomplete, the changes need to be moved inside the wrapper. maybe remove this 2nd patch and we rework it in a different PR? the first one should be mergeable as it is.

cmurphy:
@dirkmueller done, and also saved the second commit in https://github.com/cmurphy/crowbar-openstack/tree/pr_817_backup

PR #818, opened 2017-02-01
nova: fix ordering reference to cl-neutron-l3-agent

aplanas:


aplanas:
duplicated in  https://github.com/crowbar/crowbar-openstack/pull/819

PR #819, opened 2017-02-01
Fix ordering constraint for ha-compute

dirkmueller:
This gives

Constraint 'o-nova-evacuate': Invalid reference to 'neutron-l3-agent'

otherwise


dirkmueller:
(not tested in a clean env)

aplanas:
tested

PR #822, opened 2017-02-02
keystone: Start fernet-rotate HA resource

toabctl:
Otherwise there is no symlink to execute the cronjob and
the tokens are not rotated.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -443,7 +443,7 @@ if node[:keystone][:signing][:token_format] == "fernet"
       "backup_suffix" => ".orig"
     })
     op node[:keystone][:ha][:fernet][:op]
-    action :create
+    action [:create, :start]
     only_if { ha_enabled && CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
 end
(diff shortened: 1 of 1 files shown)


toabctl:
I tried fernet (switched from UUID to fernet) and the pacemaker resource was not started:

```
# crm status|grep fernet
keystone-fernet-rotate (ocf::heartbeat:symlink):       Stopped (disabled)
```

This PR fixes that.

toabctl:
> I am really not sure. I was under the impression that is pacemaker who start it

@aplanas It is pacemaker. This is in the pacemaker_primitive block.

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -443,7 +443,7 @@
       "backup_suffix" => ".orig"
     })
     op node[:keystone][:ha][:fernet][:op]
-    action :create
+    action [:create, :start]
     only_if { ha_enabled && CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
vuntz: We're missing sync marks around the creation of this pacemaker primitive. This means it could potentially fail on some node on creation. Can you also add that?

toabctl: @vuntz done

PR #824, opened 2017-02-02
kesytone: Add fernet ha improvements for propagating keys

mjura:


mjura:
This change is still needed

dirkmueller:
it has no description, no link to a bugreport and no rationale.. what is it supposed to do?

sayalilunkad:
typo in the commit message

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -405,22 +405,27 @@
...
 
-    # We would like to propagate fernet keys to all nodes in the cluster
-    execute "propagate fernet keys to all nodes in the cluster" do
-      command rsync_command
-      ignore_failure true
-      action :run
-      only_if { ha_enabled && CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  # We would like to propagate fernet keys to all nodes in the cluster
+  execute "propagate_fernet_keys" do
+    command "#{rsync_command} touch /var/lib/keystone/.fernet-sync.lock"
```
dirkmueller: you're assuming that the fernet-token master is the founder, which might not be the case. in the case where it isn't you're changing the code to effectively always sync keys from the cluster founder to other nodes *and* the fernet-token master will sync the other way. thats not a good idea.

mjura:
The problem is that, there is race condition and during applying Keystone proposal first time in HA mode. 
```
cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node, "keystone-server")
```
will return empty value.

This change probably fix already this bug 
 https://bugzilla.suse.com/show_bug.cgi?id=1023522

dirkmueller:
@mjura right, but isn't that just hiding the error then? we're having then the issue tha twe have setup keystone on two nodes, and one of them has no keys som 50% of all requests will fail..

dirkmueller:
@mjura in order for this to become a fix, you need to delay the creation of the pacemaker cronjob resource then also behind the (only if File.exist?(..llockfile) to make sure that no other node is accidentally trashing our fernet root keys

vuntz:
> The problem is that, there is race condition and during applying Keystone proposal first time in HA mode.
> 
> cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node, "keystone-server")
> will return empty value.

Really, just use `cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node)` in that case.

(I'm surprised that there's a race here, however)

dirkmueller:
I had the same idea just a few min ago.. see https://github.com/crowbar/crowbar-openstack/pull/837 for that fix

mjura:
Ok, I think that @dirkmueller figured out how to this better, let's continue in https://github.com/crowbar/crowbar-openstack/pull/837

PR #825, opened 2017-02-02
tempest: Add storage_protocol option for Manila

toabctl:
storage_protocol needs to be set to "CEPHFS" when only cephfs is available
as manila backend.
Otherwise tests like "test_get_share_with_share_type" fail with:

Failed to schedule create_share: No valid host was found.

### DIFF
--- a/chef/data_bags/crowbar/template-tempest.json
+++ b/chef/data_bags/crowbar/template-tempest.json
@@ -24,7 +24,8 @@
         "enable_ip_rules_for_protocols": "nfs, cifs",
         "run_consistency_group_tests": false,
         "run_snapshot_tests": true,
-        "enable_protocols": "nfs, cifs"
+        "enable_protocols": "nfs, cifs",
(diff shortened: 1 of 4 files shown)


chef/data_bags/crowbar/template-tempest.json
```diff
@@ -24,7 +24,8 @@
         "enable_ip_rules_for_protocols": "nfs, cifs",
         "run_consistency_group_tests": false,
         "run_snapshot_tests": true,
-        "enable_protocols": "nfs, cifs"
+        "enable_protocols": "nfs, cifs",
+        "storage_protocol": "NFS_CIFS"
```
Itxaka: was the underscore meant to be a comma?

Itxaka: Doesnt seem to, where does this default value comes from?

toabctl: @Itxaka underscore is correct. The value comes from Manila

PR #827, opened 2017-02-02
Make cinder volume HA work again

dirkmueller:
Unfortunately as part of the Hyper Agile Super config rediffing effort the feature was accidentally disabled. Revert the incorrect change and switch to a more robust implementation that does not suffer from deprecation warnings .


dirkmueller:
Ah, so that script.is broken already.. when we cherry pick only the first commit then that would make it work agsin. 

Can you adjust the script? It doesn't look bery robust

dirkmueller:
ok, I moved the breaking cleanup change into a separate pull request, #828 so this one should be good to go. it will actually fix the volume migration script issue that @s-t-e-v-e-n-k  pointed out as well

PR #828, opened 2017-02-03
Cinder volume ha rework

dirkmueller:


chef/cookbooks/cinder/recipes/common.rb
```diff
@@ -116,12 +116,21 @@
   include_recipe "crowbar-openstack::common"
 end
 
+# set a backend_host override when volume-ha is enabled
+if CrowbarPacemakerHelper.cluster_enabled?(node) &&
+    node[:cinder][:elements]["cinder-volume"].include?(
+      "cluster:#{CrowbarPacemakerHelper.cluster_name(node)}"
+    )
+  backend_host = CrowbarPacemakerHelper.cluster_vhostname(node)
+end
```
vuntz: I'd prefer to explicitly have `backend_host` set to `nil`. We can just rewrite:

```ruby
backend_host = if CrowbarPacemakerHelper.cluster_enabled?(node) &&
    node[:cinder][:elements]["cinder-volume"].include?(
        "cluster:#{CrowbarPacemakerHelper.cluster_name(node)}"
    )
  CrowbarPacemakerHelper.cluster_vhostname(node)
end

File: chef/cookbooks/cinder/attributes/default.rb [hunk 2]
```diff
@@ -17,7 +17,6 @@
 default[:cinder][:verbose] = false
 default[:cinder][:max_header_line] = 16384
 default[:cinder][:config_file] = "/etc/cinder/cinder.conf.d/100-cinder.conf"
-default[:cinder][:config_file_cinder_volume] = "/etc/cinder/cinder-volume.conf.d/100-cinder-volume.conf"
```
vuntz: Shouldn't we remove this file from the filesystem, for people who had deployed things before this patch?

aspiers:
Should this be closed or revived?

PR #830, opened 2017-02-03
crowbar_openstack_wsgi: reload apache configuration

aplanas:
Instead of restarting the apache service when a new vhost is
created, try to reload the configuration.

This can resolve issues in HA when apache is restarted, but a
barclamp in a different node is making queries to a service that
is behing apache (like keystone)

### DIFF
--- a/chef/cookbooks/crowbar-openstack/providers/wsgi.rb
+++ b/chef/cookbooks/crowbar-openstack/providers/wsgi.rb
@@ -70,9 +70,7 @@ action :create do
         error_log: current_resource.error_log,
         apache_log_dir: node[:apache][:log_dir],
       )
-      # NOTE(aplanas) a :reload is not enough for apache, we need a
-      # full restart to read the new vhost files
-      notifies :restart, resources(service: "apache2")
+      notifies :reload, resources(service: "apache2"), :delayed
     end
     Chef::Log.info "#{@new_resource} created / updated"
   end
@@ -84,9 +82,7 @@ action :delete do
       file _get_vhost_name do
         action :delete
         only_if { File.exist?(_get_vhost_name) }
-        # NOTE(aplanas) a :reload is not enough for apache, we need a
-        # full restart to read the new vhost files
-        notifies :restart, resources(service: "apache2")
+        notifies :reload, resources(service: "apache2"), :delayed
       end
       Chef::Log.info "#{@new_resource} deleted"
     end


aplanas:
For now I submit this to test it against Jenkins

chef/cookbooks/crowbar-openstack/providers/wsgi.rb
```diff
@@ -72,7 +72,7 @@ def whyrun_supported?
       )
       # NOTE(aplanas) a :reload is not enough for apache, we need a
       # full restart to read the new vhost files
-      notifies :restart, resources(service: "apache2")
+      notifies :reload, resources(service: "apache2"), :delayed
```
dirkmueller: see the note above from this aplanas guy, he might have a point ;)

aplanas:
yes sure, this is why I didn't remove the comment, but lets see how the scenario fails badly

aplanas:
Is surprisingly green. Trying ha_compute scenario here

dirkmueller:
yeah, maybe we should consider this, as the restart is definitely disruptive..

aplanas:
Looks like that apache is restarted in two different places:

* Keystone: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/keystone/recipes/server.rb#L245
* Horizon: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/horizon/recipes/server.rb#L369

I think that the Horizon one makes sense

vuntz:
@aplanas are we sure that a reload will make the wsgi app take into consideration the changed config files? I would actually expect it's not enough.

aplanas:
@vuntz I was totally under the same impression, also I am quite sure that this was the case. But I can't deny that it is working properly.

I did two kind of tests in `ve1`:

* Change manually the DB configuration in 100-keystone.conf and reload apache. 
* Change via crowbar the configuration of the database connection

In both cases the connection was lost and keystone was unable to access the database, indicating that reload worked. Also I was able to recover the connection fixing the configuration file and doing a reload again.

Looks like that around Kilo this was integrated:

https://bugs.launchpad.net/oslo-incubator/+bug/1276694

IIUC this takes care also of managing the SIGHUP signal for WSGI.

So for now I will remove the WIP label

vuntz:
@aplanas I'll be a pain, but can you double-check that it works for all services that we deploy as a wsgi app? I'm not sure if they're all using oslo.service.

Also, reload doesn't send SIGHUP, but SIGUSR1: https://httpd.apache.org/docs/2.2/en/stopping.html#graceful

vuntz:
> @aplanas I'll be a pain, but can you double-check that it works for all services that we deploy as a wsgi app? I'm not sure if they're all using oslo.service.

And the reason for being a pain is that we won't have lots of time for QA here.

aplanas:
@vuntz sure! Testing only keystone is not a good idea. I will try to add some test script in qa_crowbarsetup.sh to test this too.

aplanas:
Manually tested:

* keystone
* aodh
* ceilometer
* horizon

I can't test barbican because bsc#1024554. The PR for automation is almost ready

aplanas:
The test PR is here: https://github.com/SUSE-Cloud/automation/pull/1685

PR #832, opened 2017-02-03
nova: Fix creation of trusted flavors (bsc#1023417)

dirkmueller:


toabctl:
The fix is already included in https://github.com/crowbar/crowbar-openstack/pull/832 . Should we merge this one first and then rebase PR#832 ?

dirkmueller:
not needed anymore

PR #833, opened 2017-02-03
trove: add sync_marks for database creation step (bsc#1023426)

dirkmueller:


dirkmueller:
HA not supported here

dirkmueller:
Ok, this reallt needs more work

PR #834, opened 2017-02-06
nova: remove now unused trusted_flavors recipe

dirkmueller:
This was merged into the flavors.rb recipe


Itxaka:
Seems like the branch was ok but something is still including the trusted_flavor recipe, no idea what as I cant find it. May come from the automation repo?

dirkmueller:
@Itxaka can you re-review? gating passed..

PR #835, opened 2017-02-06
nova: change the not_if check for flavor creation

Itxaka:
Seems that there may be corner cases in which the not_if
will not work properly and try to retrigger the flavor creation

### DIFF
--- a/chef/cookbooks/nova/recipes/flavors.rb
+++ b/chef/cookbooks/nova/recipes/flavors.rb
@@ -96,12 +96,13 @@ default_flavors = flavors.select{ |key, value| !value["name"].match(/\.trusted\.
 if node[:nova][:trusted_flavors]
   trusted_flavors.keys.each do |id|
     execute "register_#{flavors[id]["name"]}_flavor" do
+      retries 5
       command <<-EOF
   #{novacmd} flavor-create #{flavors[id]["name"]} #{id} #{flavors[id]["mem"]} \
   #{flavors[id]["disk"]} #{flavors[id]["vcpu"]}
   #{novacmd} flavor-key #{flavors[id]["name"]} set trust:trusted_host=trusted
   EOF
-      not_if { system("#{novacmd} flavor-show #{flavors[id]["name"]}", out: File::NULL) }
+      not_if "#{novacmd} flavor-list | grep -q #{flavors[id]["name"]}"
       action :nothing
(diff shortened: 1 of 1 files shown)


chef/cookbooks/nova/recipes/flavors.rb
```diff
@@ -101,7 +101,7 @@
   #{flavors[id]["disk"]} #{flavors[id]["vcpu"]}
   #{novacmd} flavor-key #{flavors[id]["name"]} set trust:trusted_host=trusted
   EOF
-      not_if { system("#{novacmd} flavor-show #{flavors[id]["name"]}", out: File::NULL) }
+      not_if "#{novacmd} flavor-list| grep -q #{flavors[id]["name"]}"
```
dirkmueller: I'd prefer if we'd use flavor-list --all and then grep for " #{} " to make sure that we don't find substring matches..

and maybe we only do that once instead of for each flavor? otherwise every chef run will do 10 lists which doesn't seem to be the most performant way..

sayalilunkad: also missing space before pipe

Itxaka:
HA failing :(

sayalilunkad:
@Itxaka but testing the not_if command manually on that machine works :/

Itxaka:
`01:11:41.622 STDERR: ERROR (Conflict): Flavor with ID 5 already exists. (HTTP 409) (Request-ID: req-427e6ed2-60d1-4007-9a38-975a99da8211)`

not_if should have jumped so either something created the flavor (what) or the not_if was checked before the resource execution.

Itxaka:
Somehow, the ceilometer barclamp is setting up one of the pacemaker nodes into maintenance mode, which causes a failure in the nova-api, which turns unavailable during a few seconds.

If during those seconds, the flavor check is run, then it gets an unexpected 503, which it understands as the flavor not being there, so it tries to create it thus failing to create as the Id is in use.

Now, Im not sure, is that supposed to happen??? Should the services suddenly turn to unmanaged and start restarting/migrating and/or be unavailable for a few seconds?

Feels like it trashes the concept of the HA to suddenly have no service availability.

dirkmueller:
@Itxaka is there a sync timeout on the ceilometer proposal? can you grep for ========= in /var/log/crowbar/chef-client/ ?

PR #836, opened 2017-02-06
[TEST] nova: use lazy for not_if check on flavor creation

Itxaka:
I have a feeling that the delay of the flavor creation
is not syncyng with the check on the not_if code block.
Use lazy to delay the not_if evaluation until the
execution phase.

* Im not even sure if chef allows using lazy on the not_if part, it should but who knows. test PR!


Itxaka:
doesnt work :(

PR #837, opened 2017-02-06
keystone: fernet token sanity fixes

dirkmueller:
Move the pacemaker sync mark more up and make sure the cluster_nodes are only queried on 2nd phase, as on 1st phase it might be empty (???). also add a sanity check to see if it actually happens.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -440,6 +440,12 @@ if node[:keystone][:signing][:token_format] == "fernet"
     )
   end
 
+  unless ha_enabled
+    link "/etc/cron.hourly/openstack-keystone-fernet" do
+      to "/var/lib/keystone/keystone-fernet-rotate"
+    end
+  end
+
   crowbar_pacemaker_sync_mark "wait-keystone_fernet_rotate" if ha_enabled
 
   unless File.exist?("/etc/keystone/fernet-keys/0")
--- a/chef/data_bags/crowbar/template-keystone.json
+++ b/chef/data_bags/crowbar/template-keystone.json
@@ -58,7 +58,7 @@
         "driver": "sql"
       },
       "signing": {
-        "token_format": "uuid",
+        "token_format": "fernet",
         "certfile": "/etc/keystone/ssl/certs/signing_cert.pem",
         "keyfile": "/etc/keystone/ssl/private/signing_key.pem",
         "ca_certs": "/etc/keystone/ssl/certs/ca.pem"


chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -434,7 +438,6 @@
     )
   end
 
-  crowbar_pacemaker_sync_mark "wait-keystone_fernet_rotate"
```
vuntz: I'm not sure yet if this should be moved earlier; I however know fore sure that we need here a "sync-keystone_fernet_rotate" sync mark (this can be added here only if the wait doesn't move, as sync must always be before wait)

dirkmueller: I am not sure what we gain by a sync-. we need the founder to do some one time setup action, thats why there is a wait-/create- combination. could you elaborate?

dirkmueller:
@vuntz once we created the pacemaker resource we should not do anything in the chef cookbooks anymore, otherwise there is a race. the issue with waiting for that is that we need the fernet keys immediately aftewards, so there is no point in continuing without them. 

Besides that, the order is what you mentioned:

- install rsync
- setup fernet keys if founder
- sync fernet keys to all other nodes in the cluster
- create template
- wait mark
- create pacemaker resources for rotation
- create mark

That way the founder node will initially distribute keys, and all non-founder will have a the cron template that potentially rotates the key at some later point.

dirkmueller:
@vuntz hah, closer look at the logfiles indicate that there was a sync missing before installing rsync, so the initial keystone token sync failed with "rsync: command not found".

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -17,6 +17,10 @@
   package_name "openstack-keystone" if %w(rhel suse).include?(node[:platform_family])
 end
 
+if node[:keystone][:signing][:token_format] == "fernet"
```
Itxaka: Im not sure why moving the package install to the top? I cannot see rsync used before line 406

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -392,19 +396,32 @@
...
   if ha_enabled
-    cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node, "keystone-server")
+    cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node)
     cluster_nodes.map do |n|
       next if node.name == n.name
       node_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address
-      rsync_command += "rsync -a --delete-after /etc/keystone/fernet-keys " \
-                                "#{node_address}:/etc/keystone/;"
+      rsync_command += \
+        "rsync -a --timeout=300 --delete-after /etc/keystone/fernet-keys " \
```
Itxaka: how big can the fernet-keys be? Timeout 300 seems way too high for a rsync, unless we expect the nodes to not be available for rsync at this time?

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -392,19 +396,32 @@
...
+        "#{node_address}:/etc/keystone/; "
     end
+    raise "No other cluster members found" if rsync_command.empty?
```
Itxaka: Maybe Im not understanding this correctly but it sems that on a non-HA install with fernet tokens this will always raise. Wont a raise break the chef-run?

mjura: This is ok, above there is check `if ha_enabled`

Itxaka: ahh indeed, mixed the end of the  `cluster_nodes.map` :+1:

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -392,19 +396,32 @@
...
+    source "keystone-fernet-rotate.erb"
+    owner "root"
+    group node[:keystone][:group]
+    mode "0750"
+    variables(
+      rsync_command: rsync_command
+    )
   end
 
+  crowbar_pacemaker_sync_mark "wait-keystone_fernet_rotate"
```
vuntz: You still need to add a sync mark of type `sync` before this, as you will create a pacemaker resource, which means that you need the template resource from above to be done on all nodes before creating the pacemaker resource.

vuntz: That also means that you don't need to more the installation of the rsync package, btw (it's a bit confusing to have it at the top of the recipe anyway).

chef/cookbooks/keystone/recipes/server.rb
dirkmueller: it needs to be before the first sync, see commit message

aspiers:
#toomanyreviewers

nicolasbock:
@dirkmueller I didn't want to trample on your PR here and copied your commits to #874. Please have a look.

chef/cookbooks/keystone/recipes/server.rb
nicolasbock: @vuntz Could you have a look at #874? I think we added the sync mark at the right place, but with HA one can never be sure :wink:.

PR #840, opened 2017-02-08
Introduce a service to provide L3 HA

matelakat:
A service script will be responsible for periodically checking the
system's state and replicating DHCP servers/migrating routers by calling
neutron-ha-tool.py. Systemd will be used to run the service script.
Pacemaker uses the systemd unit as a resource.

## TODOS

- [x] make TimeoutStopSec (systemd option) to be a parameter - given that it depends heavily on how much time do we give to the script
- [ ] exponential back-off algorithm for errors

# why?
Calling neutron-ha-tool directly from pacemaker resource agent caused
problems:
 - given that the runtime depends on the number of routers, it was
 impossible to come up with a timeout that works for all deployments.
 - on timeouts the process spawned by the resource agent would be
 killed potentially causing routers to be removed but not added.
 - whenever router migration was needed, that resulted the monitor
 operation on the resource to fail, bumping the failure counter
This change adds a service script that calls neutron-ha-tool in an
infinite loop, and also adds systemd configuration for the
service.

Having a service to run the neutron-ha-tool separates the domain
operations (router migration, dhcp replication) from the resource
actions (starting and stopping the ongoing monitor-act cycle). Also
it enables graceful stop of the domain operations, as signal sending
mechanisms can be used in the ha-tool to stop operations at boundaries.

### DIFF
--- a/chef/cookbooks/neutron/files/default/neutron-l3-ha-service.rb
+++ b/chef/cookbooks/neutron/files/default/neutron-l3-ha-service.rb
@@ -1,0 +1,422 @@
+#!/usr/bin/env ruby
+#
+# Copyright 2017, SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require 'yaml'
+require 'open3'
+require 'logger'
+
+$LOG = Logger.new(STDERR)
+
+
+def main
+  terminator = Terminator.new
+  terminator.register_term_signal_handler
+  service_options = ServiceOptions.load ARGV[0]
+  ha_functions = HAFunctions.new(service_options)
+  error_counter = ErrorCounter.new service_options.max_errors_tolerated
+
+  while true
+    status = ha_functions.check_l3_agents
+    if status.exit_status == 2
+      error_counter.reset!
+
+      register_errors_to(error_counter, ha_functions.replicate_dhcp_servers)
+
+      register_errors_to(
+        error_counter,
+        terminator.dont_exit_until_block_finished do
+          ha_functions.migrate_routers_away_from_dead_agents
+        end
+      )
+    elsif status.exit_status == 0
+      error_counter.reset!
+    else
+      error_counter.bump!
+    end
+
+    $LOG.info('Sleeping until next run')
+    sleep service_options.seconds_to_sleep_between_checks
+  end
+end
+
+
+def register_errors_to error_counter, result
+  if result.exit_status == 0
+    error_counter.reset!
+  else
+    error_counter.bump!
+  end
+end
+
+
+class Terminator
+  def initialize
+    @immediate_exit_enabled = true
+    @exit_requested = false
+  end
+
+  def register_term_signal_handler
+    Signal.trap 'TERM' do
+      exit 0 if @immediate_exit_enabled
+      @exit_requested = true
+    end
+  end
+
+  def dont_exit_until_block_finished
+    @immediate_exit_enabled = false
+    result = yield
+    @immediate_exit_enabled = true
+    if @exit_requested
+      $LOG.info('TERM signal received while waiting for block to finish, exiting now')
+      exit 0
+    end
+    result
+  end
+end
+
+
+class HAFunctions
+  def initialize(service_options)
+    @service_options = service_options
+    @hatool = HATool.new service_options.hatool
+    # TODO(mlakat): Check if we can skip this step
+    # maybe we should just assume that the tool supports --retry?
+    @hatool.discover_capabilities
+  end
+
+  def check_l3_agents
+    $LOG.info('checking for dead agents')
+    run_supervised(
+      @hatool.status_command,
+      @service_options.status_timeout
+    )
+  end
+
+  def migrate_routers_away_from_dead_agents
+    $LOG.info('migrating routers away from dead agents')
+    run_supervised(
+      @hatool.migration_command,
+      @service_options.router_migration_timeout
+    )
+  end
+
+  def replicate_dhcp_servers
+    $LOG.info('replicating DHCP servers')
+    run_supervised(
+      @hatool.replicate_dhcp_command,
+      @service_options.dhcp_replication_timeout
+    )
+  end
+
+  def run_supervised(command, timeout)
+    subprocess = Subprocess.new *command
+    subprocess.env.merge! @service_options.hatool.env
+    supervisor = Supervisor.new(subprocess, timeout)
+    supervisor.run_subprocess
+  end
+end
+
+
+class TimedOut < StandardError
+end
+
+
+class MaximumErrorsReached < StandardError
+end
+
+
+class AlreadyCompleted < StandardError
+end
+
+
+class RunningHelpFailed < StandardError
+end
+
+
+class LoggingStreamReader
+  def initialize
+    @lines = []
+    @thread = nil
+  end
+
+  def read stream, log_msg
+    @thread = Thread.new do
+      stream.each do |line|
+        @lines << line
+        $LOG.info("#{log_msg} #{line}")
+      end
+    end
+  end
+
+  def content
+    @lines.join("\n")
+  end
+
+  def join
+    # Leave 1 second for the thread to join, otherwise raise an exception
+    result = @thread.join 1
+    if result.nil?
+      raise StandardError
+    end
+  end
+end
+
+
+class Subprocess
+  attr_reader :env
+
+  def initialize *args
+    @args = args
+    @completed = false
+    @env = {}
+    @stdout_reader = LoggingStreamReader.new
+    @stderr_reader = LoggingStreamReader.new
+  end
+
+  def environment
+    Hash[@env.map { |key, val| [key, val.to_s] }]
+  end
+
+  def start
+    stdin, stdout, stderr, @wait_thr = Open3.popen3(environment, *@args)
+    @pid = @wait_thr.pid
+    @stdout_reader.read stdout, "#{self} stdout:"
+    @stderr_reader.read stderr, "#{self} stderr:"
+    stdin.close
+  end
+
+  def wait timeout
+    if @completed
+      raise AlreadyCompleted
+    end
+    wait_result = @wait_thr.join(timeout)
+    if wait_result.nil?
+      raise TimedOut
+    else
+      @completed = true
+      @stdout_reader.join
+      @stderr_reader.join
+      output = @stdout_reader.content
+      error = @stderr_reader.content
+      exitstatus = @wait_thr.value.exitstatus
+      RunResult.new output, error, exitstatus
+    end
+  end
+
+  def send_signal signal
+    begin
+      Process.kill(signal, @pid)
+    rescue Errno::ESRCH
+      # Process already killed
+    end
+  end
+
+  def to_s
+    status = @pid.nil? ? 'not-started-yet' : @pid.to_s
+    status = 'already-exited' if @completed
+    "Subprocess(#{@args.join ' '})[#{status}]"
+  end
+end
+
+
+class ServiceOptions
+  attr_reader :status_timeout
+  attr_reader :dhcp_replication_timeout
+  attr_reader :router_migration_timeout
+  attr_reader :hatool
+  attr_reader :seconds_to_sleep_between_checks
+  attr_reader :max_errors_tolerated
+
+  def initialize(status_timeout, dhcp_replication_timeout, router_migration_timeout, hatool_options, sleep_time, max_errors_tolerated)
+    @status_timeout = status_timeout
+    @dhcp_replication_timeout = dhcp_replication_timeout
+    @router_migration_timeout = router_migration_timeout
+    @hatool = hatool_options
+    @seconds_to_sleep_between_checks = sleep_time
+    @max_errors_tolerated = max_errors_tolerated
+  end
+
+  def self.load(path)
+    File.open path do |file|
+      data = YAML.load file.read
+      ServiceOptions.new(
+        TimeoutOptions.from_hash(data['timeouts']['status']),
+        TimeoutOptions.from_hash(data['timeouts']['dhcp_replication']),
+        TimeoutOptions.from_hash(data['timeouts']['router_migration']),
+        HAToolOptions.from_hash(data['hatool']),
+        data['seconds_to_sleep_between_checks'].to_i,
+        data['max_errors_tolerated'].to_i
+      )
+    end
+  end
+end
+
+
+class ErrorCounter
+  attr_reader :errors
+
+  def initialize max_errors_tolerated
+    @max_errors_tolerated = max_errors_tolerated
+    @errors = 0
+  end
+
+  def reset!
+    $LOG.info('error counter: re-set to 0')
+    @errors = 0
+  end
+
+  def bump!
+    @errors += 1
+    $LOG.info("error counter: bumped to #{@errors}")
+    if @errors > @max_errors_tolerated
+      $LOG.error("error counter: exceeded limit #{@max_errors_tolerated}")
+      raise MaximumErrorsReached
+    end
+  end
+end
+
+
+class Supervisor
+  def initialize subprocess, timeout_options
+    @subprocess = subprocess
+    @timeout_options = timeout_options
+  end
+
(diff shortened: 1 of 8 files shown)


matelakat:
[Turn neutron-ha-tool into a service](https://trello.com/c/qWAjaaan/70-turn-neutron-ha-tool-into-a-service)

matelakat:
> I really think that the .service file must live in the neutron package, that is where ha-tool lives too

@aplanas we had a discussion about it with @aspiers and we came to the decision that we put everything here for now. This way we have more visibility as the changes are not spread across several systems, and it makes development easier.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -230,7 +244,7 @@
       "os_insecure"    => keystone_settings["insecure"] || node[:neutron][:ssl][:insecure]
     })
     op node[:neutron][:ha][:neutron_ha_tool][:op]
-    action :update
+    action :stop
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
```
vuntz: You can also remove the other pacemaker bits we do related to this primitive.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -257,6 +271,63 @@
...
+  ha_service_primitive_name = "neutron-l3-ha-service"
+
+  pacemaker_primitive ha_service_primitive_name do
+    agent "systemd:neutron-l3-ha-service"
+    op node[:neutron][:ha][:neutron_l3_ha_resource][:op]
+    action :update
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+  ha_service_transaction_objects << "pacemaker_primitive[#{ha_service_primitive_name}]"
+  ha_service_location_name = openstack_pacemaker_controller_only_location_for ha_service_primitive_name
```
vuntz: Empty line before this one (like we do usually in such cases).

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -257,6 +271,63 @@
     action :create
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
+
+  # Install service script
+  cookbook_file "neutron-l3-ha-service.rb" do
+    source "neutron-l3-ha-service.rb"
+    path "/usr/bin/neutron-l3-ha-service.rb"
```
vuntz: Can we avoid executables with a .rb suffix? :-)

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -257,6 +271,63 @@
     action :create
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
+
+  # Install service script
+  cookbook_file "neutron-l3-ha-service.rb" do
+    source "neutron-l3-ha-service.rb"
```
vuntz: (for the record, your filename is using _ instead of -, and you have - here -- I prefer the use of -)

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
```diff
@@ -0,0 +1,326 @@
...
+end
+
+
+class Supervisor
+  def initialize subprocess, timeout_options
+    @subprocess = subprocess
+    @timeout_options = timeout_options
+  end
+
+  def run_subprocess
```
aspiers: Again the class name and method name don't clarify how this type of subprocess running is any different from what the `Subprocess` class already offer.

spec/neutron_l3_ha_service_spec.rb
```diff
@@ -1,3 +1,4 @@
+require 'tmpdir'
```
aspiers: Why does Travis need this?  Please make the commit message describe the root cause of the "why", not the "what".

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
```diff
@@ -1,6 +1,4 @@
-# Neutron L3 HA Service
-#
-# This script is responsible for running neutron-ha-tool
+#!/usr/bin/env ruby
```
aspiers: This can be squashed.

chef/cookbooks/neutron/files/default/neutron-l3-ha-service.service
```diff
@@ -1,3 +1,19 @@
+#
```
aspiers: These could be squashed, although I don't mind too much if they're not.

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
```diff
@@ -0,0 +1,326 @@
...
+
+  def self.from_hash(hash)
+    HatoolOptions.new hash['program'], hash['env'], to_bool(hash['insecure'])
+  end
+end
+
+
+if __FILE__ == $0
+  main
+end
```
aspiers: Missing newline at EOF

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
```diff
@@ -0,0 +1,326 @@
...
+    @kill = kill_timeout.to_i
+  end
+
+  def self.from_hash(hash)
+    TimeoutOptions.new hash['terminate'], hash['kill']
+  end
+end
+
+
+class HatoolOptions
```
aspiers: This capitalization is not consistent with your `HATool` class. I think `HAToolOptions` would be better.

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
```diff
@@ -0,0 +1,326 @@
...
+    @hatool = hatool_options
+    @seconds_to_sleep_between_checks = sleep_time
+    @max_errors_tolerated = max_errors_tolerated
+  end
+
+  def self.load(path)
+    File.open path do |file|
+      data = YAML.load file.read
+      ServiceOptions.new(
+          TimeoutOptions.from_hash(data['timeouts']['status']),
```
aspiers: wrong indentation?

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
```diff
@@ -0,0 +1,326 @@
...
+        error_counter.bump!
+      elsif router_migration_result.exit_status == 0
+        error_counter.reset!
+      else
+        error_counter.bump!
+      end
+    elsif status.exit_status == 0
+      error_counter.reset!
+    end
+    sleep service_options.seconds_to_sleep_between_checks
```
aspiers: I don't think you want to sleep if there was an error in the previous run.  So maybe add `unless error_counter.errors > 0`.

spec/neutron_helpers_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    config = NeutronHelper.make_l3_ha_service_config defaults, true do |env|
+      env['other_key'] = 'other_value'
+    end
+
+    YAML.load config
+
+    expect(defaults).to eq( {:hatool => {:env => {'somekey' => 'somevalue'}}})
+  end
+end
```
aspiers: Missing newline at EOF

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
matelakat: What I was thinking is that if we don't sleep in between errors, an outage might quickly make the tool fail: Let's say api is down for 5 seconds, and if we don't sleep, we'll use all our available error possibilities within that 5 seconds. Whereas if we do sleep in between, the tool will survive?

chef/cookbooks/neutron/files/default/neutron-l3-ha-service.service
```diff
@@ -0,0 +1,11 @@
...
+Description=Neutron L3 HA Service
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/neutron-l3-ha-service.rb /etc/neutron/neutron-l3-ha-service.yaml
+# Time to wait for the process to react to the TERM signal
+TimeoutStopSec=120
+
+[Install]
+WantedBy=multi-user.target
```
matelakat: That's a good point, we don't need any Install section

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
matelakat: Yes, as I mentioned in the pull request's main message, this is a preview, so the commit structure will be changed. The end result will be 2 commits, one to disable the old resource, and another for the new one.

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
```diff
@@ -0,0 +1,326 @@
...
+  error_counter = ErrorCounter service_options.max_errors_tolerated
+
+  while true
+    status = ha_functions.check_l3_agents
+    if status.timed_out
+      error_counter.bump!
+    elsif status.exit_status == 2
+      error_counter.reset!
+      dhcp_replication_result = ha_functions.replicate_dhcp_servers
+      if dhcp_replication_result.timed_out
```
matelakat: On it

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
matelakat: Ah, I trusted rubymine to format my code. What code conventions are we following btw? This one maybe? https://github.com/bbatsov/ruby-style-guide

Maybe we shall add rubocop checks? I recall we have some automated checks for other projects?

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
```diff
@@ -0,0 +1,326 @@
...
+    File.open path do |file|
+      data = YAML.load file.read
+      ServiceOptions.new(
+          TimeoutOptions.from_hash(data['timeouts']['status']),
+          TimeoutOptions.from_hash(data['timeouts']['dhcp_replication']),
+          TimeoutOptions.from_hash(data['timeouts']['router_migration']),
+          HatoolOptions.from_hash(data['hatool']),
+          data['seconds_to_sleep_between_checks'].to_i,
+          data['max_errors_tolerated'].to_i,
+
```
matelakat: fixed

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
```diff
@@ -0,0 +1,326 @@
...
+  def bump!
+    @errors += 1
+    if @errors > @max_errors_tolerated
+      raise MaximumErrorsReached
+    end
+  end
+end
+
+
+class ScriptRunner
```
matelakat: Yeah, I'll make this one a single function instead - the responsibility is really to run something with a time limit.

matelakat: I'm having a hard time figuring out how can I stub out "global" methods, so I'll leave this class here for now.

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
aspiers: OK thanks.

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
aspiers: In that case I think an exponential back-off algorithm is the best approach.

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
aspiers: We use Hound CI as I'm sure you already know, and it's already configured for this repository.  I don't know why it didn't flag this as a violation but you could look into its config.

chef/cookbooks/neutron/files/default/neutron_l3_ha_service.rb
matelakat: onit

chef/cookbooks/neutron/files/default/neutron-l3-ha-service.rb
```diff
@@ -55,6 +46,18 @@ def main
 end
 
 
+def register_errors_to error_counter
+  result = yield
```
aspiers: This is slightly overkill since you could just pass the result as a normal parameter - but I guess it allows for a nice multi-line structure for the callers, so I don't object to it.

spec/neutron_l3_ha_service_spec.rb
matelakat: My ruby env was newer than the one we have at travis, that's why I had to add it. This commit will be squashed. As I said, I added this separate commit so that pending reviewers don't get updates from forced pushes.

spec/neutron_helpers_spec.rb
matelakat: onit

chef/cookbooks/neutron/files/default/neutron-l3-ha-service.rb
matelakat: Yes, I was hesitating about it - maybe as you mentioned it as well, I might switch to a more traditional structure.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
matelakat: done

chef/cookbooks/neutron/recipes/network_agents_ha.rb
matelakat: right

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -231,44 +232,18 @@
...
   pacemaker_primitive ha_tool_primitive_name do
     agent node[:neutron][:ha][:network][:ha_tool_ra]
-    params ({
-      "os_auth_url"    => os_auth_url_v2,
-      "os_region_name" => keystone_settings["endpoint_region"],
-      "os_tenant_name" => keystone_settings["admin_tenant"],
-      "os_username"    => keystone_settings["admin_user"],
-      "os_insecure"    => keystone_settings["insecure"] || node[:neutron][:ssl][:insecure]
-    })
     op node[:neutron][:ha][:neutron_ha_tool][:op]
```
vuntz: You can remove that line too, it's not needed for stop & delete.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -231,44 +232,18 @@
...
-
-  ha_tool_location_name = openstack_pacemaker_controller_only_location_for ha_tool_primitive_name
-  ha_tool_transaction_objects << "pacemaker_location[#{ha_tool_location_name}]"
-
-  pacemaker_transaction "neutron ha tool" do
-    cib_objects ha_tool_transaction_objects
-    # note that this will also automatically start the resources
-    action :commit_new
+    action [:stop, :delete]
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
vuntz: You need to check if the resource exists in `only_if`...

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -231,44 +232,18 @@
...
   end
 
+  # Remove ordering
   crowbar_pacemaker_order_only_existing "o-#{ha_tool_primitive_name}" do
```
vuntz: I'd change this to `pacemaker_order`

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -231,44 +232,18 @@
...
-      "os_insecure"    => keystone_settings["insecure"] || node[:neutron][:ssl][:insecure]
-    })
     op node[:neutron][:ha][:neutron_ha_tool][:op]
-    action :stop
-    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
-  end
-  ha_tool_transaction_objects << "pacemaker_primitive[#{ha_tool_primitive_name}]"
-
-  ha_tool_location_name = openstack_pacemaker_controller_only_location_for ha_tool_primitive_name
-  ha_tool_transaction_objects << "pacemaker_location[#{ha_tool_location_name}]"
```
vuntz: There's a location constraint created here that needs to be removed too

matelakat:
testing deployment...

matelakat:
Tested on my local deployment, works nicely. Plans are to test it on scale with QA next week hopefully.

rsalevsky:
Do we want to have this for master as well?

matelakat:
dhcp replication needs to be removed from this script.

vuntz:
Replaced by https://github.com/crowbar/crowbar-openstack/pull/840

PR #841, opened 2017-02-08
horizon: Configure stickiness with haproxy

vuntz:
We use cookie-based stickiness for non-SSL, and we also expire the
stickiness based on the session timeout. This should make the
haproxy/horizon combo to work well, with each session sticking to one
backend, thus avoiding potential issues where requests are spread
between multiple backends.

Depends on https://github.com/crowbar/crowbar-ha/pull/179


matelakat:
re-basing it on master

PR #843, opened 2017-02-08
magnum: Deployment fixes for HA

dirkmueller:


chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,39 +42,63 @@
...
-  create_magnum_domain = "#{openstack_command} domain create -f value -c id"
-  create_magnum_domain << ' --description "Owns users and projects created by magnum"'
-  create_magnum_domain << " #{magnum_domain_name}"
+create_magnum_domain = "#{openstack_command} domain create -f value -c id"
+create_magnum_domain << ' --description "Owns users and projects created by magnum"'
+create_magnum_domain << " #{magnum_domain_name}"
+
+if is_founder && (magnum_domain_id.nil? || magnum_domain_id.empty?)
   magnum_domain_id = Mixlib::ShellOut.new(create_magnum_domain).run_command.stdout.chomp
 end
```
vuntz: Can't we just use `openstack domain create --or-show` ? This would simplify things to:

```ruby
create_show_magnum_domain = "#{openstack_command} domain create --or-show -f value -c id"
create_show_magnum_domain << ' --description "Owns users and projects created by magnum"'
create_show_magnum_domain << " #{magnum_domain_name}"

magnum_domain_id = Mixlib::ShellOut.new(create_show_magnum_domain).run_command.stdout.chomp
```

(a side-benefit is that you don't need to have different behavior between founder and non-founder, since the founder would go first and create things, and the others would just show the domain)

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -32,6 +29,8 @@
   end
 end
 
+crowbar_pacemaker_sync_mark "sync-magnum_setup_client_installation"
```
vuntz: I believe you don't need the `sync` sync mark, as the founder doesn't need to wait for all non-founders to reach that point before proceeding.

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,39 +42,63 @@
...
   magnum_domain_id = Mixlib::ShellOut.new(create_magnum_domain).run_command.stdout.chomp
 end
 
+crowbar_pacemaker_sync_mark "create-magnum_setup_domain"
+
+unless is_founder
+  magnum_domain_id = Mixlib::ShellOut.new(get_magnum_domain_id).run_command.stdout.chomp
+end
+
+crowbar_pacemaker_sync_mark "wait-magnum_setup_domain_admin"
```
vuntz: Don't use two sync marks, just use one that covers the two blocks.

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,39 +42,63 @@
...
+if is_founder
+  magnum_domain_admin_id = Mixlib::ShellOut.new(get_magnum_domain_admin_id).run_command.stdout.chomp
+
+  if magnum_domain_admin_id.nil? || magnum_domain_admin_id.empty?
+    magnum_domain_admin_id = Mixlib::ShellOut.new(create_magnum_domain_admin).run_command.stdout.chomp
+
+    set_magnum_domain_admin = "#{openstack_command} user set --password #{magnum_domain_admin_pass}"
+    set_magnum_domain_admin << ' --description "Manages users and projects created by magnum"'
+    set_magnum_domain_admin << " #{magnum_domain_admin}"
+    Mixlib::ShellOut.new(set_magnum_domain_admin).run_command
```
vuntz: You can do this in `user create` too. This would mean that we would combine everything in one big `user create --or-show --password --description`, and then we can have code similar to what I suggested above for the domain.

chef/cookbooks/magnum/recipes/setup.rb
jgrassler: Can we rely on the `--or-show` operation being atomic (e.g. by virtue of being baked into the API or something along those lines)? For if it is implemented as a `domain list` followed by a `domain create` if the domain to be created isn't in the list, we'll create a potential race condition that way (uniqueness constraints on domains _should_ prevent that from going through, but stranger things happen at sea. Also, that will break the current chef run if it happens.).

chef/cookbooks/magnum/recipes/setup.rb
jgrassler: Again, can we rely on `--or-show` to be atomic (see first comment on that for a detailed discussion of the problem)?

jgrassler:
Alright, provided the uniqueness constraints are being enforced by the Keystone API, it should be atomic:

* https://github.com/openstack/python-openstackclient/blob/master/openstackclient/identity/v3/domain.py#L79
* https://github.com/openstack/python-openstackclient/blob/master/openstackclient/identity/v3/user.py#L129

(tl;dr: `python-openstackclient` just forges ahead and tries to create the resource in both cases and ignores a `Conflict` exception from the Keystone client if `--or-show` was specified).

So I'm happy with this once it's switched to `--or-show`.

dirkmueller:
@vuntz addressed the comments

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,41 +37,35 @@
...
+create_magnum_domain_admin << " --or-show -f value -c id #{magnum_domain_admin}"
 
-  set_magnum_domain_admin = "#{openstack_command} user set --password #{magnum_domain_admin_pass}"
-  set_magnum_domain_admin << ' --description "Manages users and projects created by magnum"'
-  set_magnum_domain_admin << " #{magnum_domain_admin}"
-  Mixlib::ShellOut.new(set_magnum_domain_admin).run_command
-end
+magnum_domain_admin_id = Mixlib::ShellOut.new(create_magnum_domain_admin).run_command.stdout.chomp
 
 check_magnum_domain_role = "#{openstack_command} role list -f csv --column Name"
```
vuntz: For the record:

```
openstack role list -f csv --column Name --user admin --domain default
Listing assignments using role list is deprecated. Use role assignment list --user <user-name> --domain <domain-name> --names instead.
```

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,41 +37,35 @@
...
 check_magnum_domain_role << " --domain #{magnum_domain_id} --user #{magnum_domain_admin_id}"
+
 magnum_domain_role = Mixlib::ShellOut.new(check_magnum_domain_role).run_command.stdout
 
-unless magnum_domain_role.include? "admin"
+unless magnum_domain_role.include?("admin")
```
vuntz: Looks to me like the check is not strict enough and would succeed for roles like `foobaradmin`.

dirkmueller:
not yet retested after all the change requests, and also still need to fix x509

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,41 +37,35 @@
...
-end
+create_magnum_domain = "#{openstack_command} domain create -f value -c id"
+create_magnum_domain << " --description 'Owns users and projects created by magnum'"
+create_magnum_domain << " --or-show"
+create_magnum_domain << " #{magnum_domain_name}"
 
-get_magnum_domain_admin_id = "#{openstack_command} user show #{magnum_domain_admin}"
-get_magnum_domain_admin_id << " -f value -c id --domain #{magnum_domain_name}"
-magnum_domain_admin_id = Mixlib::ShellOut.new(get_magnum_domain_admin_id).run_command.stdout.chomp
+magnum_domain_id = Mixlib::ShellOut.new(create_magnum_domain).run_command.stdout.chomp
```
toabctl: Can we do some error checking here?. The following example passed even if the dir does not exist:
```
chef > Mixlib::ShellOut.new("ls xxx").run_command.stdout
 => "" 
chef > 
```

You need something like:
```
cmd = Mixlib::ShellOut.new(create_magnum_domain)
cmd.run_command
cmd.error! # this raises an exception if there was an error
output = cmd.stdout
```

Newer mixlib-shellout gems have a ```cmd.error?``` to check for an error.

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,41 +37,35 @@
...
+create_magnum_domain_admin << " --description 'Manages users and projects created by magnum'"
+create_magnum_domain_admin << " --password #{magnum_domain_admin_pass}"
+create_magnum_domain_admin << " --or-show -f value -c id #{magnum_domain_admin}"
 
-  set_magnum_domain_admin = "#{openstack_command} user set --password #{magnum_domain_admin_pass}"
-  set_magnum_domain_admin << ' --description "Manages users and projects created by magnum"'
-  set_magnum_domain_admin << " #{magnum_domain_admin}"
-  Mixlib::ShellOut.new(set_magnum_domain_admin).run_command
-end
+magnum_domain_admin_id = Mixlib::ShellOut.new(create_magnum_domain_admin).run_command.stdout.chomp
```
toabctl: same here. error checking would be nice.

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,41 +37,35 @@
...
-unless magnum_domain_role.include? "admin"
+unless magnum_domain_role.include?('"admin"')
   create_magnum_domain_role = "#{openstack_command} role add --user #{magnum_domain_admin_id}"
   create_magnum_domain_role << " --domain #{magnum_domain_id} admin"
   Mixlib::ShellOut.new(create_magnum_domain_role).run_command
 end
 
+crowbar_pacemaker_sync_mark "create-magnum_setup_domain"
+
 node.set["magnum"]["trustee"]["domain_id"] = magnum_domain_id
```
toabctl: why are we storing theses 2 variables in the node? Can' we just get them with every run(we do that anyway) and pass them into the config template?

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,41 +37,35 @@
...
 node.set["magnum"]["trustee"]["domain_admin_id"] = magnum_domain_admin_id
 node.save
```
toabctl: Does that trigger a save on every chef-client run?

dirkmueller:
@toabctl solved your request for error handling via excessive nesting..

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,41 +37,41 @@
...
+create_magnum_domain = "#{openstack_command} domain create -f value -c id"
+create_magnum_domain << " --description 'Owns users and projects created by magnum'"
+create_magnum_domain << " --or-show"
+create_magnum_domain << " #{magnum_domain_name}"
 
-if magnum_domain_id.nil? || magnum_domain_id.empty?
-  create_magnum_domain = "#{openstack_command} domain create -f value -c id"
-  create_magnum_domain << ' --description "Owns users and projects created by magnum"'
-  create_magnum_domain << " #{magnum_domain_name}"
+unless node.set["magnum"]["trustee"]["domain_id"] && node.set["magnum"]["trustee"]["domain_admin_id"]
```
toabctl: This is not executed even if ```["magnum"]["trustee"]["domain_id"]``` is empty. That's wrong, isn't it?

dirkmueller: Yeah, this is broken..

chef/cookbooks/magnum/recipes/setup.rb
```diff
@@ -43,41 +37,41 @@
 openstack_command << " --os-tenant-name #{keystone_settings["admin_tenant"]}"
 openstack_command << " --os-auth-url #{auth_url} #{insecure}"
 
-get_magnum_domain_id = "#{openstack_command} domain show #{magnum_domain_name} -f value -c id"
-magnum_domain_id = Mixlib::ShellOut.new(get_magnum_domain_id).run_command.stdout.chomp
+crowbar_pacemaker_sync_mark "wait-magnum_setup_domain"
+
+create_magnum_domain = "#{openstack_command} domain create -f value -c id"
```
toabctl: shouldn't that be done only by the cluster founder?

chef/cookbooks/magnum/recipes/setup.rb
dirkmueller: fixed

chef/cookbooks/magnum/recipes/setup.rb
dirkmueller: adjusted

toabctl:
[(3) S54P5 upload magnum image into glance store by default](https://trello.com/c/QOkOA9XT/442-3-s54p5-upload-magnum-image-into-glance-store-by-default)

PR #848, opened 2017-02-17
database: Add ability to log slow queries in postgresql

vuntz:
This is useful to understand why some OpenStack API calls may be slow,
if this turns out to be caused by the database.

The setting is in ms; -1 means that it's disabled.

Closes https://github.com/sap-oc/crowbar-openstack/issues/8


aplanas:
I do not see how those parameters are exported in the configuration file of postgresql

vuntz:
> @vuntz - are we intentionally ignoring hound errors?

which one?

vuntz:
> I do not see how those parameters are exported in the configuration file of postgresql

It's magic, check https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/postgresql/templates/default/postgresql.conf.erb. In short, we simply put all attributes in the config file.

vuntz:
> The hound violations are around migrations, aren't they visible for you?

They were about an old commit, and are already fixed. See the status check :-)

PR #851, opened 2017-02-24
nova: Add attribute to configure default_schedule_zone

### LINKED ISSUE #12: nova: Add default_schedule_zone to nova proposal
This is used in the wrapper cookbook, and as this changes the schema, this actually breaks PTF. So we should push that here.

@tpatzig @mkoderer do you have more details about this?

vuntz:
Quoting the doc:

> Availability zone to use when user doesn't specify one.
>
> This option is used by the scheduler to determine which availability
> zone to place a new VM instance into if the user did not specify one
> at the time of VM boot request.
>
> Possible values:
>
> * Any string representing an availability zone name
> * Default value is None.
>  (string value)

We don't set it by default, to keep the current behavior.

Closes https://github.com/sap-oc/crowbar-openstack/issues/12

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -56,6 +56,9 @@ config_drive_format = vfat
 <% if @force_config_drive || @libvirt_type.eql?('zvm') %>
 force_config_drive = True
 <% end %>
+<% unless node[:nova][:default_schedule_zone].empty? -%>
+default_schedule_zone = <%= node[:nova][:default_schedule_zone] %>
+<% end %>
 debug = <%= node[:nova][:debug] %>
 log_dir = /var/log/nova
 use_syslog = <%= node[:nova][:use_syslog] ? 'True' : 'False' %>
--- a/chef/data_bags/crowbar/migrate/nova/206_add_default_schedule_zone.rb
+++ b/chef/data_bags/crowbar/migrate/nova/206_add_default_schedule_zone.rb
@@ -1,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "default_schedule_zone"
+    a["default_schedule_zone"] = ta["default_schedule_zone"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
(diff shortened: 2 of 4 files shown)


chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -63,6 +63,9 @@ config_drive_format = vfat
 <% if @force_config_drive || @libvirt_type.eql?('zvm') %>
 force_config_drive = True
 <% end -%>
+<% unless node[:nova][:default_schedule_zone].empty? -%>
+default_schedule_zone = <%= node[:nova][:default_schedule_zone %>
```
cmurphy: The mkcloud test is complaining because of the missing ']' here

cmurphy (changes requested):

chef/cookbooks/nova/templates/default/nova.conf.erb
vuntz: Doh, thanks. Fixed.

sayalilunkad (dismissed):
Looks good!

cmurphy (dismissed):

cmurphy:
Looks fine to me too but hound is upset

rsalevsky (dismissed):

nicolasbock (dismissed):

vuntz:
Changed to fix the hound comment.

dirkmueller (approved):

nicolasbock (approved):

toabctl (approved):

chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -56,6 +56,9 @@ config_drive_format = vfat
 <% if @force_config_drive || @libvirt_type.eql?('zvm') %>
 force_config_drive = True
 <% end %>
+<% unless node[:nova][:default_schedule_zone].empty? -%>
```
cmurphy: You could actually omit the '-' from the '-%>' here (see https://github.com/crowbar/crowbar-openstack/pull/1186). Not important of course.

cmurphy (approved):

PR #852, opened 2017-02-24
[3.0] nova: Add attribute to configure default_schedule_zone

vuntz:
Quoting the doc:

> Availability zone to use when user doesn't specify one.
>
> This option is used by the scheduler to determine which availability
> zone to place a new VM instance into if the user did not specify one
> at the time of VM boot request.
>
> Possible values:
>
> * Any string representing an availability zone name
> * Default value is None.
>  (string value)

We don't set it by default, to keep the current behavior.

Closes https://github.com/sap-oc/crowbar-openstack/issues/12

(cherry picked from commit fa0ad588f1fba630a0858e61ac57e0482a95f207)

Backport of https://github.com/crowbar/crowbar-openstack/pull/851


rsalevsky:
Closing this PR as we won't pickup new changes for Cloud 6 maintenance updates. If this is still important for newer versions please feel free to reopen this PR against master.

PR #853, opened 2017-02-28
magnum: expose trust/cluster_user_trust.

jgrassler:
[This upstream commit](https://github.com/openstack/magnum/commit/2d4e617a529ea12ab5330f12631f44172a623a14) in Magnum introduced the setting `cluster_user_trust` which defaults to `False` (the secure choice) but must be set to `True` for some clusters that run their own registry and/or the `rexray` volume drivers. In order to allow users (and functional tests that test the problematic cluster types) to make that choice we need to expose this setting in Crowbar.


jgrassler:
Oops. Fixed the revision bump now. Thanks :-)

PR #856, opened 2017-03-01
nova: Always set live_migration_flag/block_migration_flag for kvm

vuntz:
It's wrong to not set live_migration_flag when no shared storage is
used because it's actually still the flag used when volume-backed
instances are migrated.

This becomes visible since we now set live_migration_uri, and we get
this error:
  Live Migration failure: argument unsupported: migration URI is not supported by tunnelled migration

That's because the default value of live_migration_flag contains
VIR_MIGRATE_TUNNELLED.


dirkmueller:
is the a bugreport about what we're trying to fix here? I'm a bit confused by the commit message alone.

dirkmueller:
Am

vuntz:
@dirkmueller the intent was really to fix it in Cloud 6, tbh. I just pushed to master to have the change there too. So let's ignore master and focus on https://github.com/crowbar/crowbar-openstack/pull/857

There's no bug report, but it's hit by QAM.

PR #857, opened 2017-03-01
nova: Always set live_migration_flag/block_migration_flag for kvm

vuntz:
It's wrong to not set live_migration_flag when no shared storage is
used because it's actually still the flag used when volume-backed
instances are migrated.

This becomes visible since we now set live_migration_uri, and we get
this error:
  Live Migration failure: argument unsupported: migration URI is not supported by tunnelled migration

That's because the default value of live_migration_flag contains
VIR_MIGRATE_TUNNELLED.

(cherry picked from commit 096cb6f3c8df50857fe39b5f75745399dbe13e01)

Backport of https://github.com/crowbar/crowbar-openstack/pull/856

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -2596,15 +2596,13 @@ live_migration_flag = VIR_MIGRATE_UNDEFINE_SOURCE, VIR_MIGRATE_LIVE
 live_migration_uri = qemu+tcp://<%= @migration_host %>/system
 
   <% if @libvirt_migration -%>
-    <% if @shared_instances -%>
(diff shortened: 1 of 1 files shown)


vuntz:
no code change, just removed the cherry-pick mention as https://github.com/crowbar/crowbar-openstack/pull/856 got closed

vuntz:
@dirkmueller for the record, that's the code we're hitting https://git.openstack.org/cgit/openstack/nova/tree/nova/virt/libvirt/driver.py?h=liberty-eol#n5695

And the tempest test that fails is test_volume_backed_live_migration.

The point, in the end, is really to never set VIR_MIGRATE_TUNNELLED.

rsalevsky:
Let's merge this to unblock the crowbar-openstack maintenance update.

PR #858, opened 2017-03-02
barbican: Add SSL support

cmurphy:
Add a "protocol" setting and "ssl" configuration options to the barbican
barclamp.

This patch also removes the "api" "ssl" config flag in favor of the "protocol"
string setting, as this was the only barclamp to use such a setting. This
brings it into conformance with other barclamps and makes it possible to test
the SSL scenario without any special treatment[1].

[1] https://github.com/SUSE-Cloud/automation/blob/93fac15bdaa15751362f5856fa348ba3c632e8dd/scripts/qa_crowbarsetup.sh#L2242

### DIFF
--- a/chef/data_bags/crowbar/migrate/barbican/101_add_ssl_attributes.rb
+++ b/chef/data_bags/crowbar/migrate/barbican/101_add_ssl_attributes.rb
@@ -1,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  a["ssl"] = ta["ssl"]
+  a["api"]["protocol"] = a["api"]["ssl"] ? "https" : "http"
+  a["api"].delete("ssl")
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("ssl")
+  a["api"]["ssl"] = a["api"]["protocol"] == "https" ? true : false
+  a["api"].delete("protocol")
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-barbican.schema
+++ b/chef/data_bags/crowbar/template-barbican.schema
@@ -16,11 +16,11 @@
               "required": true,
               "type": "map",
               "mapping": {
+                  "protocol": { "required": true, "type": "str" },
                   "bind_host": { "required": true, "type": "str" },
                   "bind_port": { "required": true, "type": "int" },
                   "logfile": { "required": true, "type": "str" },
                   "processes": { "required": true, "type": "int" },
-                  "ssl": { "required": true, "type": "bool" },
                   "threads": { "required": true, "type": "int" }
                 }
               },
@@ -31,6 +31,16 @@
                   "database": { "required": true, "type": "str" },
                   "password": { "required": true, "type": "str" },
                   "user": { "required": true, "type": "str" }
+              }
+            },
+            "ssl": {
+              "type": "map", "required": true, "mapping": {
+                "certfile": { "type" : "str", "required" : true },
+                "keyfile": { "type" : "str", "required" : true },
+                "generate_certs": { "type" : "bool", "required" : true },
+                "insecure": { "type" : "bool", "required" : true },
+                "cert_required": { "type" : "bool", "required" : true },
+                "ca_certs": { "type" : "str", "required" : true }
               }
             },
             "debug": { "type": "bool" },
--- a/crowbar_framework/app/views/barclamp/barbican/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/barbican/_edit_attributes.html.haml
@@ -1,3 +1,25 @@
 = attributes_for @proposal do
   .panel-sub
     = header show_raw_deployment?, true
+
+  .panel-body
+
+    %fieldset
+      %legend
+        = t(".ssl_header")
+
+      = select_field %w(api protocol),
+        :collection => :api_protocols_for_barbican,
+        "data-sslprefix" => "ssl",
+        "data-sslcert" => "/etc/barbican/ssl/certs/signing_cert.pem",
+        "data-sslkey" => "/etc/barbican/ssl/private/signing_key.pem"
+
+      #ssl_container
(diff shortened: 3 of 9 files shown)


sayalilunkad:
Can you please update the commit message to prefix it with Barbican:. We normally follow that for all commit messages.

chef/data_bags/crowbar/migrate/barbican/101_add_ssl_attributes.rb
```diff
@@ -0,0 +1,20 @@
+def upgrade(ta, td, a, d)
+  a["ssl"] = {}
+  a["ssl"]["certfile"]       = ta["ssl"]["certfile"]
+  a["ssl"]["keyfile"]        = ta["ssl"]["keyfile"]
+  a["ssl"]["generate_certs"] = ta["ssl"]["generate_certs"]
+  a["ssl"]["insecure"]       = ta["ssl"]["insecure"]
+  a["ssl"]["cert_required"]  = ta["ssl"]["cert_required"]
+  a["ssl"]["ca_certs"]       = ta["ssl"]["ca_certs"]
+
+  a["api"]["protocol"] = ta["api"]["protocol"]
```
s-t-e-v-e-n-k: You could almost do the same thing that you added for the downgrade and set protocol to the correct value depending on if a["api"]["ssl"] is true. However, that's totally a nice to have.

chef/data_bags/crowbar/migrate/barbican/101_add_ssl_attributes.rb
```diff
@@ -0,0 +1,20 @@
+def upgrade(ta, td, a, d)
+  a["ssl"] = {}
```
toabctl: You could just do ```a["ssl"] = ta["ssl"]```

crowbar_framework/app/views/barclamp/barbican/_edit_attributes.html.haml
```diff
@@ -1,3 +1,25 @@
...
+  .panel-body
+
+    %fieldset
+      %legend
+        = t(".ssl_header")
+
+      = select_field %w(api protocol),
+        :collection => :api_protocols_for_barbican,
+        "data-sslprefix" => "ssl",
+        "data-sslcert" => "/etc/barbican/ssl/certs/signing_cert.pem",
```
toabctl: shouldn't that be ```%w(ssl certfile)``` ?

crowbar_framework/app/views/barclamp/barbican/_edit_attributes.html.haml
```diff
@@ -1,3 +1,25 @@
...
+        "data-sslkey" => "/etc/barbican/ssl/private/signing_key.pem"
```
toabctl: same here

crowbar_framework/app/views/barclamp/barbican/_edit_attributes.html.haml
cmurphy: I was looking to other barclamps for examples and this is consistent with how the others do it:
https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/views/barclamp/glance/_edit_attributes.html.haml#L50
https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/views/barclamp/cinder/_edit_attributes.html.haml#L244
https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/views/barclamp/aodh/_edit_attributes.html.haml#L21

toabctl: Ok. If that works..

cmurphy: If I change it to %w(ssl certfile) and similar for keyfile, then in the UI set "Generate self-signed certificates" to true, the fields for certificate file and key file change to the strings "ssl certfile" and "ssl keyfile", and then trying to apply the proposal fails because those are not real file paths. So, the way I have it is correct.

chef/data_bags/crowbar/migrate/barbican/101_add_ssl_attributes.rb
cmurphy: done

Itxaka:
@toabctl @sayalilunkad Could you re-review this? This is blocking the correct jenkins run for SSL jobs so it would be good to have it merged asap: https://ci.suse.de/view/Cloud/view/Cloud7/job/cloud-mkcloud7-job-ssl-x86_64/

chef/data_bags/crowbar/template-barbican.schema
```diff
@@ -33,6 +33,14 @@
...
               }
             },
+            "ssl": { "type": "map", "required": true, "mapping": {
+              "certfile": { "type" : "str", "required" : true },
+              "keyfile": { "type" : "str", "required" : true },
+              "generate_certs": { "type" : "bool", "required" : true },
+              "insecure": { "type" : "bool", "required" : true },
+              "cert_required": { "type" : "bool", "required" : true },
+              "ca_certs": { "type" : "str", "required" : true }
+            }},
```
sayalilunkad: nit: Can we move the second brace on the next line?

sayalilunkad:
[(3) S56P8 Add SSL support to barbican barclamp](https://trello.com/c/ibqvVB1Y/474-3-s56p8-add-ssl-support-to-barbican-barclamp)

sayalilunkad:
Rebuilding the ha gating with ssl and barbican enabled.

sayalilunkad:
I guess the gating failure on the ha job is not related so we can merge this.

cmurphy:
I'm kind of concerned that I don't see other failures like it, @sayalilunkad is it a known issue? Should I try rebuilding it?

cmurphy:
Latest failure is different and seems like a network failure while trying to create an s3 bucket, going to go ahead and merge

PR #859, opened 2017-03-02
keystone: Update keystone endpoint (bsc#1004219)

nicolasbock:
Currently the keystone endpoint is created during bootstrapping
keystone but can not be updated. This means that the keystone barclamp
can not be switched from http to https once it is already deployed
because the endpoints are not updated.

This patch introduces a new keystone resource, `update_endpoint`, which
updates the endpoints of a service.


vuntz:
Should there be some magic that changes old_protocol/old_insecure to the right values after a successful apply?

nicolasbock:
@vuntz I was hoping that [this block](https://github.com/crowbar/crowbar-openstack/pull/859/files#diff-28b6adf7bd1f81183dace3201a2720a5R243) would do it. In my local experiments it seems to work.

vuntz:
> @vuntz I was hoping that this block would do it. In my local experiments it seems to work.

Oh, I missed that. So, a couple of issues :-)

- if it's attributes set on the node by the cookbook, then you don't want the attribute in the proposal
- that being said, I don't think it should be attributes on the node, as there could be some edge case like "I change from non-ssl to ssl and at the same time, I decide to deploy keystone on another node"

vuntz:
I just realized one thing: this is actually not what we want :-) We have a similar bug where if the public name of the keystone node changes, the endpoint won't be updated, and your patch won't fix it.

So instead of remembering the old protocol and so on, we should actually remember the old URL and the insecure bit.

nicolasbock:
@vuntz Ok, at least that's a simple change :smile: I'll fix it up and test it.

sayalilunkad:
[(3) S56P1 bsc#1004219 - L3: Keystone fails when changing ssl configuration from default to ssl true](https://trello.com/c/jFwH1U0m/477-3-s56p1-bsc-1004219-l3-keystone-fails-when-changing-ssl-configuration-from-default-to-ssl-true)

nicolasbock:
@vuntz Please have another look. I finally figured out a way to communicate between the recipe and crowbar codes. The code still contains some debugging output which I will remove. Thanks!

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -219,8 +219,8 @@
   my_service_id, error = _find_id(http, headers, new_resource.endpoint_service, path, dir)
   unless my_service_id
       Chef::Log.error "Couldn't find service #{new_resource.endpoint_service} in keystone"
-      raise "Failed to talk to keystone in add_endpoint_template" if error
       new_resource.updated_by_last_action(false)
+      raise "Failed to talk to keystone in add_endpoint_template" if error
```
cmurphy: What's the reason for moving this?

chef/data_bags/crowbar/migrate/keystone/109_add_admin_url.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["api"]["internal_url_host"] = ta["api"]["internal_url_host"]
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["api"].delete("internal_url_host")
+  return a, d
+end
```
cmurphy: I'm curious why this parameter was added, was missing it part of the problem? I'm also seeing this when I try to change the endpoint from http to https:

```
WARN -- key 'internal_url_host:' is required.
```

followed by a proposal validation traceback.

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -109,6 +109,22 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+
+    unless old_role.nil?
+      old_elements = old_role.override_attributes["keystone"]
+      attributes = role.override_attributes["keystone"]
+
+      # Copy old keystone endpoints to new role.
+      @logger.debug("old_elements: #{old_elements.pretty_inspect}")
+      if old_elements.key?("endpoint")
+        @logger.debug("copying old endpoint")
+        attributes["endpoint"] = Marshal.load(Marshal.dump(old_elements["endpoint"]))
```
cmurphy: I've never used Marshal, what is the benefit of dumping and then loading rather than just using old_elements["endpoint"] directly?

chef/data_bags/crowbar/migrate/keystone/109_add_admin_url.rb
cmurphy: I think I had a stale data bag from testing something else, I don't see the warning any more. I still think it deserves a note in the commit message or as a comment in the recipe.

dirkmueller:
@nicolasbock needs rebase

chef/cookbooks/keystone/providers/register.rb
nicolasbock: In case the exception is caught, the `new_resource` would not reflect the fact that something failed. I don't think this is currently necessary, it was just more of a precautionary step.

chef/data_bags/crowbar/migrate/keystone/109_add_admin_url.rb
nicolasbock: Thanks for checking! I have added a note in the recipe. I hope it explains better why we need this parameter.

crowbar_framework/app/models/keystone_service.rb
nicolasbock: I wanted to deep copy the whole data structure. There doesn't seem to be another way in Ruby to do this though.

nicolasbock:
@dirkmueller rebased

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -213,6 +213,48 @@
...
+else
+  endpoint_protocol = node[:keystone][:api][:protocol]
+  endpoint_insecure = node[:keystone][:ssl][:insecure]
+  endpoint_host = my_admin_host
+  endpoint_port = node[:keystone][:api][:admin_port]
+end
+
+Chef::Log.info("current endpoint: #{endpoint_protocol}://" \
+               "#{endpoint_host}:" \
+               "#{endpoint_port}")
```
cmurphy: This could be demoted to a debug log I think?

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -643,12 +681,20 @@
...
-    keystone_settings: KeystoneHelper.keystone_settings(node, @cookbook_name)
+    keystone_settings: keystone_settings
     )
 end
+
+# Set new endpoint URL.
+node.set[:keystone][:api][:internal_url_host] = keystone_settings["internal_url_host"]
+node.save
+Chef::Log.info("setting new endpoint host to " \
+               "#{node[:keystone][:api][:internal_url_host]}")
```
cmurphy: This could also probably be debug

chef/data_bags/crowbar/migrate/keystone/109_add_admin_url.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
```
cmurphy: rebase issue, this file needs to be renamed 110_... and the revision number needs to be re-added to the template

crowbar_framework/app/models/keystone_service.rb
cmurphy: Thanks!

chef/data_bags/crowbar/migrate/keystone/109_add_admin_url.rb
nicolasbock: Fixed. Thanks!

chef/cookbooks/keystone/recipes/server.rb
nicolasbock: done.

chef/cookbooks/keystone/recipes/server.rb
nicolasbock: done

nicolasbock:
Hmm, the `suse/mkcloud` CI job is no reporting back?

crowbar_framework/app/models/keystone_service.rb
Itxaka: I think rails provides a method for deep copy: http://edgeguides.rubyonrails.org/active_support_core_extensions.html#deep-dup

cmurphy:
Needs a rebase again :(

Itxaka:
rebased and removing my review as Im taking over this :D

Itxaka:
Ummm, doesnt seem to work on HA.

Seems like it gets stuck in the `keystone_install_rsync` while deploying the ceph cookbook. Weird that it doesnt fails before on the keystone cookbook, so this may point to a different issue. Seems like the ceph nodes get stuck getting the secret key...

Itxaka:
Does not seem to work due to the check for `bootstrap` on teh resource, which means that it will only trigger the `update_endpoint` on the first run. Changing http to https afterwards does not trigger that resource, which actually breaks the CLI

Itxaka:
Seems like the issue comes from the ceph barclamp including one controller node in the ceph-mon role, thus resulting in only one of the controllers cluster nodes to run the keystone recipe, while the other doesnt.

In that case, its impossible for the sync mark to be...sync across the other nodes, as they are not running the chef client and will never set the mark.

Thus the HA issue is not caused by this PR :D

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -217,6 +217,48 @@
...
+
+Chef::Log.debug("current endpoint: #{endpoint_protocol}://" \
+                "#{endpoint_host}:" \
+                "#{endpoint_port}")
+
+# Update keystone endpoints (in case we switch http/https this will update the
+# endpoints to the correct ones). This needs to be done _before_ we switch
+# protocols on the keystone api.
+keystone_register "update keystone endpoint" do
+  protocol endpoint_protocol
```
jsuchome: Don't we need to have the original endpoint_protocol here?

Itxaka: I would say no due to using the `endpoint_host` for the request path.

Itxaka:
This works on HA but doesnt on non-ha. re-testing.

Itxaka:
re-test this yesterday on non-ha and it seems to work. I pushed a small change to the migration file as it was wrongly named, thus probably causing it to not be applied correctly.

jsuchome:
It does not work when I apply the patch on a system that had keystone already deployed.

Itxaka:
Indeed, after so many tries it seems that if this is deployed in an existing install with keystone already deployed and then the proposal http/https is changed it will fail.

This is due https://github.com/crowbar/crowbar-openstack/pull/859/files#diff-a4bc2321f7569042f7df4a60561b94adR119 not finding the `endpoint` key in the old_role and using the current_role one, which means is using the changed http/https to connect to keystone, no matter what the real value is.

I wonder if this would be possible to be changed on the migration, for existing installations, so the `endpoint` key is already in there with the old values.

Itxaka:
@jsuchome just re-tested this with the new migration and it seems to work.

 - Deployed cloud7 
 - Deployed keystone barclamp with http protocol
 - updated crowbar-openstack with the generated package from this PR
 - restarted crowbar (Very important!)
 - Changed protocol to https
 - ????????
 - profit!


Do you want to retest it yourself?

Also cc @vuntz to check if the migration is proper

chef/data_bags/crowbar/migrate/keystone/112_add_admin_url.rb
```diff
@@ -0,0 +1,29 @@
+def upgrade(ta, td, a, d)
+  a["api"]["internal_url_host"] = ta["api"]["internal_url_host"]
+  # Search for all keystone nodes and update their attributes
+  # based on the current endpoint data.
+  # this prevents an issue where the keystone barclamp is already deployed,
+  # the migration applied, and the http/https changed just after the migration,
+  # making the cookbook ignore that the protocol migth have changed which
```
jsuchome: Typo (*migth*)

chef/data_bags/crowbar/migrate/keystone/112_add_admin_url.rb
```diff
@@ -0,0 +1,29 @@
...
+  # leads to a failure to connect to keystone
+  # with this new attribute, the keystone cookbook will know that the protocol
+  # has changed since the last deployment and will connect to the proper endpoint
+  nodes = NodeObject.find("roles:keystone-server")
+  nodes.each do |node|
+    next if node.normal_attrs["keystone"].nil?
```
jsuchome: Can't it be empty?

Itxaka: mmmh, I dont think so, it either exists or it doesn't I believe, there is nothing that could create the keystone key and leave it empty.

jsuchome: I guess it's good than, I was just curious.

jsuchome:
OK, so that `internal_URL_host` seems like a weird key that is not touched by this PR so we should probably remove it some other time.

toabctl:
@Itxaka sorry, but needs another rebase

Itxaka:
rebased!

cmurphy:
hmm I think this broke the testbuild job:

```
08:19:00 [   42s] Hunk #1 FAILED at 18.
08:19:00 [   42s] 1 out of 1 hunk FAILED -- saving rejects to file chef/cookbooks/keystone/resources/register.rb.rej
08:19:00 [   42s] patching file chef/data_bags/crowbar/migrate/keystone/112_add_admin_url.rb
08:19:00 [   42s] patching file chef/data_bags/crowbar/template-keystone.json
08:19:00 [   42s] patching file chef/data_bags/crowbar/template-keystone.schema
08:19:00 [   42s] patching file chef/roles/keystone-server.rb
08:19:00 [   42s] patching file crowbar_framework/app/models/keystone_service.rb
08:19:00 [   42s] error: Bad exit status from /var/tmp/rpm-tmp.aqlKhV (%prep)
```

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -347,6 +347,63 @@
   end
 end
 
+action :update_endpoint do
+  http, headers = _build_connection(new_resource)
+
+  path = "/v2.0/OS-KSADM/services"
```
cmurphy: It would really be better to use v3 but we can fix that in a followup

Itxaka:
This may be because there is 2 recent patches that modify the keystone/resources/register.rb file, so until the first one (which is already merged) gets into the package for jenkins, the testbuild will fail.

I believe I have seen this before and the solution was to wait until the next day for the package to be updated ¯\\\_(ツ)_/¯

chef/cookbooks/keystone/resources/register.rb
```diff
@@ -18,7 +18,7 @@
 #
 
 actions :add_service, :add_endpoint_template, :add_tenant, :add_domain, :add_domain_role, :add_user,
-        :add_role, :add_access, :add_ec2, :wakeup
+        :add_role, :add_access, :add_ec2, :wakeup, :add_domain_role
```
cmurphy: This looks related to a different PR, maybe a rebase error?

chef/data_bags/crowbar/template-keystone.json
```diff
@@ -32,6 +32,7 @@
         "api_host": "0.0.0.0",
         "admin_port": 35357,
         "admin_host": "0.0.0.0",
+        "internal_url_host": "0.0.0.0",
```
cmurphy: Why internal_url_host and not internal_host? (like admin_host and api_host above)

crowbar_framework/app/models/keystone_service.rb
cmurphy: @Itxaka since you've taken this over do you want to change it to use that?

chef/cookbooks/keystone/resources/register.rb
Itxaka: shiiiii, good catch. I borked the rebase somehow :thinking:

chef/data_bags/crowbar/template-keystone.json
Itxaka: No idea, part of the initial PR which I was not involved in :D

crowbar_framework/app/models/keystone_service.rb
Itxaka: done!

chef/data_bags/crowbar/template-keystone.json
Itxaka: Im guessing it may have something to do with keeping consistency with the `keystone_settings["internal_url_host"]` used in other cookbooks, so its easier to see what this var represents in you check the keystone_settings.

But maybe its easier to go to the source @nicolasbock can you comment on the var name?

chef/data_bags/crowbar/migrate/keystone/112_add_admin_url.rb
```diff
@@ -0,0 +1,29 @@
...
+    node.normal_attrs["keystone"]["endpoint"] = {
+      insecure: a["ssl"]["insecure"],
+      protocol: a["api"]["protocol"],
+      internal_url_host: a["api"]["internal_url_host"],
```
cmurphy: I think this is causing problems for me, or maybe the change in the model is causing it. This updates the node to have node[:keystone][:endpoint] set, which makes the `if node[:keystone].key?(:endpoint)` condition in the keystone server recipe evaluate to true, causing the internal url to be set to 0.0.0.0 and breaking all the barclamps that use the internal url to auth to keystone.

jsuchome: I think other barclamps are not using internal_url_host value directly, they use the one constructed by neutron_settings. (But maybe I misunderstood your comment)

cmurphy: Maybe I misunderstand the cause of the problem but I am seeing http://paste.suse.de/18879 after applying this change when trying to re-run an existing barclamp with no proposal modifications.

jsuchome: Ha! That indeed looks related. It breaks HA setup

cmurphy: I quick fix is to check in the recipe if the endpoint is still set to 0.0.0.0 (we could update the default value to nil to make it cleaner) and if it is, use the admin url like always.

Itxaka: I updated this to make endpoint_host always be CrowbarHelper.get_host_for_admin_url(node, ha_enabled) as it seems that the admin url should always be the same, nothing to do with a protocal change

cmurphy:
> I'm now seeing something different with this one

We think this is a separate issue.

PR #860, opened 2017-03-06
keystone: Enable domain specific configuration

cmurphy:
In SOC6 we had the "hybrid" keystone identity and assignment backend
which allowed LDAP users to authenticate with keystone. Recent versions
of keystone allow domain-specific identity backends to be configured,
essentially providing the same functionality. This patch adds support
for domain specific configuration by adding the parameter
"domain_specific_config", which provides a mapping from domain names to
their [identity] and [ldap] settings, as well as adding a new add_domain
action to keystone_register. Instead of setting the identity and
assignment drivers to "hybrid" and setting LDAP parameters in the main
LDAP section, users should now leave the default drivers as "sql",
enable "domain_specific_drivers", and set the LDAP parameters under
"domain_specific_config" for a domain of their choosing (by default
named "ldap_users"). They also have to enable multi_domain_support in
the horizon barclamp.

LDAP parameters that were removed from keystone in Ocata[1] were not
transferred to the new domain_specific_config section.

This patch does not address migrating from the old "hybrid" backend to
using domain specific identity backends but this can be implemented in a
followup patch.

[1] https://review.openstack.org/#/c/423572/

### DIFF
--- a/chef/cookbooks/keystone/providers/register.rb
+++ b/chef/cookbooks/keystone/providers/register.rb
@@ -84,6 +84,29 @@ action :add_tenant do
   end
 end
 
+# :add_domain specific attributes
+# attribute :tenant_name, :kind_of => String
+action :add_domain do
+  http, headers = _build_connection(new_resource)
+
+  # Construct the path
+  path = "/v3/domains"
+  dir = "domains"
+
+  # Lets verify that the domain does not exist yet
+  item_id, error = _find_id(http, headers, new_resource.domain_name, path, dir)
+  if item_id || error
+    raise "Failed to talk to keystone in add_domain" if error
+    Chef::Log.info "Domain '#{new_resource.domain_name}' already exists. Not creating." unless error
+    new_resource.updated_by_last_action(false)
+  else
+    # Domain does not exist yet
+    body = _build_domain_object(new_resource.domain_name)
+    ret = _create_item(http, headers, path, body, new_resource.domain_name)
+    new_resource.updated_by_last_action(ret)
+  end
+end
+
 # :add_user specific attributes
 # attribute :user_name, :kind_of => String
 # attribute :user_password, :kind_of => String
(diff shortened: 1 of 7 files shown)


cmurphy:
This is not backwards compatible with the "hybrid" driver from SOC6 but I've written up thoughts on how one could migrate from one to the other: https://etherpad.nue.suse.com/p/keystone-hybrid-to-multidomain-migration

cmurphy:
Testing: https://github.com/SUSE-Cloud/automation/pull/1738

cmurphy:
@aplanas Set domain_specific_drivers to true, then in the domain_specific_config section in the ldap_users domain (or set a different domain) set these ldap settings: https://gist.github.com/cmurphy/11bd80d61ab95de5065825ae1b6624c1

You should then be able to list all the users in our LDAP with `openstack user list --domain ldap_users`

aplanas:
[(5) S56P6 Migrate keystone hybrid to multidomain setup](https://trello.com/c/iPVee2r7/289-5-s56p6-migrate-keystone-hybrid-to-multidomain-setup)

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -473,6 +496,18 @@ def _build_tenant_object(tenant_name)
...
+def _build_domain_object(domain_name)
+  svc_obj = {}
+  svc_obj.store("name", domain_name)
+  svc_obj.store("enabled", true)
+  ret = {}
+  ret.store("domain", svc_obj)
+  ret
+end
+
+private
```
aplanas: as I understand this is not needed, isn't? There is a private in L498

cmurphy: It's not, but every other private function here has an extraneous 'private' declaration so I was going for consistency.

aplanas: Right, thanks!

PR #861, opened 2017-03-07
Disable image_cache_manager_interval(bsc#1015324)

sayalilunkad:
Disabling this option as a precaution to avoid the race wherin
the cache_manager removes the images which are actually in use.
This can be enabled by setting to a value greater than -1.

### DIFF
--- a/chef/data_bags/crowbar/migrate/nova/115_add_cache_manager_interval.rb
+++ b/chef/data_bags/crowbar/migrate/nova/115_add_cache_manager_interval.rb
@@ -1,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "image_cache_manager_interval"
+    a["image_cache_manager_interval"] = ta["image_cache_manager_interval"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless ta.key? "image_cache_manager_interval"
+    a.delete("image_cache_manager_interval")
+  end
+  return a, d
+end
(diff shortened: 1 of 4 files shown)


sayalilunkad:
[(5) S56P4 bsc#1015324 - base image deleted before the migration](https://trello.com/c/lRCRzptA/386-5-s56p4-bsc-1015324-base-image-deleted-before-the-migration)

chef/data_bags/crowbar/migrate/nova/115_add_cache_manager_interval.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "image_cache_manager_interval"
+    a["image_cache_manager_interval"] = ta["image_cache_manager_interval"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless ta.key? "image_cache_manager_interval"
```
jsuchome: Wouldn't it safer to check `if a.key?` here?

sayalilunkad: not sure but if you think so I can change it.

jsuchome: It's probably OK, I have not done this for a while

jsuchome:
Are we doing this for master only, or do you plan a fix for Cloud6?

sayalilunkad:
@jsuchome I will backport it to C6 as well

PR #863, opened 2017-03-09
neutron: make rpc_workers attribute configurable

stefannica:
The rpc_workers neutron.conf configurable parameter was introduced in OpenStack Icehouse. It controls the number of RabbitMQ/ZeroMQ RPC listener processes spawned by neutron-server (one by default).

More info:
   * OS blueprint: https://blueprints.launchpad.net/neutron/+spec/multiple-rpc-workers
   * subsequent change in default value (from 0 to 1): https://bugs.launchpad.net/neutron/+bug/1463129
   * official Mitaka documentation: https://docs.openstack.org/mitaka/config-reference/networking/networking_options_reference.html#configure-messaging

### DIFF
--- a/chef/data_bags/crowbar/template-neutron.json
+++ b/chef/data_bags/crowbar/template-neutron.json
@@ -11,6 +11,7 @@
       "verbose": true,
       "create_default_networks": true,
       "dhcp_domain": "openstack.local",
+      "rpc_workers": 1,
       "use_lbaas": true,
       "lbaasv2_driver": "haproxy",
       "use_l2pop": false,
@@ -121,7 +122,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 107,
+      "schema-revision": 108,
       "element_states": {
         "neutron-server": [ "readying", "ready", "applying" ],
         "neutron-network": [ "readying", "ready", "applying" ]
--- a/chef/data_bags/crowbar/template-neutron.schema
+++ b/chef/data_bags/crowbar/template-neutron.schema
@@ -16,6 +16,7 @@
                     "keystone_instance": { "type": "str", "required": true },
                     "create_default_networks": { "type": "bool", "required": true },
                     "dhcp_domain": { "type": "str", "required": true },
+                    "rpc_workers": { "type": "int", "required": true },
                     "use_lbaas": { "type": "bool", "required": true },
                     "lbaasv2_driver": { "type": "str", "required": true },
                     "use_l2pop": { "type": "bool", "required": true },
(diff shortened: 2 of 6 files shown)


stefannica:
Closes [sap-oc#14](https://github.com/sap-oc/crowbar-openstack/issues/14)

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -11,6 +11,7 @@
       "verbose": true,
       "create_default_networks": true,
       "dhcp_domain": "openstack.local",
+      "rpc_workers": 1,
```
vuntz: Any reason you're changing the default, compared to what we had before? Probably best to put 0 as default value.

stefannica: I did this to reflect the current official OpenStack default value for this attribute, which has been changed in OS Liberty from 0 to 1. The explanation has something to do with the fact that at least one RPC worker needs to be running.

Do you think we should we keep the 0 default ? neutron-server will ignore 0 and use 1 anyway, so maybe also using 1 as a default makes more sense.

chef/data_bags/crowbar/template-neutron.schema
```diff
@@ -16,6 +16,7 @@
                     "keystone_instance": { "type": "str", "required": true },
                     "create_default_networks": { "type": "bool", "required": true },
                     "dhcp_domain": { "type": "str", "required": true },
+                    "rpc_workers": { "type": "int", "required": false },
```
toabctl: Why is it not required? The migration is adding it anyway and you use it in the config template..

stefannica: This was based on my (faulty) assumption that the attribute is not required if it has a default value. I will switch it back to true.

chef/data_bags/crowbar/template-neutron.json
s-t-e-v-e-n-k: @vuntz, Neutron since 7.0.0 has overridden rpc_workers < 1 to be 1, so I agree with @stefannica that the default should be 1.

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -121,7 +122,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 107,
+      "schema-revision": 108,
```
toabctl: We definitly need a rebase. master has already revison 116 so this needs to be changed (and the migration filename, too)

stefannica: @toabctl I rebased the branch, but the neutron revision and migration filename seem to still be valid in master.

stefannica:
Rebased. Hopefully, the HA job will pass now.

chef/data_bags/crowbar/template-neutron.json
toabctl: oh. I looked at nova. Pardon...

PR #865, opened 2017-03-13
cinder: set backups_enabled to false

sidthe:
Operators that do not have Cinder backup service deployed in their cloud should set this option to False.


sidthe:
[(1) heat: Set option to disable cinder backups](https://trello.com/c/Yvhm16zF/2030-1-heat-set-option-to-disable-cinder-backups)

PR #867, opened 2017-03-15
magnum: Use correct values for keystone authentication

toabctl:
This removes the deprecation warning

WARNING keystonemiddleware.auth_token [-] Use of the auth_admin_prefix, \
        auth_host, auth_port, auth_protocol, identity_uri, admin_token, \
        admin_user, admin_password, and admin_tenant_name configuration \
        options was deprecated in the Mitaka release in favor of an \
        auth_plugin and its related options. This class may be removed in \
        a future release.

when starting the magnum-api service.

### DIFF
--- a/chef/cookbooks/magnum/templates/default/magnum.conf.erb
+++ b/chef/cookbooks/magnum/templates/default/magnum.conf.erb
@@ -46,12 +46,14 @@ insecure = <%= @keystone_settings['insecure'] %>
 auth_uri = <%= @keystone_settings['public_auth_url'] %>
 auth_url = <%= @keystone_settings['internal_auth_url'] %>
 auth_version = <%= @keystone_settings['api_version_for_middleware'] %>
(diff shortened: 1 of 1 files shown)


dirkmueller:
its not the only barclamp that is affected by that, right? I think this warning is also printed by others.. can you check them as well?

toabctl:
> its not the only barclamp that is affected by that, right? I think this warning is also printed by others.. can you check them as well?

@dirkmueller I can. But can I do that in another PR then? Or would you prefer to get this in in a single PR?

dirkmueller:
> @dirkmueller I can. But can I do that in another PR then? Or would you prefer to get this in in a single PR?

@toabctl single PR definitely preferred, but note I didn't -1 this so I'm not objecting to get this one only merged.

dirkmueller:
thanks to steven for taking that over..

PR #868, opened 2017-03-15
nova: Fix ipaddress attribute

sayalilunkad:


chef/cookbooks/nova/attributes/default.rb
```diff
@@ -86,7 +86,7 @@
 # Shared Settings
 #
 default[:nova][:hostname] = "nova"
-default[:nova][:my_ip] = ipaddress
+default[:nova][:my_ip] = node["ipaddress"]
```
nicolasbock: @sayalilunkad Why is this an issue now? The code has been there for quite a while.

dirkmueller:
hmm, that basically means the config.rb wasn't executed beforehand.. which is I think the issue we should fix instead, as this is now hiding the bug (assigning the ohai ipaddress attribute which could be anything but not necessarily the admin net ip).

I think this papers over a problem that will then now be very hard to find. I'm not objecting but I'm not convinced we should merge this.

cmurphy:
This seems relevant: http://tickets.opscode.com/browse/OHAI-87

toabctl:
Closing in favor of https://github.com/crowbar/crowbar-openstack/pull/876

PR #869, opened 2017-03-16
Disable cinder snapshots in tempest when using the nfs backend

rhafer:
As of now (newton) the nfs backend in cinder does not support snapshots.


rhafer:
[S57P4 (5) Adapt mkcloud / virtual upgrade jobs to setup and use nfs server for cinder](https://trello.com/c/202p1JSP/360-s57p4-5-adapt-mkcloud-virtual-upgrade-jobs-to-setup-and-use-nfs-server-for-cinder)

PR #870, opened 2017-03-17
Fix BSC#1029442

aplanas:
horizon: create a dummy service for reload (bsc#1029442)

Apache reload (graceful) will signal the apache process, that will
restart the child process (WSGI).  This can cause a race, were an
OpenStack client try to contact to one of the WSGI services, and
the old WSGI process can answer.  This process will fail because
the WSGI socket is not anymore in the file system.

The aproach here is to create a dummy service for 'horizon', what
will SIGUSR1 the wsgi:horizon childs from apache, signaling the
reload of the configuration file.

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -27,6 +27,18 @@ if node[:horizon][:apache][:ssl]
   include_recipe "apache2::mod_ssl"
 end
 
+# Fake service to take control of the WSGI process from apache that
+# runs Horizon.  We replace the `reload` action, sending manually the
+# signal SIGUSR1 to all the process that are part of `wsgi:horizon`
+service "horizon" do
+  service_name "apache2"
+  if node[:platform_family] == "suse"
+    reload_command "pkill --signal SIGUSR1 -f '(wsgi:horizon)' && sleep 1"
+  end
+  supports reload: true, restart: true, status: true
+  ignore_failure true
+end
+
 case node[:platform_family]
 when "debian"
   # Explicitly added client dependencies for now.
@@ -54,7 +66,7 @@ else
   unless node[:horizon][:site_theme].empty?
     package "openstack-dashboard-theme-#{node[:horizon][:site_theme]}" do
       action :install
-      notifies :reload, resources(service: "apache2")
+      notifies :reload, "service[horizon]"
     end
   end
 end
@@ -73,7 +85,7 @@ unless neutron_lbaas_ui_pkgname.nil?
   unless neutron_servers.empty?
     package neutron_lbaas_ui_pkgname do
       action :install
-      notifies :reload, resources(service: "apache2")
+      notifies :reload, "service[horizon]"
     end
   end
 end
@@ -92,7 +104,7 @@ unless manila_ui_pkgname.nil?
   unless manila_servers.empty?
     package manila_ui_pkgname do
       action :install
-      notifies :reload, resources(service: "apache2")
+      notifies :reload, "service[horizon]"
     end
   end
 end
@@ -111,7 +123,7 @@ unless magnum_ui_pkgname.nil?
   unless magnum_servers.empty?
     package magnum_ui_pkgname do
       action :install
-      notifies :reload, resources(service: "apache2")
+      notifies :reload, "service[horizon]"
     end
   end
 end
@@ -130,7 +142,7 @@ unless trove_ui_pkgname.nil?
   unless trove_servers.empty?
     package trove_ui_pkgname do
       action :install
-      notifies :reload, resources(service: "apache2")
+      notifies :reload, "service[horizon]"
     end
   end
 end
@@ -149,7 +161,7 @@ unless sahara_ui_pkgname.nil?
   unless sahara_servers.empty?
     package sahara_ui_pkgname do
       action :install
-      notifies :reload, resources(service: "apache2")
+      notifies :reload, "service[horizon]"
     end
   end
 end
--- a/chef/cookbooks/horizon/templates/default/openstack-dashboard.conf.erb
+++ b/chef/cookbooks/horizon/templates/default/openstack-dashboard.conf.erb
@@ -19,7 +19,7 @@ RewriteRule / https://%{HTTP_HOST}%{REQUEST_URI} [L,R]
 <% else %>
 <VirtualHost <%= @bind_host %>:<%= @bind_port %>>
 <% end %>
-    WSGIDaemonProcess horizon user=<%= @user %> group=<%= @group %> processes=3 threads=10 home=<%= @horizon_dir %> 
+    WSGIDaemonProcess horizon user=<%= @user %> group=<%= @group %> processes=3 threads=10 home=<%= @horizon_dir %>  display-name=%{GROUP}
     WSGIScriptAlias / <%= @horizon_dir %>/openstack_dashboard/wsgi/django.wsgi
     SetEnv APACHE_RUN_USER  <%= @user %>
     SetEnv APACHE_RUN_GROUP <%= @group %>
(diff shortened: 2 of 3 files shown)


stefannica:
[bsc#1020922 - L3-Question: Limits on OpenStack services and pacemaker](https://trello.com/c/S2GzSg95/495-bsc-1020922-l3-question-limits-on-openstack-services-and-pacemaker)

stefannica:
[bsc#1029442 - mod_wsgi (pid=5120): Unable to connect to WSGI daemon process 'keystone-admin' on '/run/wsgi.1158.2.2.sock' as user with uid=30](https://trello.com/c/HoKMdqRw/493-bsc-1029442-mod-wsgi-pid-5120-unable-to-connect-to-wsgi-daemon-process-keystone-admin-on-run-wsgi-1158-2-2-sock-as-user-with-uid)

aplanas:
* Delegate the new service to the apache2 one

aplanas:
The HA fail seems unrelated. Is a recipe compile error in nova when deploying ceilometer.

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -27,6 +27,17 @@
   include_recipe "apache2::mod_ssl"
 end
 
+service "horizon" do
+  service_name "apache2"
```
stefannica: I'm curious, is there a reason why you need to override the service name to "apache2" ? is there a trick to it ?

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -27,6 +27,17 @@
   include_recipe "apache2::mod_ssl"
 end
```
stefannica: could you please add a small comment here explaining the need for this service resource ?

chef/cookbooks/horizon/recipes/server.rb
aplanas: Actually yes. service is using systemd to control the services, and 'horizon' is not a valid service. This means that if I try to reload the service, this will never succeed, because `systectl is-active` will return `false` for horizon, but true for apache2. Also is a way to delegate the rest of the commands (restart, status, etc)

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -54,7 +65,7 @@
   unless node[:horizon][:site_theme].empty?
     package "openstack-dashboard-theme-#{node[:horizon][:site_theme]}" do
       action :install
-      notifies :reload, resources(service: "apache2")
+      notifies :reload, "service[horizon]"
```
aplanas: is what is documented here: https://docs.chef.io/resource_service.html

chef/cookbooks/horizon/recipes/server.rb
aplanas: done

stefannica: this is great, thanks !

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -111,7 +125,7 @@
   unless magnum_servers.empty?
     package magnum_ui_pkgname do
       action :install
-      notifies :reload, resources(service: "apache2")
+      notifies :reload, "service[horizon]"
```
stefannica: this may be nothing or it may be a rebase mishap: the trove reload notification line below (line 143) still references the "apache2" service:
```ruby
   trove_servers = search(:node, "roles:trove-server") || []
   unless trove_servers.empty?
     package trove_ui_pkgname do
       action :install
     end
   end
 end
```

chef/cookbooks/horizon/recipes/server.rb
stefannica: this is better, thanks !

chef/cookbooks/horizon/recipes/server.rb
aplanas: Ah true. Fixed.

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -27,6 +27,20 @@
   include_recipe "apache2::mod_ssl"
 end
 
+# Fake service to take control of the WSGI process from apache that
+# runs Horizon.  We replace the `reload` action, sending manually the
+# signal SIGUSR1 to all the process that are part of `wsgi:horizon`
+service "horizon" do
+  service_name "apache2"
+  if node[:platform_family] == "suse"
+    reload_command "/usr/bin/kill -s SIGUSR1 $(ps ax | sed 's/^ *//;s/ *$//' "\
```
toabctl: I might miss something here, but why are we not just executing this kill command with a subscribes to the apache service? Something like:
```
execute 'sigusr1 for wsgi processes' do
  command '...'
  subscribes :reload, resources(service: "apache2")
end
```

Not sure if that works and I have not tested it.

aplanas: I do not want to change an external cookbook that we will, eventually, update from upstream (apache), but also your proposal will not work, because it will reload the apache service that what is generating the bug.

toabctl: ups. deleted my comment. sorry

toabctl: @aplanas it could be added to an extra file. Not sure if we ever sync with upstream.
And why is it reloading the apache service?

aplanas: I updated the apache cookbook to add the WSGI LWRP code, but was reviewed to avoid this situation. But adding the code in a different file will be cool.

If I am not doing any mistake, this is not reloading apache, because replace the reload_command with something different, at least in the SUSE case.

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -27,6 +27,20 @@
...
+                   "| grep wsgi:horizon | grep -v grep | cut -d' ' -f1) "\
```
dirkmueller: geez, this is ultra scary. why not use 

```
systemctl kill -s SIGUSR1 apache
``` 
instead?

aplanas: No, this full restart the WSGI process (all of them), recreate the sockets and we go back to the original error.

dirkmueller: @aplanas so the thing is that we need to only restart the wsgi processes but not apache itself? thats not very sustainable.. wow.

aplanas: Right, this patch only restart the subprocess from apache that takes care for horizon.

The upstream bug reported in bsc#1029442 says that the socket management from mod_wsgi will change eventually

dirkmueller: @aplanas I see, thanks.

cmurphy:
The HA failure is 

```
19:36:18 +(qa_crowbarsetup.sh:3935) oncontroller_testsetup(): crm_mon --failcounts -1
19:36:18    apache2: migration-threshold=3 fail-count=1 last-failure='Tue Mar 28 17:42:45 2017'
```

which seems quite related to me.

aplanas:
@cmurphy I am not sure. If the time is correct at `17:42:45` was installing keystone:

```
17:40:37 +(qa_crowbarsetup.sh:3149) update_one_proposal(): crowbar keystone proposal commit default
17:42:53 Committed default
```

This is way before this code form Horizon was executed for first time

```
18:13:35 +(qa_crowbarsetup.sh:3176) do_one_proposal(): crowbar horizon proposal create default
18:13:36 Created default
```

chef/cookbooks/horizon/recipes/server.rb
aplanas: I like a lot, I will use this. Thanks!

aplanas: done

PR #872, opened 2017-03-21
ec2-api: Add SSL support

s-t-e-v-e-n-k:
Add support to deploy ec2-api with SSL enabled endpoints, optionally
generating self-signed certificates.

Co-Authored-By: Steve Kowalik <steven@wedontsleep.org>

This replaces https://github.com/crowbar/crowbar-openstack/pull/773 .

### DIFF
--- a/chef/cookbooks/ec2-api/recipes/ec2api.rb
+++ b/chef/cookbooks/ec2-api/recipes/ec2api.rb
@@ -23,10 +23,9 @@ package "openstack-ec2-api-s3"
 
 # NOTE: ec2 is deployed via the nova barclamp
 ha_enabled  = node[:nova]["ec2-api"][:ha][:enabled]
-# SSL support is for the entire barclamp
-ssl_enabled = node[:nova][:ssl][:enabled]
+ssl_enabled = node[:nova]["ec2-api"][:ssl][:enabled]
+api_protocol = ssl_enabled ? "https" : "http"
 db_settings = fetch_database_settings "nova"
-api_protocol = node[:nova][:ssl][:enabled] ? "https" : "http"
 ec2_api_port = node[:nova][:ports][:ec2_api]
 ec2_metadata_port = node[:nova][:ports][:ec2_metadata]
 if ha_enabled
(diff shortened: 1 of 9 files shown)


nicolasbock:
@s-t-e-v-e-n-k I should have asked whether you tried to deploy this. Have you?

s-t-e-v-e-n-k:
I have successfully deployed this on vp4. The reason for my branch over Sayali's is that this one won't behave unexpectedly if you set ec2-api to HTTPS and nova to HTTP (although, why you would is beyond me)

nicolasbock:
@s-t-e-v-e-n-k thanks!

dirkmueller:
@s-t-e-v-e-n-k I would have preferred to continue the discussion in the original PR instead of forking it, and thereby loosing all review history..

chef/cookbooks/ec2-api/recipes/ec2api.rb
```diff
@@ -23,10 +23,9 @@
 
 # NOTE: ec2 is deployed via the nova barclamp
 ha_enabled  = node[:nova]["ec2-api"][:ha][:enabled]
-# SSL support is for the entire barclamp
-ssl_enabled = node[:nova][:ssl][:enabled]
```
dirkmueller: This is used in other places within this file as well, and that means you've now changed ssl enablement from the global flag to the ec2-api flag. that should be reverted.

chef/data_bags/crowbar/template-nova.json
```diff
@@ -42,6 +42,15 @@
...
           "database": "ec2"
+        },
+        "ssl": {
+          "enabled": false,
+          "certfile": "/etc/ec2api/ssl/certs/signing_cert.pem",
+          "keyfile": "/etc/ec2api/ssl/certs/signing_key.pem",
+          "generate_certs": false,
+          "insecure": false,
+          "cert_required": false,
+          "ca_certs": "/etc/barbican/ssl/certs/ca.pem"
```
dirkmueller: barbican?

chef/cookbooks/ec2-api/recipes/ec2api.rb
s-t-e-v-e-n-k: Global setting? ec2-api can not use nova's ssl certificates due to file permissions, so why should it make use of nova's ssl setting, when we could potentially be in the situation where nova is deployed  with SSL enabled and ec2-api is not.

chef/data_bags/crowbar/template-nova.json
s-t-e-v-e-n-k: Good catch.

chef/cookbooks/ec2-api/recipes/ec2api.rb
dirkmueller: indeed, this behavior change is actually making it better, retracting my comment

dirkmueller:
[(5) S57P2 S56P9 (S55P8) (S51P11 S50P12 S49P14) Re-add Nova EC2 API that got removed in Mitaka and added as new project](https://trello.com/c/x51hX6aZ/269-5-s57p2-s56p9-s55p8-s51p11-s50p12-s49p14-re-add-nova-ec2-api-that-got-removed-in-mitaka-and-added-as-new-project)

PR #873, opened 2017-03-21
rabbitmq: Add SSL support

cmurphy:
Add SSL support to rabbitmq barclamp

This patch adds the option to configure some of the SSL settings for
rabbitmq and ensure all OpenStack services are using the SSL-secured
port. There are some important notes regarding how this differs from
enabling SSL for the REST APIs:
    
- The SSL port is configurable, since typically rabbitmq has different
   ports designated for regular and SSL traffic and it would be confusing
   to only allow one port.
- The ceilometer pipeline for swift does not have the ability to read
   the oslo.messaging SSL-specific options from oslo.config[1]. Rather
   than send ceilometer traffic over an unsecured channel, we disable
   this pipeline when rabbitmq is enabled.
- Disabling the "insecure" option requires the client services to
   configure a CA to validate the rabbitmq server certificate, which must
   have the same path on all clients. There is no way to tell the client
   services to require a certificate be validated against the
   system-trusted CAs.
- The cert_required option is configurable, but it currently does not do
   the work of setting up client certs so it is somewhat toothless.
 
[1] https://bugs.launchpad.net/ceilometermiddleware/+bug/1673738

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -123,15 +123,28 @@ class CrowbarOpenStackHelper
       if rabbit.nil?
         Chef::Log.warn("No RabbitMQ server found!")
       else
+        if rabbit[:rabbitmq][:ssl][:enabled]
+          port = rabbit[:rabbitmq][:ssl][:port]
+          use_ssl = true
+        else
+          port = rabbit[:rabbitmq][:port]
+          use_ssl = false
+        end
+        client_ca_certs = if rabbit[:rabbitmq][:ssl][:enabled] && \
+            !rabbit[:rabbitmq][:ssl][:insecure]
+          rabbit[:rabbitmq][:ssl][:client_ca_certs]
+        end
         @rabbitmq_settings[instance] = {
           address: rabbit[:rabbitmq][:address],
-          port: rabbit[:rabbitmq][:port],
+          port: port,
           user: rabbit[:rabbitmq][:user],
           password: rabbit[:rabbitmq][:password],
           vhost: rabbit[:rabbitmq][:vhost],
+          use_ssl: use_ssl,
+          client_ca_certs: client_ca_certs,
           url: "rabbit://#{rabbit[:rabbitmq][:user]}:" \
             "#{rabbit[:rabbitmq][:password]}@" \
-            "#{rabbit[:rabbitmq][:address]}:#{rabbit[:rabbitmq][:port]}/" \
+            "#{rabbit[:rabbitmq][:address]}:#{port}/" \
             "#{rabbit[:rabbitmq][:vhost]}"
         }
 
--- a/chef/cookbooks/glance/templates/default/glance-api.conf.erb
+++ b/chef/cookbooks/glance/templates/default/glance-api.conf.erb
@@ -84,6 +84,12 @@ lock_path = /var/run/glance
 [oslo_messaging_notifications]
 driver = messaging
 
+[oslo_messaging_rabbit]
+rabbit_use_ssl = <%= @rabbit_settings[:use_ssl] %>
+<% if @rabbit_settings[:client_ca_certs] -%>
+kombu_ssl_ca_certs = <%= @rabbit_settings[:client_ca_certs] %>
+<% end -%>
+
 [paste_deploy]
 <% if node[:glance][:enable_caching] and node[:glance][:use_cachemanagement] -%>
 flavor = keystone+cachemanagement
--- a/chef/cookbooks/trove/recipes/conductor.rb
+++ b/chef/cookbooks/trove/recipes/conductor.rb
@@ -30,7 +30,8 @@ template node[:trove][:conductor][:config_file] do
   variables(
     keystone_settings: KeystoneHelper.keystone_settings(node, :trove),
     sql_connection: sql_connection,
-    rabbit_trove_url: rabbit_trove_url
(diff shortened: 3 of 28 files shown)


chef/cookbooks/trove/recipes/conductor.rb
```diff
@@ -30,7 +30,8 @@
   variables(
     keystone_settings: KeystoneHelper.keystone_settings(node, :trove),
     sql_connection: sql_connection,
-    rabbit_trove_url: rabbit_trove_url
+    rabbit_trove_url: rabbit_trove_url,
+    rabbit_default_settings: fetch_rabbitmq_settings,
```
cmurphy: ugh really :(

cmurphy:
My earlier assumptions about insecure mode were wrong, I've corrected it but the configuration is a little more complicated now

nicolasbock:
Hmm, trove is not happy.

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -123,15 +123,27 @@ def self.rabbitmq_settings(node, barclamp)
       if rabbit.nil?
         Chef::Log.warn("No RabbitMQ server found!")
       else
+        if rabbit[:rabbitmq][:ssl][:enabled]
+          port = rabbit[:rabbitmq][:ssl][:port]
+          use_ssl = true
+          insecure = rabbit[:rabbitmq][:ssl][:insecure]
+          client_ca_certs = insecure ? nil : rabbit[:rabbitmq][:ssl][:client_ca_certs]
```
cmurphy: I really feel like this will be less readable if I try to reorganize the code to avoid this nesting :(

cmurphy:
trove should be happier now, there were some copypaste errors

nicolasbock:
@cmurphy nice! The HA failures are caused by what #868 is trying to address.

sayalilunkad:
@cmurphy is there a reason for not exposing the SSL settings in the UI?

cmurphy:
@sayalilunkad they should be exposed: https://github.com/crowbar/crowbar-openstack/pull/873/files#diff-5ac9ea0391298c99aea873751910b146

sayalilunkad:
@cmurphy ah yes. I needed to restart crowbar to see them. Thanks

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -123,15 +123,28 @@ def self.rabbitmq_settings(node, barclamp)
...
+          rabbit[:rabbitmq][:ssl][:client_ca_certs]
+        end
         @rabbitmq_settings[instance] = {
           address: rabbit[:rabbitmq][:address],
-          port: rabbit[:rabbitmq][:port],
+          port: port,
           user: rabbit[:rabbitmq][:user],
           password: rabbit[:rabbitmq][:password],
           vhost: rabbit[:rabbitmq][:vhost],
+          use_ssl: use_ssl,
```
dirkmueller: NIT: we could use ```rabbit[:rabbitmq][:ssl][:enabled]``` here, removing the need for two lines of use_ssl setting above.

cmurphy: I'll submit another change to clean that up

chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -33,7 +44,7 @@
     "rabbitmq": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 11,
+      "schema-revision": 12,
```
vuntz: Should have been bumped to 100.

chef/cookbooks/glance/templates/default/glance-api.conf.erb
```diff
@@ -84,6 +84,12 @@ lock_path = /var/run/glance
 [oslo_messaging_notifications]
 driver = messaging
 
+[oslo_messaging_rabbit]
+rabbit_use_ssl = <%= @rabbit_settings[:use_ssl] %>
+<% if @rabbit_settings[:client_ca_certs] -%>
+kombu_ssl_ca_certs = <%= @rabbit_settings[:client_ca_certs] %>
+<% end -%>
```
vuntz: This change is missing in glance-registry.conf.

chef/cookbooks/trove/templates/default/trove.conf.erb
```diff
@@ -65,3 +65,9 @@ identity_uri = <%= @keystone_settings['admin_auth_url'] %>
 admin_user = <%= @keystone_settings['service_user'] %>
 admin_password = <%= @keystone_settings['service_password'] %>
 admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
+
+[oslo_messaging_rabbit]
+rabbit_use_ssl = <%= @rabbit_default_settings[:use_ssl] %>
+<% if @rabbit_default_settings[:client_ca_certs] -%>
+kombu_ssl_ca_certs = <%= @rabbit_default_settings[:client_ca_certs] %>
```
vuntz: Nearly sure that `@rabbit_default_settings` is not defined here.

cmurphy: It is defined here: https://github.com/crowbar/crowbar-openstack/blob/56b009c337b38a8bbd59551bb6790f6c1670aad1/chef/cookbooks/trove/recipes/api.rb#L145

PR #875, opened 2017-03-21
upgrade: Check for remote nodes platform version before creating virtlogd primitive

jsuchome:
Do not create virtlogd primitive for remote resource if all nodes are not yet upgraded to SLES12SP2.


jsuchome:
[(8) S49P16 Orchestrate cloud upgrade with compute-ha setup](https://trello.com/c/2qIRkJMO/2060-8-s49p16-orchestrate-cloud-upgrade-with-compute-ha-setup)

PR #878, opened 2017-03-23
neutron: avoid adding a g-neutron group (bsc#1030318)

dirkmueller:
We can not move primitives in/out a group in a maintenance update.
As we don't actually want groups anymore, just remove the bit from

https://github.com/crowbar/crowbar-openstack/pull/588

### DIFF
--- a/chef/cookbooks/neutron/recipes/server_ha.rb
+++ b/chef/cookbooks/neutron/recipes/server_ha.rb
@@ -42,29 +42,9 @@ end
 transaction_objects << "pacemaker_primitive[#{server_primitive_name}]"
 primitives << server_primitive_name
 
-if node[:neutron][:use_infoblox]
-  infoblox_primitive_name = "infoblox-agent"
-  pacemaker_primitive infoblox_primitive_name do
-    agent node[:neutron][:ha][:infoblox][:infoblox_ra]
-    op node[:neutron][:ha][:infoblox][:op]
-    action :update
-    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
-  end
-  transaction_objects << "pacemaker_primitive[#{infoblox_primitive_name}]"
-  primitives << infoblox_primitive_name
-end
-
-group_name = "g-neutron"
-pacemaker_group group_name do
-  members primitives
-  action :update
-  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
-end
-transaction_objects << "pacemaker_group[#{group_name}]"
-
-clone_name = "cl-#{group_name}"
+clone_name = "cl-#{server_primitive_name}"
 pacemaker_clone clone_name do
-  rsc group_name
+  rsc server_primitive_name
   meta ({
     "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node),
     "interleave" => "true",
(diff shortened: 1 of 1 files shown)


dirkmueller:
[#bsc1030318 Infoblox HA breakage on existing installations](https://trello.com/c/z5jZ0ecW/496-bsc1030318-infoblox-ha-breakage-on-existing-installations)

dirkmueller:
So the mkcloud/hacloud=1 failure is

```
Failed Actions:
* vip-admin-database-default-data_monitor_10000 on d52-54-77-77-01-01 'not running' (7): call=32, status=complete, exitreason='none',
    last-rc-change='Thu Mar 23 22:21:18 2017', queued=0ms, exec=0ms
* vip-admin-rabbitmq-default-data_monitor_10000 on d52-54-77-77-01-01 'not running' (7): call=55, status=complete, exitreason='none',
    last-rc-change='Thu Mar 23 22:21:15 2017', queued=0ms, exec=0ms
```

which appears unrelated to this change (unless I'm missing something obvious)

cmurphy:
Correct, that failure is not related, we're trying to address it with https://github.com/SUSE-Cloud/automation/pull/1802

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -52,27 +64,19 @@
...
-clone_name = "cl-#{group_name}"
-pacemaker_clone clone_name do
-  rsc group_name
-  meta ({
-    "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node),
-    "interleave" => "true",
-  })
-  action :update
-  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  clone_name = "cl-#{infoblox_primitive_name}"
```
rhafer: Doesn't this break the creation of the ```crowbar_pacemaker_order_only_existing "o-#{clone_name}"``` ordering constraint below (line 91)?

BTW, I guess the infoblox resource needs a separate ordering constraint now. At least that's what we do in master.

rhafer:
> Correct, that failure is not related, we're trying to address it with SUSE-Cloud/automation#1802

@cmurphy I just created https://bugzilla.suse.com/show_bug.cgi?id=1030822 to track the underlying problem. (I think adding a workaround in mkcloud as you do is good as a quick fix, but longer term we'll probably need a real solution)

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -42,6 +42,18 @@
...
+pacemaker_clone clone_name do
+  rsc server_primitive_name
+  meta ({
+    "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node),
+    "interleave" => "true",
+  })
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects << "pacemaker_clone[#{clone_name}]"
```
vuntz: We will need a call to `openstack_pacemaker_controller_only_location_for` for this resource.

chef/cookbooks/neutron/recipes/server_ha.rb
vuntz: Correct.

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -85,7 +90,7 @@
 end
 
 crowbar_pacemaker_order_only_existing "o-#{clone_name}" do
-  ordering ["postgresql", "rabbitmq", "cl-keystone", clone_name]
+  ordering ["postgresql", "rabbitmq", "cl-keystone", clone_name, infoblox_clone_name]
```
rhafer: I don't think this works, when the ```infoblox_clone_name```resource does not exist (i.e.   when ```node[:neutron][:use_infoblox]``` is ```false```), does it?

dirkmueller: @rhafer I think ```crowbar_pacemaker_order_only_existing``` automatically removes constraints that do not exist on that cluster (because e.g. they're on another cluster)

rhafer: Hm, true. Sorry I missed that.

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -77,6 +57,31 @@
...
+  pacemaker_clone infoblox_clone_name do
+    rsc infoblox_primitive_name
+    meta ({
+      "clone-max" => CrowbarPacemakerHelper.num_corosync_nodes(node),
+      "interleave" => "true",
+    })
+    action :update
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+  transaction_objects << "pacemaker_clone[#{infoblox_clone_name}]"
```
rhafer: You're still missing the ```openstack_pacemaker_controller_only_location_for``` for the infoblox_clone resource as mentioned in @vuntz other comment.

dirkmueller: @rhafer yep, thanks. done

chef/cookbooks/neutron/recipes/server_ha.rb
```diff
@@ -42,29 +42,9 @@
 transaction_objects << "pacemaker_primitive[#{server_primitive_name}]"
 primitives << server_primitive_name
```
vuntz: For the record, you can kill the use of `primitives` in the file.

PR #880, opened 2017-03-28
upgrade: Check for run_list_map when searching for ceph nodes.

jsuchome:
When ceph node is prepared for the upgrade, it cannot be find by
the search for roles, but search for run_list_map still works.

This is actually consistent with the (upgrade unrelated) search method
that is already used in ceph recipes.

### DIFF
--- a/chef/cookbooks/glance/recipes/ceph.rb
+++ b/chef/cookbooks/glance/recipes/ceph.rb
@@ -18,7 +18,7 @@
 #
 
 ceph_env_filter = " AND ceph_config_environment:ceph-config-default"
-ceph_servers = search(:node, "roles:ceph-osd#{ceph_env_filter}") || []
+ceph_servers = search(:node, "run_list_map:ceph-osd#{ceph_env_filter}") || []
 if ceph_servers.length > 0
   include_recipe "ceph::keyring"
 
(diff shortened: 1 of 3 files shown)


Itxaka:
Do we have an explanation of why is that? Do we remove the roles when preparing the nodes for them when we prepare them for the upgrade?

Just curious.

tserong:
Somehow if nodes aren't up, the role search fails, but the run_list_map search still works fine (see https://github.com/crowbar/crowbar-ceph/pull/47)

Itxaka:
> Somehow if nodes aren't up, the role search fails, but the run_list_map search still works fine (see crowbar/crowbar-ceph#47)

Very interesting, thanks @tserong !

Itxaka:
not that fast!

```
[2017-03-28T08:45:38+00:00] INFO: HTTP Request Returned 400 Bad Request: invalid search query: 'run_list_map::ceph-osd AND ceph_config_environment:ceph-config-default' Parse error at offset: 12 Reason: Expected one of AND, OR, NOT, ?, \, ", (, [, { at line 1, column 14 (byte 14) after :
```

chef/cookbooks/glance/recipes/ceph.rb
```diff
@@ -18,7 +18,7 @@
 #
 
 ceph_env_filter = " AND ceph_config_environment:ceph-config-default"
-ceph_servers = search(:node, "roles:ceph-osd#{ceph_env_filter}") || []
+ceph_servers = search(:node, "run_list_map::ceph-osd#{ceph_env_filter}") || []
```
Itxaka: extra `:` in the search query!

jsuchome: good!

PR #881, opened 2017-03-28
[cinder]Support to remove dependence of volume on its snapshot

sjamgade:


chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -207,6 +207,9 @@ volume_driver = cinder.volume.drivers.lvm.LVMVolumeDriver
   <% end -%>
 
   <% if volume['backend_driver'] == 'rbd' -%>
+    <% if volume['rbd']['flatten_volume_from_snapshot']  -%>
```
aplanas: extra space here at the end

chef/data_bags/crowbar/template-cinder.schema
```diff
@@ -246,7 +246,8 @@
                       "admin_keyring": { "type": "str", "required": true },
                       "pool": { "type": "str", "required": true },
                       "user": { "type": "str", "required": true },
-                      "secret_uuid" : { "type": "str", "required": true }
+                      "secret_uuid" : { "type": "str", "required": true },
+                      "flatten_volume_from_snapshot": { "type": "bool", "required": false }
```
aplanas: why is not required? I wonder if is not better to do a required one, add the migration and set as a default value 'false' here

dirkmueller:
@sjamgade I think the style of the commit message summary should be ```cinder: ...``` not ```[cinder] ...```

chef/cookbooks/cinder/templates/default/cinder.conf.erb
sjamgade: done https://github.com/crowbar/crowbar-openstack/pull/883

chef/data_bags/crowbar/template-cinder.schema
sjamgade: done https://github.com/crowbar/crowbar-openstack/pull/883

sjamgade:
recreated https://github.com/crowbar/crowbar-openstack/pull/883 from a branch

PR #882, opened 2017-03-29
magnum: Stop encoding spaces into image URL

s-t-e-v-e-n-k:
The curl command to download the image uses one long string with inner
line continuations, which also includes the spaces used to keep Hound
happy. Split the image url string to the outer block so it can be
constructed without spaces while keeping Hound silent.

### DIFF
--- a/chef/cookbooks/magnum/recipes/post_install.rb
+++ b/chef/cookbooks/magnum/recipes/post_install.rb
@@ -62,12 +62,15 @@ openstack_args += " --os-endpoint internalURL"
 
 openstack_cmd = "openstack #{openstack_args}"
 
+image_url = "http://#{provisioner_address}:8091/files/" \
+  "#{service_sles_image_name}/" \
+  "#{service_sles_image_name}.#{node[:kernel][:machine]}.qcow2"
 
 execute "create_magnum_image" do
-  command "curl http://#{provisioner_address}:8091/files/#{service_sles_image_name}/\
-  #{service_sles_image_name}.#{node[:kernel][:machine]}.qcow2 | \
+  command "curl #{image_url} | \
   #{openstack_cmd} #{openstack_args_glance} image create --disk-format qcow2 \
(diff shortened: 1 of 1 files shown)


s-t-e-v-e-n-k:
Failed twice with the known drbd issue in the HA job, but not non-HA job looks fine.

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -62,12 +62,15 @@
...
 openstack_cmd = "openstack #{openstack_args}"
 
+image_url = "http://#{provisioner_address}:8091/files/" \
+  "#{service_sles_image_name}/" \
+  "#{service_sles_image_name}.#{node[:kernel][:machine]}.qcow2"
 
 execute "create_magnum_image" do
-  command "curl http://#{provisioner_address}:8091/files/#{service_sles_image_name}/\
-  #{service_sles_image_name}.#{node[:kernel][:machine]}.qcow2 | \
+  command "curl #{image_url} | \
```
dirkmueller: not your problem, but i'd prefer some timeouts and maybe ```-s``` in the curl invocation

s-t-e-v-e-n-k:
Double-checking my testing, I couldn't reproduce the failure I saw in the job, however:

Before:

root@d52-54-77-77-77-01:~ # glance image-download --file magnum-test f4426773-73e3-447b-bce0-84f5f8dc0456
root@d52-54-77-77-77-01:~ # file magnum-test 
magnum-test: HTML document, ASCII text, with very long lines

Which makes me think the image is the generated HTML of the directory listing based on where the line continuation is.

With above patch:

root@d52-54-77-77-77-01:~ # glance image-download --file magnum-test dbe574d9-ff74-436b-b397-bf771774ac48
root@d52-54-77-77-77-01:~ # file magnum-test 
magnum-test: QEMU QCOW Image (v3), 1909456896 bytes

PR #883, opened 2017-03-29
cinder:Support to remove dependence of volume on its snapshot

sjamgade:


aplanas:
why did you closed the old PR? I am a bit lost.

sjamgade:
Well I made a mistake when I created it on master,
So recreated this on on branch.

chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -207,6 +207,9 @@ volume_driver = cinder.volume.drivers.lvm.LVMVolumeDriver
   <% end -%>
 
   <% if volume['backend_driver'] == 'rbd' -%>
+    <% if volume['rbd']['flatten_volume_from_snapshot'] -%>
+rbd_flatten_volume_from_snapshot = true
```
s-t-e-v-e-n-k: I'd personally prefer this written as

rbd_flatten_volume_from_snapshot = <%= volume['rbd']['flatten_volume_from_snapshot'] %>

sjamgade:
I have incorporated your suggestions. I like it. 
Is the change to the commit good enough. Or should I add some description too ?

By putting in keys with default values in our conf, are we cluttering the conf, which was the reason it was split ?

chef/data_bags/crowbar/template-cinder.schema
```diff
@@ -246,7 +247,8 @@
                       "admin_keyring": { "type": "str", "required": true },
                       "pool": { "type": "str", "required": true },
                       "user": { "type": "str", "required": true },
-                      "secret_uuid" : { "type": "str", "required": true }
+                      "secret_uuid" : { "type": "str", "required": true },
+                      "flatten_volume_from_snapshot": { "type": "bool", "required": true }
```
s-t-e-v-e-n-k: This setting is present twice in the schema, but only one in the settings, which is correct?

sjamgade: [volume_defaults provide the value, when rendering in ui](https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/assets/javascripts/barclamps/cinder/application.js#L104)

s-t-e-v-e-n-k: Objection removed.

chef/data_bags/crowbar/migrate/cinder/106_add_rbd_flatten_snapshots_from_volume.rb
```diff
@@ -0,0 +1,10 @@
+def upgrade(ta, td, a, d)
```
toabctl: you are adding a migration but you don't bump the revision in template-cinder.json . that looks wrong.

toabctl:
Is there a bug report for this? If so, please update the commit msg...

sjamgade:
This is actually from the internal [jenkins card](https://trello.com/c/vqidhZU4/290-cloud-mkcloud7-job-4nodes-linuxbridge-tempestfull-x86-64)

chef/data_bags/crowbar/migrate/cinder/107_add_rbd_flatten_snapshots_from_volume.rb
```diff
@@ -0,0 +1,10 @@
+def upgrade(ta, td, a, d)
+  a["volume_defaults"]["rbd"]["flatten_volume_from_snapshot"] = \
+    ta["volume_defaults"]["rbd"]["flatten_volume_from_snapshot"]
```
toabctl: Hm. Thinking about this - the new attribute is required so you need to loop over all available volumes (in addition to  ```volume_defaults```) and add the new attribute. Otherwise applying the barclamp would complain that the key is missing. There are examples for that in previous migration scripts.

sjamgade:
So if I am not wrong. Had I tested this in an `upgrade`, where the deployment used a `rbd` `backend`. then the `cinder` barclamp would have failed during the application, because the param is `required` and would be absent in the `committed` proposal.

PR #884, opened 2017-03-30
cinder: netapp: add max_over_subscription_ratio

matelakat:
Enable users to specify the max_over_subscription_ratio when using
netapp driver. Please note that originally this is a float value, but we
are interpreting it as an integer. If other drivers wish to override the
value, they only need to add max_over_subscription_ratio to their
configuration, the template should hanle the rest.


dirkmueller:
Can you use git cherry-pick -x to link to the master commit?

matelakat:
> Can you use git cherry-pick -x to link to the master commit?

@dirkmueller We do not have this in upstream, I directly created this PR against 3.0. Let me check what the fix in master would look like so we can go down the path of upstream -> backport.

matelakat:
Closing this down, and wait fo #898 to be merged, so we can backport that.

PR #885, opened 2017-03-30
Neutron and Nova barclamp changes to deploy HA config with Cisco ACI properly

mfriesenegger:


mmnelemane:
Covered in https://github.com/crowbar/crowbar-openstack/pull/638

PR #887, opened 2017-04-05
rabbitmq: Implement HA with native rabbitmq clustering

vuntz:
This removes the need for shared storage and also enables active/active HA as well as some load balancing.


chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -8,6 +8,7 @@
       "password": "",
       "user": "nova",
       "vhost": "/nova",
+      "cluster": true,
```
s-t-e-v-e-n-k: Given the migration explicitly sets it to false, I'd feel safer if the default in the template was also false.

vuntz: For the record, the migration sets it to false to not disrupt existing deployments. We want it for new deployments :-)

matelakat:
`transport_url` is missing from lot of components I think.

PR #891, opened 2017-04-07
Clean up extraneous variables

cmurphy:
This patch reduces LOC by simplifying the use_ssl template variable
definition.


cmurphy:
The patch this is cleanup up only merged a few minutes ago, I think that's why the testbuild pr is failing to build

PR #894, opened 2017-04-11
heat,manila,trove: Switch to new keystone settings

s-t-e-v-e-n-k:
Building on an earlier commit that fixed Magnum, fix Heat, Manila
and Trove configuration templates to use the New World Order of
Keystone settings. As a drive-by, remove the lots of lines checking
the insecure setting in the heat template.

Builds on the work started in https://github.com/crowbar/crowbar-openstack/pull/867, but doesn't require it.

### DIFF
--- a/chef/cookbooks/heat/templates/default/heat.conf.erb
+++ b/chef/cookbooks/heat/templates/default/heat.conf.erb
@@ -23,88 +23,60 @@ control_exchange = heat
 
 [clients]
 endpoint_type = internalURL
-<% if @keystone_settings['insecure'] -%>
 insecure = <%= @keystone_settings['insecure'] %>
-<% end -%>
 
 [clients_barbican]
 endpoint_type = internalURL
-<% if @keystone_settings['insecure'] -%>
 insecure = <%= @keystone_settings['insecure'] %>
-<% end -%>
 
 [clients_ceilometer]
 endpoint_type = internalURL
-<% if @keystone_settings['insecure'] -%>
 insecure = <%= @keystone_settings['insecure'] %>
-<% end -%>
 
 [clients_cinder]
 endpoint_type = internalURL
-<% if @keystone_settings['insecure'] -%>
 insecure = <%= @keystone_settings['insecure'] %>
-<% end -%>
 
 [clients_glance]
 endpoint_type = internalURL
-<% if @keystone_settings['insecure'] -%>
 insecure = <%= @keystone_settings['insecure'] %>
-<% end -%>
 
 [clients_heat]
 endpoint_type = publicURL
-<% if @keystone_settings['insecure'] -%>
 insecure = <%= @keystone_settings['insecure'] %>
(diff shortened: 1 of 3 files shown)


chef/cookbooks/heat/templates/default/heat.conf.erb
```diff
@@ -145,14 +117,14 @@ max_header_line = <%= node[:heat][:max_header_line] %>
...
 auth_uri = <%= @keystone_settings['public_auth_url'] %>
 auth_version = <%= @keystone_settings['api_version_for_middleware'] %>
-<% if @keystone_settings['insecure'] -%>
+project_name = <%= @keystone_settings['service_tenant'] %>
+username = <%= @keystone_settings['service_user'] %>
+password = <%= @keystone_settings['service_password'] %>
 insecure = <%= @keystone_settings['insecure'] %>
-<% end -%>
 region_name = <%= @keystone_settings['endpoint_region'] %>
-identity_uri = <%= @keystone_settings['admin_auth_url'] %>
```
cmurphy: Heat fails for me here - we need to set auth_url

http://lists.openstack.org/pipermail/openstack-dev/2016-March/088636.html

Same for trove

s-t-e-v-e-n-k:
I was expecting the barclamps to fail to apply if the service doesn't start -- manila does this, but heat and trove continue merrily along. I've tested this *properly* now, and made sure the services are running and answer requests.

chef/cookbooks/heat/templates/default/heat.conf.erb
```diff
@@ -143,16 +115,16 @@ workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
 max_header_line = <%= node[:heat][:max_header_line] %>
 
 [keystone_authtoken]
-auth_uri = <%= @keystone_settings['public_auth_url'] %>
+auth_url = <%= @keystone_settings['public_auth_url'] %>
```
cmurphy: no no - both auth_uri and auth_url are required

I know. I'm sorry.

s-t-e-v-e-n-k: I want nice things! Requiring both is not nice things. :-)

chef/cookbooks/heat/templates/default/heat.conf.erb
```diff
@@ -144,15 +116,16 @@ max_header_line = <%= node[:heat][:max_header_line] %>
 
 [keystone_authtoken]
 auth_uri = <%= @keystone_settings['public_auth_url'] %>
+auth_url = <%= @keystone_settings['public_auth_url'] %>
```
cmurphy: It's probably best to keep consistent with other services and use the internal url here, e.g. https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/neutron/templates/default/neutron.conf.erb#L46

PR #895, opened 2017-04-11
Fix mistakes from rabbitmq SSL addition

cmurphy:
This patch bumps the schema revision for the rabbitmq databag to 100 and
adds missing oslo_messaging_rabbitmq settings to the glance registry
configuration.


dirkmueller:
Rename the migration file to 100_... as well?

cmurphy:
@dirkmueller woops, done

dirkmueller:
Git commit message summary should start with ```glance: ...``` and maybe more generic (the summary appears in the changelog)

cmurphy:
Split into two commits to address the mistakes separately

rsalevsky:
LGTM, the original "mistake" is not in the current crowbar-openstack update so this blocks nothing.

PR #897, opened 2017-04-12
upgrade: Do not reboot XEN node from the first chef-client run

jsuchome:
During node upgrade, we're executing chef-client from crowbar_join.
If node reboots in this stage, upgrade script that executed this
crowbar_join will be interrupted and cannot finish.

The reboot is indeed needed, but it could be delayed to the next regular chef-client run, I hope.


jsuchome:
[Handle XEN nodes during upgrade](https://trello.com/c/sflBGV96/238-handle-xen-nodes-during-upgrade)

jsuchome:
Yes, @gosipyan tested it with XEN setup

PR #898, opened 2017-04-12
cinder: netapp: add max_over_subscription_ratio

matelakat:
Enable users to specify the max_over_subscription_ratio when using
netapp driver. If other drivers needed this configuration value they
would add max_over_subscription_ratio to their configuration and the
template will render the value to the configuration file. So if for a
specific volume

    volume[volume['backend_driver']]['max_over_subscription_ratio']

exists, then it's value will be included.

Please note that an integer is used for this number whereas OpenStack
will interpret this as a float.

### DIFF
--- a/chef/cookbooks/cinder/templates/default/cinder.conf.erb
+++ b/chef/cookbooks/cinder/templates/default/cinder.conf.erb
@@ -269,6 +269,10 @@ iscsi_write_cache = off
 image_upload_use_cinder_backend = true
   <% end -%>
 
+  <% if volume[volume['backend_driver']] && volume[volume['backend_driver']]['max_over_subscription_ratio'] -%>
+max_over_subscription_ratio = <%= volume[volume['backend_driver']]['max_over_subscription_ratio'] %>
+  <% end -%>
+
   <% break unless @use_multi_backend -%>
 <% end -%>
 
--- a/chef/data_bags/crowbar/migrate/cinder/106_add_max_over_subscription_ratio.rb
+++ b/chef/data_bags/crowbar/migrate/cinder/106_add_max_over_subscription_ratio.rb
@@ -1,0 +1,26 @@
+def upgrade(ta, td, a, d)
+  unless a["volume_defaults"]["netapp"].key? "max_over_subscription_ratio"
+    a["volume_defaults"]["netapp"]["max_over_subscription_ratio"] = \
+      ta["volume_defaults"]["netapp"]["max_over_subscription_ratio"]
+
+    a["volumes"].each do |volume|
+      next if volume["backend_driver"] != "netapp"
+      volume["netapp"]["max_over_subscription_ratio"] = \
+        ta["volume_defaults"]["netapp"]["max_over_subscription_ratio"]
+    end
+  end
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless ta["volume_defaults"]["netapp"].key? "max_over_subscription_ratio"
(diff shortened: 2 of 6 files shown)


chef/data_bags/crowbar/migrate/cinder/106_add_max_over_subscription_ratio.rb
```diff
@@ -0,0 +1,20 @@
+def upgrade(ta, td, a, d)
+  a["volume_defaults"]["netapp"]["max_over_subscription_ratio"] = \
+    ta["volume_defaults"]["netapp"]["max_over_subscription_ratio"]
```
vuntz: Since there'll be a backport, you should check if the attribute already exists before doing that.

(same on downgrade, for the template attribute)

matelakat: Ah, ok!

chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -269,6 +269,12 @@ iscsi_write_cache = off
 image_upload_use_cinder_backend = true
   <% end -%>
 
+  <% unless volume[volume['backend_driver']].nil? -%>
+  <% unless volume[volume['backend_driver']]['max_over_subscription_ratio'].nil? -%>
```
dirkmueller: that blows my mind. why not simply use

```ruby
<% if  volume[volume['backend_driver']] && volume[volume['backend_driver']]['max_over_subscription_ratio'] -%>
```

chef/data_bags/crowbar/template-cinder.schema
```diff
@@ -71,7 +71,8 @@
                     "netapp_password": { "type": "str", "required": true },
                     "netapp_vfiler": { "type": "str", "required": true },
                     "netapp_transport_type": { "type": "str", "required": true },
-                    "netapp_volume_list": { "type": "str", "required": true }
+                    "netapp_volume_list": { "type": "str", "required": true },
+                    "max_over_subscription_ratio": { "type": "int", "required": true }
```
dirkmueller: its a float actually upstream

crowbar_framework/config/locales/cinder/en.yml
```diff
@@ -105,6 +105,7 @@ en:
               netapp_password: 'Password for accessing NetApp'
               netapp_server_hostname: 'Server host name'
               netapp_server_port: 'Server port'
+              max_over_subscription_ratio: 'Max Over Subscription Ratio'
```
dirkmueller: Maximum Oversubscription Ratio

chef/data_bags/crowbar/template-cinder.schema
matelakat: @dirkmueller float did not seem to work with the UI - i think that's the reason we are not having any floats in barclamps. Do you think we can get by using an int, or are you suggesting to fix float upstream instead?

chef/cookbooks/cinder/templates/default/cinder.conf.erb
matelakat: Right, let's use that form.

crowbar_framework/config/locales/cinder/en.yml
matelakat: Onit

chef/data_bags/crowbar/template-cinder.schema
dirkmueller: works for me. thanks

PR #899, opened 2017-04-12
sahara: Add MapR and Storm to plugins enabled by default (fate#323275)

vuntz:
This is tech preview, so it doesn't hurt to expose more things by
default.

### DIFF
--- a/chef/data_bags/crowbar/template-sahara.json
+++ b/chef/data_bags/crowbar/template-sahara.json
@@ -16,7 +16,7 @@
       "cinder_instance": "none",
       "service_user": "sahara",
       "service_password": "",
-      "plugins": "vanilla,spark,cdh,ambari",
+      "plugins": "vanilla,spark,storm,cdh,ambari,mapr",
       "api": {
         "protocol": "http",
         "bind_open_address": true,


toabctl:
@vuntz not sure but iirc there was a reason why theses are not enabled. maybe tempest will fail with theses plugins enabled...

Itxaka:
I cant remember why I set only those plugins, @toabctl may be rigth on the tempest failures. I trigered a run with sahara enabled to see if it passes

vuntz:
@Itxaka @toabctl mkcloud seems happy...

chef/data_bags/crowbar/template-sahara.json
```diff
@@ -16,7 +16,7 @@
       "cinder_instance": "none",
       "service_user": "sahara",
       "service_password": "",
-      "plugins": "vanilla,spark,cdh,ambari",
+      "plugins": "vanilla,spark,storm,cdh,ambari,mapr",
```
dirkmueller: can we keep the upstream ordering or is there meaning in why we changed it?

https://github.com/openstack/sahara/blob/master/sahara/plugins/opts.py#L20

toabctl:
> @Itxaka @toabctl mkcloud seems happy...

@vuntz Then I'm happy, too

chef/data_bags/crowbar/template-sahara.json
vuntz: The rationale is to have the items sorted alphabetically (except for the vailla one).

PR #907, opened 2017-04-18
Ansible installer integration for Monasca

dirkmueller:


chef/cookbooks/monasca/files/default/ansible.cfg
```diff
@@ -0,0 +1,232 @@
+# Copyright 2017 FUJITSU LIMITED
```
toabctl: I wonder what options we really need from that file...

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -0,0 +1,120 @@
...
+  source "ansible.cfg"
+  owner "root"
+  group "root"
+  mode "0644"
+end
+
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+monasca_hosts = MonascaHelper.monasca_hosts(search(:node, "roles:monasca-server"))
+
+raise "no nodes with monasca-server role found" if monasca_hosts.nil? || monasca_hosts.empty?
```
toabctl: Can we move that check up? There is no need to install packages and configure ansible if the server role is not there

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+template "/opt/monasca-installer/monasca-hosts" do
+  source hosts_template
+  owner "root"
+  group "root"
+  mode "0644"
+  variables(
+    monasca_host: monasca_hosts[0],
+    monasca_hosts: monasca_hosts,
```
toabctl: are we allowing more than one monasca-server role? If not, one variable would be enough.

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -0,0 +1,120 @@
...
+    ansible_ssh_user: "root",
```
toabctl: would we ever change this to something non-root?

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -0,0 +1,120 @@
...
+  mode "0644"
+  variables(
+    keystone_settings: keystone_settings
+  )
+  notifies :run, "execute[run ansible]", :delayed
+end
+
+# This file is used to mark that ansible installer run successfully.
+# Without this, it could happen that the installer was not re-tried
+# after a failed run.
```
toabctl: Why can that happen? When ansible-playbook is executed (see https://github.com/crowbar/crowbar-openstack/pull/907/files#diff-75448b1e14110b2431151f11121aff35R114 ), the file is removed before the run and created after the run. I don't get the idea behind that.

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -0,0 +1,120 @@
...
+file "/opt/monasca-installer/.installed" do
+  content "monasca installed"
+  owner "root"
+  group "root"
+  mode "0644"
+  notifies :run, "execute[run ansible]", :delayed
+  action :create_if_missing
+end
+
+monasca_node = search(:node, "roles:monasca-server")[0]
```
toabctl: we already did the search at https://github.com/crowbar/crowbar-openstack/pull/907/files#diff-75448b1e14110b2431151f11121aff35R27 . Please reuse the result (searches are expensive afaik)

chef/cookbooks/monasca/templates/default/all_group.erb
```diff
@@ -0,0 +1,97 @@
...
+# Copyright 2017 FUJITSU LIMITED
+# file: group_vars/all_group
+
+# ports
+monasca_api_client_port: 8070
+monasca_api_admin_port: 8081
+persister_app_connector_port: 8190
+persister_admin_connector_port: 8191
+monasca_log_api_port: 5607
+keystone_port: 5000
```
toabctl: would be good to use the ports defined in the keystone cookbook

chef/data_bags/crowbar/template-monasca.json
```diff
@@ -69,17 +81,20 @@
...
       "element_order": [
         [ "monasca-server" ],
         [ "monasca-metric-agent" ],
+        [ "monasca-master" ],
         [ "monasca-log-agent" ]
       ],
       "element_run_list_order": {
         "monasca-server": 110,
         "monasca-metric-agent": 120,
+        "monasca-master": 115,
```
toabctl: not sure but shouldn't have monasca-master the lowest number?

chef/cookbooks/monasca/templates/default/monasca-hosts-cluster.erb
```diff
@@ -0,0 +1,278 @@
...
+keystone-node                    ansible_ssh_host=<%= @keystone_host %> ansible_ssh_user=<%= @ansible_ssh_user %>
+
+# Load balancer hosts
+# TODO
+#monasca-log-api-balancer-node    ansible_ssh_host= ansible_ssh_user=<%= @ansible_ssh_user %>
+#monasca-api-balancer-node        ansible_ssh_host= ansible_ssh_user=<%= @ansible_ssh_user %>
+#kibana-balancer-node             ansible_ssh_host= ansible_ssh_user=<%= @ansible_ssh_user %>
+
+# ElasticSearch nodes for cluster
+<% @monasca_hosts.each_with_index do |host, index| %>
```
jgrassler: Just a thought for later: if we eventually switch to a big 9 node setup with dedicated elasticsearch nodes and all the other bells and whistles we will have to revisit everything that iterates over `monasca_hosts` in this file and introduce dedicated parameters, e.g. `elasticsearch_hosts`, `monasca_api_hosts`. But let's leave this alone for now and get this thing merged...

crowbar_framework/app/models/monasca_service.rb
```diff
@@ -31,7 +31,7 @@ def role_constraints
       {
         "monasca-metric-agent" => {
           "unique" => false,
-          "admin" => true,
+          "admin" => false,
```
jgrassler: Why do we disallow running monitoring agents on the Crowbar node?

crowbar_framework/app/models/monasca_service.rb
```diff
@@ -40,7 +40,7 @@ def role_constraints
         },
         "monasca-log-agent" => {
           "unique" => false,
-          "admin" => true,
+          "admin" => false,
```
jgrassler: Why do we disallow running logging agents on the Crowbar node?

chef/cookbooks/monasca/recipes/master.rb
dirkmueller: done

chef/cookbooks/monasca/recipes/master.rb
dirkmueller: I remoed monasa_host and changed everything to use monasca_hosts

chef/cookbooks/monasca/recipes/master.rb
dirkmueller: done

chef/data_bags/crowbar/template-monasca.json
dirkmueller: I think it doesn't matter, they will not be executed on the same node

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -0,0 +1,121 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitation.
+#
+
+monasca_servers = search(:node, "roles:monasca-server")
+monasca_hosts = MonascaHelper.monasca_hosts(monasca_servers)
+raise "no nodes with monasca-server role found" if monasca_hosts.nil? || monasca_hosts.empty?
```
toabctl: Thinking about it - this could be done in crowbar itself to get a nice error message when a master is deployed but no server. But this can be done later so not a blocker...

PR #909, opened 2017-04-19
RFC: Oscm barclamp

dirkmueller:


chef/cookbooks/oscm/recipes/server.rb
```diff
@@ -116,11 +116,15 @@
   })
 end
 
-bash "add keypair" do
+bash "add oscm keypair" do
```
sjamgade: This looks fine. 
But `template` resource can get the job done and looks more clean

dirkmueller:
closing until further input from PM/business

PR #910, opened 2017-04-19
rabbitmq: Fix ordering of ssl setup

cmurphy:
The ha recipe of the rabbitmq cookbook forcibly changes the IDs of the
rabbitmq user and group after rabbitmq is installed. If we try to change
group ownership of a file after rabbitmq is installed but before the
rabbitmq group's gid is changed, the file will be left with just an
apparently-random gid as its group. This was happening to the rabbitmq
crowbar-generated private SSL key, which caused rabbitmq to be unable to
read its own key.  Subsequently, the memcached package was installed
which scooped up the gid and the SSL key then became apparently owned by
the memcached group.

This patch simply delays generation of the SSL key pair until after the
ha recipe has been able to muck with the rabbitmq user and group.

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/rabbit.rb
+++ b/chef/cookbooks/rabbitmq/recipes/rabbit.rb
@@ -34,6 +34,17 @@ end
 
 include_recipe "rabbitmq::default"
 
+if ha_enabled
+  log "HA support for rabbitmq is enabled"
+  include_recipe "rabbitmq::ha"
+  # All the rabbitmqctl commands are local, and can only be run if rabbitmq is
+  # local
+  service_name = "rabbitmq"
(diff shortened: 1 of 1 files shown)


cmurphy:
Rebuilding the ha job with want_all_ssl=1, so if that's green then this should be good to go

cmurphy:
@s-t-e-v-e-n-k it's just how the diff rendered, the ha section and the ssl section swapped places

cmurphy:
Okay the HA + SSL job is failing because of https://github.com/SUSE-Cloud/automation/pull/1901 but it finished the deployment, which was the bug this was fixing, and it made it past all the crm resource failure checks that now happen after every proposal so I think this is ready to merge.

PR #911, opened 2017-04-19
monasca: choose proper monitoring role node for monasca-server

dirkmueller:


crowbar_framework/app/models/monasca_service.rb
```diff
@@ -31,7 +31,7 @@ def role_constraints
       {
         "monasca-metric-agent" => {
           "unique" => false,
-          "admin" => true,
```
jgrassler: Why does this pull request make changes to the monasca-metric-agent role?

crowbar_framework/app/models/monasca_service.rb
```diff
@@ -40,7 +40,7 @@ def role_constraints
         },
         "monasca-log-agent" => {
           "unique" => false,
-          "admin" => true,
```
jgrassler: Why does this pull request make changes to the monasca-log-agent role?

crowbar_framework/app/models/monasca_service.rb
dirkmueller: so that select_nodes_for_role properly selects from the available nodes (excludes admin until it is proven that admin node is safe to include all the time)

crowbar_framework/app/models/monasca_service.rb
dirkmueller: so that select_nodes_for_role properly selects from the available nodes (excludes admin until it is proven that admin node is safe to include all the time)

crowbar_framework/app/models/monasca_service.rb
jgrassler: Ah, ok.

crowbar_framework/app/models/monasca_service.rb
jgrassler: Ah, ok.

PR #915, opened 2017-04-20
keystone: Allow updating the admin pass after barclamp deployment

Itxaka:
Before this, we could not update the password of the admin user in keystone
due using the same var for auth and the new password.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -514,6 +514,36 @@ register_auth_hash = { user: node[:keystone][:admin][:username],
                        password: node[:keystone][:admin][:password],
                        tenant: node[:keystone][:admin][:tenant] }
 
+updated_password = node[:keystone][:admin][:updated_password]
+
+unless updated_password.empty? ||
+    updated_password.nil? ||
+    updated_password == node[:keystone][:admin][:password]
+
+  if !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)
+    keystone_register "update admin password" do
+      protocol node[:keystone][:api][:protocol]
+      insecure keystone_insecure
+      host my_admin_host
+      port node[:keystone][:api][:admin_port]
+      auth register_auth_hash
+      user_name node[:keystone][:admin][:username]
+      user_password updated_password
+      tenant_name node[:keystone][:admin][:tenant]
+      action :nothing
+    end.run_action(:add_user)
+  end
+
+  ruby_block "update admin password on node attributes" do
+    block do
+      node.set[:keystone][:admin][:password] = updated_password
+      node.save
+      register_auth_hash[:password] = updated_password
+    end
+    action :nothing
+  end.run_action(:create)
+end
+
 # Silly wake-up call - this is a hack; we use retries because the server was
 # just (re)started, and might not answer on the first try
 keystone_register "wakeup keystone" do
--- a/crowbar_framework/app/models/keystone_service.rb
+++ b/crowbar_framework/app/models/keystone_service.rb
@@ -131,5 +131,25 @@ class KeystoneService < PacemakerServiceObject
 
     @logger.debug("Keystone apply_role_pre_chef_call: leaving")
   end
+
+  def update_proposal_status(inst, status, message, bc = @bc_name)
+    @logger.debug("update_proposal_status: enter #{inst} #{bc} #{status} #{message}")
+
+    prop = Proposal.where(barclamp: bc, name: inst).first
+    unless prop.nil?
+      prop["deployment"][bc]["crowbar-status"] = status
+      prop["deployment"][bc]["crowbar-failed"] = message
+      # save the updated_password into the password field to update the raw_view
+      if status == "success" && !prop["attributes"][bc]["admin"]["updated_password"].blank?
+        prop["attributes"][bc]["admin"]["password"] = prop["attributes"][bc]["admin"]["updated_password"]
+      end
+      res = prop.save
+    else
+      res = true
+    end
+
+    @logger.debug("update_proposal_status: exit #{inst} #{bc} #{status} #{message}")
+    res
+  end
 end
 
--- a/crowbar_framework/config/locales/keystone/en.yml
+++ b/crowbar_framework/config/locales/keystone/en.yml
@@ -32,6 +32,8 @@ en:
         admin:
           username: 'Administrator Username'
           password: 'Administrator Password'
+          updated_password: 'Update Administrator Password'
+          updated_password_hint: 'Changing the admin password directly in OpenStack its not supported. You can change the admin password directly using this field.'
         ssl_header: 'SSL Support'
         api:
           protocol: 'Protocol'
--- a/chef/data_bags/crowbar/migrate/keystone/110_add_updated_password.rb
+++ b/chef/data_bags/crowbar/migrate/keystone/110_add_updated_password.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["admin"]["updated_password"] = ta["admin"]["updated_password"]
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["admin"].delete("updated_password")
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-keystone.json
+++ b/chef/data_bags/crowbar/template-keystone.json
@@ -38,7 +38,8 @@
       "admin": {
         "tenant": "admin",
         "username": "admin",
-        "password": "crowbar"
+        "password": "crowbar",
+        "updated_password": ""
       },
       "service": {
         "tenant": "service",
@@ -179,7 +180,7 @@
     "keystone": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 109,
+      "schema-revision": 110,
       "element_states": {
         "keystone-server": [ "readying", "ready", "applying" ]
       },
--- a/chef/data_bags/crowbar/template-keystone.schema
+++ b/chef/data_bags/crowbar/template-keystone.schema
@@ -43,7 +43,8 @@
                     "admin": { "type": "map", "required": true, "mapping": {
                       "tenant": { "type" : "str", "required" : true },
                       "username": { "type" : "str", "required" : true },
-                      "password": { "type" : "str", "required" : true }
+                      "password": { "type" : "str", "required" : true },
+                      "updated_password": { "type" : "str", "required" : false }
                     }},
                     "service": { "type": "map", "required": true, "mapping": {
                       "tenant": { "type" : "str", "required" : true },
--- a/crowbar_framework/app/views/barclamp/keystone/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/keystone/_edit_attributes.html.haml
@@ -14,7 +14,12 @@
 
       = string_field %w(default tenant)
       = string_field %w(admin username)
-      = password_field %w(admin password)
+      - if @proposal.active?
+        = password_field %w(admin updated_password)
+        .alert.alert-info
+          = t('.admin.updated_password_hint')
+      - else
+        = password_field %w(admin password)
 
       = boolean_field %w(default create_user),
         "data-showit" => "true",


Itxaka:
Unfortunately, while this works and it updates the password and saves it properly, there is still the issue of other cookbooks.

As they are setting up the register_auth_hash out of a resource (https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/cinder/recipes/api.rb#L36), it gets created during the chef compile phase, so once we update the register_auth_hash with the new password, the other vars available in the other cookbooks still have the old password, making any keystone calls fail.

The only workaround in this case, I think, its to force a KeystoneHelper cache invalidation, and make the register_auth_hash a method of KeystoneHelper so it gets resolved on the execution phase instead of on the compilation.

Itxaka:
[P3: bsc#1006301 - Unable to change keystone admin password after initial deployment](https://trello.com/c/MxL2m7or/230-p3-bsc-1006301-unable-to-change-keystone-admin-password-after-initial-deployment)

Itxaka:
Seems that it was easier than expected. We can move the resources for the new password into the compilation phase, as we can use the updated_password var as a flag. 

This way, the admin user is the first updated before all other cookbooks use the keystone credentials, thus using the proper ones, even when they are updated.

Itxaka:
ping @nicolasbock 

Maybe you can reuse some of this code for https://github.com/crowbar/crowbar-openstack/pull/859 ? Not sure if you encountered the same issues in there?

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -514,6 +514,34 @@
...
+    updated_password == node[:keystone][:admin][:password]
+
+  keystone_register "update admin password" do
+    protocol node[:keystone][:api][:protocol]
+    insecure keystone_insecure
+    host my_admin_host
+    port node[:keystone][:api][:admin_port]
+    auth register_auth_hash
+    user_name node[:keystone][:admin][:username]
+    user_password node[:keystone][:admin][:updated_password]
```
jsuchome: That's just silly nitpick, but you can pass `updated_password` here

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -514,6 +514,34 @@
...
+    tenant_name node[:keystone][:admin][:tenant]
+    action :nothing
+  end.run_action(:add_user)
+
+  ruby_block "update admin password on node attributes" do
+    block do
+      node.set[:keystone][:admin][:password] = node[:keystone][:admin][:updated_password]
+      node.save
+      register_auth_hash[:password] = node[:keystone][:admin][:password]
```
jsuchome: Shouldn't this be already changed to new password as well?

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -514,6 +514,34 @@
...
+    protocol node[:keystone][:api][:protocol]
+    insecure keystone_insecure
+    host my_admin_host
+    port node[:keystone][:api][:admin_port]
+    auth register_auth_hash
+    user_name node[:keystone][:admin][:username]
+    user_password node[:keystone][:admin][:updated_password]
+    tenant_name node[:keystone][:admin][:tenant]
+    action :nothing
+  end.run_action(:add_user)
```
jsuchome: I was thinking - in case of HA setup, does this make sense for all keystone cluster nodes? (Definitely the node attributes part)
But actually it seems we do not have any HA guards for other keystone_register calls, so it's probably OK (?).

Itxaka:
> This looks good to me, I'm wondering if we want to clear the node updated_password attribute at the end of the block so we don't run it every chef run.

It wont due the check for equality between the normal password and updated_password :D https://github.com/crowbar/crowbar-openstack/pull/915/files#diff-28b6adf7bd1f81183dace3201a2720a5R521

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -514,6 +514,34 @@
...
+
+  ruby_block "update admin password on node attributes" do
+    block do
+      node.set[:keystone][:admin][:password] = node[:keystone][:admin][:updated_password]
```
Itxaka: same, will change!

chef/cookbooks/keystone/recipes/server.rb
Itxaka: woops, leftover from a different approach, will change!

chef/cookbooks/keystone/recipes/server.rb
Itxaka: nope, unfortunately on the compile phase, the register_auth_hash is already compiled(its above this resource in the file thus it gets compiled before) and in memory, so we need to directly change it unless we want all other resources to use the old password.

chef/cookbooks/keystone/recipes/server.rb
Itxaka: well, actually it wont matter I think due to the resource internally checking for the validity of the password before trying to update it (https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/keystone/providers/register.rb#L141) so it will only be changed once, but still let me test it locally and maybe add an extra check to launch it only on the founder.

jsuchome: I was not worried about the password value, more about the possibility of running the code for each cluster member, but it probably is not an issue (as we do not care in other places).

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -514,6 +514,34 @@
...
+      node.save
+      register_auth_hash[:password] = node[:keystone][:admin][:password]
+    end
+    action :nothing
+  end.run_action(:create)
+end
+
 # Silly wake-up call - this is a hack; we use retries because the server was
 # just (re)started, and might not answer on the first try
 keystone_register "wakeup keystone" do
```
cmurphy: Should the update come after this wakeup? Or is the .run_action() what makes that unnecessary?

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -514,6 +514,36 @@
...
+      auth register_auth_hash
+      user_name node[:keystone][:admin][:username]
+      user_password updated_password
+      tenant_name node[:keystone][:admin][:tenant]
+      action :nothing
+    end.run_action(:add_user)
+
+    ruby_block "update admin password on node attributes" do
+      block do
+        node.set[:keystone][:admin][:password] = updated_password
```
cmurphy: In the raw view in the UI the "password" field never gets the updated password, it just keeps the old password, is that by design? It would also be nice of the proposal validated that the "password" field isn't changed since it has no effect.

Itxaka: ahhh I forgot about the raw view :(

True, it doesnt update the password field. I guess that I would need to update the proposal somehow, but avoid that field being passed to the chef run :(

chef/cookbooks/keystone/recipes/server.rb
Itxaka: The run_action makes it run during the chef compile phase, so at the beginning of the chef run instead of waiting to the execute/converge phase. This is due the necessity of having the keystone_settings["admin"]["password"] key updated asap so other cookbooks use the updated password instead of trying to use the old password.

Itxaka:
Tested on HA and it works properly :D

chef/cookbooks/keystone/recipes/server.rb
jsuchome: Do you think we need it? Once the password is updated, the real password value is changed, so you're talking about the time user changes it in the barclamp and switches to the raw view before applying the proposal?

cmurphy: No, I'm saying after the proposal is applied the raw view still shows the updated_password and the password being different values, even though internally the password value has been updated. I don't think it's that important, though.

chef/cookbooks/keystone/recipes/server.rb
cmurphy: Thanks for clarifying

Itxaka:
@cmurphy updated with proper raw_view update after the proposal success!

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -131,5 +131,25 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
 
     @logger.debug("Keystone apply_role_pre_chef_call: leaving")
   end
+
+  def update_proposal_status(inst, status, message, bc = @bc_name)
+    @logger.debug("update_proposal_status: enter #{inst} #{bc} #{status} #{message}")
+
+    prop = Proposal.where(barclamp: bc, name: inst).first
+    unless prop.nil?
```
jsuchome: You might need to additionally check for `prop.active?` just to be sure

Itxaka: I think `update_proposal_status` is only called from apply_role. Plus status of `success` can only come from a proper chef run.

Plus the rest of the code is lifted directly from `ServiceObject` so its the code that is being run normally.

cmurphy:
> updated with proper raw_view update after the proposal success!

Neat! I think it still needs the note about changing the password directly in openstack, as per the card's acceptance criteria?

crowbar_framework/app/models/keystone_service.rb
jsuchome: You're saying there should not be any proposal that is not active. That's right, I assume

Itxaka: Yep, that code will only run after the apply button has been clicked. Plus its the code that was already in there, Im only adding line 134 to the original function.

Itxaka:
> Neat! I think it still needs the note about changing the password directly in openstack, as per the card's acceptance criteria?

Yep, still looking for a decent description to add :/

crowbar_framework/config/locales/keystone/en.yml
```diff
@@ -32,6 +32,8 @@ en:
         admin:
           username: 'Administrator Username'
           password: 'Administrator Password'
+          updated_password: 'Update Administrator Password'
+          updated_password_hint: 'If you have changed the admin password directly in OpenStack, you need to update it here and apply the barclamp so that the deployment tool can use the updated password. Otherwise, you can change the admin password directly using this field.'
```
cmurphy: This isn't quite true. If I change the password with `openstack user set --password newpass admin`, the crowbar cannot operate any more because it only knows about the old password. In order to get it to work again I need to go into the raw view and set the "password" field to the password I just changed it to.

Itxaka: That its true and it sucks :(

Let me try to rework it somehow to offer a decent way of dealing wiht this.

cmurphy: I think it would be okay just to document that you have to go into raw mode to fix it.

cmurphy: It shouldn't be a very common use-case, it's just good to know there's a way to fix it if someone didn't know any better and broke it.

Itxaka: tbh I dont know how to make this work properly.

Stupid chef and its attributes stuff, its a mess. node attributes have more precedence over roles attributes, so changing the password here to match a `openstack user set` does not work because it will try to use the attributes in the nodes.

Now, how do we deal with the rest of the attributes overrriding the nodes attributes, I dont know. Need to dig deeper into our apply methods.

jsuchome: Isn't it really enough just to put a warning here to use crowbar and not openstack?

cmurphy: ++ I think that would be fine

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -514,6 +514,36 @@
                        password: node[:keystone][:admin][:password],
                        tenant: node[:keystone][:admin][:tenant] }
 
+updated_password = node[:keystone][:admin][:updated_password]
+
+unless updated_password.empty? ||
+    updated_password.nil? ||
+    updated_password == node[:keystone][:admin][:password]
+
+  if (ha_enabled && CrowbarPacemakerHelper.is_cluster_founder?(node)) || !ha_enabled
```
jsuchome: `if !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)`

PR #920, opened 2017-04-21
Monasca: Default grafana-monasca dashboard

kejc87:
This patch updates the grafana monasca dashboard to
reflect the changes on the Metrics Api replacement
from java to python


dirkmueller:
@aplanas the file should not be here actually but pregenerated by check depending on the environment that is installed but that's subject for another PR..

PR #922, opened 2017-04-21
monasca: listen on all interfaces

dirkmueller:
Configure monasca-api and monasca-log-api to listen on all
interfaces.


toabctl:
@dirkmueller this broke the monasca-log-api healthcheck:
```
TASK [monasca-log-api : Check monasca-log-api healthcheck] *********************
fatal: [monasca-log-api-node]: FAILED! => {"changed": false, "connection": "close", "content": "", "content_type": "text/html; charset=iso-8859-1", "date": "Sun, 23 Apr 2017 06:23:59 GMT", "failed": true, "msg": "Status code was not [204]: HTTP Error 400: Bad Request", "redirected": false, "server": "Apache", "status": 400, "url": "http://*:5607/healthcheck"}
        to retry, use: --limit @/opt/monasca-installer/monasca.retry
```

toabctl:
But the commit msg does not mention why we needed this change. hm.

dirkmueller:
@toabctl ah, sorry i didn't test it :-(

 The issue is that this is used for the vhost statement, and listening only on the public net means the internal url and admin url in the keystone catalog are not handled by apache. So the normal thing for non-ha deployment is to listen on any ip, e.g. 0.0.0.0

If you use above string, is that better?

toabctl:
> If you use above string, is that better?

@dirkmueller You mean if I use ```0.0.0.0```? Not tested but the check is done from the master node and monasca-log-api is running on a different node. So this can not work. I guess we need an extra variable in the ansible installer.

dirkmueller:
@toabctl. Right, wow this is broken :-(

I'm pretty frustrated by the ansible layer, it adds a lot of bugs :-(

We could change thr playbook to use ```monasca_log_api_url``` instead

toabctl:
> We could change thr playbook to use monasca_log_api_url instead

```log_api_ip``` is the var we want. I'll prepare a fix for the installer...

toabctl:
Fix is https://gitlab.com/monasca-installer/ansible-monasca-log-api/merge_requests/2 (tested locally - seems to work)

PR #923, opened 2017-04-22
monasca: Rename metric-agent to agent

toabctl:
monasca-agent is the upstream project name. It is confusing to have

- monasca-agent
- monasca-log-agent
- monasca-metric-agent

In fact there are only 2 (the 2 first ones).


toabctl:
Not sure if we want to have that at this point in time. But after the release, migrations are needed...

toabctl:
> Better now than layer when we need to deal with migration

@dirkmueller I have not tested it. Is the CI testing this automatically?

dirkmueller:
@toabctl yes, want_monasca=1 nodenumber=3 should test it.

PR #925, opened 2017-04-23
horizon: Fix pkill process search (bsc#1035270)

cmurphy:
Without this patch, the horizon service reload produces strange
non-fatal errors in the chef logs:

```
  ERROR: service[horizon] (horizon::server line 33) had an error: Expected process to exit with [0], but received ''
```

This happens because while pkill is smart enough not to include its own
PID in its matching processes, it is not smart enough not to include its
parent process. When chef executes a command, it shells out using 'sh
-c', then because the command includes a '&&' it must fork and exec the
pkill command rather than simply execing, so the command "sh -c pkill
--signal SIGUSR1 '(wsgi:horizon)' && sleep 1" is in the process table
and gets caught by pkill. You can see this in action by comparing these
commands:

```
  $ sh -c "pgrep -alf foobar && true"
  586 sh -c pgrep -alf foobar && true
  $ sh -c "pgrep -alf foobar"
  $
```

The sh -c then terminates because SIGUSR1's default action is "Term".
Chef uses Process.waitpid2 to get the exit code of the command, but
because it was terminated by a signal rather than exiting normally, the
status return value gets populated by the signal code rather than the
exit code (see si_status of the waitpid man page[1]). Chef doesn't
handle this properly and ends up setting the exit code to nil, which is
why it reports '' as the exit code.

We can fix this by narrowing down pkill's matches. pkill accepts regexes
and the wsgi process group is known to come at the beginning of the
process's command string, so we just alter the pkill command to look
only for matches at the beginning of the string, which will exclude
processes matching 'sh -c pgrep ...'.

We also add another sleep to the beginning of the command because on the
first chef run the installation of the horizon vhost, the apache reload,
and the horizon reload happen in quick succession and the wsgi workers
may not have had time to start between apache being refreshed and the
horizon reload happening. This way we can remove the ignore_failure
setting and start treating failures as fatal.

[1] https://linux.die.net/man/2/waitpid

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -33,10 +33,9 @@ end
 service "horizon" do
   service_name "apache2"
   if node[:platform_family] == "suse"
-    reload_command "pkill --signal SIGUSR1 -f '(wsgi:horizon)' && sleep 1"
+    reload_command 'sleep 1 && pkill --signal SIGUSR1 -f "^\(wsgi:horizon\)" && sleep 1'
   end
   supports reload: true, restart: true, status: true
-  ignore_failure true
 end
 
 case node[:platform_family]


dirkmueller:
According to man page the exit code could be 1 in case of no process found that matches the criteria. I guess we could say we don't care about that.

cmurphy:
@dirkmueller I think if the process is not running and we're expecting it to then that should cause a failure. But I can also add the ignore_failure back and this would still be good enough to close the bug.

cmurphy:
I guess pacemaker was the reason for the ignore_failure

aplanas:
I am not sure here. The sleep + ignore behavior comes from the apache recipe:

https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/apache2/recipes/default.rb#L199

There we have a comment that this will protect from an apache problem when we reload is frequently.

cmurphy:
The comment in that recipe (https://github.com/crowbar/crowbar-core/blame/master/chef/cookbooks/apache2/recipes/default.rb#L191-L193) indicates to me that the sleep is only applicable to apache itself since many resources depend on it and may send notifications to it in quick succession. The horizon wsgi processes can start much faster and very few resources notify it so it shouldn't be needed.

But the fact that the sleep is already there for apache and hasn't been causing a problem makes me think I'm wrong about why this error is appearing for horizon, so will have to do some more digging.

cmurphy:
Updated the patch and the description. It's now evident to me why removing the && sleep seemed to fix the issue for me. I still think the && sleep is unnecessary but it is easy enough to work around it.

PR #928, opened 2017-04-24
monasca: Configure and enable monasca-agent

toabctl:
Use Chef to configure monasca-agent and its basic plugins. That way we
don't have to use (and patch) monasca-setup which does currently not
work for SUSE systems.
Also this has advantages when we implement HA.

### DIFF
--- a/chef/cookbooks/keystone/recipes/monitor_monasca.rb
+++ b/chef/cookbooks/keystone/recipes/monitor_monasca.rb
@@ -1,0 +1,39 @@
+#
+# Copyright 2017 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+return unless node["roles"].include?("monasca-agent")
+
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+
+if node[:keystone][:ha][:enabled]
+  admin_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+  bind_admin_host = admin_address
+  bind_admin_port = node[:keystone][:ha][:ports][:admin_port]
+else
+  bind_admin_host = node[:keystone][:api][:admin_host]
+  bind_admin_port = node[:keystone][:api][:admin_port]
+end
+
+# we want to monitor the service that is locally available (*not* the port bind to haproxy)
+monitor_url = "#{node[:keystone][:api][:protocol]}://#{bind_admin_host}:#{bind_admin_port}/"
+
+monasca_agent_plugin_http_check "http_check for keystone" do
+  built_by "keystone-server"
+  name "identity-api"
+  url monitor_url
+  dimensions "service" => "identity-api"
+  match_pattern ".*v#{keystone_settings["api_version"]}.*"
+end
--- a/chef/cookbooks/monasca/recipes/agent.rb
+++ b/chef/cookbooks/monasca/recipes/agent.rb
@@ -28,42 +28,73 @@ if monasca_server.nil?
 end
 
 monasca_api_url = MonascaHelper.api_public_url(monasca_server)
-
 agent_dimensions = { service: "monitoring" }
-service = "monitoring"
 
-monasca_reconfigure_file = "/usr/sbin/monasca-reconfigure"
-
-template monasca_reconfigure_file do
+# the monasca-reconfigure script. if you use that script, the chef settings
+# will be overwritten and after the next chef-client run, the settings
+# from monasca-reconfigure will be overwritten. So DO NOT USE IT!
+template "/usr/sbin/monasca-reconfigure" do
   source "monasca-reconfigure.erb"
   owner "root"
   group "root"
-  mode 0o750
+  mode "0750"
   variables(
     monasca_api_url: monasca_api_url,
-    service: service,
+    service: "monitoring",
     agent_settings: agent_settings,
     agent_keystone: agent_keystone,
     keystone_settings: keystone_settings,
     agent_dimensions: agent_dimensions,
     install_plugins_only: false
   )
-  notifies :run, "execute[monasca-setup detect services]", :delayed
 end
 
-execute "monasca-setup detect services" do
-  command monasca_reconfigure_file
-  user "root"
-  group "root"
-  action :nothing
+# the monasca-agent configuration
+template "/etc/monasca/agent/agent.yaml" do
+  source "monasca-agent_agent.yaml.erb"
+  owner agent_settings[:user]
+  group agent_settings[:group]
+  mode "0640"
+  variables(
+    hostname: node[:hostname],
+    monasca_api_url: monasca_api_url,
+    agent_dimensions: agent_dimensions,
+    log_dir: agent_settings["log_dir"],
+    log_level: agent_settings[:debug] ? "DEBUG": "INFO",
+    keystone_settings: keystone_settings,
+    agent_settings: agent_settings
+  )
 end
 
-service "monasca-agent" do
-  service_name agent_settings[:service_name]
+# monasca-agent uses supervisord to start multiple processes (forwarder, statsd, collector)
+template "/etc/monasca/agent/supervisor.conf" do
+  source "monasca-agent_supervisor.conf.erb"
+  owner agent_settings[:user]
+  group agent_settings[:group]
+  mode "0640"
+end
+
+# enable and start the monasca-agent
+service agent_settings[:agent_service_name] do
+  service_name agent_settings[:agent_service_name]
   supports status: true, restart: true, start: true, stop: true
   action [:enable, :start]
-  ignore_failure true
   # provider Chef::Provider::CrowbarPacemakerService if ha_enabled
+  subscribes :restart, resources(template: "/etc/monasca/agent/agent.yaml")
+  subscribes :restart, resources(template: "/etc/monasca/agent/supervisor.conf")
 end
 
-node.save
+##########################################################
+# plugin config
+##########################################################
+
+# configure basic system plugins
+["cpu", "memory", "load", "network", "disk"].each do |plugin|
+  file "/etc/monasca/agent/conf.d/#{plugin}.yaml" do
+    content lazy { IO.read("/usr/share/monasca/agent/conf.d/#{plugin}.yaml") }
+    owner agent_settings[:user]
+    group agent_settings[:group]
+    mode "0640"
+    notifies :restart, resources(service: agent_settings[:agent_service_name]), :delayed
+  end
+end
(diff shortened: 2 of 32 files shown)


matrixik:
I think monasca-reconfigure should still be generated to allow someone to run it by hand to start monitoring more services than only default ones.

toabctl:
> I think monasca-reconfigure should still be generated to allow someone to run it by hand to start monitoring more services than only default ones.

I would configure more plugins in another PR. Doing that with chef is just a lot easier (and doesn't take ages and might even work on SUSE)

toabctl:
> monasca-reconfigure.conf.erb from templates/default dir needs to be deleted

@dirkmueller done

chef/cookbooks/monasca/templates/default/monasca-agent_supervisor.conf.erb
```diff
@@ -0,0 +1,48 @@
+[supervisorctl]
+serverurl = unix:///var/tmp/monasca-agent-supervisor.sock
```
dirkmueller: Why is this on /var/tmp/ and not on /run?

chef/cookbooks/monasca/templates/default/monasca-agent_supervisor.conf.erb
```diff
@@ -0,0 +1,48 @@
...
+
+[unix_http_server]
+file=/var/tmp/monasca-agent-supervisor.sock
```
dirkmueller: Dito

chef/cookbooks/monasca/templates/default/monasca-agent_supervisor.conf.erb
```diff
@@ -0,0 +1,48 @@
...
+
+[rpcinterface:supervisor]
+supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
+
+[supervisord]
+minfds = 1024
+minprocs = 200
+loglevel = info
+logfile = /var/log/monasca-agent/supervisord.log
```
dirkmueller: Iirc in the package it is /var/log/monasca/agent ? I think this location might be more consistent but we need to double check with support config and other plaves

toabctl: It is /var/log/monasca-agent/

chef/cookbooks/monasca/templates/default/monasca-agent_supervisor.conf.erb
toabctl: done

chef/cookbooks/monasca/templates/default/monasca-agent_supervisor.conf.erb
toabctl: done

chef/cookbooks/monasca/templates/default/monasca-agent_agent.yaml.erb
```diff
@@ -8,34 +8,21 @@ Api:
...
   # Monitoring API URL: URL for the monitoring API, if undefined it will be pulled from the keystone service catalog
   # Example: https://region-a.geo-1.monitoring.hpcloudsvc.com:8080/v2.0
-  url: <%= @monasca_url %>
+  url: <%= @monasca_api_url %>
   # Keystone Username
-  username: <%= @username %>
+  username: <%= @agent_settings["keystone"]["service_user"] %>
   # Keystone Password
-  password: "<%= @password %>"
+  password: <%= @agent_settings["keystone"]["service_password"] %>
```
dirkmueller: Quoting might be more appropriate in yaml

chef/cookbooks/monasca/templates/default/monasca-agent_agent.yaml.erb
```diff
@@ -8,34 +8,21 @@ Api:
   #
   # Monitoring API URL: URL for the monitoring API, if undefined it will be pulled from the keystone service catalog
   # Example: https://region-a.geo-1.monitoring.hpcloudsvc.com:8080/v2.0
-  url: <%= @monasca_url %>
+  url: <%= @monasca_api_url %>
```
dirkmueller: Quoting

chef/cookbooks/monasca/templates/default/monasca-agent_agent.yaml.erb
```diff
@@ -8,34 +8,21 @@ Api:
...
   # Keystone Username
-  username: <%= @username %>
+  username: <%= @agent_settings["keystone"]["service_user"] %>
```
dirkmueller: Quoting

chef/cookbooks/monasca/templates/default/monasca-agent_agent.yaml.erb
```diff
@@ -8,34 +8,21 @@ Api:
...
-  password: "<%= @password %>"
+  password: <%= @agent_settings["keystone"]["service_password"] %>
   # Keystone API URL: URL for the Keystone server to use
   # Example: https://region-a.geo-1.identity.hpcloudsvc.com:35357/v3/
-  keystone_url: <%= @keystone_url %>
-  # Domain id to be used to resolve username
-  user_domain_id: <%= @user_domain_id %>
-  # Domain name to be used to resolve username
-  user_domain_name: <%= @user_domain_name %>
+  keystone_url: <%= @keystone_settings["admin_auth_url"] %>/v<%= @keystone_settings["api_version"] %>
```
dirkmueller: Quoting

chef/cookbooks/monasca/templates/default/monasca-agent_agent.yaml.erb
toabctl: done

chef/cookbooks/monasca/templates/default/monasca-agent_agent.yaml.erb
toabctl: done

chef/cookbooks/monasca/templates/default/monasca-agent_agent.yaml.erb
toabctl: done

chef/cookbooks/monasca/templates/default/monasca-agent_agent.yaml.erb
toabctl: done

artur-ba:
I think configuring agent by hand is not really a good idea. What if somebody after installation run manually monasca-setup? You configuration in agent.yml can be overwritten and plugins detection also.
If I understand correctly this way chef-client will monitor monasca-agent configuration files and this will stop any configuration that can be done via monasca-setup which is a core functionality of monasca-agent.

toabctl:
> I think configuring agent by hand is not really a good idea. What if somebody after installation run manually monasca-setup? You configuration in agent.yml can be overwritten and plugins detection also.
> If I understand correctly this way chef-client will monitor monasca-agent configuration files and this will stop any configuration that can be done via monasca-setup which is a core functionality of monasca-agent.

Yes. But monasca-agent needs a lot of time to run and most of the checks don't work anyway on SLES. And it gets more complicated when we think about monitoring in the HA case.

toabctl:
And if you want to run monasca-setup by hand, just do not assign the monasca-agent role to that node (then chef will not do anything there)

chef/cookbooks/monasca/recipes/agent.rb
```diff
@@ -26,41 +26,81 @@
...
+  mode "0750"
   variables(
+    hostname: node[:hostname],
     monasca_api_url: monasca_api_url,
-    service: service,
-    agent_settings: agent_settings,
-    agent_keystone: agent_keystone,
+    agent_dimensions: { service: "monitoring" },
+    log_dir: agent_settings["log_dir"],
+    log_level: agent_settings[:debug] ? "DEBUG": "INFO",
```
dirkmueller: does that "debug" setting really exist? I was looking for it some time ago and couldn't find it..

I'm fine with removing the log_level granularity but that would need to be cleaned up then in the barclamp template.

toabctl:
> It lgtm. I am wondering if there is a better way to avoid the duplication of the config files in chef.. could we inherit the packaged versions of the templates?

@dirkmueller done

chef/cookbooks/monasca/recipes/agent.rb
toabctl: It exists. it is using the usual logging log levels.

chef/cookbooks/monasca/recipes/agent.rb
```diff
@@ -29,41 +29,52 @@
...
-  command monasca_reconfigure_file
-  user "root"
-  group "root"
-  action :nothing
+# monasca-agent uses supervisord to start multiple processes (forwarder, statsd, collector)
+template "/etc/monasca/agent/supervisor.conf" do
+  source "monasca-agent_supervisor.conf.erb"
+  owner agent_settings[:user]
+  group agent_settings[:group]
+  mode "0750"
```
toabctl: done (also for the agent.yaml file)

chef/cookbooks/keystone/recipes/monitor_monasca.rb
```diff
@@ -0,0 +1,52 @@
...
+node[:monasca] ||= {}
+node[:monasca][:agent_plugin_config] ||= {}
+node[:monasca][:agent_plugin_config][:http_check_instances] ||= {}
+
+# http_check plugin config
+node[:monasca][:agent_plugin_config][:http_check_instances]["keystone-server"] =
+  { "built_by" => "keystone-server",
+    "dimensions" => {"service"=>"identity-api"},
+    "match_pattern" => ".*v#{keystone_settings["api_version"]}.*",
+    "timeout" => 10,
```
jgrassler: We could have this one as a default in the resource provider.

chef/cookbooks/keystone/recipes/monitor_monasca.rb
```diff
@@ -0,0 +1,52 @@
...
+    "name" => "identity-api",
+    "collect_response_time" => true,
```
jgrassler: Same here, We could have this one as a default in the resource provider.

chef/cookbooks/keystone/recipes/monitor_monasca.rb
```diff
@@ -0,0 +1,52 @@
...
+    "url" => monitor_url,
+    "use_keystone" => true }
```
jgrassler: Same here, We could have this one as a default in the resource provider (it'll be `true` for all OpenStack services which will save a lot of typing for these and add an extra line for those services that do not use Keystone authentication).

chef/cookbooks/monasca/providers/agent_plugins.rb
```diff
@@ -0,0 +1,55 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "json"
+require "yaml"
+
+# LWRP action to update the monasca-agent 'http_check' plugin config
+#
+# the action looks for a hash in node[:monasca][:agent_plugin_config][:http_check_instances]
```
jgrassler: While dumping this hash to YAML is very elegant and straightforward we should document at least the most common/useful keys here, since this is very people will go looking first if they want to use this resource.

chef/cookbooks/keystone/recipes/monitor_monasca.rb
```diff
@@ -0,0 +1,52 @@
...
+  bind_admin_host = admin_address
+  bind_admin_port = node[:keystone][:ha][:ports][:admin_port]
+else
+  bind_admin_host = node[:keystone][:api][:admin_host]
+  bind_admin_port = node[:keystone][:api][:admin_port]
+end
+
+# we want to monitor the service that is locally available (*not* the port bind to haproxy)
+monitor_url = "#{node[:keystone][:api][:protocol]}://#{bind_admin_host}:#{bind_admin_port}/" \
+              "v#{keystone_settings["api_version"]}"
```
dirkmueller: it should access / (and check whether the return value includes the expected api_version) imho..

toabctl: done

toabctl:
@dirkmueller do you have more comments here?

sjamgade:
Please correct me if I am wrong. 
1. In case of HA, the agent will run all nodes of the cluster, so all the nodes would do an `http_check`. And the checks would point to the local `address` from admin network.

2. So, should we not disable agent on all nodes except, `cluster_founder` in case of HA.

toabctl:
rebased...

PR #929, opened 2017-04-24
keystone: Set OS_PROJECT_NAME in openrc

dirkmueller:
The previously set OS_TENANT_NAME is deprecated with the switch
to Keystone v3, so it is time to switch to the new setting..


toabctl:
I think we need https://github.com/SUSE-Cloud/automation/pull/1913 to get the CI passing...

PR #933, opened 2017-04-26
monasca: Re-execute installer on variables change

matrixik:
Save MD5 of variables for Ansible to lock file and compare it on every
Chef run.

### DIFF
--- a/chef/cookbooks/monasca/recipes/master.rb
+++ b/chef/cookbooks/monasca/recipes/master.rb
@@ -19,7 +19,9 @@ monasca_hosts = MonascaHelper.monasca_hosts(monasca_servers)
 raise "no nodes with monasca-server role found" if monasca_hosts.nil? || monasca_hosts.empty?
 
 package "ansible"
-package "monasca-installer"
+package "monasca-installer" do
+  notifies :run, "execute[force running ansible]", :delayed
+end
 
 cookbook_file "/etc/ansible/ansible.cfg" do
   source "ansible.cfg"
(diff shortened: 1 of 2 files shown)


matrixik:
WIP but check logic and provide feedback, please.

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -50,76 +52,67 @@
...
-  elasticsearch_hosts: monasca_net_ip,
-  monasca_api_log_level: node[:monasca][:api][:log_level],
-  log_api_log_level: node[:monasca][:log_api][:log_level]
-}.to_json
+ansible_cmd =
+  "rm -f #{lock_file} " \
+  "&& ansible-playbook " \
+    "-i monasca-hosts -e '@/opt/monasca-installer/crowbar_vars.yml' " \
+    "monasca.yml " \
+  "&& #{get_versions} > #{lock_file}"
```
jgrassler: Either this...

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -50,76 +52,67 @@
...
 
 execute "run ansible" do
-  command "rm -f /opt/monasca-installer/.installed " \
-          "&& ansible-playbook -i monasca-hosts -e '#{ansible_vars}' monasca.yml " \
-          "&& touch /opt/monasca-installer/.installed"
+  command ansible_cmd
+  cwd "/opt/monasca-installer"
+  only_if { actual_versions != previous_versions }
+  notifies :create, "file[#{lock_file}]", :immediately
```
jgrassler: ...or this is redundant: both the `lock_file` resource and the last command in `ansible_cmd` create the lock file if the Ansible run is successful. Since the `lock_file` resource no longer notifies an Ansible run we can get rid of it (I'd rather get rid of this resource than the command because that makes the code shorter/more readable).

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -50,76 +52,67 @@
...
+end
+
+execute "force running ansible" do
```
sjamgade: We can do without this block also. 
If I am not wrong. `previous_versions` will be empty in case file is absent.

sjamgade: `previous_versions` will be empty so forcing is not really required.

chef/cookbooks/monasca/recipes/master.rb
matrixik: Done

chef/cookbooks/monasca/recipes/master.rb
matrixik: But on updating `/opt/monasca-installer/crowbar_vars.yml` both versions will be the same.
Also not sure if
```ruby
package "monasca-installer" do
  notifies :run, "execute[force running ansible]", :delayed
end
```
here I need to execute force update, if it's delayed then `rpm -qa` should probably return newer versions than in lock file.

matrixik:
Rebased

matrixik:
Rebased

chef/cookbooks/monasca/templates/default/crowbar_vars.yml.erb
```diff
@@ -0,0 +1,51 @@
...
+smtp_port: <%= @master_settings['smtp_port'] %>
+smtp_user: <%= @master_settings['smtp_user'] %>
+smtp_password: <%= @master_settings['smtp_password'] %>
+smtp_from_address: <%= @master_settings['smtp_from_address'] %>
+
+memcached_listen_ip: <%= @monasca_net_ip %>
+kafka_host: <%= @monasca_net_ip %>
+kibana_host: <%= @pub_net_ip %>
+kibana_plugins: {
+  "monasca-kibana-plugin" => {
```
jgrassler: The problem's here. While most of the Ruby code in the recipe looks very YAML like, this attribute's contents do not (the rebase introduced this problem).

toabctl:
retriggered CI

PR #939, opened 2017-04-28
cinder: Provide NFS mount options to cinder volume shares (bsc#1036369)

jsuchome:
Cinder uses specific mount options only when they are explicitely specified in nfs_shares_config file.

Alternative PR: https://github.com/crowbar/crowbar-openstack/pull/972

### DIFF
--- a/chef/cookbooks/cinder/recipes/volume.rb
+++ b/chef/cookbooks/cinder/recipes/volume.rb
@@ -212,8 +212,17 @@ node[:cinder][:volumes].each_with_index do |volume, volid|
     end
 
   when volume[:backend_driver] == "nfs"
+    shares_content = volume[:nfs][:nfs_shares]
+    nfs_opts = volume[:nfs][:nfs_mount_options].strip
+    unless nfs_opts.empty?
+      # Add global NFS options to each mount
+      shares_content = shares_content.split("\n").collect do |l|
+        opt = l.index(" -o ").nil? ? "-o" : ""
+        l + " #{opt} #{nfs_opts}"
+      end.join("\n")
+    end
     file "/etc/cinder/nfs_shares-#{backend_id}" do
-      content volume[:nfs][:nfs_shares]
+      content shares_content
       owner "root"
       group node[:cinder][:group]
       mode "0640"
--- a/crowbar_framework/config/locales/cinder/en.yml
+++ b/crowbar_framework/config/locales/cinder/en.yml
@@ -53,14 +53,14 @@ en:
           netapp_parameters: 'NetApp Parameters'
           netapp_volume_driver: 'NetApp'
           netapp:
-            nfs_shares_config_hint: 'Format one entry per line in the form hostname:/vol/path optional-nfs-mount-options'
+            nfs_shares_config_hint: 'Format one entry per line in the form hostname:/vol/path -o optional-nfs-mount-options'
             vfiler_hint: 'Only Use this option when utilizing the MultiStore feature on the NetApp storage with iSCSI'
             vserver_hint: 'If using the NFS storage protocol, this parameter is mandatory for storage service catalog support'
             volume_list_hint: 'Optional comma separated list of NetApp controller volume names'
           nfs_parameters: 'NFS Parameters'
           nfs_volume_driver: 'NFS'
           nfs:
-            nfs_shares_config_hint: 'Format one entry per line in the form hostname:/vol/path optional-nfs-mount-options'
+            nfs_shares_config_hint: 'Format one entry per line in the form hostname:/vol/path -o optional-nfs-mount-options'
           raw_parameters: 'Raw Devices Parameters'
           raw_volume_driver: 'Raw Devices'
           rbd_parameters: 'RADOS Parameters'


jsuchome:
[P2: Bug 1036369 - L3: nfs_mount_options are ignored by cinder-volume](https://trello.com/c/5CnUcfzy/257-p2-bug-1036369-l3-nfs-mount-options-are-ignored-by-cinder-volume)

chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -213,7 +213,7 @@ def make_volume(node, backend_id, volume)
 
   when volume[:backend_driver] == "nfs"
     file "/etc/cinder/nfs_shares-#{backend_id}" do
-      content volume[:nfs][:nfs_shares]
+      content "#{volume[:nfs][:nfs_shares]} -o #{volume[:nfs][:nfs_mount_options]}"
```
nicolasbock: This works if the `nfs_mount_options` are non-empty, but fails if they are. I confirmed that the following works in either case:

~~~
    content = "#{volume[:nfs][:nfs_shares]}"
    content += " -o #{volume[:nfs][:nfs_mount_options]}" unless volume[:nfs][:nfs_mount_options].strip.empty?
    file "/etc/cinder/nfs_shares-#{backend_id}" do
      content "#{content}"
~~~

Not sure if that's the prettiest solution thought :smile:.

jsuchome: It's probably best we can have. Thanks for spotting it!

chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -212,8 +212,11 @@ def make_volume(node, backend_id, volume)
     end
 
   when volume[:backend_driver] == "nfs"
+    shares_content = volume[:nfs][:nfs_shares]
```
vuntz: Isn't nfs_shares a list? The way we append the option may not work in that case.

jsuchome: No, it's a string. See that we were adding it directly as `content` previously.

jsuchome:
Adding do-not-merge label just to wait for customer response

Itxaka:
Just wondering, if the field for nfs mount options in the barclamp does not work as expected, should we remove it as part of this pr as to avoid confusion on whre to set the nfs options?

jsuchome:
@Itxaka but it should start to work with this PR

Itxaka:
Woops, was thinking of a different approach to this issue, not sure if I dreamed about it or we tested it before. Sorry!

chef/cookbooks/cinder/recipes/volume.rb
toabctl: @jsuchome but it is a string of lines. Looking at the help text from crowbar (https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/config/locales/cinder/en.yml#L56 ) I wonder why we have the ```nfs_mount_options``` at all. Given that we have it currently, the fix would be imo to loop over the lines from ```nfs_shares``` and append ```nfs_mount_options``` on every line

jsuchome: Oh. That's true. But to me it looks like the help text is wrong, as the nfs options written in the /etc/cinder/nfs_shares-* files need to be prefixed with `-o`.

so what are you actually requesting?

jsuchome: Or we can keep both variables and change the help text so that it does not mention options...  then we would need to add the same options to each line, like you're suggesting.

nicolasbock: Since a customer might want to have different mount options for each mount, adding the same everywhere seems like a bad idea. I vote for dropping the mount options field.

jsuchome: @nicolasbock True, but does it actually make sense to have different options for each mount? What about https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/cinder/templates/default/cinder.conf.erb#L194 , this seems to be per-volume?

nicolasbock: Hmm, true. But the options set there didn't seem to work, or am I remembering this incorrectly?

jsuchome: Right, definitely not the way it was expected.

jsuchome:
So I've adapted this PR according to @toabctl suggestion and created an alternative one for removing the options key here: https://github.com/crowbar/crowbar-openstack/pull/972

chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -212,8 +212,13 @@ def make_volume(node, backend_id, volume)
...
+    nfs_opts = volume[:nfs][:nfs_mount_options].strip
+    unless nfs_opts.empty?
+      shares_content = shares_content.split("\n").collect { |l| l + " -o #{nfs_opts}" }.join("\n")
```
toabctl: I would leave the ```-o``` out. Not sure if you can use ```-o``` multiple times but if you can not, this will prevent the usecase where you add  "per share" options and also globally applied options. Like:
```
List of NFS Exports:
hostname:/vol/path -o foo1
hostname:/vol/path2 -o foo2
```
and
```
Mount Options:
bar1
```

toabctl:
please squash the commits

chef/cookbooks/cinder/recipes/volume.rb
jsuchome: I think `-o` is needed. 
And you can see that in this PR I assume there are only global options, that's why I also adapted help text in `cinder/en.yml`.

See the alternative PR https://github.com/crowbar/crowbar-openstack/pull/972 for a way with different options

toabctl: But customers might already use the "options per line". I don't think we can change the behavior here.

jsuchome:
@toabctl here there's an older commit describing the way 
nfs_shares_config file should look like (so -o is needed):
https://github.com/openstack/cinder/commit/553e0d92c40c73aa1680743c4287f31770131c97

jsuchome:
closing in favor of https://github.com/crowbar/crowbar-openstack/pull/972

PR #940, opened 2017-04-28
monasca: fix Kibana dashboard

jgrassler:
This pull request should fix our current issues with the Monasca Kibana dashboard. I haven't fully tested this yet and I'm still unsure why we do not see logs when the dashboard _is_ working (even though curling the elasticsearch index does show some), so marking this WIP for now.

### DIFF
--- a/chef/cookbooks/monasca/recipes/master.rb
+++ b/chef/cookbooks/monasca/recipes/master.rb
@@ -96,6 +96,16 @@ ansible_vars = {
   memcached_listen_ip: monasca_net_ip,
   kafka_host: monasca_net_ip,
   kibana_host: pub_net_ip,
+  kibana_plugins: {
+    "monasca-kibana-plugin" => {
+      "url" => "file:///usr/share/monasca-kibana-plugin/monasca-kibana-plugin.tar.gz",
+      "configuration" => {
+        "monasca-kibana-plugin.enabled" => true,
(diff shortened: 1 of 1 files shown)


kornicameister:
Oh, I see it is using new auth_uri. Like it already.

jgrassler:
Please do not merge this, yet. At least on post-deployment runs it still broke the `monasca-kibana` role.

jgrassler:
I removed the role assignment bits from this pull request now, since this is likely to take longer and I don't want to block them with this. You'll find the role assignments here: https://github.com/crowbar/crowbar-openstack/pull/952

toabctl:
@jgrassler could you squash and fix hound issues please?

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -96,6 +96,16 @@
   memcached_listen_ip: monasca_net_ip,
   kafka_host: monasca_net_ip,
   kibana_host: pub_net_ip,
+  kibana_plugins: {
+    'monasca-kibana-plugin' => {
+      'url' => '/monasca-kibana-plugin-0.0.5.tar.gz',
+      'configuration' => {
+        'monasca-kibana-plugin.enabled' => true,
+        'monasca-kibana-plugin.auth_uri' => keystone_settings["public_auth_url"],
```
matrixik: Should be probably `internal_auth_url` instead of `public_auth_url` to make it work also in cloud environment.

PR #941, opened 2017-05-02
monasca: log config and cleanup (Pr 919 rework)

toabctl:


chef/cookbooks/monasca/recipes/log_agent.rb
```diff
@@ -26,7 +26,6 @@
 end
 
 log_agent_dimensions = {
-  service: "monitoring",
```
dirkmueller: is this intentional? not mentioned in the git commit message on why?

toabctl: @dirkmueller fixed

PR #942, opened 2017-05-02
tempest: Install monasca packages with tempest tests

nicolasbock:


chef/cookbooks/tempest/recipes/install.rb
```diff
@@ -35,3 +35,7 @@
  "heat", "ceilometer", "sahara"].each do |component|
   package "python-#{component}client"
 end
+
+["api", "log-api"].each do |component|
+  package "python-monasca-#{component}"
```
dirkmueller: These packages only exist on x86 and should only be installed if Monsaca is being deployed

nicolasbock: Please have another look.

toabctl:
commit msg should start with ```tempest: ...```

nicolasbock:
@toabctl done

PR #943, opened 2017-05-03
cinder: Provide NFS mount options to cinder volume shares (bsc#1036369)

jsuchome:
Cinder uses specific mount options only when they are
explicitely specified in nfs_shares_config file.

(cherry picked from commit 15d333d7351441e2f4d13fe44cc6a6d33aea7441)


jsuchome:
The original approach was probably wrong

PR #944, opened 2017-05-03
monasca: Deploy monasca-persister (python) via chef

toabctl:
This commit does 2 things:

1) Use the python version of monasca-persister. There is a java version
which is currently deployed via the monasca-installer (ansible based)
but there are known problems with that version. Also the python version
is more inline with the other OpenStack services and easier to debug.
2) Use chef to deploy monasca-persister. Instead of using
monasca-installer (ansible based) use directly chef which makes the
integration way easier.


toabctl:
rebased

toabctl:
rebased

toabctl:
We go with the java version deployed via monasca-installer for now.

PR #947, opened 2017-05-03
Monasca agent postgres checks

toabctl:


chef/cookbooks/monasca/providers/agent_plugin_http_check.rb
```diff
@@ -36,14 +36,9 @@
...
-  # not yet installed)
-  directory "/etc/monasca/agent/conf.d/" do
-    owner node[:monasca][:agent][:user]
-    group node[:monasca][:agent][:group]
-    mode "0755"
-    recursive true
-  end
+  # be sure the package is installed. that way "/etc/monasca/agent/conf.d/" is available
+  # and also the user and group are there
+  package "openstack-monasca-agent"
```
jgrassler: Just in case anyone considers this double package declaration dodgy (it happens again in the Monasca cookbook's `agent.rb` recipe: while this amounts a double declaration of resources, we just tested this on a QA cloud and it solves the chicken-egg problem of all the service cookbooks running before the Monasca cookbook installs the `openstack-monasca-agent` package (which creates the files/users this provider acts on). This kind of thing _is_ possible in chef (surprised me, coming from a Puppet background where this is a lot stricter).

PR #952, opened 2017-05-04
Add missing keystone roles for Monasca

jgrassler:
This pull request adds the roles we need to use the Kibana/Grafana dashboards in Monasca.


dirkmueller:
Don't we need to create the monasca-user role first? I had a pr for that..

jgrassler:
Ok, creation of the monasca-user role is included now.

toabctl:
I tested it manually in the CI env. Works. Thanks!

PR #953, opened 2017-05-04
Monasca: add monasca-agent role nodes with monasca-server roles also.

sjamgade:


toabctl:
@sjamgade ```monasca-log-api needs monasca-agent running on the same node.```. If that is true, could we express this dependency in crowbar somehow then?

crowbar_framework/app/models/monasca_service.rb
```diff
@@ -95,6 +95,8 @@ def create_proposal
     monasca_server = select_nodes_for_role(nodes, "monasca-server", "monitoring") || []
     log_agent_nodes = select_nodes_for_role(nodes, "monasca-log-agent", "compute") || []
     agent_nodes = select_nodes_for_role(nodes, "monasca-agent", "compute") || []
+    agent_nodes = agent_nodes + monasca_server
```
toabctl: Why no installing the agent on all nodes (except admin maybe)?

dirkmueller:
We could add a check for it in the barclamp save hook. Or we could (somewhat ugly) make the server role also expand to the agent role automatically

crowbar_framework/app/models/monasca_service.rb
dirkmueller: Allowing admin probably makes sense for the metrics agent, but that's a topic for a separate pr imho.

jgrassler:
I think having this visible in `create_proposal` is the right way to go. Including it in the server role is a bit too sneaky for my taste. Better to have this visible (and yes, add a warning in the `barclamp_save` hook to prevent people from removing the `monasca-agent` role from the `monasca-server` nodes manually.

crowbar_framework/app/models/monasca_service.rb
sjamgade: The administrator can always add that role if required. But it is necessary to have it on `monasca_server` nodes.

sjamgade:
@jgrassler  do you mean as `validation_error`  in `save_proposal` ?

jgrassler:
@sjamgade: yes.

toabctl:
@sjamgade nit: commit msgs should start with lower case ```monasca:``` . See ```git log``` for common style.

crowbar_framework/app/models/monasca_service.rb
toabctl: The current assignment does not make sense. We want to have the agent installed on the controller. Otherwise we are not monitoring the OpenStack services.
So I would suggest:
- install the agent on all nodes except admin
- in the crowbar hook, verify that the agent is installed on the node where the server is running

sjamgade:
@toabctl done

toabctl:
> @toabctl done

@sjamgade the first commit msg is still upper-case. Also the agent role assignment ist still wrong imo.

crowbar_framework/app/models/monasca_service.rb
```diff
@@ -94,7 +94,7 @@ def create_proposal
 
     monasca_server = select_nodes_for_role(nodes, "monasca-server", "monitoring") || []
     log_agent_nodes = select_nodes_for_role(nodes, "monasca-log-agent", "compute") || []
-    agent_nodes = select_nodes_for_role(nodes, "monasca-agent", "compute") || []
+    agent_nodes = nodes
```
dirkmueller: I believe you can just remove the "compute" argument.

sjamgade: Yes, it makes the nodes pass through validations.

PR #954, opened 2017-05-04
keystone: Raise and expose number of wsgi workers

dirkmueller:
It seems our default was way too low, especially for the UUID
backend. In experiments a chef-client on the controller node
was speed up almost by factor two by raising numbers accordingly.

### DIFF
--- a/chef/data_bags/crowbar/template-keystone.json
+++ b/chef/data_bags/crowbar/template-keystone.json
@@ -33,7 +33,9 @@
         "admin_port": 35357,
         "admin_host": "0.0.0.0",
         "version": "3",
-        "region": "RegionOne"
+        "region": "RegionOne",
+        "processes" : 8,
+        "threads" : 8
       },
       "admin": {
         "tenant": "admin",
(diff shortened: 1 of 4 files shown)


chef/data_bags/crowbar/template-keystone.json
```diff
@@ -33,7 +33,9 @@
         "admin_port": 35357,
         "admin_host": "0.0.0.0",
         "version": "3",
-        "region": "RegionOne"
+        "region": "RegionOne",
+        "processes" : 8,
+        "threads" : 8
```
cmurphy: The default in mod_wsgi is 15, is 8 really a reasonable default here?

http://modwsgi.readthedocs.io/en/develop/configuration-directives/WSGIDaemonProcess.html

dirkmueller: our LWRP defaults to 10 threads already and the documentation is pretty clear that increasing threads only makes sense for I/O bound usage. I believe keystone would not fit that pattern, but I haven't had done hard measurements. 

https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/crowbar-openstack/resources/wsgi.rb#L10

PR #956, opened 2017-05-05
monasca: Set proper conditions value for services

artur-ba:


toabctl:
nit: commit msg should start with ```monasca: ```

toabctl:
nit again. it is lower case ```monasca:```. But I'm fine with merging it as it is.

PR #957, opened 2017-05-05
Re-add ignore_failure param for horizon service

cmurphy:
In commit a64765b1 we removed the ignore_failure setting for the
horizon wsgi worker reload resource, intending to treat command failures
as fatal. This became a problem because there are still cases where the
refresh might fail, such as when apache is starting and there are no
wsgi processes yet, and we don't want that to cause the chef run to
abort. While we search for these corner cases, this patch adds back the
ignore_failure parameter so that chef can continue to run even if the
refresh fails.

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -38,6 +38,7 @@ service "horizon" do
     reload_command 'sleep 1 && pkill --signal SIGUSR1 -f "^\(wsgi:horizon\)" && sleep 1'
   end
   supports reload: true, restart: true, status: true
+  ignore_failure true
 end
 
 case node[:platform_family]


cmurphy:
https://trello.com/c/VpSTIGA1/347-c7-wsgi-reload-regression

cmurphy:
This failure happens right after

```
service[pacemaker-reload-of-apache2] reloaded
```

so I suspect the apache reload is killing the workers and there isn't enough time for them to start up again before the pkill tries to kill them again, which the sleep at the beginning of the command was supposed to fix.

toabctl:
@cmurphy could you add a commit sha instead of writing ```in a previous commit``` please?

cmurphy:
@toabctl done, sorry about that

dirkmueller:
somehow this change series feels like we're barking up the wrong tree. there must be a more sustainable solution..

cmurphy:
I think my mistake was trying to fix two different things in a64765b. I think changing the regex was necessary to fix the "Expected process to exit with [0], but received ''" problem but then trying to remove the ignore_failure exposed a different issue that I don't have a good answer for yet.

PR #959, opened 2017-05-08
Ironic horizon ui

skazi0:
Horizon plugin for Ironic baremetal service will be installed when ironic services are detected when configuring Horizon.

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -168,6 +168,25 @@ unless sahara_ui_pkgname.nil?
   end
 end
 
+# install horizon ironic plugin if needed
+ironic_ui_pkgname =
+  case node[:platform_family]
+  when "suse"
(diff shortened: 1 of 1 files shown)


skazi0:
[(3) S60P3 Deploy Ironic UI package for Horizon during barclamp install](https://trello.com/c/BPVrE3cv/529-3-s60p3-deploy-ironic-ui-package-for-horizon-during-barclamp-install)

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -168,6 +168,25 @@
   end
 end
 
+# install horizon ironic plugin if needed
+ironic_ui_pkgname =
```
rhafer: I somehow think this should live in the default attributes. But as all the other horizon plugins are handled in a similar way I guess it's ok

skazi0:
rebased to trigger checks

PR #960, opened 2017-05-08
monasca: Configure data retention for InfluxDB

artur-ba:


chef/data_bags/crowbar/template-monasca.json
```diff
@@ -62,6 +62,10 @@
         "smtp_password": "",
         "smtp_from_address": "monasca@localhost"
       },
+      "influxdb": {
```
toabctl: Why not adding theses 2 params to the master section where we already have eg ```influxdb_mon_api_password``` ?

artur-ba: I thought it would be better to have those params in separate section but if you like I will move it to master section.

dirkmueller: It should be consistent - so you can also move the other influxdb_* keys into that subsection

chef/data_bags/crowbar/template-monasca.json
```diff
@@ -62,6 +62,10 @@
...
+        "influxdb_period": 60,
+        "influxdb_duration": "d"
```
dirkmueller: does it make sense to have two keys for that? the only place where it is used it is concatenated again. Having two user visible config keys that can only be used together, what is the point of that?

artur-ba:
@jgrassler I didn't know about this pull request. But in my change configuration of data retention policy is more detailed according to the InfluxDB documentation. 

Lets discuss it and choose the best solution.

chef/data_bags/crowbar/template-monasca.json
toabctl: +1. a single key to configure days should be enough.

artur-ba: The point of that was that data retention for InfluxDB can be set in hours, days, weeks etc. The first variable set integer value of the retention and the second duration time so for example days, but if you will like to have retention policy only in days I can change it.

PR #972, opened 2017-05-19
cinder: Remove nfs_mount_options key as redundant (bsc#1036369)

jsuchome:
Hint for nfs_shares value encourages user to put the options there
already, so we do not need extra field for specifying Mount Options.

An alternative approach to https://github.com/crowbar/crowbar-openstack/pull/939

UPDATE: not only that it is redundant, nfs_mount_options actually does not work

### DIFF
--- a/chef/data_bags/crowbar/migrate/cinder/107_remote_nfs_mount_options.rb
+++ b/chef/data_bags/crowbar/migrate/cinder/107_remote_nfs_mount_options.rb
@@ -1,0 +1,16 @@
+def upgrade(ta, td, a, d)
+  a["volume_defaults"]["nfs"].delete("nfs_mount_options")
+  a["volumes"].each do |volume|
+    next if volume["backend_driver"] != "nfs"
+    volume["nfs"].delete("nfs_mount_options")
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["volume_defaults"]["nfs"]["nfs_mount_options"] = ""
+  a["volumes"].each do |volume|
+    volume["nfs"]["nfs_mount_options"] = "" if volume["backend_driver"] == "nfs"
+  end
+  return a, d
+end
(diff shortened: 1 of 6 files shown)


jsuchome:
Schema would need to be adapted as well...

nicolasbock:
@jsuchome Just to give you a heads up: I am testing it, but I am in Provo this week for a SOC 7 training and will not have a lot of time to do that. The testing might take a little :smile:.

jsuchome:
@nicolasbock thanks for update!

jsuchome:
[P2: Bug 1036369 - L3: nfs_mount_options are ignored by cinder-volume](https://trello.com/c/5CnUcfzy/257-p2-bug-1036369-l3-nfs-mount-options-are-ignored-by-cinder-volume)

jsuchome:
@s-t-e-v-e-n-k I fixed the typo (thanks for spotting), squashed the commits... will restart the HA job again at some point

jsuchome:
Could I ask you for a new review? Old ones were dismissed by my force push...

jsuchome:
@matelakat you've had similar situation; do you have anything to say here? A veto?

chef/data_bags/crowbar/migrate/cinder/107_remote_nfs_mount_options.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["volume_defaults"]["nfs"].delete("nfs_mount_options")
```
dirkmueller: this is not enough, you also need to iterate over the volumes attributes and remove it there if it is the nfs type.

chef/data_bags/crowbar/migrate/cinder/107_remote_nfs_mount_options.rb
```diff
@@ -0,0 +1,9 @@
...
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["volume_defaults"]["nfs"]["nfs_mount_options"] = ""
```
dirkmueller: and readd it here..

chef/data_bags/crowbar/migrate/cinder/107_remote_nfs_mount_options.rb
jsuchome: @dirkmueller adapted

PR #974, opened 2017-05-24
Add resource limits controls (bsc#1020922)

cmurphy:

1) Create a systemd_override LWRP and use it in the cinder barclamp.
2) Use the same LWRP in the keystone and horizon barclamps to demonstrate
3) Add the parameter to the rabbitmq barclamp, which will either use the systemd_override resource or /etc/security/limits.conf depending on whether it is deployed with pacemaker or not.
4) Add the parameter to the database barclamp, which will either use the systemd_override resource or /etc/security/limits.conf depending on whether it is deployed with pacemaker or not.

Changes to fix the rabbitmq and postgresql pacemaker services were pulled into #1029 and those are required for this pull request to work.

### DIFF
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -252,3 +252,4 @@ default[:neutron][:ha][:neutron_l3_ha_service][:hatool][:program] = "/usr/bin/ne
 default[:neutron][:ha][:neutron_l3_ha_service][:hatool][:env] = {}
 default[:neutron][:ha][:neutron_l3_ha_service][:seconds_to_sleep_between_checks] = 10
 default[:neutron][:ha][:neutron_l3_ha_service][:max_errors_tolerated] = 10
+default[:neutron][:ha][:neutron_l3_ha_service][:log_file] = "/var/log/neutron/neutron-l3-ha-service.log"
--- a/chef/cookbooks/neutron/files/default/neutron-l3-ha-service.rb
+++ b/chef/cookbooks/neutron/files/default/neutron-l3-ha-service.rb
@@ -27,6 +27,10 @@ class HAToolLog
     end
     @logger
   end
+
+  def self.file=(file_path)
+    @logger = Logger.new(file_path, "daily") unless file_path.nil?
+  end
 end
 
 def main
@@ -35,6 +39,7 @@ def main
   service_options = ServiceOptions.load ARGV[0]
   ha_functions = HAFunctions.new(service_options)
   error_counter = ErrorCounter.new service_options.max_errors_tolerated
+  HAToolLog.file = service_options.log_file
 
   while true
     status = ha_functions.check_l3_agents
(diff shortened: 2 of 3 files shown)


cmurphy:
Still working out some issues.

cmurphy:
> my only concern is the name of the key 'security_limits'

I totally agree, it's not a good name. "Limits" on its own is a very overloaded term. security_limits comes from what used to be /etc/security/limits.conf but we're not using that any more. Could also say pam_limits but again that's not what's being modified anymore. May be resource_limits, resource_restrictions, service_limits, resource_controls...

cmurphy:
Will update this to use the name resource_limits instead

Itxaka:
This looks pretty good and that commit msg :1st_place_medal: 

I tested it with an HA env and it worked to create but then failed to delete (restart of the service failed with an exit error 3), may have been my environment. Deploying a normal env to test it but it looks really good.

chef/data_bags/crowbar/template-keystone.json
```diff
@@ -175,14 +175,19 @@
...
+        }
       }
     }
   },
   "deployment": {
     "keystone": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 111,
+      "schema-revision": 112,
```
Itxaka: May need to bump this to 113 and change the migration filename due to merging of the http/https patch! sorry!

cmurphy: Yeah I expected that :)

cmurphy:
> I tested it with an HA env and it worked to create but then failed to delete

Looking again I think I didn't properly implement delete for the apache resources so it might not just be your env

chef/cookbooks/crowbar-openstack/providers/systemd_override.rb
```diff
@@ -0,0 +1,104 @@
...
+        action :nothing
+      end
+
+      override_file_path = _get_override_file_path
+      service_resource_name = _get_service_resource_name
+      file override_file_path do
+        action :delete
+        only_if { ::File.exist?(override_file_path) }
+        notifies :run, resources(execute: "systemctl daemon-reload"), :delayed
+        notifies :reload, resources(service: service_resource_name), :delayed
```
Itxaka: Ahh, seems like reload its not supported for some (all?) services!

cmurphy: Woops - yes it should be :restart, it was right for the :create case

cmurphy:
Implemented delete for the apache overrides, fixed the bug @Itxaka pointed out, and changed the name to resource_limits for @s-t-e-v-e-n-k

cmurphy:
Also now fixed the migration file names and the commit message.

cmurphy:
Needs more work after feedback from the bug reporter.

aspiers:
@cmurphy Presumably one of 3. and 4. in the original PR description is supposed to refer to the DB?  Currently they are identical.

cmurphy:
@aspiers yes, fixed

cmurphy:
With https://github.com/crowbar/crowbar-openstack/pull/1301 this should now no longer force restarts of postgresql and rabbitmq.

Itxaka:
@cmurphy what is the status of this? SAP is also in need of something like this so it would be great if this could get merged

cmurphy:
@Itxaka I've been waiting for feedback from the requester, but they have been busy with other fires and this is lower on their priority list. This is ready to merge, so if it's useful to someone else then let's go for it. I think @aspiers's review is no longer applicable since this won't restart the database/message queue any more.

nicolasbock:
I think this is a great change!

We already have a `systemd` resource provider in https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/utils/providers/systemd_service_restart.rb and combining your new provider with the existing one would reduce the additional code and centralize `systemd` changes. (I like the way you handle the `systemd` paths better here than how it is done in `crowbar-core`, so merging your change with the existing plumbing would have the benefit that it would clean up what's already there. :smile:)

Itxaka:
[increase max open file limits](https://trello.com/c/E0gqtDEF/151-increase-max-open-file-limits)

cmurphy:
@Itxaka for this new use case, are the ideal limits known? Could we hard-code them, maybe even in the package? If we have a good idea of what they should be for everybody then I think that would be even better than allowing manual tuning.

Itxaka:
@cmurphy not for the generic ones, no. They initially asked just o increase and the example was the maximum allowed(?) of 65536

Then on a further card they asked for the limits for memcached and the example there was 4096.

I can talk to them and ask what the ideal limits would be for them, but as this is the huge cloud of 250+ nodes, they may just want to set the maximum and forget, so I cannot promise anything there.

Itxaka:
@cmurphy they use 65536 in prod by default to not have issues and are thinking of setting 4096 for memcached (which should be in a different PR as it requires changes to the memcached system file)

Itxaka:
Extracted the LWRP into the utils cookbook at https://github.com/crowbar/crowbar-core/pull/1365

If that is the way we want to go, I'll remove those files from this PR.

Itxaka:
wtf happened here github???? I did not closed this, I just pushed to it????

Itxaka:
I have no idea on how I did broke this, so I reopened at https://github.com/crowbar/crowbar-openstack/pull/1374

PR #977, opened 2017-05-26
Make HA timeouts configurable

nicolasbock:
The defaults for HA resource timeouts can be too strict depending on the
deployment size and hardware. Tuning these timeouts currently requires
either to patch an existing barclamp (which does not survive a crowbar
package update) or to create a separate cookbook to override the
hardcoded values.

This patch introduces tunable timeout values.


nicolasbock:
This issue came up during the SOC 7 Training in Provo. This patch is still incomplete in the sense that it only touches the database barclamp. A full patch would have to expose these timeout values for other barclamps as well.

@vuntz and @aspiers, please let me know what you think. Thanks!

vuntz:
Do you have examples of why these timeouts are too strict? I'm slightly surprised here, tbh.

nicolasbock:
@vuntz Unfortunately I don't. During the Provo Training Ivan Pencak brought this up as an issue he is seeing once in a while. I contacted him and unfortunately he doesn't have a current example for this right now. But he said that he sees this issue from time to time with customers and then has to increase timeouts by hand. Being able to configure those instead directly through the barclamp would be a lot more convenient.

Given the somewhat nebulous nature of this issue, I could ask Ivan to ping me the next he sees this, or to open a bug with some more documentation. Or, on the other hand, we could go ahead with this change under the assumption that it won't do any harm :smile:.

What do you think?

PR #980, opened 2017-05-29
keystone: grant admin role for default domain to admin (bsc#1040822)

Itxaka:
We were missing giving out the admin role for the default admin
to the admin user, thus horizon would not display the domains
panel.

In order to do this a new action is created on the keystone helpers
add_domain_role that would take a user, a role and a domain and grant
that user the given role in that domain.

Unfortunately, the calls to the domain role modify will return
"204 - No content" on create/update so the "_update_item" method on
the keystone provider needed an update to also accept a http 204 as
"success". As to not affect any other calls that may receive a 204
but turning that response into a success is not wanted, a new parameter
was introduced into the function to activate this behaviour. This
safely allows the introduction of the new response as valid only for
those calls that need them

### DIFF
--- a/chef/cookbooks/keystone/providers/register.rb
+++ b/chef/cookbooks/keystone/providers/register.rb
@@ -107,6 +107,38 @@ action :add_domain do
   end
 end
 
+# :add_domain_role specific attributes
+# attribute :domain_name, :kind_of => String
+# attribute :user_name, :kind_of => String
+# attribute :role_name, :kind_of => String
+action :add_domain_role do
+  http, headers = _build_connection(new_resource)
+
+  # get user_id
+  path = "/v3/users"
+  dir = "users"
+  user_id, uerror = _find_id(http, headers, new_resource.user_name, path, dir)
+  # get role_id
+  path = "/v3/roles"
+  dir = "roles"
+  role_id, rerror = _find_id(http, headers, new_resource.role_name, path, dir)
+  # get domain_id
+  path = "/v3/domains"
+  dir = "domains"
+  domain_id, derror = _find_id(http, headers, new_resource.domain_name, path, dir)
+
+  if uerror || rerror || derror
+    Chef::Log.info "Could not obtain the proper ids from keystone"
+    raise "Failed to talk to keystone in add_domain_role"
+  end
+
+  # Construct the path
+  path = "/v3/domains/#{domain_id}/users/#{user_id}/roles/#{role_id}"
+
+  ret = _update_item(http, headers, path, nil, new_resource.domain_name)
+  new_resource.updated_by_last_action(ret)
+end
+
 # :add_user specific attributes
 # attribute :user_name, :kind_of => String
 # attribute :user_password, :kind_of => String
@@ -346,6 +378,10 @@ def _update_item(http, headers, path, body, name)
     return true
   elsif resp.is_a?(Net::HTTPCreated)
     Chef::Log.info("Created keystone item '#{name}'")
+    return true
+  # several APIs use 204 on v3 as success response
+  elsif resp.is_a?(Net::HTTPNoContent)
+    Chef::Log.info("Created/Updated keystone item #{name}")
     return true
   else
     Chef::Log.error("Unable to updated item '#{name}'")
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -610,6 +610,19 @@ if node[:keystone][:domain_specific_drivers]
   end
 end
 
+# Create default role admin for admin user in default domain
+keystone_register "add default admin role for domain default" do
+  protocol node[:keystone][:api][:protocol]
+  insecure keystone_insecure
+  host my_admin_host
+  port node[:keystone][:api][:admin_port]
+  auth register_auth_hash
+  user_name node[:keystone][:admin][:username]
+  role_name "admin"
+  domain_name "Default"
+  action :add_domain_role
+end
+
 # Create default user
 if node[:keystone][:default][:create_user]
   keystone_register "add default #{node[:keystone][:default][:username]} user" do
--- a/chef/cookbooks/keystone/resources/register.rb
+++ b/chef/cookbooks/keystone/resources/register.rb
@@ -17,8 +17,8 @@
 # limitations under the License.
 #
 
-actions :add_service, :add_endpoint_template, :add_tenant, :add_domain, :add_user, :add_role,
-        :add_access, :add_ec2, :wakeup
+actions :add_service, :add_endpoint_template, :add_tenant, :add_domain, :add_domain_role, :add_user,
+        :add_role, :add_access, :add_ec2, :wakeup
 
 attribute :protocol, kind_of: String
 attribute :insecure, kind_of: [TrueClass, FalseClass], default: false


Itxaka:
[P2: Bug 1040822 - #L3 Domain tab missing in Identity section of Horizon with multi_domain_support](https://trello.com/c/UK9KezfP/291-p2-bug-1040822-l3-domain-tab-missing-in-identity-section-of-horizon-with-multi-domain-support)

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -335,7 +367,7 @@ def _create_item(http, headers, path, body, name)
 
 # Return true on success
 private
-def _update_item(http, headers, path, body, name)
+def _update_item(http, headers, path, body, name, no_content=false)
```
cmurphy: 204 (2XX) is always a success, I'm not sure why it needs special treatment?

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -347,6 +379,10 @@ def _update_item(http, headers, path, body, name)
   elsif resp.is_a?(Net::HTTPCreated)
     Chef::Log.info("Created keystone item '#{name}'")
     return true
+  # Domain role creation/update returns 204 for success (¬_¬)
```
cmurphy: It's the same for projects - https://developer.openstack.org/api-ref/identity/v3/index.html#assign-role-to-user-on-project

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -107,6 +107,38 @@
...
+  # get role_id
+  path = "/v2.0/OS-KSADM/roles"
+  dir = "roles"
+  role_id, rerror = _find_id(http, headers, new_resource.role_name, path, dir)
+  # get domain_id
+  path = "/v3/domains"
+  dir = "domains"
+  domain_id, derror = _find_id(http, headers, new_resource.domain_name, path, dir)
+
+  if uerror || rerror || derror
```
jsuchome: I know it this is used elsewhere, but what about checking for specific error immediately after each `_find_id`?

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -107,6 +107,38 @@
...
+    Chef::Log.info "Could not obtain the proper ids from keystone"
+    raise "Failed to talk to keystone in add_domain_role"
+  end
+
+  # Construct the path
+  path = "/v3/domains/#{domain_id}/users/#{user_id}/roles/#{role_id}"
+
+  ret = _update_item(http, headers, path, nil, new_resource.domain_name, no_content=true)
```
jsuchome: Why not just `(..., true)` ?

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -610,6 +610,19 @@
...
+# Create default role admin for admin user in default domain
+keystone_register "add default admin role for domain default" do
+  protocol node[:keystone][:api][:protocol]
+  insecure keystone_insecure
+  host my_admin_host
+  port node[:keystone][:api][:admin_port]
+  auth register_auth_hash
+  user_name node[:keystone][:admin][:username]
+  role_name "admin"
+  domain_name "Default"
```
cmurphy: If this is hardcoding the default domain then how does it solve the issue? The admin user might then see the box but only be able to select the default domain.

chef/cookbooks/keystone/providers/register.rb
Itxaka: explicit rather than implicit I guess, out of habit, I dont mind changing it :)

chef/cookbooks/keystone/providers/register.rb
Itxaka: Umm, then it seems to be a side issue of us using the v2 paths everywhere instead of the v3 directly, as it seems that nothing else expects a 204 in our code.

chef/cookbooks/keystone/providers/register.rb
Itxaka: _find_id already logs a full message for any errors found, so unless this is for performance reasons (fail fast?) I dont think there is too much value into doing the check just after the call.

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -107,6 +107,38 @@
...
 
+# :add_domain_role specific attributes
+# attribute :domain_name, :kind_of => String
+# attribute :user_name, :kind_of => String
+# attribute :role_name, :kind_of => String
+action :add_domain_role do
+  http, headers = _build_connection(new_resource)
+
+  # get user_id
+  path = "/v2.0/users"
```
Itxaka: :+1:

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -107,6 +107,38 @@
...
+  dir = "users"
+  user_id, uerror = _find_id(http, headers, new_resource.user_name, path, dir)
+  # get role_id
+  path = "/v2.0/OS-KSADM/roles"
```
Itxaka: :+1:

chef/cookbooks/keystone/providers/register.rb
jsuchome: I didn't check :-) OK, then

chef/cookbooks/keystone/recipes/server.rb
Itxaka: Isnt keystone the one creating the default domain "default" automatically?

That is the reason it is hardcoded here, because there is always gonna be a default domain id "default"

cmurphy: It is created automatically, but granting the admin user a role in the default domain doesn't really do anything. The admin user already has permissions to do whatever it wants (see https://bugs.launchpad.net/keystone/+bug/968696). So this would make one domain show up in the dropdown but that's not useful.

Itxaka:
> I haven't had a chance to test this fully but I left some initial thoughts.
> 
> I'm a little wary of adding this by default - it doesn't follow the upstream install guide and it isn't done automatically by keystone-manage bootstrap.

Mmmh, upstream guide + bootstrap also dont add extra users and we do it ourselves, not saying this should be the default upstream behavior but the Horizon Workflow mentions that you need to grant admin role to the domain in order to make it work: https://wiki.openstack.org/wiki/Horizon/DomainWorkFlow

> 
> I also don't think it really solves any problems - the admin user can already add themselves and other users to whatever domains they would like, at which point the domains tab should be populated in horizon. 

For one, I think that as you mention the admin can always make the appropriate changes manually in order to get the admin role for the domain. But on the other hand, if a customer wants to use multidomain features from horizon, they still would need to manually add their admin role to the default domain in order to see the panel, so we could fix that for them so there is no extra steps there for them.

I dont think we lose anything for defaulting to making the admin user a admin role for the default domain.

> The multi_domain_support config option in horizon is just to enable the initial domain field in the login window, not the dropdown tab.

Yes, this part was mixed on the bug report as part of being the same issue and it has nothing to do with this patch, the result of the config option is as expected :)

Itxaka:
woops

Itxaka:
rebased so hound run properly

toabctl:
Ups. Merged to early. Wanted to wait for customer feedback. Sorry...

PR #982, opened 2017-05-30
neutron: Avoid conflict between br-fixed and br-public (bsc#1040396)

vuntz:
With ovs, when the vlan type driver is used, we usually have a br-fixed
network to deal with vlan networks.

However, when the floating network is not using vlan and is on the same
conduit as the fixed network, then the underlying interface for br-fixed
and br-public is going to be the same. And of course it can't be in both
bridges. This results in the interface moving from one to the other
during a chef run, creating some network outage.

The solution is to simply merge br-fixed in br-public for that case.

We only need fixes on the rails side, as the cookbook side was fixed
with https://github.com/crowbar/crowbar-openstack/pull/128

https://bugzilla.suse.com/show_bug.cgi?id=1040396

### DIFF
--- a/crowbar_framework/app/models/neutron_service.rb
+++ b/crowbar_framework/app/models/neutron_service.rb
@@ -360,9 +360,38 @@ class NeutronService < PacemakerServiceObject
         # need the bridges)
         ovs_bridge_networks = ["nova_floating", "ironic"]
         ovs_bridge_networks.concat attributes["additional_external_networks"]
-        if ml2_type_drivers.include?("vlan")
-          ovs_bridge_networks << "nova_fixed"
+        ovs_bridge_networks.push("nova_fixed") if ml2_type_drivers.include?("vlan")
+
+        # next block doesn't apply for non-DVR compute nodes, hence the check
+        # on nova_floating
+        floating_net_def = node.get_network_by_type("nova_floating")
+        if ml2_type_drivers.include?("vlan") && !floating_net_def.nil?
+          # If fixed & floating are on the same interface, and floating is not
+          # using a vlan, then br-fixed and br-floating will fight for the
+          # underlying network interface; in that case, we actually only need
+          # one bridge; we'll pick br-public
+          # Note that we assume that all relevant nodes will have a similar
+          # bridge mapping in the end (and not something like one node with an
+          # entry for both floating and physnet1, and some other node with only
+          # physnet1 -- that kind of setups simply won't work).
+          fixed_net_def = node.get_network_by_type("nova_fixed")
+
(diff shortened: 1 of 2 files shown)


aplanas:
[P1: bsc#1040396 - Network interruption during chef run when "vlan:false: true" on nova_floating and public](https://trello.com/c/gCvZbPX3/285-p1-bsc-1040396-network-interruption-during-chef-run-when-vlan-false-true-on-nova-floating-and-public)

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -362,6 +362,24 @@ def update_ovs_bridge_attributes(attributes, node)
...
+
+          # if fixed & floating are on the same conduit, and floating is not
+          # using a vlan, then br-fixed and br-floating will fight for the
+          # underlying network interface; in that case, we actually only need
+          # one bridge; we'll pick br-public
+          fixed_net_def = node.get_network_by_type("nova_fixed")
+          floating_net_def = node.get_network_by_type("nova_floating")
+
+          same_bridge = !floating_net_def["use_vlan"] &&
+            fixed_net_def["conduit"] == floating_net_def["conduit"]
```
rhafer: Hm, comparing the conduits might not be enough. I think you really need to resolve the physical devices that the conduit resolves to. Different conduits can still map to the same physical interface. And where it gets really tricky is that a conduit mappings can be different for different nodes. I.e. on one node you might have floating and fixed use the same interface, while on another one (e.g. with more nics) it might map to different interfaces.

vuntz: @rhafer good point. I'm resolving the conduits and using the list of slaves now (since using the bond interface name is not safe, in case it doesn't exist yet)

toabctl:
@vuntz CI failed with:
```
14:05:12 +(mkcloud-common.sh:108) safely(): crowbarctl proposal commit nova default
14:05:18 {"error":"Failed to apply the proposal: exception before calling chef (undefined method `[]' for nil:NilClass)"}
```

vuntz:
> @vuntz CI failed with:
> ```
> 14:05:12 +(mkcloud-common.sh:108) safely(): crowbarctl proposal commit nova default
> 14:05:18 {"error":"Failed to apply the proposal: exception before calling chef (undefined method `[]' for nil:NilClass)"}
> ```

Thanks, fixed. The code was only tested on DVR setups, so good that the CI caught the bug for non-DVR :-)

toabctl:
Triggered the CI with crowbar-teaming mode (feel free to ignore the result. it is the first run outside of my local testing)

PR #983, opened 2017-05-30
Ironic: add option to disable autoclean + conf update

skazi0:
A new attribute was added to disable automatic cleaning of baremetal nodes. This is especially useful during testing and development as the cleaning procedure can take a lot of time. For production it is not recommended but there are also some cases where it can be useful (e.g. when one tenant wants to quickly re-cycle hardware).

Second commit is update of nova conf to remove some deprecation warnings.

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -145,11 +145,13 @@ protocol = <%= @glance_server_protocol %>
 
 <% unless @ironic_settings.nil? -%>
 [ironic]
-api_endpoint = <%= @ironic_settings[:api_protocol] %>://<%= @ironic_settings[:public_host] %>:<%= @ironic_settings[:api_port] %>
-admin_username = <%= @ironic_settings[:service_user] %>
-admin_password = <%= @ironic_settings[:service_password] %>
-admin_url = <%= @keystone_settings['protocol'] %>://<%= @keystone_settings['internal_url_host'] %>:<%= @keystone_settings['admin_port'] %>/<%= @ironic_settings[:keystone_version] %>/
-admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
+auth_type = password
+auth_url = <%= @keystone_settings['internal_auth_url'] %>
+username = <%= @ironic_settings[:service_user] %>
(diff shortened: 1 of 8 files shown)


chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -145,11 +145,13 @@ protocol = <%= @glance_server_protocol %>
...
-admin_password = <%= @ironic_settings[:service_password] %>
-admin_url = <%= @keystone_settings['protocol'] %>://<%= @keystone_settings['internal_url_host'] %>:<%= @keystone_settings['admin_port'] %>/<%= @ironic_settings[:keystone_version] %>/
-admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
+auth_type = password
+auth_url = <%= @keystone_settings['internal_auth_url'] %>
+username = <%= @ironic_settings[:service_user] %>
+password = <%= @ironic_settings[:service_password] %>
+project_name = <%= @keystone_settings['service_tenant'] %>
+project_domain_name = <%= @keystone_settings['admin_domain']%>
+user_domain_name = <%= @keystone_settings['admin_domain'] %>
```
jsuchome: Are those last three items really needed? They are not specific to ironic and we're already saving them in keystone section...

(I do not know about ironic, so I'm just curious.. in general the second commit could be better with better message)

jsuchome:
First commit looks good, I think second commit should be explained better.

chef/cookbooks/nova/templates/default/nova.conf.erb
skazi0: 1. They are needed. I tried without them first and it failed to authenticate.
2. This seems to be a new recommendation from upstream to define auth parameters in each section separately with a standardized naming. I think this is due to migration from `keystoneclient` to `keystoneauth` library. See "Configure credentials for accessing other OpenStack services." under https://docs.openstack.org/project-install-guide/baremetal/draft/install-obs.html#configuring-ironic-conductor-service).

skazi0: Relevant reno: https://github.com/openstack/nova/blob/master/releasenotes/notes/deprecate-old-auth-parameters-948d70045335b312.yaml

PR #984, opened 2017-05-31
monasca: log monasca-installer output

jgrassler:
By popular demand: log full debug output from monasca-installer to `/var/log/monasca-installer.log`.

### DIFF
--- a/chef/cookbooks/monasca/files/default/monasca-installer.logrotate
+++ b/chef/cookbooks/monasca/files/default/monasca-installer.logrotate
@@ -1,0 +1,10 @@
+/var/log/monasca-installer.log {
+        weekly
+        rotate 10
+        size 1M
+        copytruncate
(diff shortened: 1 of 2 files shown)


chef/cookbooks/monasca/files/default/monasca-installer.logrotate
```diff
@@ -0,0 +1,10 @@
+/var/log/monasca-installer.log {
```
toabctl: we usually have the logrotate configs in the package...

jgrassler: Usually but not in all cases. For Horizon we have it in Crowbar as well. And I believe in this case it's quite appropriate since the chef recipe determines the log file's location, not the package. If you insist I can move it to the package, though...

PR #985, opened 2017-06-01
tempest: run resources on the execute phase

Itxaka:
For some reason the keystone resources used on the
tempest cookbook are forced to run on the compile phase.

Not only there seems to be no reasoning behind this written
anywhere, but also causes issues in case we modify the keystone
protocol or endpoints, as the resources will run before anything
in the keystone cookbook.

This change makes the resources run at the proper execution phase
so this respects the cookbook ordering, making keystone resources
run always first.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -55,7 +55,7 @@ keystone_register "tempest tempest wakeup keystone" do
   port keystone_settings["admin_port"]
   auth register_auth_hash
   action :wakeup
-end.run_action(:wakeup)
+end
 
 keystone_register "create tenant #{tempest_comp_tenant} for tempest" do
   protocol keystone_settings["protocol"]
@@ -65,7 +65,7 @@ keystone_register "create tenant #{tempest_comp_tenant} for tempest" do
   auth register_auth_hash
   tenant_name tempest_comp_tenant
   action :add_tenant
-end.run_action(:add_tenant)
+end
 
 auth_url = KeystoneHelper.service_URL(
     keystone_settings["protocol"], keystone_settings["internal_url_host"],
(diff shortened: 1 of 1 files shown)


Itxaka:
now to wait and see what breaks :dancer:

cmurphy:
:crossed_fingers:

Itxaka:
wont be enough, there are several resources that are evaluated during compile phase and require a kesytone connection

Itxaka:
This is not the correct syntax for `lazy` in template variables :o

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -334,8 +337,16 @@
...
-public_network_id = `neutron --insecure --os-user-domain-name Default --os-project-domain-name Default --os-username #{tempest_comp_user} --os-password #{tempest_comp_pass} --os-tenant-name #{tempest_comp_tenant} --os-auth-url #{keystone_settings["internal_auth_url"]} net-list -f csv -c id -- --name floating | tail -n 1 | cut -d'"' -f2`.strip
-raise("Cannot fetch ID of floating network") if public_network_id.empty?
+ruby_block "get public network id" do
+  block do
+    cmd = "neutron #{insecure} --os-user-domain-name Default --os-project-domain-name Default"
+    cmd << " --os-username #{tempest_comp_user} --os-password #{tempest_comp_pass}"
+    cmd << " --os-tenant-name #{tempest_comp_tenant}"
+    cmd << " --os-auth-url #{keystone_settings["internal_auth_url"]}"
+    cmd << " net-list -f value -c id --name floating"
+    node[:tempest][:public_network_id] = `#{cmd}`.strip
```
Itxaka: needs a check for the network_id as it could be empty due to neutron being down (seen in the different tests)

Itxaka:
[P2: bsc#1004219 - L3: Keystone fails when changing ssl configuration from default to ssl true](https://trello.com/c/jFwH1U0m/222-p2%3A-bsc%231004219---l3%3A-keystone-fails-when-changing-ssl-configuration-from-default-to-ssl-true)

Itxaka:
Added guards for the new vars to not be empty

PR #986, opened 2017-06-01
nova: disable memory usage statistics (bsc#1020906)

bmwiedemann:
to avoid VMs being stopped during live+block migration
https://bugzilla.suse.com/show_bug.cgi?id=1020906

This change helped on SOC6. Not sure if it also applies to SOC7.
And not sure if it has negative side effects.


toabctl:
@bmwiedemann please prefix the commit msg with ```nova:```

sjamgade:
The change looks good.
but [But will probably fail telemetry](http://specs.openstack.org/openstack/nova-specs/specs/juno/implemented/enabled-qemu-memballoon-stats.html) for memory stats

bmwiedemann:
@sjamgade then the question is, if that telemetry is needed for our customers.

Then we could make this a crowbar option, so I can set it to 0 for my deployment.

sjamgade:
@bmwiedemann, as discussed. This PR can be closed as this was fixed [upstream](https://bugzilla.suse.com/show_bug.cgi?id=1020928)

PR #987, opened 2017-06-01
heat: move keystone-dependant resource to execute phase

Itxaka:
We were doing calls to keystone during the compile phase
to obtain the domain id for heat. This can cause issues
when changing the keystone endpoint protocol.

This patch moves the gathering of the domain id to the
execution phase and makes the variables for the heat
config template to be lazy, so the execution is delayed
until the template is written.

Here its the example for the proper use of `lazy` for template variables: https://github.com/chef-boneyard/chef-web-docs-2016/blob/master/includes_resources_common/includes_resources_common_lazy_evaluation.rst

Needed for https://github.com/crowbar/crowbar-openstack/pull/859

### DIFF
--- a/chef/cookbooks/heat/recipes/server.rb
+++ b/chef/cookbooks/heat/recipes/server.rb
@@ -341,17 +341,23 @@ end
 
 crowbar_pacemaker_sync_mark "create-heat_register" if ha_enabled
 
-shell_get_stack_user_domain = <<-EOF
-  export OS_URL="#{keystone_settings["protocol"]}://#{keystone_settings["internal_url_host"]}:#{keystone_settings["service_port"]}/v3"
-  eval $(openstack \
-    --os-username #{keystone_settings["admin_user"]} \
-    --os-password #{keystone_settings["admin_password"]} \
-    --os-tenant-name #{keystone_settings["admin_tenant"]} \
-    --os-auth-url=$OS_URL \
-    --os-region-name='#{keystone_settings["endpoint_region"]}' \
-    --os-identity-api-version=3 #{insecure} domain show -f shell --variable id #{stack_user_domain_name});
-  echo $id
-EOF
+ruby_block "get stack user domain" do
+  block do
+    url = "#{keystone_settings["protocol"]}://#{keystone_settings["internal_url_host"]}"
+    url << ":#{keystone_settings["service_port"]}/v3"
+    stack_user_domain_id = `openstack \
+--os-username #{keystone_settings["admin_user"]} \
+--os-password #{keystone_settings["admin_password"]} \
+--os-tenant-name #{keystone_settings["admin_tenant"]} \
+--os-auth-url=#{url} \
+--os-region-name='#{keystone_settings["endpoint_region"]}' \
+--os-identity-api-version=3 \
+#{insecure} \
+domain show -f value -c id #{stack_user_domain_name}`
+    raise "Could not obtain the stack user domain id" if stack_user_domain_id.empty?
+    node[:heat][:stack_user_domain_id] = stack_user_domain_id.strip
+  end
+end
 
 rabbit_settings = fetch_rabbitmq_settings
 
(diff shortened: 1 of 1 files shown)


Itxaka:
[P2: bsc#1004219 - L3: Keystone fails when changing ssl configuration from default to ssl true](https://trello.com/c/jFwH1U0m/222-p2%3A-bsc%231004219---l3%3A-keystone-fails-when-changing-ssl-configuration-from-default-to-ssl-true)

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -361,25 +364,29 @@
...
-    heat_metadata_server_url: "#{node[:heat][:api][:protocol]}://#{my_public_host}:#{node[:heat][:api][:cfn_port]}",
-    heat_waitcondition_server_url: "#{node[:heat][:api][:protocol]}://#{my_public_host}:#{node[:heat][:api][:cfn_port]}/v1/waitcondition",
-    heat_watch_server_url: "#{node[:heat][:api][:protocol]}://#{my_public_host}:#{node[:heat][:api][:cloud_watch_port]}",
-    stack_user_domain: %x[ #{shell_get_stack_user_domain} ].chomp,
-    stack_domain_admin: node[:heat]["stack_domain_admin"],
-    stack_domain_admin_password: node[:heat]["stack_domain_admin_password"],
-    trusts_delegated_roles: node[:heat][:trusts_delegated_roles],
-    insecure: keystone_settings["insecure"],
-    heat_ssl: node[:heat][:ssl]
+    lazy {
```
Itxaka: unavoidable issue, this is the syntax for lazy

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -341,17 +341,20 @@
...
-    --os-identity-api-version=3 #{insecure} domain show -f shell --variable id #{stack_user_domain_name});
-  echo $id
-EOF
+ruby_block "get stack user domain" do
+  block do
+    stack_user_domain_id = `openstack \
+--os-username #{keystone_settings["admin_user"]} \
+--os-password #{keystone_settings["admin_password"]} \
+--os-tenant-name #{keystone_settings["admin_tenant"]} \
+--os-auth-url=#{keystone_settings["protocol"]}://#{keystone_settings["internal_url_host"]}:#{keystone_settings["service_port"]}/v3 \
```
Itxaka: really really like to not have to split this as it would look terrible and confusing

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -361,25 +364,29 @@
...
+        verbose: node[:heat][:verbose],
+        rabbit_settings: rabbit_settings,
+        keystone_settings: keystone_settings,
+        database_connection: db_connection,
+        bind_host: bind_host,
+        api_port: api_port,
+        cloud_watch_port: cloud_watch_port,
+        cfn_port: cfn_port,
+        auth_encryption_key: node[:heat][:auth_encryption_key][0, 32],
+        heat_metadata_server_url: "#{node[:heat][:api][:protocol]}://#{my_public_host}:#{node[:heat][:api][:cfn_port]}",
```
Itxaka: really really like to not have to split this as it would look terrible and confusing

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -361,25 +364,29 @@
...
+        heat_waitcondition_server_url: "#{node[:heat][:api][:protocol]}://#{my_public_host}:#{node[:heat][:api][:cfn_port]}/v1/waitcondition",
```
Itxaka: really really like to not have to split this as it would look terrible and confusing

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -361,25 +364,29 @@
...
+        heat_watch_server_url: "#{node[:heat][:api][:protocol]}://#{my_public_host}:#{node[:heat][:api][:cloud_watch_port]}",
```
Itxaka: really really like to not have to split this as it would look terrible and confusing

chef/cookbooks/heat/recipes/server.rb
cmurphy: You could create shorter variables to make this work on one line

aplanas:
A note: I am having problems with heat in v3.0: I can't create stacks there, but I can list. Moving back to v2 works.

jsuchome:
Looks correct, I'll test it together with the keystone PR

Itxaka:
@vuntz I could use the `node.run_state` to store this vars, but storing them in the node seems like nicer to me, so we can use guards to not run those resources in the next runs (I dont think they change?)

I thought we may be able to directly set the vars, as theoretically they end up in the recipe context, but it doesnt seem to work, not sure if its due to the version of chef used or what ¯\\\_(ツ)_/¯

Itxaka:
Added a guard to avoid stack_user_domain_id being empty and a shorter url line for hound pleasure

PR #988, opened 2017-06-01
monasca: fix double monasca-installer run

jgrassler:
The dual execute resources in master.rb caused monasca-installer to be run twice upon initial deployment. This commit replaces them by a single execute resource. It performed well for initial deployment, but please do not merge just yet, since I'd like to test all the corner cases we originally added the dual execute resources for.


jgrassler:
Ok, tested the following corner cases now:

* `monasca-installer` is not run if both crowbar generated configuration is unchanged and `crowbar-openstack`/`monasca-installer` packages have not been updated since last run.
* `monasca-installer`is run if `crowbar-openstack` package has been updated since last run (applies to `monasca-installer` version change as well)
* `monasca-installer` is run if crowbar generated configuration is changed

So no regressions and ready for merge.

PR #990, opened 2017-06-06
Do not try to update keystone endpoint during upgrade, when keystone is not running yet

jsuchome:
Check for "done_os_upgrade" value of upgrade_step.
This is still present when the first chef-client run is executed at the end of upgrade,
it will be deleted once the upgrade is fully finished.


jsuchome:
[Fix keystone endpoint change during the upgrade](https://trello.com/c/KeweyF1p/309-fix-keystone-endpoint-change-during-the-upgrade)

jsuchome:
Merged as part of https://github.com/crowbar/crowbar-openstack/pull/993

PR #991, opened 2017-06-06
keystone: Avoid executing update_endpoint for no reason

Itxaka:
Due to not enough guards, the update_endpoint resource in the
keystone cookbook was executing on every run, causing several
extra calls to the keystone endpoints, including updating
the endpoint on every chef run.

This adds an extra guard to avoid executing it in case
the protocol nor the insecure or port have changed, thus
only updating the endpoints when they need to be updated.

As a side effect it also removes a log line that was not needed.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -234,10 +234,6 @@ end
 
 endpoint_host = my_admin_host
 
-Chef::Log.info("current endpoint: #{endpoint_protocol}://" \
-                "#{endpoint_host}:" \
-                "#{endpoint_port}")
-
 # Update keystone endpoints (in case we switch http/https this will update the
 # endpoints to the correct ones). This needs to be done _before_ we switch
 # protocols on the keystone api.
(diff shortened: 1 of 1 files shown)


chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -255,7 +251,10 @@
   action :update_endpoint
   only_if do
     node[:keystone][:bootstrap] &&
-      (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node))
+      (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)) &&
+      (node[:keystone][:endpoint][:protocol] != node[:keystone][:api][:protocol] ||
+      node[:keystone][:endpoint][:insecure] != node[:keystone][:ssl][:insecure] ||
+      node[:keystone][:endpoint][:port] != node[:keystone][:api][:admin_port])
```
jsuchome: Would it make sense to put these lines into some function?

Itxaka: Well, as far as I know you cannot call functions from a not_if/only_if block (I think?) so this should be a helper instead, which to be used only once..I rather not as I dont see the advantage. Sure its a bit ugly, but its simpler to see what it does on a glance :)

Open to move it to a helper if requested thought.

cmurphy: or maybe in the provider? https://github.com/crowbar/crowbar-openstack/blob/d645606fc9f4f7eb5c4014beaff4cd086a28eb42/chef/cookbooks/keystone/providers/register.rb#L651

Itxaka: Well, I dont really like to deviate too much from the chef workflow as scary stuff happens when you do,  and executing the resource to then do nothing during the execution is a bad workflow, IMHO. I rather do the helper than in the provider.

cmurphy: Should this check endpoint_host too?

Itxaka: Endpoint host is always `endpoint_host = my_admin_host`

Plus I dont think the keystone ssl pr dealt with changing hosts ;)

Itxaka:
[P2: bsc#1004219 - L3: Keystone fails when changing ssl configuration from default to ssl true](https://trello.com/c/jFwH1U0m/222-p2-bsc1004219---l3-keystone-fails-when-changing-ssl-configuration-from-default-to-ssl-true)

PR #993, opened 2017-06-07
keystone: protect against non existing key

Itxaka:
When checking the resource not_if, there could be cases where the
endpoint key is not defined, thus leading to an error.


Itxaka:
Will be included in https://github.com/crowbar/crowbar-openstack/pull/990

Itxaka:
[P2: bsc#1004219 - L3: Keystone fails when changing ssl configuration from default to ssl true](https://trello.com/c/jFwH1U0m/222-p2-bsc1004219---l3-keystone-fails-when-changing-ssl-configuration-from-default-to-ssl-true)

PR #994, opened 2017-06-08
[3.0] Add resource limit controls (bsc#1020922)

cmurphy:

1. Create a systemd_override LWRP and use it in the cinder barclamp.
2. Use the same LWRP in the keystone and horizon barclamps to demonstrate
3. Fix the rabbitmq service resource, which was unable to restart when deployed with pacemaker.
4. Add the parameter to the rabbitmq barclamp, which will either use the systemd_override resource or /etc/security/limits.conf depending on whether it is deployed with pacemaker or not.

Depends on #1034


cmurphy:
Needs more work after feedback from bug reporter.

cmurphy:
This is good to go now

cmurphy:
@Itxaka could you re-review?

PR #996, opened 2017-06-12
monasca, tempest: create missing Monasca roles

jgrassler:
This pull request creates two roles required by the Monasca tempest tests:

* Create monasca-read-only-user (allows read-only access to monasca-api) in
  Monasca barclamp
* Create role `anotherrole` as part of tempest setup (used by Monasca tempest
  tests)

I split them across these two barclamps because it makes sense to generally create  `monasca-read-only-user` in the Monasca barclamp.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -91,6 +91,8 @@ users = [
           {"name" => tempest_adm_user, "pass" => tempest_adm_pass, "role" => "admin" }
         ]
 
+roles = [ 'anotherrole' ]
+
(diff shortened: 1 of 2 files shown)


chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -91,6 +91,8 @@
           {"name" => tempest_adm_user, "pass" => tempest_adm_pass, "role" => "admin" }
         ]
 
+roles = [ 'anotherrole' ]
```
dirkmueller: how is anotherrole related to tempest tests?

PR #997, opened 2017-06-12
cinder: Fix latest migration - delete nfs_mount_options from schema

jsuchome:


jsuchome:
Hm, I do not understand how it appeared in my master branch. But we have a different problem here, probably - mixed schema versions in https://github.com/crowbar/crowbar-openstack/pull/883 and https://github.com/crowbar/crowbar-openstack/pull/972

PR #999, opened 2017-06-13
monasca: Collect logs from monasca components

samirjorina:


dirkmueller:
How do we avoid that there is a loop of some $action triggering a log message which triggers the kibana upload which then triggers the same log message again?

samirjorina:
Log agent will not collect its own logs, which should ensure there is no 'logs-loop'. Not sure if there is any other case this may happen...

samirjorina:
Commits squashed, commit message updated

samirjorina:
Please do not merge it yet - I need to check one more thing.

PR #1001, opened 2017-06-13
Use keystone v3 everywhere (SCRD-781)

cmurphy:
The Identity v2.0 API has been deprecated since Mitaka and will be going away in T. As of the Pike release, usage of the v2.0 API started causing deprecation warnings in the keystone logs. This patch series does not turn off the v2.0 API (so customers can still use it) but it transforms the commands in the keystone_register provider to use the v3 endpoint so that crowbar creating keystone resources will stop causing deprecation warnings.

Along with choosing a different endpoint, I've started renaming parameters and variable names from "tenant" to "project", which is the v3 name for the same idea. This already very large patch series would be even more massive if I updated it everywhere, so this only updates it in the keystone barclamp, keystone_register provider and the keystone_settings helper and we can keep going from there little by little.

Where possible, this also updates the auth sections of config files for various services to use the v3  URL, which also usually means adding project_domain_name and user_domain_name parameters to those sections.

This also does a bit of cleanup of the keystone_register provider to enhance readability and DRY-ness.

### DIFF
--- a/chef/cookbooks/aodh/recipes/aodh.rb
+++ b/chef/cookbooks/aodh/recipes/aodh.rb
@@ -69,7 +69,7 @@ keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
 register_auth_hash = { user: keystone_settings["admin_user"],
                        password: keystone_settings["admin_password"],
-                       tenant: keystone_settings["admin_tenant"] }
+                       project: keystone_settings["admin_project"] }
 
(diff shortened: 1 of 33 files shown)


cmurphy:
Added the do-not-merge label since if we want to backport #1223 it will be easier to get that in first and then rebase on top of it.

cmurphy:
The testbuild job wasn't kicking off the mkcloud job, attempted a force push to start it over

PR #1003, opened 2017-06-13
Fix mysql cookbook for SLES12

rhafer:
Some simple fixes to make mysql/mariadb work again on SLES 12.

This is currently not really required an just the result of a small experiment with mariadb on Cloud 7. It might be interesting for our way to Cloud 8 though.

(There is no HA support)


jsuchome:
I do not know if all those changes are correct/needed, but the relevant code changes look fine.

PR #1004, opened 2017-06-14
Hide the rabbitmq management plugin from admin network

dirkmueller:
As it is not behind SSL and the management plugin has some security
issues, it might become an additional attack vector.


dirkmueller:
if new deployment works, we need an upgrade story

PR #1005, opened 2017-06-14
heat: revert to using older config options

sayalilunkad:
Tempest seems to pass with this change.


rsalevsky:
Can you mention in the commit message that this partly reverts eef7a80fe8835cd9e8ea2401039bd9d532799268?

sayalilunkad:
@rsalevsky @s-t-e-v-e-n-k edited the commit message, and the revert is the fix.

PR #1011, opened 2017-06-16
monasca: fix monitoring for HA clouds

jgrassler:
This commit changes the Openstack service HTTP checks to

  1) Monitor the service on the node itself for non-HA clouds
  2) Monitor the service on the pacemaker VIP for HA enabled
     clouds.

Previously, HA clouds would show all services as down because only the active
node would report any given service as up whereas the others would report it as
down, resulting in a net status of down.

Marking as WIP since this is as yet untested.

### DIFF
--- a/chef/cookbooks/barbican/recipes/monitor_monasca.rb
+++ b/chef/cookbooks/barbican/recipes/monitor_monasca.rb
@@ -17,8 +17,14 @@
 return unless node["roles"].include?("monasca-agent")
 
 network_settings = BarbicanHelper.network_settings(node)
-bind_host = network_settings[:api][:ha_bind_host]
-bind_port = network_settings[:api][:bind_port]
+
+bind_port = node[:barbican][:api][:bind_port]
+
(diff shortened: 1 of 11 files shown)


toabctl:
@jgrassler hm. I don't get this. I thought we want to monitor (in the HA case) **all** services behind haproxy and the VIP.

jgrassler:
Unfortunately most services run on the active node only in our HA architecture. So if we monitor the services we get 1 node reporting the service as up and n-1 nodes reporting it as down, which gives us a net result of "the service is down".

jgrassler:
Abandoning this in favour of https://github.com/crowbar/crowbar-openstack/pull/1011: If we monitor the VIP the way we do it here we may still get `n` conflicting results as well, since any of the `n` agents monitoring it might fail to reach it for some reason. Hence monitoring the VIP has no real benefit over monitoring individual backend services. Also, we lose the information "one of the backends is to blame for the DOWN status".

PR #1012, opened 2017-06-19
Allow deployment of OpenStack roles to SP3 nodes only.

jsuchome:
Since we allow deployments of 2 different SP versions, make sure only SP3 are used for openStack.

(SP2 nodes are for ceph)

### DIFF
--- a/crowbar_framework/app/models/aodh_service.rb
+++ b/crowbar_framework/app/models/aodh_service.rb
@@ -32,7 +32,7 @@ class AodhService < PacemakerServiceObject
           "unique" => false,
           "count" => 1,
           "exclude_platform" => {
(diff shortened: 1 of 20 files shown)


jsuchome:
[#cloud8 Make OpenStack barclamps work for SP3](https://trello.com/c/DxMDobCg/33-cloud8-make-openstack-barclamps-work-for-sp3)

jsuchome:
@aplanas these values are not in "exclude_platform" but just "platform"

aplanas:
Ah no, for some reason I was thinking that this was JSON instead of a Ruby hash, and the extra comma can cause a problem in the former, but not on the latter.

PR #1013, opened 2017-06-19
heat: revert to using older config options

sayalilunkad:
Tempest seems to pass with this change.

partial revert: https://github.com/crowbar/crowbar-openstack/commit/eef7a80fe8835cd9e8ea2401039bd9d532799268

(cherry picked from commit d0e707c)


aplanas:
is this a cherry-pick from master? if so can we do the -x?

sayalilunkad:
@aplanas I actually didn't need to cherry-pick because the stable branch was recently created. So I just created a new PR from my original branch. Would it be better to cherry-pick from master?

aplanas:
No need to create a new PR, you can reuse this.

IMHO cherry-picks from master to stable provide better code management, so +1 for doing in this way for me

sayalilunkad:
@aplanas done, thanks

PR #1018, opened 2017-06-21
keystone: better keystone http/https change logic (bsc#1044266)

Itxaka:
Before this patch, we modified the role and saved it which
caused the crowbar-revision to inscrease, which could lead to
desynchronization when applying a barclamp to a subset of
the keystone-server nodes.

This patch avoid that by directly saving the endpoint information
to the keystone-server nodes after the chef run, plus as a side
effect simplifies the code by avoiding any role modification
before the chef run is done.

### DIFF
--- a/crowbar_framework/app/models/keystone_service.rb
+++ b/crowbar_framework/app/models/keystone_service.rb
@@ -110,20 +110,6 @@ class KeystoneService < PacemakerServiceObject
 
     server_elements, server_nodes, ha_enabled = role_expand_elements(role, "keystone-server")
 
-    unless old_role.nil?
-      old_elements = old_role.override_attributes["keystone"]
-      attributes = role.override_attributes["keystone"]
-
-      # Copy old keystone endpoints to new role.
-      @logger.debug("old_elements: #{old_elements.pretty_inspect}")
-      if old_elements.key?("endpoint")
-        @logger.debug("copying old endpoint")
-        attributes["endpoint"] = old_elements["endpoint"].deep_dup
-      else
-        @logger.debug("old role did not have endpoints; taking current endpoint settings")
-      end
-    end
-
     reset_sync_marks_on_clusters_founders(server_elements)
     Openstack::HA.set_controller_role(server_nodes) if ha_enabled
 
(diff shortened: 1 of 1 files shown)


Itxaka:
[Keystone's crowbar-revsion increasing after chef-client run](https://trello.com/c/triedtIl/322-keystones-crowbar-revsion-increasing-after-chef-client-run)

Itxaka:
Tested on non-HA and works, deploying HA env to test.

Also, wow, hound didnt complain, thats gotta be a first for me! :dog:

cmurphy:
Is this the fix for the ceph synchronization problem?

jsuchome:
That ha job fails in manila deployment, because there's no ceph-mds role. This is quite complicated in current SOC8 code ... and not relevant to this PR.

aplanas:
@Itxaka can you try with this scenario? https://ci.suse.de/view/Cloud/view/Cloud7/job/cloud-mkcloud7-job-ha-linuxbridge-x86_64/

nicolasbock:
@Itxaka When I wrote the code to store the endpoints in the role, I was concerned with what would happen if a user would change out the set of Keystone servers and replace them. If the endpoint information is stored in the nodes, wouldn't that mean that this information is lost then?

Itxaka:
> Is this the fix for the ceph synchronization problem?
@cmurphy yes, it may fix it :ok_hand: 

> @Itxaka can you try with this scenario? https://ci.suse.de/view/Cloud/view/Cloud7/job/cloud-mkcloud7-job-ha-linuxbridge-x86_64/

@aplanas trigerred and added to the checklist :clap: 

> @Itxaka When I wrote the code to store the endpoints in the role, I was concerned with what would happen if a user would change out the set of Keystone servers and replace them. If the endpoint information is stored in the nodes, wouldn't that mean that this information is lost then?

@nicolasbock I would need more context to understand this correctly. Do you mean like a customer just trashing their keystone nodes and adding newer ones? I guess that in that case, yes it would be lost, but it would not matter as kesytone would be clean configured from zero so no changes needed to the endpoint, just the creation of newer endpoints.

nicolasbock:
@Itxaka Yes, that's what I was thinking of. But I think you have a point. If the customer really replaces keystone then they would have to re-bootstrap the database anyhow. Thanks!

Itxaka:
Actually the normal mkcloud job worked correctly: https://ci.suse.de/job/openstack-mkcloud/68664/

But for some reason, the jobs in github are mixed, probably a jenkins context issue :/

Will try to retrigger them properly.

Itxaka:
swift was already deployed on the HA job so it seems that its working as expected, only that the jenkins context issue leads to look like it doesnt :)

Itxaka:
```Error (105): No suitable node(s) for nova-compute-kvm found.``` :thinking:

jsuchome:
> Error (105): No suitable node(s) for nova-compute-kvm found. ￼

That's likely because small number of nodes - 3 get used for ceph and there's noting left for compute. I was thinging https://github.com/SUSE-Cloud/automation/pull/2026 could fix it

kbaikov:
@Itxaka  I applied this patch manually but i still get the error on applying swift barclamp: 

RuntimeError: crowbar-pacemaker_sync_mark[sync-swift_before_ha] (swift::proxy_ha line 29) had an error: RuntimeError: Some cluster nodes didn't set swift_before_ha to 14!

This is develcloud7. It is currently installed on qa4 hardware, if you are interested in debugging that please ssh to 10.162.66.10 and take a look.

Thanks.

Itxaka:
@kbaikov After checking it out it seems a completely different issue :)

PR #1019, opened 2017-06-22
tempest: add support for vmware images

stefannica:
This PR updates the tempest cookbook to create
and/or use qcow or vmware/vmdk images instead of the regular
ami type images, when VMware compute nodes are enabled.

This allows a user to control what is targeted by tempest
when both kvm and vmware compute nodes are present:
  - if an 'ami' type image file is supplied through the
  'tempest_test_images' barclamp attribute (the previous
  and current default behavior), tempest will only target the
  kvm backend and exclude vmware functionality from its scope.
  In this configuration, tempest will test if the kvm backend
  is functioning properly in a kvm-vmware multi-backend environment.
  - if a raw or vmdk image is supplied, tempest will do the opposite:
  it will target the vmware backend and exclude kvm functionality
  from its scope. A raw image file will also be automatically
  converted to its vmdk counterpart.

If no vmware compute nodes are present, the behaviour is unchanged.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -175,6 +175,11 @@ alt_machine_id_file = node[:tempest][:tempest_path] + "/alt_machine.id"
 
 insecure = "--insecure"
 
+kvm_compute_nodes = search(:node, "roles:nova-compute-kvm") || []
+vmware_compute_nodes = search(:node, "roles:nova-compute-vmware") || []
+vmware_enabled = vmware_compute_nodes.any?
+vmware_only = vmware_enabled && kvm_compute_nodes.empty?
+
 tempest_test_image = node[:tempest][:tempest_test_images][node[:kernel][:machine]]
 
 cirros_version = File.basename(tempest_test_image).split("-")[1]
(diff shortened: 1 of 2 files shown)


toabctl:
@stefannica Shouldn't be the goal to use a working image for VMWare instead of trying to avoid the vmware compute nodes?

stefannica:
@toabctl, I gave it some thought, but I thought it would break some of the tempest testcases that aren't prepared to handle VMware's compute peculiarities, not to mention the added non-deterministic effect of nova's scheduler. The fact that this was never previously attempted in VMware QA jobs wasn't looking good either (I will attempt to find out why).
But now that I know I'm not the only one thinking about it, I will give it try. Thanks for suggesting it.

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -233,7 +235,9 @@
     --name="$IMG_NAME-tempest-machine-alt" \
     --visibility public --container-format ami --disk-format ami \
     --property kernel_id=$KERNEL_ID \
-    --property ramdisk_id=$RAMDISK_ID < $(findfirst '*.img') | extract_id)
+    --property ramdisk_id=$RAMDISK_ID \
+    #{vmware_compute_nodes.empty? ? "--property hypervisor_type='qemu'" : ""} \
```
dirkmueller: Isn't that the reversed condition of what you want to do ? _

PR #1020, opened 2017-06-27
monasca: Fix the host configuration of curator

knakazono:
This change move the double quotes into the brackets.


knakazono:
This patch requires backporting to stable/4.0 branch.

knakazono:
Thanks for reviewing!
I updated the commit message.

PR #1021, opened 2017-06-27
[4.0] keystone: better keystone http/https change logic (bsc#1044266)

Itxaka:
Before this patch, we modified the role and saved it which
caused the crowbar-revision to increase, which could lead to
desynchronization when applying a barclamp to a subset of
the keystone-server nodes.

This patch avoids that by directly saving the endpoint information
to the keystone-server nodes after the chef run, plus as a side
effect, simplifies the code by avoiding any endpoint attributes
modification before the chef run is done as its not really needed.

(cherry picked from commit 4d8a928af1be899394b3f9ef56187a060676611c)

### DIFF
--- a/crowbar_framework/app/models/keystone_service.rb
+++ b/crowbar_framework/app/models/keystone_service.rb
@@ -110,20 +110,6 @@ class KeystoneService < PacemakerServiceObject
 
     server_elements, server_nodes, ha_enabled = role_expand_elements(role, "keystone-server")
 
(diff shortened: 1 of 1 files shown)


Itxaka:
[Keystone's crowbar-revsion increasing after chef-client run](https://trello.com/c/triedtIl/322-keystones-crowbar-revsion-increasing-after-chef-client-run)

PR #1025, opened 2017-06-28
postgresql: Bump the database version for SLES12SP3

jsuchome:


jsuchome:
I assume Leap version needs to be updated as well...

skazi0:
@jsuchome this is only for new installations, right? It will not force 9.4->9.6 upgrade when upgrading node to SP3?

jsuchome:
@skazi0  I think relevant chef recipe will install the package if it is not there, so it actually affects potential upgrade as well

skazi0:
@jsuchome if there was no postgresql before then it should not hurt. The problem is that afaik postgresql doesn't do "in-place" upgrades so if there was 9.4 with data before, installing 9.6 would not migrate the data and we will have two engines running in parallel. For upgrade case we will probably need to handle the data migration anyway so I think this PR is fine for now.

jsuchome:
> For upgrade case we will probably need to handle the data migration anyway 
Exactly

jsuchome:
[#cloud8 Make OpenStack barclamps work for SP3](https://trello.com/c/DxMDobCg/33-cloud8-make-openstack-barclamps-work-for-sp3)

PR #1026, opened 2017-06-28
monasca: show Monasca barclamp

jgrassler:
I think we can flick the switch on this one now.


dirkmueller:
definitly good to go for cloud8

PR #1029, opened 2017-06-28
ha: Fix postgresql and rabbitmq crm services

cmurphy:
Needed for #974


dirkmueller:
(rebased to retrigger gating)

PR #1035, opened 2017-07-03
monasca: Hardwire grafana timezone to UTC

samirjorina:
Requires backporting to stable/4.0


toabctl:
@jgrassler could you mention in the commit msg why we need this?

jgrassler:
@toabctl: s/jgrassler/samirjorina/ (I didn't create this pull request) :-)

@samirjorina: can you please add a body to the commit message that outlines the problem we're solving here? (something like "If the browser time zone differs from the time zone of the `monasca-server` node we may get 'no results found' until local time in the browser catches up with the `monasca-server node's time zone for the first time.")

samirjorina:
Commit message updated

PR #1038, opened 2017-07-04
horizon: Fix SSLCipherSuite for SOC6 (bsc#1043484)

aplanas:
Update SSLCipherSuite to avoid the wrong expansion from mod_ssl in
httpd-2.4.16 on SP1 (for a better explanation check the bug report)

https://bugzilla.suse.com/show_bug.cgi?id=1043484

### DIFF
--- a/chef/cookbooks/horizon/templates/suse/openstack-dashboard.conf.erb
+++ b/chef/cookbooks/horizon/templates/suse/openstack-dashboard.conf.erb
@@ -26,7 +26,7 @@
 
 <VirtualHost <%= @bind_host %>:<%= @bind_port_ssl %>>
     SSLEngine On
-    SSLCipherSuite DEFAULT_SUSE
+    SSLCipherSuite EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH
     SSLProtocol all -SSLv2 -SSLv3
     SSLCertificateFile <%= @ssl_crt_file %>
     SSLCertificateKeyFile <%= @ssl_key_file %>


aplanas:
Lets wait and test in the SSL scenario before merging

dirkmueller:
```
   curl -L -m 120 -s -S -k http://d52-54-77-77-01-01.vf2.cloud.suse.de
```

looks like the horizon test is not adjusting to ssl, could you check?

aplanas:
I am on it

bmwiedemann:
@dirkmueller I had made a PR for that last week, but then noticed that there should be a redirect page on port 80, so we closed the PR

aplanas:
Well, clearly the problem is still there. The append still happens, a is not working with DEFAULT and DEFAULT_SUSE, but is OK with ALL and HIGH, and the rest.

openssl is patched for DEFAULT and DEFAULT_SUSE, so I think that a workaround for now is manually expand DEFAULT, or better, go to the same direction that https://github.com/crowbar/crowbar-openstack/commit/643b4461669479480ef571409c2cc9c1e1cb92a5

aplanas:
Retesting in https://ci.suse.de/job/openstack-mkcloud/71009/

aplanas:
https://ci.suse.de/job/openstack-mkcloud/71146/console shows a green SSL deployment with this new version

rsalevsky:
We should wait for some feedback from the security team as they recommended the previous change.

rsalevsky:
Seams to work now.

PR #1041, opened 2017-07-05
[4.0] Add resource limit controls (bsc#1020922)

cmurphy:
Backport of #974

Create a systemd_override LWRP and use it in the cinder barclamp.
Use the same LWRP in the keystone and horizon barclamps to demonstrate
Fix the rabbitmq service resource, which was unable to restart when deployed with pacemaker.
Add the parameter to the rabbitmq barclamp, which will either use the systemd_override resource or /etc/security/limits.conf depending on whether it is deployed with pacemaker or not.

Depends on #1040


jsuchome:
> Backport of #1029

Looks like copy & paste error. Could you please link correct PR?

cmurphy:
@jsuchome fixed

cmurphy:
Closing in favor of #1393

PR #1042, opened 2017-07-06
neutron: Setting apic to required:false

mmnelemane:
This needs to be done  because
- default ml2_mechanism_driver is not apic 
- the json template does not have the default settings
- setting it to true causes error while applying barclamp (no 'apic' object)

### DIFF
--- a/chef/data_bags/crowbar/template-neutron.schema
+++ b/chef/data_bags/crowbar/template-neutron.schema
@@ -45,7 +45,7 @@
                       "ovsdb_interface": { "type": "str", "required": true },
                       "of_interface": { "type": "str", "required": true }
                     }},
-                    "apic": { "type": "map", "required": true, "mapping": {
+                    "apic": { "type": "map", "required": false, "mapping": {
                       "hosts": { "type" : "str", "required" : true },
                       "system_id": { "type" : "str", "required" : true },
                       "username": { "type" : "str", "required": true },


dirkmueller:
@mmnelemane change commit message to start with "neutron:"

Itxaka:
[P2: Bug 1045783 - #L3 Keystone SSL error on on control node after applying PTF for BZ#1032756](https://trello.com/c/Q9KXh7yl/329-p2-bug-1045783-l3-keystone-ssl-error-on-on-control-node-after-applying-ptf-for-bz1032756)

mmnelemane:
@toabctl Since "apic" is not a default feature, its not supplied in the default JSON template. So if "required" is set to true and the JSON template doesn't provide default parameters, it throws an error <Object "apic" not found>. It is implemented on the lines similar to "cisco_switches". The other 2 ways to handle this could be 1. To use a "use_apic" variable 2. To add dummy values to template-neutron.json for the "apic" field. Please suggest.

s-t-e-v-e-n-k:
But apic is supplied in the default JSON template?

https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/template-neutron.json#L43

mmnelemane:
@s-t-e-v-e-n-k : yes. I probably found the bug in a slightly different setup. Closing this PR since it doesn't fix anything.

mmnelemane:
@s-t-e-v-e-n-k : Yes. I probably tested on a slightly different setup. Closing this PR since the fix is not required.

PR #1046, opened 2017-07-06
monasca: fix health check URL for monasca-log-api.

jgrassler:
The monasca-log-api health check had an extra "v3.0" in its URL which
caused it to always fail. This commit generates a proper URL.

### DIFF
--- a/chef/cookbooks/monasca/libraries/helper.rb
+++ b/chef/cookbooks/monasca/libraries/helper.rb
@@ -127,6 +127,17 @@ module MonascaHelper
     return monasca_log_api_url
   end
 
+  # Returns a log API health check URL for use by a Monasca agent's http_check
+  # plugin
+  def self.log_api_healthcheck_url(node)
+    my_net = node[:monasca][:network]
+    port = node[:monasca][:log_api][:bind_port]
+    listen_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(
+      node, my_net).address
+
+    "http://#{listen_ip}:#{port}/healthcheck"
+  end
+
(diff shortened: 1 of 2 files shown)


chef/cookbooks/monasca/libraries/helper.rb
```diff
@@ -127,6 +127,17 @@ def self.log_api_network_url(node)
     return monasca_log_api_url
   end
 
+  # Returns a log API health check URL for use by a Monasca agent's http_check
+  # plugin
+  def self.log_api_healthcheck_url(node)
```
toabctl: hm. This changes the behavior a lot. Previously (while using ```MonascaHelper.log_api_network_url()```, ```node[:monasca][:log_api][:url]``` was taken into account. If that is no longer needed you should at least reuse ```log_api_public_url()``` (with setting version to nil)

jgrassler: I was toying with that idea, but eventually I opted for this approach anyway. Two reasons:

1) It abuses version for something it's not really meant for. If the API ever changes to have an extra component in front of the version or something like that, this would unexpectedly break `log_api_healthcheck_url()` as well.
2) Currently, `node[:monasca][:log_api][:url]` is user set. It shouldn't be that way, but it is. Looking up the IP address is not pretty either, but it's more reliable than user input.

If you really prefer `node[:monasca][:log_api][:url]` I can use it here as well, but I'd rather use it as little as possible (preferably not at all, but I guess that's something for the second release).

toabctl:
Backport is to stable/4.0 is https://github.com/crowbar/crowbar-openstack/pull/1047

PR #1050, opened 2017-07-07
monasca: set API URLs automatically

jgrassler:
This commit points the API URLs to the Monasca node's IP
address on the network selected in node[:monasca][:network] if
they were are not specified explicitly.

Note: I haven't tested this, yet, so please hold off on merging until I've verified it actually works as intended.


jgrassler:
Alright, provided the Hound fixes didn't break anything substantial this one is ready for review/testing.

To test this:

1) Create a cloud with two dedicated networks for Monasca
2) Ensure that your monasca-server node has interfaces with an IP address assigned on both these networks
3) Ensure all nodes with the `monasca{,-log}-agent` roles have an interface on the network named in `node[:monasca][:network][:clients]` (technically a route through which they can reach the `monasca-server` node's interface on that network, but that would defeat the purpose of having a dedicated network in the first place)

jgrassler:
I tested this against a fresh cloud and after the small fix[0] I added it both deploys Monasca just fine and complains in the situations where it's supposed to complain (missing network interface/allocated IP on that interface).

[0] missing question mark from making Hound happy.

PR #1051, opened 2017-07-10
ceilometer: Use ceilometer-upgrade command

toabctl:
Since Ocata, 'ceilometer-dbsync' is no longer
available. 'ceilometer-upgrade' is the new command.

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/server.rb
+++ b/chef/cookbooks/ceilometer/recipes/server.rb
@@ -137,10 +137,12 @@ my_public_host = CrowbarHelper.get_host_for_public_url(node,
                                                        ceilometer_protocol == "https",
                                                        ha_enabled)
 
-crowbar_pacemaker_sync_mark "wait-ceilometer_db_sync" if ha_enabled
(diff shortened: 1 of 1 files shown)


chef/cookbooks/ceilometer/recipes/server.rb
```diff
@@ -137,10 +137,10 @@
                                                        ceilometer_protocol == "https",
                                                        ha_enabled)
 
-crowbar_pacemaker_sync_mark "wait-ceilometer_db_sync" if ha_enabled
+crowbar_pacemaker_sync_mark "wait-ceilometer_upgrade" if ha_enabled
 
-execute "ceilometer-dbsync" do
-  command "ceilometer-dbsync"
+execute "ceilometer-upgrade" do
+  command "ceilometer-upgrade --skip-gnocchi-resource-types"
```
jsuchome: Why is `--skip-gnocchi-resource-types` needed? Could you add a comment?

toabctl: Done

PR #1055, opened 2017-07-10
neutron: Correct schema revision. Avoid duplicate versions.

mmnelemane:
Corrected  schema revision. Two migration scripts with the same version number (109) were found. The later of them is moved to 111 and the schema_revision is updated accordingly.


s-t-e-v-e-n-k:
This broadly looks fine, but you should update the commit message body to include roughly the same content as the PR description to help people looking over the history at a later date.

jsuchome:
Should not be needed after https://github.com/crowbar/crowbar-openstack/pull/1058

toabctl:
Pr#1058 merged so this is no longer needed.

PR #1057, opened 2017-07-10
Enable Ocata for SOC8

toabctl:
The nova-placement-api service is mandatory in Ocata.

### DIFF
--- a/chef/cookbooks/nova/recipes/database.rb
+++ b/chef/cookbooks/nova/recipes/database.rb
@@ -31,7 +31,7 @@ crowbar_pacemaker_sync_mark "wait-nova_database" do
   only_if { ha_enabled }
 end
 
-[node[:nova][:db], node[:nova][:api_db]].each do |d|
+[node[:nova][:db], node[:nova][:api_db], node[:nova][:placement_db]].each do |d|
   # Creates empty nova database
   database "create #{d[:database]} database" do
     connection db_settings[:connection]
--- a/chef/cookbooks/nova/recipes/database.rb
+++ b/chef/cookbooks/nova/recipes/database.rb
@@ -103,8 +103,8 @@ execute "nova-manage db online_data_migrations" do
   end
 end
 
-# Update Nova DB to revision 334 (most recent Newton migration)
-execute "nova-manage db sync (continue)" do
+# Update Nova DB to latest revision
+execute "nova-manage db sync" do
   user node[:nova][:user]
   group node[:nova][:group]
   command "nova-manage db sync"
--- a/chef/cookbooks/nova/recipes/database.rb
+++ b/chef/cookbooks/nova/recipes/database.rb
@@ -63,6 +63,29 @@ end
   end
 end
 
+# create the nova_cell0 database (similar to nova_api) and give the
+# nova_api DB user the correct privileges
+database "create nova_cell0 database" do
+  connection db_settings[:connection]
+  database_name "nova_cell0"
+  provider db_settings[:provider]
+  action :create
+  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+database_user "grant privileges to the #{node[:nova][:api_db][:user]} database user" do
+  connection db_settings[:connection]
+  database_name "nova_cell0"
+  username node[:nova][:api_db][:user]
+  password node[:nova][:api_db][:password]
+  host "%"
+  privileges db_settings[:privs]
+  provider db_settings[:user_provider]
+  action :grant
+  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+
 execute "nova-manage api_db sync" do
   user node[:nova][:user]
   group node[:nova][:group]
@@ -72,6 +95,33 @@ execute "nova-manage api_db sync" do
   # are the founder of the HA cluster (so that it's really only done once).
   only_if do
     !node[:nova][:api_db_synced] &&
+      (!node[:nova][:ha][:enabled] || CrowbarPacemakerHelper.is_cluster_founder?(node))
+  end
+end
+
+# handle cell0 and cell1 before doing the db sync
+execute "nova-manage create cell0" do
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage cell_v2 map_cell0"
+  action :run
+  # TODO: Does not work on Newton (14.x.x). Remove when switched to Ocata
+  not_if 'rpm -qa --qf "%{VERSION}\n" openstack-nova|grep ^14\.'
+  only_if do
+    !node[:nova][:db_synced] &&
+      (!node[:nova][:ha][:enabled] || CrowbarPacemakerHelper.is_cluster_founder?(node))
+  end
+end
+
+execute "nova-manage create cell1" do
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage cell_v2 create_cell --name cell1 --verbose"
+  action :run
+  # TODO: Does not work on Newton (14.x.x). Remove when switched to Ocata
+  not_if 'rpm -qa --qf "%{VERSION}\n" openstack-nova|grep ^14\.'
+  only_if do
+    !node[:nova][:db_synced] &&
       (!node[:nova][:ha][:enabled] || CrowbarPacemakerHelper.is_cluster_founder?(node))
   end
 end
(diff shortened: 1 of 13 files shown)


chef/cookbooks/nova/templates/default/nova-placement.conf.erb
```diff
@@ -0,0 +1,14 @@
...
+auth_type = password
+auth_url = <%= @keystone_settings['admin_auth_url'] %>
+project_name = <%= @keystone_settings['service_tenant'] %>
+project_domain_name = <%= @keystone_settings["admin_domain"] %>
+username = <%= @username %>
+user_domain_name = <%= @keystone_settings['admin_domain'] %>
+password = <%= @keystone_settings['service_password'] %>
+
+[placement_database]
+<% if @placement_database_connection -%>
```
s-t-e-v-e-n-k: I'm curious about this guard -- from my reading of the code, it will always be set to something, which means this isn't required?

toabctl: iirc it is not set on the compute nodes where no DB connection (and credentials) are needed.

s-t-e-v-e-n-k: That makes sense, in which case perhaps move the guard one line up if possible so the section isn't emitted either on compute nodes.

toabctl: Done

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -76,6 +99,29 @@
   end
 end
 
+# handle cell0 and cell1 before doing the db sync
+execute "nova-manage create cell0" do
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage cell_v2 map_cell0"
```
cmurphy: No --verbose for this one? (I see it's not in the install guide but we could still be consistent here)

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -76,6 +99,29 @@
...
+  action :run
+  only_if do
+    !node[:nova][:db_synced] &&
```
cmurphy: This seems like an error, there should be a separate flag that subscribes to "execute[nova-manage create cell0]", the db_synced flag subscribes to a different exec

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -76,6 +99,29 @@
...
+  end
+end
+
+execute "nova-manage create cell1" do
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage cell_v2 create_cell --name cell1 --verbose"
+  action :run
+  only_if do
+    !node[:nova][:db_synced] &&
```
cmurphy: Same with this one, should be using a different flag

chef/data_bags/crowbar/migrate/nova/201_discover_hosts_in_cells_interval.rb
```diff
@@ -0,0 +1,10 @@
+def upgrade(ta, td, a, d)
+  a["scheduler"]["discover_hosts_in_cells_interval"] = \
+    ta["scheduler"]["discover_hosts_in_cells_interval"]
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["scheduler"].delete("discover_hosts_in_cells_interval")
```
cmurphy: No need to guard here?

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -103,8 +149,8 @@
   end
 end
 
-# Update Nova DB to revision 334 (most recent Newton migration)
-execute "nova-manage db sync (continue)" do
```
cmurphy: Looks like line 119 needs to be fixed too

chef/cookbooks/nova/recipes/database.rb
cmurphy: I guess it happens to work because the db sync happens later in the sequence but I still think it would be better if it has its own in case they get out of sync.

chef/cookbooks/nova/recipes/placement_api.rb
```diff
@@ -0,0 +1,131 @@
...
+node.normal[:apache][:listen_ports_crowbar][:nova] = { plain: [bind_port] }
+
+ruby_block "setup nova-placement-api wsgi script" do
+  block do
+    Dir.mkdir("/srv/www/nova-placement-api", 0o755) \
+      unless Dir.exist?("/srv/www/nova-placement-api")
+
+    # The wsgi script is installed in /usr/bin. We copy that script to make
+    # its name consistent with what the crowbar_openstack_wsgi resource
+    # expects. We should think about doing that in the package instead.
```
toabctl: Done

chef/data_bags/crowbar/migrate/nova/200_placement_api.rb
```diff
@@ -0,0 +1,15 @@
...
+  if a["placement_db"]["password"].nil? || a["placement_db"]["password"].empty?
+    service = ServiceObject.new "fake-logger"
+    a["placement_db"]["password"] = service.random_password
+  end
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("placement_db")
```
toabctl: No. I don't think we want to backport the placement API for SOC7

chef/cookbooks/nova/recipes/database.rb
toabctl: There is no --verbose for that subcommand. Just the global --verbose.

chef/cookbooks/nova/recipes/database.rb
toabctl: Hm. That is done by intention. After the cell setup, ```nova-manage db sync``` needs to be called to init the cell DB. That's why I bound the cell commands to the ```node[:nova][:db_synced]``` flag.
It can get out-of-sync, but if we introduce antother flag for cells, it can get out-of-sync, too (when cells failed for some reason but ```node[:nova][:db_synced]``` is set to ```true```. Then the cell setup is broken, too

chef/cookbooks/nova/recipes/database.rb
toabctl: same reason

chef/data_bags/crowbar/migrate/nova/201_discover_hosts_in_cells_interval.rb
toabctl: no backport needed so no guard needed.

chef/cookbooks/nova/recipes/database.rb
toabctl: Hm. Why? Nothing subscribes to ```nova-manage db online_data_migrations (continue)```. Or do you mean something different?

cmurphy: Ah I guess I just assumed it was a mistake

chef/cookbooks/nova/recipes/database.rb
cmurphy: hmm okay

chef/cookbooks/nova/recipes/database.rb
```diff
@@ -76,6 +99,29 @@
...
+  only_if do
+    !node[:nova][:db_synced] &&
+      (!node[:nova][:ha][:enabled] || CrowbarPacemakerHelper.is_cluster_founder?(node))
+  end
+end
+
+execute "nova-manage create cell1" do
+  user node[:nova][:user]
+  group node[:nova][:group]
+  command "nova-manage cell_v2 create_cell --name cell1 --verbose"
```
toabctl: Hm. That command is not available on Newton. I'll try the simple setup command which should be available for Newton and Ocata...

toabctl: Using the ```simple_cell_setup``` command with Newton does not work. So I guess we need to merge this and then move the packages to D:C:8 . Otherwise the CI will be broken...

dirkmueller: can you add a "only_if { %x{ nova-manage --version } =~ magicverison }" to fix newton and ocata?

toabctl: @dirkmueller hopefully fixed (but not tested yet)

toabctl:
@cmurphy @dirkmueller could you please have another look?

toabctl:
Also tested locally with ocatacloud8 and 2-node scenario and all tempest tests passed.

PR #1058, opened 2017-07-10
Fix schema migrations

toabctl:


chef/data_bags/crowbar/template-neutron.json
```diff
@@ -172,7 +172,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 110,
```
jsuchome: `add_cisco_apic` was revision 109, didn't you rename wrong file?

toabctl: Hm. True. and there might be more fixes needed.

s-t-e-v-e-n-k:
https://github.com/crowbar/crowbar-openstack/pull/1055 also deals with the neutron schema change, however, I like this better due to headroom for backports.

PR #1060, opened 2017-07-11
glance: Restrict image_location metadata (bsc#1023507)

aplanas:
When `show_multiple_locations` is enabled in Glance, any user
can rewrite the metadata information for locations, causing
a securiy breach.

This patch limit to admin the editing of image_location metadata
when this feature is enabled.

### DIFF
--- a/chef/cookbooks/glance/recipes/api.rb
+++ b/chef/cookbooks/glance/recipes/api.rb
@@ -103,6 +103,26 @@ template "/etc/glance/glance-swift.conf" do
   notifies :restart, "service[#{node[:glance][:api][:service_name]}]"
 end
 
+# If `show_multiple_locations` is enabled we need to update
+# policy.json in a more restrictive way.  Only admin (or image owner)
+# can now update the `image_location` metadate field.
+#
+# For more information check bsc#1023507, and upstream bug:
+# https://bugs.launchpad.net/ossn/+bug/1549483
+#
+# NOTE(aplanas) -- policy.json needs to be synchronized to current
+# version of OpenStack.
+template "/etc/glance/policy.json" do
+  source "policy.json.erb"
+  owner "root"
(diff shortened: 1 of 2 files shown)


aplanas:
[P2: Bug 1023507 - VUL-0: openstack-glance: Users of Glance may be able to replace active image data (OSSN 0065)](https://trello.com/c/XIKMt2au/345-p2-bug-1023507-vul-0-openstack-glance-users-of-glance-may-be-able-to-replace-active-image-data-ossn-0065)

cmurphy:
If anyone has a custom policy.json for glance this will completely break their policies, which is a major security concern.

Could we somehow do a find and replace in the existing file instead of managing the whole file?

Could this maybe be fixed in the package instead? (not sure how upgrading would work if they've modified the file)

Could we just create some messaging to users that they should do this themselves?

chef/cookbooks/glance/recipes/api.rb
```diff
@@ -103,6 +103,26 @@
...
 
+# If `show_multiple_locations` is enabled we need to update
+# policy.json in a more restrictive way.  Only admin (or image owner)
+# can now update the `image_location` metadate field.
+#
+# For more information check bsc#1023507, and upstream bug:
+# https://bugs.launchpad.net/ossn/+bug/1549483
+#
+# NOTE(aplanas) -- policy.json needs to be synchronized to current
+# version of OpenStack.
```
cmurphy: This is super important and super painful. This comment should maybe include the date or release that it was last modified to make it easier to track whether it's current.

aplanas:
The other option is to move it into the package, that have the problem that will use, by default, the most restrictive version of the policy

aplanas:
@cmurphy lets drop this and try to put it in the package for the constrained version!

Thanks!

PR #1061, opened 2017-07-12
Fix mysql cookbook

jsuchome:
Backport of https://github.com/crowbar/crowbar-openstack/pull/1003


toabctl:
@jsuchome all commit messages should start with ```mysql:```

jsuchome:
@toabctl Hm, true. But the original ones did not start this way, do you think it's correct to rename them when cherry-picking?

PR #1065, opened 2017-07-14
keystone: Remove deprecated options (SCRD-46)

cmurphy:
This patch removes settings that were deprecated in Ocata and earlier.
These changes are all compatible with the Newton release incorporated in
Cloud 7.

* Remove use of the "verbose" setting. This was deprecated in oslo.log
  in favor of always logging at the INFO level[1].
* Remove settings for [eventlet_server]. The eventlet server has been
  deprecated since kilo[2], though the config options remain in tree in
  Ocata and were not removed in Mitaka as planned. We are not making use
  of the eventlet server, we only support apache and uwsgi as frontends,
  and only test on apache. The bind port and host settings that are
  exposed to the user are still valid for these alternate frontends so
  we do not remove them from the databag, only from the
  [eventlet_server] section in the config file.
* Remove write-related [ldap] settings. Write support for LDAP as an
  identity backend was deprecated in Mitaka[3]. We never added these
  settings for the domain-specific backend options, so this just removes
  the settings for the LDAP-only setup.
* Remove the [kvs] section from the config file. It was deprecated in
  Ocata[4] and we never set any configuration here anyway.

Read keystone's Ocata release notes for more information[5].

[1] http://lists.openstack.org/pipermail/openstack-dev/2015-July/070754.html
[2] http://lists.openstack.org/pipermail/openstack-dev/2015-February/057359.html
[3] https://review.openstack.org/#/c/256257/
[4] https://blueprints.launchpad.net/keystone/+spec/deprecated-as-of-ocata
[5] https://docs.openstack.org/releasenotes/keystone/ocata.html

### DIFF
--- a/chef/data_bags/crowbar/migrate/keystone/200_remove_deprecated_opts.rb
+++ b/chef/data_bags/crowbar/migrate/keystone/200_remove_deprecated_opts.rb
@@ -1,0 +1,24 @@
+def upgrade(ta, td, a, d)
+  a.delete("verbose")
+
+  a["ldap"].delete("allow_subtree_delete") if a["ldap"].key?("allow_subtree_delete")
+  a["ldap"].delete("user_allow_create") if a["ldap"].key?("user_allow_create")
+  a["ldap"].delete("user_allow_update") if a["ldap"].key?("user_allow_update")
+  a["ldap"].delete("user_allow_delete") if a["ldap"].key?("user_allow_delete")
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["verbose"] = ta["verbose"]
+
+  ldap_a = a["ldap"]
+  ldap_ta = ta["ldap"]
+  ldap_keys = ["allow_subtree_delete",
+               "user_allow_create",
+               "user_allow_update",
+               "user_allow_delete"]
+  ldap_keys.each do |key|
+    ldap_a[key] = ldap_ta[key] unless ldap_a.key?(key)
+  end
+  return a, d
+end
--- a/chef/cookbooks/keystone/attributes/default.rb
+++ b/chef/cookbooks/keystone/attributes/default.rb
@@ -26,7 +26,6 @@ default[:keystone][:group] = "keystone"
 
 default[:keystone][:debug] = false
 default[:keystone][:frontend] = "apache"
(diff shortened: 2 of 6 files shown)


cmurphy:
@nicolasbock I had a nasty typo in the downgrade script, please re-review

chef/data_bags/crowbar/migrate/keystone/200_remove_deprecated_opts.rb
```diff
@@ -0,0 +1,25 @@
+def upgrade(ta, td, a, d)
+  a.delete("verbose") if a.key?("verbose")
```
dirkmueller: verbose used to be a required key, so the if a.key? check is slightly redundant (doesn't hurt though)

cmurphy:
> I still see usages of user_allow_create and friends in keystone.conf.erb, shouldn't they be removed as well? (and the default attributes for it)

Good catch, will do

chef/data_bags/crowbar/migrate/keystone/200_remove_deprecated_opts.rb
```diff
@@ -0,0 +1,25 @@
...
+  a["ldap"].delete("user_allow_create") if a["ldap"].key?("user_allow_create")
+  a["ldap"].delete("user_allow_update") if a["ldap"].key?("user_allow_update")
+  a["ldap"].delete("user_allow_delete") if a["ldap"].key?("user_allow_delete")
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["verbose"] = ta["verbose"]
+  unless a["ldap"].key?("allow_subtree_delete")
+    a["ldap"]["allow_subtree_delete"] = ta["allow_subtree_delete"]
```
dirkmueller: needs to be
```ruby
 a["ldap"]["allow_subtree_delete"] = ta["ldap"]["allow_subtree_delete"]
```
(I would suggest to make a local variable for ldap_a and ldap_ta and do that all in one block)

cmurphy: Fixed the error. I'm not sure I understand what you're picturing with the local variables but I took a stab at condensing it.

PR #1066, opened 2017-07-17
cinder: Update for Ocata (SCRD-41)

cmurphy:


crowbar_framework/app/helpers/barclamp/cinder_helper.rb
```diff
@@ -49,7 +49,6 @@ def raw_methods_for_cinder(selected)
     def netapp_storage_family(selected)
       options_for_select(
         [
-          ["Data ONTAP in 7-Mode", "ontap_7mode"],
           ["Clustered Data ONTAP", "ontap_cluster"]
```
dirkmueller: hmm.. having a select with only one option isn't such a nice experience. I wonder if we want to add E-Series to the drop down

cmurphy: Looking at https://docs.openstack.org/ocata/config-reference/block-storage/drivers/netapp-volume-driver.html it seems like E-series support is a little more involved and so I think it's a little out of scope for this pull request/jira issue. Agreed that it's not good UX, so I'll drop the dropdown.

chef/cookbooks/cinder/recipes/role_cinder_controller.rb
```diff
@@ -18,6 +18,7 @@
   include_recipe "cinder::api"
   include_recipe "cinder::scheduler"
   include_recipe "cinder::controller_ha"
+  include_recipe "cinder::memcached"
```
stefannica: @cmurphy I wonder if the order is relevant ? I mean if one of the services using memcache is enabled and started _before_ the memcached instance is started, won't that create problems ? I have a similar case (glance) and I wanted to get your opinion on this.

cmurphy: It probably does matter. I didn't run into a problem because keystone already deploys memcache, but if it was running on a different node I expect it wouldn't have worked. Will fix.

cmurphy: Pushed the fix for this but will update the memcached commit to use the new helper method, as discussed

crowbar_framework/app/helpers/barclamp/cinder_helper.rb
cmurphy: I've kept the dropdown but made it greyed-out and unselectable with a note that only the cluster family is supported. This way users know which family they're getting when they're filling it out, and we leave space to add new families as needed.

crowbar_framework/app/views/barclamp/cinder/_edit_attributes.html.haml
```diff
@@ -49,7 +49,15 @@
               %legend
                 = t('.volumes.netapp_parameters')
 
-              = select_field %w(volumes {{@index}} netapp storage_family), :collection => :netapp_storage_family, "data-showit" => "ontap_7mode;ontap_cluster", "data-showit-target" => "#netapp_7mode_container_{{@index}};#netapp_cmode_container_{{@index}}", "data-showit-direct" => "true"
+              %div{ :id => "netapp_storage_family_{{@index}}" }
```
cmurphy: This makes zero sense.

dirkmueller:
@cmurphy is this wip or not? jira ticket is marked as ready for review. can we remove the memcached token to a separate PR and handle that once the security concerns are analyzed? rest looks good to me.

cmurphy:
I pushed the memcached changes to https://github.com/crowbar/crowbar-openstack/pull/1099 and removed that commit from this pull request. #1099 still contains this commit because otherwise there would be merge conflicts in the template file, but I think once this is merged #1099 will be left with just the one commit.

cmurphy:
I think I need to split this change into separate commits since part of it is a fix for a schema update that landed in stable/4.0 - https://github.com/crowbar/crowbar-openstack/pull/1099#discussion_r130080220

cmurphy:
This is now split into two commits, one of which will need to be backported.

PR #1067, opened 2017-07-18
WIP: galera support

jsuchome:
Experimental PR based on @rhafer 's work

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -30,8 +30,8 @@ end
 
 unless node[:database][:galera_bootstrapped]
   mysql_service_name = "mariadb"
-  
-  bash "reload systemd after mariadb.service  update" do
+
+  bash "reload systemd after mariadb.service update" do
     code "systemctl daemon-reload"
     action :nothing
     subscribes :run,
@@ -66,14 +66,14 @@ unless node[:database][:galera_bootstrapped]
 
   # Initialize cluster on the founder node, this will start mariadb
   # via systemctl. We stop it again after the bootstrap.
-  execute  "boostrapping first mariadb galera cluster node" do
+  execute "boostrapping first mariadb galera cluster node" do
     command "galera_new_cluster"
     action :run
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
 
   execute "assign-root-password-galera" do
-    command "/usr/bin/mysqladmin -u root password \"#{node['mysql']['server_root_password']}\""
+    command "/usr/bin/mysqladmin -u root password \"#{node[:mysql][:server_root_password]}\""
     action :run
     only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
     only_if "/usr/bin/mysql -u root -e 'show databases;'"
@@ -89,7 +89,7 @@ unless node[:database][:galera_bootstrapped]
     action :start
     not_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
-  
+
   crowbar_pacemaker_sync_mark "sync-database_after_bootstrap_rest" do
     revision node[:database]["crowbar-revision"]
   end
@@ -113,7 +113,6 @@ unless node[:database][:galera_bootstrapped]
   file "delete mariadb.service" do
     path "/etc/systemd/system/mariadb.service"
     action :delete
-    only_if
   end
 end
 
@@ -128,15 +127,13 @@ service_name = "galera"
 
 pacemaker_primitive service_name do
   agent resource_agent
-  params (
-    {
-      "enable_creation" => true,
-      "wsrep_cluster_address" => "gcomm://d52-54-00-01-00-01,d52-54-00-01-00-02,d52-54-00-01-00-03",
-      "check_user" => "root",
-      "check_passwd" => node[:database][:mysql][:server_root_password],
-      "socket" => "/var/run/mysql/mysql.sock"
-    }
-  )
+  params ({
+    "enable_creation" => true,
+    "wsrep_cluster_address" => "gcomm://d52-54-00-01-00-01,d52-54-00-01-00-02,d52-54-00-01-00-03",
+    "check_user" => "root",
+    "check_passwd" => node[:database][:mysql][:server_root_password],
+    "socket" => "/var/run/mysql/mysql.sock"
+  })
   action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
@@ -147,12 +144,10 @@ ms_name = "ms-#{service_name}"
 pacemaker_ms ms_name do
   rsc service_name
   meta(
-    {
-      "master-max" => 3,
-      "ordered" => "false",
-      "interleave" => "false",
-      "notify" => "true"
-    }
+    "master-max" => 3,
+    "ordered" => "false",
+    "interleave" => "false",
+    "notify" => "true"
   )
   action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -125,11 +125,14 @@ end
 
 service_name = "galera"
 
+cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node, "database-server")
+nodes_names = cluster_nodes.map { |n| n[:hostname] }
+
 pacemaker_primitive service_name do
   agent resource_agent
   params ({
     "enable_creation" => true,
-    "wsrep_cluster_address" => "gcomm://d52-54-00-01-00-01,d52-54-00-01-00-02,d52-54-00-01-00-03",
+    "wsrep_cluster_address" => "gcomm://" + nodes_names.join(","),
     "check_user" => "root",
     "check_passwd" => node[:database][:mysql][:server_root_password],
     "socket" => "/var/run/mysql/mysql.sock"
@@ -144,7 +147,7 @@ ms_name = "ms-#{service_name}"
 pacemaker_ms ms_name do
   rsc service_name
   meta(
-    "master-max" => 3,
+    "master-max" => nodes_names.size,
     "ordered" => "false",
     "interleave" => "false",
     "notify" => "true"
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -54,7 +54,7 @@ unless node[:database][:galera_bootstrapped]
   # will be delete again further down.
   file "mariadb.service" do
     path "/etc/systemd/system/mariadb.service"
-    content ::File.open("/usr/share/mysql/systemd/mariadb.service").read
+    content lazy { ::File.open("/usr/share/mysql/systemd/mariadb.service").read }
     mode "0644"
   end
 
@@ -99,9 +99,9 @@ unless node[:database][:galera_bootstrapped]
     action :stop
   end
 
-  block "mark node for galera bootstrap" do
+  ruby_block "mark node for galera bootstrap" do
     block do
-      node.set[:database][:galera_bootstraped] = true
+      node.set[:database][:galera_bootstrapped] = true
       node.save
     end
     subscribes :create,
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -1,0 +1,175 @@
+#
+# Cookbook Name:: mysql
+# Recipe:: ha_galera
+#
+# Copyright 2017, SUSE Linux GmbH.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+resource_agent = "ocf:heartbeat:galera"
+
+package "galera-3"
+
+# Wait for all nodes to reach this point so we know that all nodes will have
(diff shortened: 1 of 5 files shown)


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,175 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+resource_agent = "ocf:heartbeat:galera"
+
+package "galera-3"
```
jsuchome: Does not seem to be needed with mariadb 10.1 (that provides galera plugin itself)

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,175 @@
...
+end
+
+service_name = "galera"
+
+pacemaker_primitive service_name do
+  agent resource_agent
+  params (
+    {
+      "enable_creation" => true,
+      "wsrep_cluster_address" => "gcomm://d52-54-00-01-00-01,d52-54-00-01-00-02,d52-54-00-01-00-03",
```
jsuchome: TODO: generate from cluster membets

chef/cookbooks/mysql/templates/default/my.cnf.erb
```diff
@@ -52,6 +52,16 @@ datadir                = <%= node['mysql']['datadir'] %>
 tmpdir                 = <%= node['mysql']['tmpdir'] %>
 skip-external-locking
 
+# GALERA
+
+wsrep_provider=/usr/lib64/galera-3/libgalera_smm.so
```
jsuchome: Seems like this could be be dropped with mariadb-10.1 (and points to old library anyway)

chef/cookbooks/mysql/templates/default/my.cnf.erb
```diff
@@ -52,6 +52,16 @@ datadir                = <%= node['mysql']['datadir'] %>
...
+wsrep_cluster_address="gcomm://d52-54-00-01-00-01,d52-54-00-01-00-02,d52-54-00-01-00-03"
```
jsuchome: TODO: get via argument

crowbar_framework/app/models/database_service.rb
```diff
@@ -73,33 +73,35 @@ def validate_proposal_after_save(proposal)
...
-        validation_error I18n.t(
-          "barclamp.#{@bc_name}.validation.drbd_not_enabled",
-          cluster_name: cluster_name(cluster)
-        ) unless role.default_attributes["pacemaker"]["drbd"]["enabled"]
+      unless db_engine == "mysql"
+        storage_mode = attributes["ha"]["storage"]["mode"]
         validation_error I18n.t(
-          "barclamp.#{@bc_name}.validation.invalid_size_drbd"
-        ) if attributes["ha"]["storage"]["drbd"]["size"] <= 0
+                           "barclamp.#{@bc_name}.validation.unknown_mode_ha",
```
s-t-e-v-e-n-k: I think this whole block needs a rethink, and could probably do with some cleanup and reflowing. Secondly, I also think we want to check if we have <3 cluster members and error if sql_engine is mysql.

jsuchome: I agree, but it might be enough for some next PR

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,175 @@
...
+    subscribes :create,
+      "file[delete mariadb.service]",
+      :immediately
+    action :nothing
+  end
+
+  file "delete mariadb.service" do
+    path "/etc/systemd/system/mariadb.service"
+    action :delete
+    only_if
```
Itxaka: empty only_if

chef/cookbooks/mysql/recipes/ha_galera.rb
Itxaka: hardcoded addresses?

chef/cookbooks/mysql/templates/default/my.cnf.erb
Itxaka: harcoded addresses?

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: yep, see my older comment

crowbar_framework/app/models/database_service.rb
```diff
@@ -60,6 +60,53 @@ def create_proposal
     base
   end
 
+  def validate_ha_attributes(attributes)
```
jsuchome: I'm afraid I can't do anything about it. Splitting the method even more would not bring clarity IMHO.

crowbar_framework/app/models/database_service.rb
```diff
@@ -60,6 +60,48 @@ def create_proposal
...
+            cluster_name: cluster_name(cluster)
+          )
+        end
+        if attributes["ha"]["storage"]["drbd"]["size"] <= 0
+          validation_error I18n.t(
+            "barclamp.#{@bc_name}.validation.invalid_size_drbd"
+          )
+        end
+      end
+    when "mysql"
```
jsuchome: No more empty body

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,168 @@
...
+# all the required packages installed before we create the pacemaker
+# resources
+crowbar_pacemaker_sync_mark "sync-database_before_ha" do
+  revision node[:database]["crowbar-revision"]
+end
+
+unless node[:database][:galera_bootstrapped]
+  mysql_service_name = "mariadb"
+
+  bash "reload systemd after mariadb.service  update" do
```
s-t-e-v-e-n-k: Nit: Double space after service

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,168 @@
...
+  end
+
+  directory "/var/run/mysql/" do
+    owner "mysql"
+    group "root"
+    mode "0755"
+    action :create
+  end
+
+  # Temporary install mariadb unit as the "galera_new_cluster" bootstrap
```
s-t-e-v-e-n-k: We could fix this by adding Alias=mariadb.service into the service file we ship in the mariadb package, making this block unnecessary.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,168 @@
...
+  file "mariadb.service" do
+    path "/etc/systemd/system/mariadb.service"
+    content ::File.open("/usr/share/mysql/systemd/mariadb.service").read
+    mode "0644"
+  end
+
+  execute "mysql_install_db" do
+    command "mysql_install_db"
+    action :run
+    # only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
jsuchome: So, founder or all cluster nodes?

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,168 @@
...
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+transaction_objects.push("pacemaker_primitive[#{service_name}]")
+
+ms_name = "ms-#{service_name}"
+pacemaker_ms ms_name do
+  rsc service_name
+  meta(
+    "master-max" => 3,
```
s-t-e-v-e-n-k: This should be the number of cluster members, not hardcoded, IMO.

chef/cookbooks/mysql/templates/default/my.cnf.erb
```diff
@@ -52,6 +52,15 @@ datadir                = <%= node['mysql']['datadir'] %>
 tmpdir                 = <%= node['mysql']['tmpdir'] %>
 skip-external-locking
 
+# GALERA
+
+wsrep_cluster_address="gcomm://d52-54-00-01-00-01,d52-54-00-01-00-02,d52-54-00-01-00-03"
```
s-t-e-v-e-n-k: This is exactly same format as the wsrep_cluster_address in the pacemaker primitive. This should be filled in by the recipe.

crowbar_framework/config/locales/database/en.yml
```diff
@@ -45,9 +45,9 @@ en:
               options: 'Mount Options'
       validation:
         invalid_db_engine: 'Invalid database engine: %{db_engine}.'
-        ha_postgresql: 'High availability support is only available for PostgreSQL.'
         unknown_mode_ha: 'Unknown mode for HA storage: %{storage_mode}.'
         no_device: 'No device specified for shared storage.'
         no_filesystem: 'No filesystem type specified for shared storage.'
         drbd_not_enabled: 'DRBD is not enabled for cluster %{cluster_name}.'
         invalid_size_drbd: 'Invalid size for DRBD device.'
+        cluster_size_even: 'The Galera cluster needs odd number of nodes.'
```
s-t-e-v-e-n-k: "needs an odd number of cluster members"

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: Or maybe there's some packaging bug?
If _"galera_new_cluster" bootstrap script doesn't work with installed "mysql.service" unit_ , that does not look correct...

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: Ah, good point

chef/cookbooks/mysql/templates/default/my.cnf.erb
jsuchome: Yep

crowbar_framework/config/locales/database/en.yml
jsuchome: OK, I'll update it

jsuchome:
@s-t-e-v-e-n-k I was finally able to deploy database barclamp with mysql and HA. It suffers from the problem already mentioned by @rhafer (problem with promoting the nodes). At least this is something you could work with I hope

chef/cookbooks/mysql/templates/default/my.cnf.erb
```diff
@@ -52,6 +52,16 @@ datadir                = <%= node['mysql']['datadir'] %>
...
 
+<% if node[:database][:ha][:enabled] %>
+# GALERA
+
+wsrep_cluster_address="gcomm://"<%= @nodes_names.join(",") %>
+innodb_autoinc_lock_mode=2
+innodb_doublewrite=1
+query_cache_size=0
+wsrep_on=ON
+binlog_format=ROW
```
rsalevsky: we should add `wsrep_sst_method=xtrabackup`

jsuchome: Why this and not anything else (and not the default)?

chef/cookbooks/mysql/recipes/ha_galera.rb
s-t-e-v-e-n-k: This is not required at all. From galera_new_cluster it uses systemctl start ${1:-mariadb} so we should call 'galera_new_cluster mysql' and all this temporary stuff can hit /dev/null.

jsuchome: I think you are wrong. I did not check this closely, but when I really removed it, mysql was not started.... Oh, it was probably on the other (non-founder) nodes, that need to start mysql during bootstrap as well.

s-t-e-v-e-n-k: Really? The service file is removed after bootstrapping. I've been testing a diff that removes all these gubbins.

jsuchome: But the database needs to be started on all nodes during bootstrapping, to set the root password

s-t-e-v-e-n-k: The only reason this service hackery exists is for galera_new_cluster, which we only run on the cluster founder, and if we call it with 'mysql' as the first argument (as opposed to no arguments), it will use that service name. The second piece of the puzzle is updating mysql_service_name to be "mysql" and it works great.

jsuchome: > The second piece of the puzzle is updating mysql_service_name to be "mysql" and it works great.

Ah, true. That could work.

chef/cookbooks/mysql/templates/default/my.cnf.erb
```diff
@@ -52,6 +52,17 @@ datadir                = <%= node['mysql']['datadir'] %>
 tmpdir                 = <%= node['mysql']['tmpdir'] %>
 skip-external-locking
 
+<% if node[:database][:ha][:enabled] %>
+# GALERA
+
+wsrep_provider=/usr/lib64/galera-3/libgalera_smm.so
```
s-t-e-v-e-n-k: This file doesn't exist without the galera-3 package installed, so should be removed at the same time as the package "galera-3" statement in the recipe.

jsuchome: I've brought "galera-3" package back (see that the commit got removed). I was unable to finish the configuration without providing `wsrep_provider`

jsuchome:
@s-t-e-v-e-n-k not that I removed again that commit that was removing that temporary service. I probably did something wrong there, but it was breaking my deployment.

chef/cookbooks/mysql/templates/default/my.cnf.erb
rsalevsky: It's the only non-blocking SST method; see http://galeracluster.com/documentation-webpages/sst.html

jsuchome: @rsalevsky  Hm. I did not test too long, but with any non-default value of `wsrep_sst_method` I was not able to start the galera cluster.

rsalevsky: Yeah, I tend to say we should do it in a separate PR. Otherwise this PR can get very messy.

jsuchome:
Closing, was moved to https://github.com/crowbar/crowbar-openstack/pull/1111

PR #1068, opened 2017-07-18
tempest: Update to use cirros 0.3.5

dirkmueller:
It's that time of the year again. It is only a security fix rebase
of 0.3.4, so it should go in pretty smoothly.


dirkmueller:
passes testing

PR #1069, opened 2017-07-18
glance: remove options deprecated in ocata

stefannica:
The following deprecated glance & related config options
are removed in this PR:

  * [default]/verbose deprecated option
  * [glance_store]/vmware_api_insecure - renamed to vmware_insecure
  * [keystone_authtoken]/service_token_roles_required is set
    to True to enable fetching expired tokens based on valid
    service tokens

The following deprecated glance config options are NOT
removed in this commit:

  * [default]/show_multiple_locations - this option is still required
    to enable cinder internal image transfer and there is no alternative
    in place (see bsc#945043). When this option is removed, it is expected
    that the underlying feature will be enabled by default, in which case
    the policy.json file is already updated in the glance package
    to protect against unauthorized access (see
    https://wiki.openstack.org/wiki/OSSN/OSSN-0065)

  * the admin_password, admin_tenant_name, admin_user, auth_region
    and auth_url options are still needed by the glance-scrubber
    service. In theory, these parameters should no longer be required
    (according to https://review.openstack.org/237742), but in
    reality this isn't the case.

  * the deprecated keystonemiddleware.auth_token in-process token
    needs to be replaced with a memcache server (done separately, due 
    to security implications - see #1090)

References:
  * Ocata release notes for glance: https://docs.openstack.org/releasenotes/glance/ocata.html
  * Ocata release notes for keystonemiddleware: https://docs.openstack.org/releasenotes/keystonemiddleware/ocata.html
  * OSSN extending the deprecated show_multiple_locations option
    to Pike: https://wiki.openstack.org/wiki/OSSN/OSSN-0065
  * OSSN deprecating the use_user_token and associated authN options:
    https://wiki.openstack.org/wiki/OSSN/OSSN-0060

### DIFF
--- a/chef/cookbooks/glance/templates/default/glance-api.conf.erb
+++ b/chef/cookbooks/glance/templates/default/glance-api.conf.erb
@@ -17,7 +17,6 @@ registry_port = <%= @registry_bind_port %>
 transport_url = <%= @rabbit_settings[:url] %>
 registry_client_insecure = <%= @keystone_settings["insecure"] %>
 debug = <%= node[:glance][:debug] ? "True" : "False" %>
-verbose = <%= node[:glance][:verbose] ? "True" : "False" %>
 log_file = <%= node[:glance][:api][:log_file] %>
 use_syslog = <%= node[:glance][:use_syslog] ? "True" : "False" %>
 use_stderr = false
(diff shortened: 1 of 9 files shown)


stefannica:
Depends on #1070

chef/cookbooks/glance/templates/default/glance-api.conf.erb
```diff
@@ -77,6 +76,9 @@ project_domain_name = <%= @keystone_settings["admin_domain"]%>
 user_domain_name = <%= @keystone_settings["admin_domain"] %>
 auth_url = <%= @keystone_settings['admin_auth_url'] %>
 auth_type = password
+service_token_roles_required = true
+service_token_roles = <%= @keystone_settings['service_tenant'] %>
+memcached_servers = <%= @memcached_servers.join(',') %>
```
dirkmueller: this isn't really removing deprecated options, does it? can you split it into a separate commit (same PR is fine)?

stefannica: that's right. I sliced the commit in two, as you suggested. it didn't complain (much).

stefannica:
moved keystonemiddleware memcached bits to #1090

PR #1070, opened 2017-07-19
memcached: add memcached servers cookbook helper

stefannica:
Created a memcached helper to elliminate the code duplicated in keystone, nova, horizon, swift (and future glance, cinder and other cookbooks) that computes the list of memcached servers.


stefannica:
@cmurphy do you think the helper should also try to minimize the memcached.rb bits ? I can tell they are duplicated in pretty much the same fashion.

cmurphy:
> @cmurphy do you think the helper should also try to minimize the memcached.rb bits ? I can tell they are duplicated in pretty much the same fashion.

Seems reasonable to me

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -57,6 +57,12 @@
 my_admin_host = CrowbarHelper.get_host_for_admin_url(node, ha_enabled)
 my_public_host = CrowbarHelper.get_host_for_public_url(node, node[:keystone][:api][:protocol] == "https", ha_enabled)
 
+memcached_servers = MemcachedHelper.get_memcached_servers(
```
stefannica: FYI, I moved this further up to ensure that the memcached resource is created/started before other keystone services

chef/cookbooks/swift/recipes/proxy.rb
```diff
@@ -46,9 +46,6 @@
   return
 end
 
-local_ip = Swift::Evaluator.get_ip_by_type(node, :admin_ip_expr)
```
stefannica: these weren't used anymore after my modifications

chef/cookbooks/swift/recipes/proxy.rb
```diff
@@ -255,12 +252,6 @@
   end
 end
 
-## Find other nodes that are swift-auth nodes, and make sure
-## we use their memcached!
-proxy_config[:memcached_ips] = node_search_with_cache("roles:swift-proxy").map do |x|
```
stefannica: Same here: this isn't used anywhere (probably just a leftover).

chef/cookbooks/swift/recipes/proxy.rb
```diff
@@ -388,7 +376,6 @@
 EOH
     action :nothing
     subscribes :run, resources(template: node[:swift][:proxy_config_file])
-    notifies :restart, resources(service: "memcached-swift-proxy")
```
stefannica: removed this because memcached_instance isn't actually a resource provider (it would have failed on debian, if still supported)

chef/cookbooks/swift/recipes/proxy.rb
stefannica: Swift disagrees with me. I stand corrected.

PR #1072, opened 2017-07-19
ceilometer: Allow enabling SSL with HA (bsc#1049153)

bhdn:


nicolasbock:
The Travis-CI job is failing:

~~~
$ rvm use 2.1.0 --install --binary --fuzzy
ruby-2.1.0 is not installed - installing.
Searching for binary rubies, this might take some time.
Requested binary installation but no rubies are available to download, consider skipping --binary flag.
Gemset '' does not exist, 'rvm ruby-2.1.0 do rvm gemset create ' first, or append '--create'.
The command "rvm use 2.1.0 --install --binary --fuzzy" failed and exited with 2 during .
~~~

I noticed that Travis has been upgrading their container OS to "Trusty". Could that be the problem? But then other PR's would fail this step as well. I'll check...

stefannica:
@nicolasbock I have two PRs, #1069 and #1070 that are failing travis-ci in the same manner. Any ideas about how we can fix travis ?

nicolasbock:
@stefannica I think the issue is the underlying version of the VM OS that Travis-CI is using. Have a look at https://github.com/crowbar/crowbar-openstack/pull/1073. This PR forces the VM back to `precise` and things work again as before.

bhdn:
Sorry, I delete the branch by mistake.

PR #1073, opened 2017-07-20
Insist on precise for Travis-CI

nicolasbock:
We seem to be running into Travis-CI issues lately. Travis-CI has
switched their default VM to the trusty release which might be reason
for the issues we are seeing. This patch is only a test to verify that
this is the case (or not).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: ruby
 sudo: false
+cache: bundler
+dist: trusty
 
-rvm:
-  - 2.1.0
+rvm: 2.1.9


nicolasbock:
@cmurphy I pushed an update running on trusty but with an updated ruby environment. Let's see how far that goes...

nicolasbock:
@cmurphy Please have another look. If we update our Travis-CI setting to `rvm 2.2` the Travis-CI tests are passing again.

rsalevsky:
Please, don't merge yet. We should try to stay at the 2.1 release because this is the version we use in the product. I'll do some research maybe I find a better solution. There was quite some changes in travis in the past few weeks.

nicolasbock:
Thanks for the update @rsalevsky.

rsalevsky:
@nicolasbock Can you update this PR based on https://github.com/crowbar/crowbar-core/pull/1273?

nicolasbock:
@rsalevsky done.

Itxaka:
This should also be ported to crowbar-{ha,hyperv} rigth?

PR #1074, opened 2017-07-20
neutron: remove options deprecated in ocata

stefannica:
The following deprecated neutron & related config options
are removed in this PR:
  * [default]/verbose deprecated option
  * [keystone_authtoken]/service_token_roles_required is set
to True to enable fetching expired tokens based on valid
service tokens
  * the DHCP agent's dhcp_domain option was renamed to dns_domain
  * removed L3 agent's deprecate send_arp_for_ha option
  * removed L3 agent's deprecated external_network_bridge option

Not included in this PR: 
  * the deprecated keystonemiddleware.auth_token in-process token
  needs to be replaced with a memcache server (done separately, due 
  to security implications - see #1090)

References:
  * Ocata release notes for neutron: https://docs.openstack.org/releasenotes/neutron/ocata.html
  * Ocata release notes for keystonemiddleware: https://docs.openstack.org/releasenotes/keystonemiddleware/ocata.html


stefannica:
Root cause behind the recent CI failures: the CI mkcloud jobs are using Newton packages, but these modifications require Ocata packages. Switching from neutron CLI to openstack CLI cannot be done in Newton because there are some commands missing from the openstack CLI (i.e. the openstack router set --external-gateway support is available only with python-openstackclient >= 3.6.0)

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -52,6 +51,10 @@ password = <%= @keystone_settings['service_password'] %>
 project_name = <%= @keystone_settings['service_tenant'] %>
 project_domain_name = <%= @keystone_settings['admin_domain'] %>
 user_domain_name = <%= @keystone_settings['admin_domain'] %>
+memcached_servers = <%= @memcached_servers.join(',') %>
+service_token_roles_required = true
+service_token_roles = <%= @service_roles.join(',') %>
```
cmurphy: Might as well just hardcode admin here the same way the option above it is.

stefannica:
thanks @cmurphy, I got rid of the extra bits

stefannica:
moved keystonemiddleware memcached bits to #1090

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
```diff
@@ -4,7 +4,7 @@ resync_interval = <%= @resync_interval %>
 dhcp_driver = <%= @dhcp_driver %>
 enable_isolated_metadata = <%= @enable_isolated_metadata %>
 enable_metadata_network = <%= @enable_metadata_network %>
-dhcp_domain = <%= @dhcp_domain %>
+dns_domain = <%= @dhcp_domain %>
```
dirkmueller: maybe we want to rename @dhcp_domain to @dns_domain as well, otherwise this looks like an error.

dirkmueller:
@stefannica also needs rebase

stefannica:
@dirkmueller done renaming all occurrences of dhcp_domain to dns_domain (except those in the nova cookbook, that will be removed by #1091 anyway)

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -172,7 +171,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 200,
+      "schema-revision": 201,
```
cmurphy: This and the migration file need to be bumped to 202

chef/data_bags/crowbar/migrate/neutron/201_remove_deprecated_opts.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  a.delete("verbose")
+  d["dns_domain"] = d["dhcp_domain"]
```
cmurphy: Is this not meant to be `a["dns_domain"] = a["dhcp_domain"]`? (I hate these variable names)

chef/data_bags/crowbar/template-neutron.json
stefannica: @cmurphy why ? I thought incrementing the revision number would be enough. Or am I missing something ?

stefannica: @cmurphy nevermind, I just got it ;)

PR #1075, opened 2017-07-20
mysql: Switch to mysql2 to support newer mysql/mariadb versions

rsalevsky:
The rubygem-mysql got deprecated by rubygem-mysql2 which supports our
mariadb 10.1.X version.

See **SCRD-593**


chef/cookbooks/database/libraries/provider_database_mysql.rb
```diff
@@ -22,82 +22,86 @@ class Chef
...
+          if new_resource.encoding
+            create_sql += " CHARACTER SET = #{new_resource.encoding}"
+          end
+          if new_resource.collation
+            create_sql += " COLLATE = #{new_resource.collation}"
           end
+          Chef::Log.debug(
+            "#{new_resource.name}: Performing query [#{create_sql}]"
+          )
+          client.query(create_sql)
```
dirkmueller: I'd prefer to use "execute" for something that isn't use to query ressources. Also, the way you're quoting values is unsafe. Maybe better use prepare with "?" and then execute the statement with the variables expanded.

chef/cookbooks/database/libraries/provider_database_mysql.rb
```diff
@@ -22,82 +22,86 @@ class Chef
...
+              schema_present = true
             end
           end
+
+          # drop
+          return unless schema_present
+          Chef::Log.info("Dropping schema '#{new_resource.database_name}'")
+          drop_sql = "DROP SCHEMA IF EXISTS `#{new_resource.database_name}`"
+          Chef::Log.debug("Performing query [#{drop_sql}]")
+          client.query(drop_sql)
```
dirkmueller: same here

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -22,63 +22,128 @@ class Chef
...
+          # create
+          return if user_present
+          Chef::Log.info(
+            "Creating user '#{new_resource.username}'@'#{new_resource.host}'"
+          )
+          create_sql = "CREATE USER '#{new_resource.username}'@'#{new_resource.host}'"
+          if new_resource.password
+            create_sql += " IDENTIFIED BY '#{new_resource.password}'"
+          end
+          client.query(create_sql)
```
dirkmueller: same here

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -22,63 +22,128 @@ class Chef
...
-              db.query(statement)
-              @new_resource.updated_by_last_action(true)
-            ensure
-              close
-            end
+          # test
+          user_present = false
+
+          test_sql = "SELECT User,Host from mysql.user"
+          test_sql += " WHERE User='#{new_resource.username}'"
```
dirkmueller: not safe way of quoting variables

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -22,63 +22,128 @@ class Chef
...
           end
+
+          # drop
+          return unless user_present
+          Chef::Log.info(
+            "Dropping user '#{new_resource.username}'@'#{new_resource.host}'"
+          )
+          drop_sql = "DROP USER"
+          drop_sql += " '#{new_resource.username}'@'#{new_resource.host}'"
+          client.query(drop_sql)
```
dirkmueller: same here

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -22,63 +22,128 @@ class Chef
...
+          grant = "GRANT #{new_resource.privileges.join(",")}"
+          grant += " ON #{db_name}.#{tbl_name}"
+          grant += " TO '#{new_resource.username}'@'#{new_resource.host}'"
+          grant += " IDENTIFIED BY '#{new_resource.password}'"
+
+          Chef::Log.debug(
+            "#{new_resource}: granting access with statement [#{grant}]"
+          )
+          client.query(grant)
+          client.query("FLUSH PRIVILEGES")
```
dirkmueller: same here

chef/cookbooks/database/libraries/provider_database_mysql.rb
```diff
@@ -22,82 +22,80 @@ class Chef
...
+          create_sql = "CREATE SCHEMA IF NOT EXISTS `#{database_name}`"
+          if new_resource.encoding
+            encoding = client.escape(new_resource.encoding)
+            create_sql += " CHARACTER SET = #{encoding}"
           end
+          if new_resource.collation
+            collation = client.escape(new_resource.collation)
+            create_sql += " COLLATE = #{collation}"
+          end
+          client.query(create_sql)
```
rhafer: Wouldn't it be better to use prepare/execute here as well?  In that case I think you could also leave out the ```escape``` calls too.

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -22,63 +22,112 @@ class Chef
...
         def action_create
-          unless exists?
-            begin
-              statement = "CREATE USER `#{@new_resource.username}`@`#{@new_resource.host}`"
-              statement += " IDENTIFIED BY '#{@new_resource.password}'" if @new_resource.password
-              db.query(statement)
-              @new_resource.updated_by_last_action(true)
-            ensure
-              close
-            end
```
rhafer: Same here about the code duplication (just for the "user" case)

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -22,63 +22,112 @@ class Chef
...
+          return if user_present
+          Chef::Log.info("Creating user '#{new_resource.username}@#{new_resource.host}'")
+          username = client.escape(new_resource.username)
+          host = client.escape(new_resource.host)
+          create_sql = "CREATE USER '#{username}'@'#{host}'"
+          if new_resource.password
+            password = client.escape(new_resource.password)
+            create_sql += " IDENTIFIED BY '#{password}'"
           end
+          client.query(create_sql)
```
rhafer: prepare/execute?

rsalevsky:
@rhafer The old mysql gem is unmaintained since end of 2014 and so it never allowed any newer version. Here is the error message: http://pastebin.nue.suse.com/23126/src

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
rsalevsky: It seams that mysql's prepare mechanism doesn't support statements with `IDENTIFIED`. It always said syntax error if I used it.

chef/cookbooks/database/libraries/provider_database_mysql.rb
rsalevsky: I converted the code already to prepare/execute but at the end it felt too complex. There are 3 possible querys (with or without CHARACTER; with or without COLLATE) and the execute parameters depend on this.

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
rsalevsky: Fixed

rhafer:
> The old mysql gem is unmaintained since end of 2014

Yeah I saw that.

> and so it never allowed any newer version. Here is the error message: http://pastebin.nue.suse.com/23126/src

That's not quite true. The error just means that the gem we ship is build against a different version of libmysql than what you have installed. It doesn't mean that the gem does not work with the newer client library. (It seems to build just fine against 10.1.x at least)

Don't get me wrong, I am all fine with having this change in master. I was just interested in some more details. And I am slightly wondering if that version update is the right thing for the Cloud 7 backport, or if we might better keep the old/outdated version there.

rsalevsky:
Mmh... Maybe we misinterpreted the error message then. But from my POV there is no difference if we ship mysql or mysql2 in Cloud 7, we anyway have to create a ECO to ship the other parts of the mariadb enablement so work wise and time wise it is the same. And from a consistency point it would be better to ship it on Cloud 7.

chef/cookbooks/database/libraries/provider_database_mysql.rb
rhafer: Hm, it should be possible to push all parameters into an array (when they are needed) and just pass that array to execute using the splat operator. Something like:

```
args = [ database_name ]
if new_resource.encoding
  args <<  new_resource.encoding
  create_sql += " CHARACTER SET = ?"
end

sql = client.prepare ....
sql.execute(*args)
```

That doesn't seem anymore complex than the current code to me.

rhafer:
> we anyway have to create a ECO to ship the other parts of the mariadb enablement so work wise and time wise it is the same.

Ok. Thinking about this again it might even more complicated to get a rubygem-mysql (the old one) build against the correct mariadb release into cloud 7. So I guess you're right.

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
dirkmueller: maybe 

``` identified by password ?``` works (and then substitute password('#{password}'). in any case you could also use quote()

File: chef/cookbooks/database/libraries/provider_database_mysql_user.rb [hunk 10]
```diff
@@ -22,63 +22,105 @@ class Chef
...
+          Chef::Log.info("Granting privileges for '#{new_resource.username}@#{new_resource.host}'")
+
+          username = client.escape(new_resource.username)
+          host = client.escape(new_resource.host)
+          password = client.escape(new_resource.password)
+
+          grant_sql = "GRANT #{new_resource.privileges.join(",")}"
+          grant_sql += " ON #{db_name}.#{tbl_name}"
+          grant_sql += " TO '#{username}'@'#{host}'"
+          grant_sql += " IDENTIFIED BY '#{password}'"
```
dirkmueller: do we need to specify the password here again? this is already existing user, right?

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
rsalevsky: mmh... `identified by password` is already deprecated. Can you give me some more context about the quote usage?

chef/cookbooks/database/libraries/provider_database_mysql.rb
```diff
@@ -22,82 +22,69 @@ class Chef
...
-              Chef::Log.debug("#{@new_resource}: Performing query [#{create_sql}]")
-              db.query(create_sql)
-              @new_resource.updated_by_last_action(true)
-            ensure
-              close
-            end
+          return if schema_present?(new_resource.database_name)
+          Chef::Log.info("Creating schema '#{new_resource.database_name}'")
+          create_args = [new_resource.database_name]
+          create_sql = "CREATE SCHEMA IF NOT EXISTS ?"
```
rhafer: Hm, this doesn't seem to work with prepare either :(. If I understand https://mariadb.com/kb/en/mariadb/prepare-statement/ correctly you can't parameterize the database name (as it's an identifier) in a prepare statement :(

The same seems to be true for the CHARACTER SET and COLLATE values.

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
rhafer: It's not (only) the ```IDENTIFIED BY``` thing that makes it fail. It already fails at the user name.

```
MariaDB [(none)]> prepare test from "CREATE USER ?";
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '?' at line 1
```

I guess this is similar to the issue with preparing ```CREATE SCHEMA``` statement.

So I guess we can't really use prepare/execute with parameters  for any of these :(

chef/cookbooks/database/libraries/provider_database_mysql.rb
```diff
@@ -22,82 +22,69 @@ class Chef
...
-              @new_resource.updated_by_last_action(true)
-            ensure
-              close
-            end
-          end
+        def action_drop
+          return unless schema_present?(new_resource.database_name)
+          Chef::Log.info("Dropping schema '#{new_resource.database_name}'")
+          drop_sql = client.prepare("DROP SCHEMA IF EXISTS ?")
+          drop_sql.execute(new_resource.database_name)
```
s-t-e-v-e-n-k: This now doesn't match the creation step, and as others have stated this might not work?

rhafer: Oh, good catch. The same is true for "DROP USER" as well.

rsalevsky:
Rebased

jsuchome:
[Update database barclamp to allow deploying mariadb with HA](https://trello.com/c/vlSc0mvA/2-update-database-barclamp-to-allow-deploying-mariadb-with-ha)

PR #1078, opened 2017-07-20
Barbican: reduce asynchronous workers to 1

abel-navarro:
During a join session with @aspiers it was found out that having more
than one asynchronous worker causes problems when shutting down the
service. That happens regardless the asynchronous queue is enabled or
not.

Having problems shutting down services is a reason for pacemaker to
fence a node. It is likely that this is what we have been observing
lately on HA nodes being fenced on reboot.

### DIFF
--- a/chef/cookbooks/barbican/templates/default/barbican.conf.erb
+++ b/chef/cookbooks/barbican/templates/default/barbican.conf.erb
@@ -19,7 +19,7 @@ kombu_ssl_ca_certs = <%= @rabbit_settings[:client_ca_certs] %>
 
 [queue]
 
-asynchronous_workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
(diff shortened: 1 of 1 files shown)


toabctl:
> So maybe we should try porting that fix to our packages instead of this workaround?

@aspiers let's do that for now

abel-navarro:
I was hoping this solution could help, at least for some manual tests. Although we did have successful service stops with this change, this doesn't seem to be the case anymore, so this change needs to be discarded.

aspiers:
> Although we did have successful service stops with this change, this doesn't seem to be the case anymore

Please can you link to specific Jenkins failures showing failed stops?  Or did you just try manually?

abel-navarro:
I did try manually. I just rebuilt my environment so I cannot post the related logs, sorry :(

PR #1079, opened 2017-07-21
WIP Galera HA

s-t-e-v-e-n-k:


jsuchome:
Please make sure to update that commit with VIP creation, I've updated mine significantly.

toabctl:
Does this superseed https://github.com/crowbar/crowbar-openstack/pull/1067 ? Or the other way arround?

s-t-e-v-e-n-k:
This was so I could work on this issue and share code with others while Jiri was on holidays.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,204 @@
...
+    mode "0755"
+    action :create
+  end
+
+  execute "mysql_install_db" do
+    command "mysql_install_db"
+    action :run
+  end
+
+  # Start mariadb on the founder. We stop it again after bootstrapping.
```
rhafer: Would you mind adding some commentary here, that this way of bootstrapping only works because we have a empty ```wsrep_cluster_address``` in my.cnf at this point in time?

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,204 @@
...
+    only_if "/usr/bin/mysql -u root -e 'show databases;'"
+  end
+
+  # Wait on all nodes for the founder to complete the bootstrap
+  crowbar_pacemaker_sync_mark "sync-database_after_bootstrap_founder" do
+    revision node[:database]["crowbar-revision"]
+  end
+
+  service "starting mysql on remaining nodes" do
+    service_name mysql_service_name
```
rhafer: Does this really work? We still have an empty ```wsrep_cluster_address``` setting in the config. Don't we? I.e. mariadb would also to the galera bootstrapping on all of the other cluster nodes. I don't think we want that.

jsuchome:
> In general it would be really nice if we could rely on the pacemaker resources to correctly bootstrap 
> the setup.

I don't think it is possible. Our "bootstraping" is basically used to set the master password and there does not seeem a way around it. But once the pacemaker resources get started, they have their own bootstrapping, executed by RA.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,204 @@
...
+crowbar_pacemaker_sync_mark "create-database_ha_resources" do
+  revision node[:database]["crowbar-revision"]
+end
+
+include_recipe "crowbar-pacemaker::haproxy"
+
+ha_servers = CrowbarPacemakerHelper.haproxy_servers(node, "database-server")
+ha_servers = ha_servers[0].each do |n|
+  n["port"] = 3306
+end
```
jsuchome: To achieve configuration where only one server is used at a time, you might need 

- https://github.com/crowbar/crowbar-ha/pull/215
- Something like
```
ha_servers = CrowbarPacemakerHelper.haproxy_servers(node, "database-server")
haproxy_servers_nodes = ha_servers[1]
ha_servers = ha_servers[0].each do |n|
  n["port"] = 3306
  this_node = haproxy_servers_nodes[n['name']]
  n["backup"] = !CrowbarPacemakerHelper.is_cluster_founder?(this_node)
end
```

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,204 @@
...
+service_name = "galera"
+
+vip_primitive = "vip-admin-#{CrowbarDatabaseHelper.get_ha_vhostname(node)}"
+ip_addr = CrowbarDatabaseHelper.get_listen_address(node)
+
+mysql_op = {}
+mysql_op["monitor"] = {}
+mysql_op["monitor"]["interval"] = "10s"
+
+pacemaker_primitive vip_primitive do
```
jsuchome: I guess we're gonna remove this (plus related colocation and order resources) when we use HAProxy, and rely on VIP created by (included) `crowbar-pacemaker::haproxy`.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,204 @@
...
+  # note that this will also automatically start the resources
+  action :commit_new
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+crowbar_pacemaker_sync_mark "create-database_ha_resources" do
+  revision node[:database]["crowbar-revision"]
+end
+
+include_recipe "crowbar-pacemaker::haproxy"
```
jsuchome: We need to combine this with using `prepare_role_for_ha_with_haproxy` in the database model. However, it's more complicated than just
```
    vip_networks = ["admin", "public"]
    dirty = prepare_role_for_ha_with_haproxy(role, ["database", "ha", "enabled"],
      database_ha_enabled,
      database_elements,
      vip_networks)
    role.save if dirty
```

because database uses a different VIP/hostname than what haproxy uses for the openstack services (and so `crowbar-pacemaker::haproxy` looks for a wrong cluster name/VIP and fails). 
(thanks to @rhafer for investigation)

jsuchome: This https://github.com/crowbar/crowbar-ha/pull/216 might help

s-t-e-v-e-n-k: I daresay we do not want or need public in vip_networks, I think just ["admin"] would be it.

jsuchome: OK (that code was copy-pasted from other places and only slightly modified and not tested)

PR #1083, opened 2017-07-25
cinder: Fix broken downgrades in schema migrations

vuntz:


toabctl:
@vuntz sorry. I missed that one and now we merged https://github.com/crowbar/crowbar-openstack/pull/1097 . pardon.

PR #1085, opened 2017-07-25
cinder: Fix broken downgrade in schema migration

vuntz:


toabctl:
@vuntz is that a backport? do we need to forward-port it?

toabctl:
Fix in master is https://github.com/crowbar/crowbar-openstack/pull/1097

toabctl:
Backpor for stable-4.0 is https://github.com/crowbar/crowbar-openstack/pull/1098

PR #1087, opened 2017-07-25
crowbar: Use databags to store config

Itxaka:
As we commonly search for the insecure value of a service in
all cookbooks, we provide a helper to quickly gather that value.

This is only the first building block, next one will save those insecure
values into the databag.

### DIFF
--- a/crowbar_framework/app/models/openstack_service_object.rb
+++ b/crowbar_framework/app/models/openstack_service_object.rb
@@ -1,0 +1,49 @@
+#
+# Copyright 2017, SUSE LINUX Products GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# This is a subclass of ServiceObject providing some helpers methods.
+# Barclamps that have roles using pacemaker clusters should subclass this.
+#
+# It also provides some helpers that ServiceObject will wrap.
+#
+
+class OpenstackServiceObject < PacemakerServiceObject
+  def apply_role_post_chef_call(old_role, role, all_nodes)
+    @logger.debug("#{@bc_name} apply_role_post_chef_call: entering")
+    # do this in post, because we depend on values that are computed in the
+    # cookbook
+    save_config_to_databag(old_role, role)
+    @logger.debug("#{@bc_name} apply_role_post_chef_call: leaving")
+  end
+
+  def save_config_to_databag(old_role, role)
+    @logger.debug("#{@bc_name} save_config_to_databag: entering")
+    if role.nil?
+      config = nil
+    else
+      insecure = Openstack::DataBagConfig.insecure(@bc_name, role)
+
+      config = {
+        insecure: insecure
+      }
+    end
+
+    instance = Crowbar::DataBagConfig.instance_from_role(old_role, role)
+    Crowbar::DataBagConfig.save("openstack", instance, @bc_name, config)
+    @logger.debug("#{@bc_name} save_config_to_databag: leaving")
+  end
+end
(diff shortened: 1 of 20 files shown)


crowbar_framework/lib/openstack/data_bag_config.rb
```diff
@@ -0,0 +1,57 @@
...
+          # nova
+          use_ssl = attributes["ssl"]["enabled"]
+        else
+          # magnum, sahara, trove
+          use_ssl = false
+        end
+
+        insecure = use_ssl && attributes["ssl"]["insecure"]
+        unless barclamp == "keystone"
+          insecure ||= keystone_insecure(barclamp, role)
```
nicolasbock: `insecure` is always defined, isn't it, and that line doesn't do anything.

crowbar_framework/lib/openstack/data_bag_config.rb
```diff
@@ -0,0 +1,57 @@
...
+        if attributes.key?("api") && attributes["api"].key?("protocol")
+          # aodh, cinder, glance, heat, keystone, manila, neutron
+          use_ssl = attributes["api"]["protocol"] == "https"
+        elsif attributes.key?("api") && attributes["api"].key?("ssl")
+          # barbican
+          use_ssl = attributes["api"]["ssl"]
+        elsif attributes.key?("ssl") && attributes["ssl"].key?("enabled")
+          # nova
+          use_ssl = attributes["ssl"]["enabled"]
+        else
```
nicolasbock: Are we really this inconsistent in how we specify whether we are using `ssl` or not? Interesting...

crowbar_framework/lib/openstack/data_bag_config.rb
```diff
@@ -0,0 +1,57 @@
...
+          # magnum, sahara, trove
+          use_ssl = false
+        end
+
+        insecure = use_ssl && attributes["ssl"]["insecure"]
```
nicolasbock: Do we _always_ have the `ssl` key?

crowbar_framework/lib/openstack/data_bag_config.rb
matelakat: I think it's the OR boolean operation isn't it?

```
insecure = insecure || keystone_insecure(barclamp, role)
```

nicolasbock: That's my understanding as well. And since `insecure` is set to something two lines above, how would it ever become `keystone_insecure()`?

Itxaka: it will if `insecure` is either `nil` or `false`

so its more like `insecure = insecure if not false or nil else keystone_insecure`

crowbar_framework/lib/openstack/data_bag_config.rb
Itxaka: Not always, magnum for example does not have that key. I still need to test this in a full deployment, So I'll come back with the results for this

crowbar_framework/lib/openstack/data_bag_config.rb
nicolasbock: You are right! This syntax definitely takes some getting used to! :smile:

Itxaka:
Seems to work as expected, storing the proper data inthe databags:

```
root@crowbar:~ # knife data bag show crowbar-config openstack
default: 
  aodh:       
    insecure:  false
  barbican:   
    insecure:  false
  ceilometer: 
    insecure:  false
  cinder:     
    insecure:  false
  glance:     
    insecure:  false
  heat:       
    insecure:  false
  horizon:    
    insecure:  false
  keystone:   
    insecure:  false
  manila:     
    insecure:  false
  neutron:    
    insecure:  false
  nova:       
    insecure:  false
  rabbitmq:   
    address:   192.168.251.81
    password:  lMxztNEM0Nlv
    port:      5672
    url:       rabbit://nova:lMxztNEM0Nlv@192.168.251.81:5672//nova
    user:      nova
    vhost:     //nova
  sahara:     
    insecure:  false
  swift:      
    insecure:  false
  trove:      
    insecure:  false
id:       openstack
```

crowbar_framework/app/models/openstack_service_object.rb
```diff
@@ -0,0 +1,49 @@
...
+
+#
+# This is a subclass of ServiceObject providing some helpers methods.
+# Barclamps that have roles using pacemaker clusters should subclass this.
+#
+# It also provides some helpers that ServiceObject will wrap.
+#
+
+class OpenstackServiceObject < PacemakerServiceObject
+  def apply_role_post_chef_call(old_role, role, all_nodes)
```
abel-navarro: do we need the `all_nodes` argument? seems it's not used in this function.

Itxaka: This is overriding the original method found in the service_object[0] which accepts those params :)

[0] https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/service_object.rb#L1344

PR #1093, opened 2017-07-26
Cinder: get insecure values from databag

Itxaka:
the databag


Itxaka:
NOT TESTED!

chef/cookbooks/cinder/recipes/common.rb
```diff
@@ -32,24 +32,23 @@
...
-if nova_apis.length > 0
-  nova_api = nova_apis[0]
-  nova_api_insecure = nova_api[:nova][:ssl][:enabled] && nova_api[:nova][:ssl][:insecure]
-else
-  nova_api_insecure = false
-end
+glance_config = BarclampLibrary::Barclamp::Config.load(
+  "openstack",
+  "glance",
+  node[:cinder][:glance_instance]
```
abel-navarro: how's glance related to cinder here? shouldn't be separate entities?

Itxaka: So this is the glance instance that its passed from the barclamp here https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/models/cinder_service.rb#L80

So its a dependency of the cinder to have glance as its linked to that instance, which comes into play in here https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/barclamp/libraries/barclamp_library.rb#L427

Basically as per Vincent comments, there is this future-proof way of dealing with barclamp instances in case in the future there could be more than one of those barclamps deployed at the same time (so instead of glance-config-default, you could have several instances like pacemaker, glance-config-proposal_1 and glance-config-proposal2)

Itxaka:
ah, this wont work properly until the other patch has been included in the crowbar-openstack package, so probably tomorrow :/

chef/cookbooks/cinder/recipes/common.rb
```diff
@@ -32,24 +32,23 @@
...
+)
+glance_insecure = glance_config["insecure"] || false
+
+nova_insecure = BarclampLibrary::Barclamp::Config.load("openstack", "nova")["insecure"] || false
```
s-t-e-v-e-n-k: You use node[:cinder][:glance_instance] above for getting the glance configuration, but do not that here for the nova configuration, is that intended?

chef/data_bags/crowbar/migrate/cinder/200_fill_databag_with_insecure.rb
```diff
@@ -0,0 +1,23 @@
...
+  # For each role, find the role, get the value of the insecure attribute,
+  # generate the config and save the databag
+  name = "cinder"
+  ["#{name}-controller", "#{name}-volume"].each do |r|
+    role = RoleObject.find_role_by_name(r)
+    insecure = Openstack::DataBagConfig.insecure(name, role)
+    config = {
+      insecure: insecure
+    }
+    # as we dont have an old_role here, we just pass the role twice, wont affect anything
```
s-t-e-v-e-n-k: Nit: wont should be won't

chef/cookbooks/cinder/recipes/common.rb
Itxaka: Yeah, I think we cant use nova_instance here as its not passed from the barclamp config, as its not a dependency. Same stuff as above, this comes from https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/models/cinder_service.rb#L80

chef/data_bags/crowbar/migrate/cinder/200_fill_databag_with_insecure.rb
Itxaka: fixed!

Itxaka:
Now its tested.

Created a develcloud8 env, deployed barclamps up to cinder, made sure that the databags were empty, applied patch, run the migrations and the databag for cinder was filled correctly with the proper value

Itxaka:
rebased!

dirkmueller:
@Itxaka needs rebase

Itxaka:
@dirkmueller done, thakns for the heads up!

PR #1095, opened 2017-07-27
Decrease the amount of node saves

scottwulf:
Only save the node when attribute(s) have changed
Rebase of PR #674

### DIFF
--- a/chef/cookbooks/barbican/recipes/api.rb
+++ b/chef/cookbooks/barbican/recipes/api.rb
@@ -135,9 +135,6 @@ else
   bind_port = node[:barbican][:api][:bind_port]
 end
 
-node.normal[:apache][:listen_ports_crowbar] ||= {}
-node.normal[:apache][:listen_ports_crowbar][:barbican] = { plain: [bind_port] }
-
 crowbar_openstack_wsgi "WSGI entry for barbican-api" do
   bind_host bind_host
   bind_port bind_port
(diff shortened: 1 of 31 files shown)


Itxaka:
@nicolasbock we are encouraging to have people restore the old PR and manage themselves as its a pain in the ass to push to other people changes (for once, you need to disable 2 factor auth to be able to push).
You would have noticed that I also have several of those from Vincent :D

nicolasbock:
@Itxaka Ah, I see. I did notice that you like to do that too, which is why I often feel like I am having a deja-vu moment when I am reviewing your PRs :wink:. And Github requires you to disable two-factor auth? That's ridiculous! Ok, so my potential objections are all gone :smile:

Itxaka:
@nicolasbock yeah, not sure why but when pushing to someone else's branch, it requires your user+password+auth code instead of using your ssh key for identification, and you cannot insert the auth key in the terminal git client (I think?) so you end up disabling and enabling 2 factor auth everytime you need to push :/

nicolasbock:
@Itxaka that's complete BS from GitHub!

abel-navarro:
@scottwulf this PR has less files changed than the original one. Is that intended?

scottwulf:
Yes, less files is intended. The memcached.rb recipes for keystone and nova no longer exist resulting in two less files. The setting of [:memcached][:listen] has been removed from horizon/recipes/server.rb resulting in one less file. And nova/recipes/docker.rb no longer exists.

Glad you asked. I had originally skipped the changes to barbican/recipes/api.rb. I wanted to understand them first by either reading the commit message and/or checking with @vuntz before applying and testing them.

PR #1096, opened 2017-07-27
heat: justify trustee credentials

mtupitsyn:
Using keystoneauth credentials for trustee user was deprecated, and
removed since ocata release. Trustee credentials must be defined in
a separate conf section now. We already have this section populated,
but it needs a few adjustments: need to provide auth_type, user and
project domain ids. Also, project_name needs to be removed, because
trustee token is scoped to a trust.

### DIFF
--- a/chef/cookbooks/heat/templates/default/heat.conf.erb
+++ b/chef/cookbooks/heat/templates/default/heat.conf.erb
@@ -137,10 +137,11 @@ key_file = <%= @heat_ssl[:keyfile] %>
 <% end %>
 
 [trustee]
+auth_type = password
 auth_url = <%= @keystone_settings['internal_auth_url'] %>
-project_name = <%= @keystone_settings['service_tenant'] %>
 username = <%= @keystone_settings['service_user'] %>
 password = <%= @keystone_settings['service_password'] %>
+user_domain_name = <%= @keystone_settings['admin_domain'] %>
 
 [oslo_messaging_rabbit]
 rabbit_use_ssl = <%= @rabbit_settings[:use_ssl] %>


chef/cookbooks/heat/templates/default/heat.conf.erb
```diff
@@ -137,10 +137,12 @@ key_file = <%= @heat_ssl[:keyfile] %>
 <% end %>
 
 [trustee]
+auth_type = password
 auth_url = <%= @keystone_settings['internal_auth_url'] %>
-project_name = <%= @keystone_settings['service_tenant'] %>
```
cmurphy: project_domain_{id,name} is only applicable when the auth is scoped to a project, so either setting still needs to be here or project_domain_id should be removed.

toabctl:
Does this solve the first tempest failure on https://ci.suse.de/job/openstack-mkcloud/74873/consoleFull ?

chef/cookbooks/heat/templates/default/heat.conf.erb
jgrassler: It should probably removed entirely. Heat creates trusts scoped to the user's project and delegates them to the user in this section. Having project information in there can only ever cause harm because all the Keystone authentication libraries will balk if you supply additional project information on top of the credentials required to consume the trust (user name, password, trust ID).

I'm currently building a cloud to test this in. I'll try and see what happens if the project information entirely. Maybe Heat uses this information for something unrelated to trust creation/consumption, but it definitely does not need it for either.

mtupitsyn:
@toabctl  Yes, I think so (sorry, I didn't find a way to reply inline). There seems to be no heat-engine log from controller stored by the Jenkins job, but error reported to Tempest looks exactly like the issue with trustee credentials.

chef/cookbooks/heat/templates/default/heat.conf.erb
jgrassler: I tested this PR now and can confirm that `project_domain_id` is unnecessary. I took the liberty of adding a commit with that change.

jgrassler: Also, feel free to meld my commit into yours. I'd rather not merge two commits when all the second one does is remove part of the first one :-)

mtupitsyn: @jgrassler Sorry, I did not find a way to meld your commit... My understanding is that you can only squash commits at the moment of pull request merge. I don't see any rebase or merge controls, neither in this pull request, nor in original branch in mtupitsyn/crowbar-openstack repo. So to address your and @cmurphy 's comments, I've added one more empty commit with updated subject. The idea is to squash all 3 commits together at the time of pull request merge, using git message from 3rd (most comprehensive) commit.

cmurphy: You can squash the commits before merging. In my opinion code review is also review of git correctness so it is worthwhile to get it in the state you want it to be in when it merges.

```
git fetch <your fork remote name>
git checkout <your fork remote name>/feature/heat-ocata
git rebase -i HEAD~3
(pick/fixup/squash commits as necessary)
git push -f <your fork remote name> HEAD:feature/heat-ocata
```

See https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

dirkmueller: I've performed that history cleanup now

toabctl:
> @toabctl Yes, I think so (sorry, I didn't find a way to reply inline). 

@mtupitsyn  Press 'r'

> There seems to be no heat-engine log from controller stored by the Jenkins job

It is in the supportconfigs. This is the standard format for SUSE customers. Whenever there's a bug, supportconfigs are usually attached for the different nodes. You can use [soc-ci](https://github.com/SUSE-Cloud/automation/blob/master/hostscripts/soc-ci/README.soc-ci.md) (soc-ci os-mkcloud-artifacts $JOB-ID) to download and extract the logs (see the README).

> but error reported to Tempest looks exactly like the issue with trustee credentials.

It doesn't solve the tempest issue. I tested it locally.

stefannica:
@toabctl it should've been 'user_domain_name' instead of 'user_domain_id'. With this correction in place, the tempest test case passes now.

PR #1099, opened 2017-07-28
cinder: Add memcached_servers for token caching

cmurphy:
The in-process token cache of keystonemiddleware  was deprecated in
Mitaka[1]. This patch uses the memcached setup helper to set up a local
memcached server for the cinder controller and configure its use in the 
[keystone_authtoken] section.

While the keystonemiddleware documentation claims (at the time of this
commit) that the keystone and memcached servers must both be in UTC[2],
that is outdated information and has been corrected[3]. Therefore this
is a safe change to make even though we don't enforce UTC in our 
deployent.

Setting up token encryption in memcached is an optional feature that we
can enable later if desired.

[1] https://docs.openstack.org/releasenotes/keystonemiddleware/mitaka.html
[2] https://docs.openstack.org/keystonemiddleware/latest/middlewarearchitecture.html#memcached-and-system-time
[3] https://bugs.launchpad.net/keystone/+bug/1221087


chef/data_bags/crowbar/template-cinder.schema
```diff
@@ -189,10 +189,10 @@
                       "san_password": { "type": "str", "required": true },
                       "san_thin_provision": { "type": "bool", "required": true },
                       "eqlx_group_name": { "type": "str", "required": true },
-                      "eqlx_use_chap": { "type": "bool", "required": true },
-                      "eqlx_chap_login": { "type": "str", "required": true },
-                      "eqlx_chap_password": { "type": "str", "required": true },
-                      "eqlx_cli_timeout": { "type": "int", "required": true },
+                      "use_chap_auth": { "type": "bool", "required": true },
```
toabctl: Is there a migration needed for ```eqlx_use_chap```, ```eqlx_chap_login``` and ```eqlx_chap_password``` ? Could be included in the 200 migration.

cmurphy: There is already a migration - https://github.com/crowbar/crowbar-openstack/commit/6aa34a2f264636044d66428bfa92783b2b26bcd0

You had done the migration but only changed the schema for the volume_defaults and not the volumes.

chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -299,6 +290,11 @@ password = <%= @keystone_settings['service_password'] %>
 project_name = <%= @keystone_settings['service_tenant'] %>
 project_domain_name = <%= @keystone_settings["admin_domain"]%>
 user_domain_name = <%= @keystone_settings["admin_domain"] %>
+service_token_roles_required = true
+service_token_roles = admin
+memcached_servers = <%= @memcached_servers.join(',') %>
+memcache_security_strategy = ENCRYPT
+memcache_secret_key = <%= node[:cinder][:memcache_secret_key] %>
```
stefannica: Just some food for thought: it might be more efficient to have a secret key that is shared by all services using memcached keystone tokens. The memcached keys for stored keystone tokens are generated from the token id, secret_key and security_strategy (see: https://github.com/openstack/keystonemiddleware/blob/master/keystonemiddleware/auth_token/_cache.py#L265).

This means that if two or more service instances need to store the same token in the same memcached instance (doesn't this happen e.g. with user tokens ?), then several copies of the same token will be stored instead of just one.

chef/data_bags/crowbar/template-cinder.schema
toabctl: Oh. So this needs a backport, right? Thanks for fixing it!

cmurphy: Hmm yes I think it actually does need a backport, so I should probably split it into its own commit.

This pull request was my attempt to split out the memcache changes from the rest of the ocata updates so the main discussion on these parameter changes should happen here https://github.com/crowbar/crowbar-openstack/pull/1066

chef/cookbooks/cinder/templates/default/cinder.conf.erb
dirkmueller: @stefannica thats the whole point of encrypting it in the first place (so that exploiting one service doesn't give you admin on another service)

cmurphy:
@toabctl could you re-review? I removed the commit you were commenting on and it was merged in https://github.com/crowbar/crowbar-openstack/pull/1066

PR #1100, opened 2017-07-28
mysql: Remove debian-sys-maint user usage

rsalevsky:
This commit removes the usage of the debian-sys-maint user and the
related server_debian_password. Both is not relevant for us and only
creates confusion.


rsalevsky:
I'm also thinking about removing the `server_repl_password`.

rsalevsky:
I don't know when it got introduced but for use I don't see a use case.

nicolasbock:
@rsalevsky Thanks!

jsuchome:
Do we need `debian.cnf.erb` now? Or even further, what about removing whole `if platform_family?("debian")` part from recipe (and relevant templates)?

rsalevsky:
@jsuchome For now I would like to limit the cleanup to this part because it will help me to get rid of the grants.sql file. But in general I agree that we can remove the other debian specific parts.

jsuchome:
[Sanitize and enhance mariadb configuration done by cookbook](https://trello.com/c/PcPTyEX7/1-sanitize-and-enhance-mariadb-configuration-done-by-cookbook)

PR #1103, opened 2017-07-31
manila: Update keystone_authtoken parameters

cmurphy:
Use the service_token_roles and service_token_roles_required parameters
to ensure properly authenticated interservice communication and to avoid
deprecation warnings[1].

Add a memcached instance for the manila controller and configure manila
to use it for token caching, since the in-process token cache is
deprecated and produces warnings[2]. This includes adding a parameter to
generate a secret key to use for cache encryption.

[2] https://docs.openstack.org/releasenotes/keystonemiddleware/mitaka.html

manila: Add memcached servers

### DIFF
--- a/chef/data_bags/crowbar/template-manila.json
+++ b/chef/data_bags/crowbar/template-manila.json
@@ -8,6 +8,7 @@
       "use_syslog": false,
       "service_user": "manila",
       "service_password": "",
+      "memcache_secret_key": "",
       "rabbitmq_instance": "none",
       "keystone_instance": "none",
       "database_instance": "none",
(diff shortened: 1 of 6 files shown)


chef/data_bags/crowbar/template-manila.json
```diff
@@ -96,7 +97,7 @@
     "manila": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 102,
+      "schema-revision": 200,
```
toabctl: The migration file is not here. maybe you missed ```git add``` ?

cmurphy: Yep, fixed.

PR #1105, opened 2017-08-01
manila: Init node hash for ceph config (bsc#1051229)

toabctl:
When adding extra sections to the ceph config, the node hash might
not be initialized which leads to:

undefined method `[]' for nil:NilClass


toabctl:
Backport is https://github.com/crowbar/crowbar-openstack/pull/1106

PR #1107, opened 2017-08-01
mysql: Remove repl user usage

rsalevsky:
We don't use replication for HA, so let's get rid of the replication
user and password.

Another step to get rid of the `grants.sql` file usage.


rsalevsky:
@jsuchome I like small PR's with single topics more then big PR's which are more difficult to review. For backporting I would combine them.

jsuchome:
[Sanitize and enhance mariadb configuration done by cookbook](https://trello.com/c/PcPTyEX7/1-sanitize-and-enhance-mariadb-configuration-done-by-cookbook)

PR #1109, opened 2017-08-01
heat: ocata cookbook update

stefannica:
The following heat configuration options deprecated in Ocata (or earlier) are removed or replaced in this PR:
    
 - [default]/verbose deprecated option removed
 - [keystone_authtoken]/service_token_roles_required is set to True to enable fetching expired tokens based on valid service tokens and [keystone_authtoken]/service_token_roles is set to the admin role
 - the following deprecated keystone_authtoken configuration options: _identity_uri_, _admin_user_,
_admin_password_, and _admin_tenant_name_ are replaced with _auth_url_, _project_name_, _project_domain_name_ and _user_domain_name_

The deprecated keystonemiddleware.auth_token in-process token cache used by heat services is replaced with a memcached server. When HA is used, several memcached servers are configured and used, one for each heat-server node. 
The keystone token data stored in memcached is authenticated and encrypted using a generated secret key.

References:  
 - Ocata release notes for heat: https://docs.openstack.org/releasenotes/heat/ocata.html
 - Mitaka release notes for keystonemiddleware: https://docs.openstack.org/r
eleasenotes/keystonemiddleware/mitaka
 - memcached protection options: https://docs.openstack.org/keystonemiddlewa
re/latest/middlewarearchitecture.html#memcache-protection

### DIFF
--- a/chef/data_bags/crowbar/template-heat.json
+++ b/chef/data_bags/crowbar/template-heat.json
@@ -4,7 +4,6 @@
   "attributes": {
     "heat": {
       "debug": false,
-      "verbose": true,
       "max_header_line": 16384,
       "rabbitmq_instance": "none",
       "database_instance": "none",
@@ -41,7 +40,7 @@
     "heat": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 200,
+      "schema-revision": 201,
       "element_states": {
         "heat-server": [ "readying", "ready", "applying" ]
       },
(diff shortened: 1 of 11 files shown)


chef/data_bags/crowbar/template-heat.json
```diff
@@ -40,7 +41,7 @@
     "heat": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 200,
+      "schema-revision": 202,
```
cmurphy: Is this misnumbered? I only see migrations for 200 and 201 here.

stefannica:
@cmurphy you were right about 'verbose' also being in the custom view. Furthermore, this is not an isolated case - many other barclamps are similar in this respect, amongst which nova and glance, for which I overlooked this aspect. I'll have to go back and correct them as well.

PR #1111, opened 2017-08-02
mysql: Add support for HA provisioning with Galera

s-t-e-v-e-n-k:
Add a ha_galera recipe that will bootstra Galera on a cluster, as
well creating Pacemaker resources for the cluster, with other support
sprinkled into the mysql::server recipe.

For reference: https://github.com/crowbar/crowbar-openstack/pull/1079 was the original PR that had more stuff but we decided to solve them separately. (EDIT by jsuchome)

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -1,0 +1,149 @@
+#
+# Cookbook Name:: mysql
+# Recipe:: ha_galera
+#
+# Copyright 2017, SUSE Linux GmbH.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+resource_agent = "ocf:heartbeat:galera"
+
+package "galera-3"
+
+# Wait for all nodes to reach this point so we know that all nodes will have
+# all the required packages installed before we create the pacemaker
+# resources
+crowbar_pacemaker_sync_mark "sync-database_before_ha" do
+  revision node[:database]["crowbar-revision"]
+end
+
+unless node[:database][:galera_bootstrapped]
+  mysql_service_name = "mysql"
+
+  directory "/var/run/mysql/" do
+    owner "mysql"
+    group "root"
+    mode "0755"
+    action :create
+  end
+
+  execute "mysql_install_db" do
+    command "mysql_install_db"
+    action :run
+  end
+
+  # Start mariadb on the founder. At this point, the configuration file has no
+  # cluster members defined, so it will listen for the other nodes to join it.
+  # We will stop it again after bootstrapping.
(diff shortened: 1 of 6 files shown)


jsuchome:
Seems to work as expected. I have to yet test the OpenStack services access, once we have fix for db_maker behavior...

jsuchome:
We need something like 
```
13c13
<     if node[:database][:ha][:enabled]
---
>     if node[:database][:ha][:enabled] && node[:database][:sql_engine] != "mysql"
```

In `cookbooks/database/libraries/crowbar.rb` until we have correct haproxy setup. @s-t-e-v-e-n-k @rhafer could we add it to this PR (as separate commit)?

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,149 @@
...
+    service_name mysql_service_name
+    action :stop
+  end
+
+  ruby_block "mark node for galera bootstrap" do
+    block do
+      node.set[:database][:galera_bootstrapped] = true
+      node.save
+    end
+    notifies :create, "template[/etc/my.cnf]"
```
rhafer: Note: This will not update the config file with the none empty cluster_address. (I guess that's what would need to happen here)

It goes unnoticed because the pacemaker resource override the cluster address on the commandline.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,149 @@
...
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+    only_if "/usr/bin/mysql -u root -e 'show databases;'"
+  end
+
+  # Wait on all nodes for the founder to complete the bootstrap
+  crowbar_pacemaker_sync_mark "sync-database_after_bootstrap_founder" do
+    revision node[:database]["crowbar-revision"]
+  end
+
+  service "starting mysql on remaining nodes" do
```
rhafer: Since we're setting ```cluster_addresses = "gcomm://"``` on all nodes is ```server.rb```. We're effectively bootstrapping 3 different galera clusters here. Some magic in the galera resource agent seems to clean this up later somehow, by forcing one node to bootstrap again. But this just works by accident it seems.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,149 @@
...
+    service_name mysql_service_name
+    action :start
+    not_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+
+  execute "assign-root-password-galera" do
+    command "/usr/bin/mysqladmin -u root password \"#{node[:mysql][:server_root_password]}\""
+    action :run
+    not_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+    only_if "/usr/bin/mysql -u root -e 'show databases;'"
```
rhafer: If the cluster were bootstrapped correctly this would not be needed. The password change should just be replicated to all nodes.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,149 @@
...
+  end
+
+  crowbar_pacemaker_sync_mark "sync-database_after_bootstrap_rest" do
+    revision node[:database]["crowbar-revision"]
+  end
+
+  service "stop mysql after boostrap completed" do
```
rhafer: Before stopping everything again we would actually need to make sure that the cluster if fully synced. (I.e. the initial state transfer happened). e.g. by looking at he ```wsrep_``` status attributes on the nodes.

rhafer:
This can be closed now since #1134 was merged

PR #1112, opened 2017-08-02
mysql: Drop test database

rsalevsky:


chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -189,6 +189,23 @@
   only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
+database "drop test database" do
+  connection db_connection
+  database_name "test"
+  provider db_settings[:provider]
+  action :drop
+  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
jsuchome: Do we have to do this for each chef-client run?

rsalevsky: Once should be enough. Is there a easy way to realize this?

jsuchome: Something like we were doing with that grants file before - touch some file after it is done for the first time. Or save some indication into the node data.
I'm not sure if it would make things better, maybe some chef expert ( @Itxaka ;-) ?) could have a hint.

rhafer: I'd prefer a node attribute. Just read during the compile phase (i.e. not inside the ```only_if``` guard) and either let ```database_user``` or ```database``` send a notification to a ruby_block that sets it to true (or write it at the end of the compile phase). Similar to what we do for the galera boostrapped attribute.

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -155,38 +155,56 @@
...
+    action :grant
+    only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+
+  database "drop test database" do
+    connection db_connection
+    database_name "test"
+    provider db_settings[:provider]
+    action :drop
+    only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
jhesketh: Wouldn't we want to drop this regardless of ha status?
Perhaps we should do an equivalent of "drop database if exists"

rhafer: > Wouldn't we want to drop this regardless of ha status?

That's true, and that's what this code actually does. But in the case of an HA setup we only want it to execute the deletion on a single node of the cluster. This is what `CrowbarPacemakerHelper.is_cluster_founder?(node)` is good for. When deploying a cluster we ensure that there's exactly one node where this condition is true.

jhesketh: Ah, thanks, sorry I missed that. Would it make sense to add an "if exists" condition to the sql? (I'm not very familiar with chef's dbal)

rhafer: > Would it make sense to add an "if exists" condition to the sql?

The `database` provider does that kind of check it self (https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/database/libraries/provider_database_mysql.rb#L54) to avoid deleting a non-existing database.

> (I'm not very familiar with chef's dbal)

There is no real database abstraction layer in chef itself :). All we have it the `database` and `database_user` resources/providers in: https://github.com/crowbar/crowbar-openstack/tree/master/chef/cookbooks/database/libraries/

jhesketh: Awesome, thanks for the pointers :-)

rsalevsky:
Rebased

PR #1113, opened 2017-08-02
Do not return VIP address as the database listening adress.

jsuchome:
This is only temporary solution, while we are missing proper
haproxy setup in database (mysql) cluster.


jsuchome:
@nicolasbock not yet (we've just tested some stuff manually), that's why we want to do this simple way first and later finetune with the proper one

nicolasbock:
@jsuchome Thanks!

jsuchome:
Closing ... it's already present in other PR's

PR #1114, opened 2017-08-02
crowbar-openstack: Use databag to search for rabbitmq settings

Itxaka:
Use the databag to provide the rabbitmq settings to other cookbooks.
Includes a rabbitmq migration to fill the rabbitmq databag in
case rabbitmq is already deployed when applying this commit.

### DIFF
--- a/chef/data_bags/crowbar/migrate/rabbitmq/200_fill_databag_with_settings.rb
+++ b/chef/data_bags/crowbar/migrate/rabbitmq/200_fill_databag_with_settings.rb
@@ -1,0 +1,45 @@
+def upgrade(ta, td, a, d)
+  # Check if the proposal is active. If its not then we dont need to do anything
+  # with the databag as it will be filled by the deployment
+  prop = Proposal.where(barclamp: "rabbitmq").first
+  unless prop.nil?
+    return a, d unless prop.active?
+  end
+
+  # Find the role, get all the values and save the databag
+  role = RoleObject.find_role_by_name("rabbitmq-config-default")
+  node = NodeObject.find("roles:rabbitmq-config-default").first
+  address = node["rabbitmq"]["address"]
+
+  client_ca_certs = if a["ssl"]["enabled"] && !a["ssl"]["insecure"]
+    a["ssl"]["client_ca_certs"]
+  end
+
+  port = if a["ssl"]["enabled"]
+    a["ssl"]["port"]
+  else
+    a["port"]
+  end
+
+  config = {
+    address: address,
+    port: port,
+    user: a["user"],
+    password: a["password"],
+    vhost: a["vhost"],
+    use_ssl: a["ssl"]["enabled"],
+    client_ca_certs: client_ca_certs,
+    url: "rabbit://#{a["user"]}:#{a["password"]}@#{address}:#{port}/#{a["vhost"]}"
+  }
+  # as we dont have an old_role here, we just pass the role twice, wont affect anything
+  # as the instance_from_role method has an || to use any of those (old_role or role)
+  instance = Crowbar::DataBagConfig.instance_from_role(role, role)
+  Crowbar::DataBagConfig.save("openstack", instance, "rabbitmq", config)
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  # There is no current way of removing the databag
+  # not even sure if we should do it as it won't affect anything
+  return a, d
+end
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -100,57 +100,12 @@ class CrowbarOpenStackHelper
 
   def self.rabbitmq_settings(node, barclamp)
     instance = node[barclamp][:rabbitmq_instance] || "default"
+    config = BarclampLibrary::Barclamp::Config.load("openstack", "rabbitmq", instance)
 
-    # Cache the result for each cookbook in an instance variable hash. This
-    # cache needs to be invalidated for each chef-client run from chef-client
-    # daemon (which are all in the same process); so use the ohai time as a
-    # marker for that.
-    if @rabbitmq_settings_cache_time != node[:ohai_time]
-      if @rabbitmq_settings
-        Chef::Log.info("Invalidating rabbitmq settings cache " \
-                       "on behalf of #{barclamp}")
-      end
-      @rabbitmq_settings = nil
-      @rabbitmq_settings_cache_time = node[:ohai_time]
-    end
+    Chef::Log.warn("No RabbitMQ server found!") if config.empty?
+    Chef::Log.info("RabbitMQ server found at #{config[:address]}") unless config.empty?
 
-    if @rabbitmq_settings && @rabbitmq_settings.include?(instance)
-      Chef::Log.info("RabbitMQ server found at #{@rabbitmq_settings[instance][:address]} [cached]")
-    else
-      @rabbitmq_settings ||= Hash.new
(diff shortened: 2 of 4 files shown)


chef/data_bags/crowbar/migrate/rabbitmq/200_fill_databag_with_settings.rb
```diff
@@ -0,0 +1,30 @@
...
+  # Check if the proposal is active. If its not then we dont need to do anything
+  # with the databag as it will be filled by the deployment
+  prop = Proposal.where(barclamp: "rabbitmq").first
+  unless prop.nil?
+    return a, d unless prop.active?
+  end
+
+  # Find the role, get all the values and save the databag
+  role = RoleObject.find_role_by_name("rabbitmq-config-default")
+  config = {
```
abel-navarro: don't we need the `use_ssl` and `client_ca_certs` anymore?

Itxaka: Good point!

cmurphy:
Could you provide the rationale for this change in the commit message? What was the problem we faced before that necessitated this change?

Itxaka:
@cmurphy you are totally rigth, I updated the commit message to provide the why behind this PR.

Thanks for the heads up!

chef/data_bags/crowbar/migrate/rabbitmq/200_fill_databag_with_settings.rb
```diff
@@ -0,0 +1,36 @@
...
+    use_ssl: a[:ssl][:enabled],
+    client_ca_certs: client_ca_certs,
+    url: "rabbit://#{a[:user]}:#{a[:password]}@#{a[:address]}:#{a[:port]}/#{a[:vhost]}"
+  }
+  # as we dont have an old_role here, we just pass the role twice, wont affect anything
+  # as the instance_from_role method has an || to use any of those (old_role or role)
+  instance = Crowbar::DataBagConfig.instance_from_role(role, role)
+  Crowbar::DataBagConfig.save("openstack", instance, "rabbitmq", config)
+  return a, d
+end
```
Itxaka: Umm, it may be possible that they are strings instead of symbols, but I tested the other migrations and they worked with the same approach.

Let me test this further and thanks!

Itxaka: oh wow, it is that. Not sure how the other migrations worked then???? ARGH.

Thanks! Fixing this and the other (already merged :() migrations

Itxaka: oh I see, it worked because I tested this one: https://github.com/crowbar/crowbar-openstack/pull/1093 which access the values via string instead of symbol. So its all my dumb mistake, because the other migrations wont trigger on a mkcloud.

Thanks again for finding this one @cmurphy !

chef/data_bags/crowbar/migrate/rabbitmq/200_fill_databag_with_settings.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  # Find the role, get all the values and save the databag
+  role = RoleObject.find_role_by_name("rabbitmq-config-default")
+
+  client_ca_certs = if a["ssl"]["enabled"] && !a["ssl"]["insecure"]
+    a["ssl"]["client_ca_certs"]
+  end
+  config = {
+    address: a["address"],
```
cmurphy: I'm still having trouble with this - seems like "address" isn't in the existing databag and so is getting set to null, which breaks the services because their transport_url ends up pointing at localhost.

Itxaka: That is true, address is not part of the template but of the node, so I should grab the node that has rabbitmq instead and get it from there.

I got confused there due to the saving of that value to the databag on teh rabbitmq_service.rb file, but it should not be there.

Itxaka:
@cmurphy done and tested with both normal deployment and HA deployment, that should get us the proper address and save it to the databag properly

nicolasbock:
@Itxaka I'll wait a little bit before I OK your PR this time :smirk:

chef/data_bags/crowbar/migrate/rabbitmq/200_fill_databag_with_settings.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+  config = {
+    address: address,
+    port: a["port"],
+    user: a["user"],
+    password: a["password"],
+    vhost: a["vhost"],
+    use_ssl: a["ssl"]["enabled"],
+    client_ca_certs: client_ca_certs,
+    url: "rabbit://#{a["user"]}:#{a["password"]}@#{a["address"]}:#{a["port"]}/#{a["vhost"]}"
```
cmurphy: the a["address"] here should just be address

Itxaka: good thing @nicolasbock waited lmao

nicolasbock: I know :smirk:

chef/data_bags/crowbar/migrate/rabbitmq/200_fill_databag_with_settings.rb
```diff
@@ -0,0 +1,38 @@
...
+  role = RoleObject.find_role_by_name("rabbitmq-config-default")
+  node = NodeObject.find("roles:rabbitmq-config-default").first
+  address = node["rabbitmq"]["address"]
+
+  client_ca_certs = if a["ssl"]["enabled"] && !a["ssl"]["insecure"]
+    a["ssl"]["client_ca_certs"]
+  end
+  config = {
+    address: address,
+    port: a["port"],
```
cmurphy: Sorry but there's still a problem. In the chef helper there was logic to use either the regular port or the ssl port depending on whether ssl is enabled, but that logic is missing here.

Itxaka: True, and actually this needs change in the service file as well, as we are only storing the normal port and not the ssl port.

Itxaka: should ahve look into the template instead of into the cookbook attributes :/

chef/data_bags/crowbar/migrate/rabbitmq/200_fill_databag_with_settings.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  config = {
+    address: address,
+    port: port,
+    user: a["user"],
+    password: a["password"],
+    vhost: a["vhost"],
+    use_ssl: a["ssl"]["enabled"],
+    client_ca_certs: client_ca_certs,
+    url: "rabbit://#{a["user"]}:#{a["password"]}@#{address}:#{a["port"]}/#{a["vhost"]}"
```
cmurphy: nope fix the port here too

Itxaka: OMG. I think I should stop working today and take a break down the park or something  ¬_¬

Itxaka:
Thanks @cmurphy for that in-depth review :+1: 
Thanks  @nicolasbock for still reviewing my patches after all the dismissed reviews :+1:

nicolasbock:
I don't give up that easily @Itxaka :smile:

PR #1116, opened 2017-08-02
Heat: use the databags to find if magnum is deployed

Itxaka:
Instead of doing a node search which is expensive, check
for the existance of the magnum databag which will mean that
magnum is deployed


sayalilunkad:
Just rebuilt the job with magnum enabled.

PR #1117, opened 2017-08-02
Horizon: Use databag searches for insecure values

Itxaka:


stefannica:
@Itxaka why is this change necessary ? is there a bug report or some other type of description that I can look at to understand the problem that this PR is trying to solve ?

Itxaka:
@stefannica https://suse-jira.dyndns.org/browse/SCRD-467 is the task that involves this.

Basically, chef searches are very costly while static items like getting the databags are pretty cheap, so in order to improve performance, especially when we have big environments (in which the searches get exponentially costlier) we are trying to move as much as possible to store and use databags for some of the configurations.

toabctl:
@Itxaka adding some description to the commit msg why you want this would be good

nicolasbock:
@stefannica I just had a nice discussion with @Itxaka about this: Have a look at 1a869e073eb63d397bba12844705c4dab1e36cbf which adds the code that creates the databag in the first place.

PR #1119, opened 2017-08-02
Neutron: get insecure values from databag

Itxaka:
the databag


Itxaka:
uhhh, how can the testbuild fail and mkcloud succeed? This was green a couple of hours ago ¬_¬

PR #1120, opened 2017-08-02
Wait for mysql to stop

jsuchome:
Includes https://github.com/crowbar/crowbar-openstack/pull/1111 , interesting is only 2nd commit.


jsuchome:
Does not seem to be needed, @rhafer has solved it differently in https://github.com/crowbar/crowbar-openstack/pull/1134

PR #1123, opened 2017-08-03
HAProxy setup for mariadb (galera) cluster

jsuchome:
Requires https://github.com/crowbar/crowbar-ha/pull/215

### DIFF
--- a/chef/cookbooks/mysql/attributes/server.rb
+++ b/chef/cookbooks/mysql/attributes/server.rb
@@ -75,3 +75,6 @@ default["mysql"]["tunable"]["expire_logs_days"]         = 10
 default["mysql"]["tunable"]["max_binlog_size"]          = "100M"
 default["mysql"]["tunable"]["auto_increment_increment"] = 10
 default["mysql"]["tunable"]["auto_increment_offset"]    = 1
+
+# Ports to bind to when haproxy is used
+default[:mysql][:ha][:ports][:admin_port] = 3306
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -142,3 +142,26 @@ end
 crowbar_pacemaker_sync_mark "sync-database_root_password" do
   revision node[:database]["crowbar-revision"]
 end
+
+include_recipe "crowbar-pacemaker::haproxy"
+
+ha_servers = CrowbarPacemakerHelper.haproxy_servers_for_service(
+  node, "mysql", "database-server", "admin_port"
+).sort_by { |s| s["name"] }
+
+# Let all nodes but one act as backup (standby) servers.
+# Backup server is only used when non-backup one is down. Thus we prevent possible deadlocks
+# from OpenStack services writing to the database on different nodes at once.
+ha_servers = ha_servers.each_with_index do |n, i|
+  n["backup"] = i > 0
+end
+
+haproxy_loadbalancer "galera" do
+  address CrowbarDatabaseHelper.get_listen_address(node)
+  port 3306
+  mode "tcp"
+  options ["mysql-check"]
+  stick ({ "on" => "dst" })
+  servers ha_servers
+  action :nothing
+end.run_action(:create)
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -160,7 +160,7 @@ haproxy_loadbalancer "galera" do
   address CrowbarDatabaseHelper.get_listen_address(node)
   port 3306
   mode "tcp"
-  options ["mysql-check"]
+  options ["mysql-check user haproxy"]
   stick ({ "on" => "dst" })
   servers ha_servers
   action :nothing
--- a/crowbar_framework/app/models/database_service.rb
+++ b/crowbar_framework/app/models/database_service.rb
@@ -137,22 +137,38 @@ class DatabaseService < PacemakerServiceObject
     database_elements, database_nodes, database_ha_enabled = role_expand_elements(role, "database-server")
     Openstack::HA.set_controller_role(database_nodes) if database_ha_enabled
 
-    prepare_role_for_ha(role, ["database", "ha", "enabled"], database_ha_enabled)
+    vip_networks = ["admin"]
+    dirty = prepare_role_for_ha_with_haproxy(role, ["database", "ha", "enabled"],
+      database_ha_enabled,
+      database_elements,
+      vip_networks)
+    role.save if dirty
+
     reset_sync_marks_on_clusters_founders(database_elements)
+
+    sql_engine = role.default_attributes["database"]["sql_engine"]
 
     if database_ha_enabled
       net_svc = NetworkService.new @logger
-      unless database_elements.length == 1 && PacemakerServiceObject.is_cluster?(database_elements[0])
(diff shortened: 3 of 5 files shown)


jsuchome:
That last commit might be totally wrong, I realize I'm playing with something I do not really understand... an earlier alternative was https://github.com/crowbar/crowbar-ha/pull/216

jsuchome:
[Confiugure HAProxy for galera](https://trello.com/c/BQJLG0lL/7-confiugure-haproxy-for-galera)

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -142,3 +142,30 @@
...
+  # Backup server is only used when non-backup one is down. Thus we prevent possible deadlocks
+  # from OpenStack services writing to the database on different nodes at once.
+  n["backup"] = n["name"] != founder_name
+end
+
+haproxy_loadbalancer "galera" do
+  address CrowbarDatabaseHelper.get_listen_address(node)
+  port node[:mysql][:api][:admin_port]
+  mode "tcp"
+  options ["tcplog", "mysql-check"]
```
jsuchome: As we're not saving haproxy logs anywhere, "tcplog" is probably not needed anyway

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -142,3 +142,30 @@
...
+  "pacemaker_founder:true AND pacemaker_config_environment:#{node[:pacemaker][:config][:environment]}"
+).first
+founder_name = founder[:hostname]
+
+ha_servers = CrowbarPacemakerHelper.haproxy_servers_for_service(
+  node, "mysql", "database-server", "admin_port"
+)
+
+ha_servers = ha_servers.each do |n|
+  # Let all nodes but one act as backup (stnadby) servers.
```
jhesketh: (nit) typo s/stnadby/standby

chef/cookbooks/mysql/attributes/server.rb
```diff
@@ -75,3 +75,6 @@
 default["mysql"]["tunable"]["max_binlog_size"]          = "100M"
 default["mysql"]["tunable"]["auto_increment_increment"] = 10
 default["mysql"]["tunable"]["auto_increment_offset"]    = 1
+
+# Ports to bind to when haproxy is used for the real ports
```
s-t-e-v-e-n-k: But this comment is a lie -- 3306 is the correct port number for mysql, and haproxy will only be listening on the VIP anyway.

crowbar_framework/app/models/database_service.rb
```diff
@@ -137,11 +137,20 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+
     reset_sync_marks_on_clusters_founders(database_elements)
 
-    if database_ha_enabled
-      net_svc = NetworkService.new @logger
+    sql_engine = role.default_attributes["database"]["sql_engine"]
+
+    net_svc = NetworkService.new @logger
+
+    if database_ha_enabled && sql_engine == "postgresql"
```
s-t-e-v-e-n-k: I sort of want to increase the level here -- have an if database_ha_enabled block and then a case block in that with a mysql and postgresql section.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -142,3 +142,30 @@
...
+  # Backup server is only used when non-backup one is down. Thus we prevent possible deadlocks
+  # from OpenStack services writing to the database on different nodes at once.
+  n["backup"] = n["name"] != founder_name
+end
+
+haproxy_loadbalancer "galera" do
+  address CrowbarDatabaseHelper.get_listen_address(node)
+  port node[:mysql][:api][:admin_port]
+  mode "tcp"
+  options ["mysql-check"]
```
s-t-e-v-e-n-k: I'm concerned about including this when you also don't create haproxy@% user to monitor the cluster health.

chef/cookbooks/mysql/attributes/server.rb
jsuchome: Yeah, I've ended up using 3306 for both VIP and mysql. It's not technically a lie, is it?

 But we still need this value, or, more specifically, the structure - because `haproxy_servers_for_service` is using it.

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: What I've read in haproxy docs, if user is not provided with this option, it does it does not use one.

chef/cookbooks/mysql/attributes/server.rb
s-t-e-v-e-n-k: It bends the truth rather than breaking it. I have no objection to the setting existing, I just want the comment to tell the complete truth.

chef/cookbooks/mysql/recipes/ha_galera.rb
s-t-e-v-e-n-k: Then I'm even more concerned, since this will just end up haproxy attempting to check the cluster health and logging over and over that it can not do so?

crowbar_framework/app/models/database_service.rb
jsuchome: I'll take a look at it.

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: It's valid setup without a username specified, check http://www.haproxy.org/download/1.6/doc/configuration.txt . However it's true they describe it as "deprecated and not recommended"..

But they have it this way e.g. here https://docs.openstack.org/ha-guide/controller-ha-haproxy.html

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -142,3 +142,30 @@
...
+ha_servers = CrowbarPacemakerHelper.haproxy_servers_for_service(
+  node, "mysql", "database-server", "admin_port"
+)
+
+ha_servers = ha_servers.each do |n|
+  # Let all nodes but one act as backup (standby) servers.
+  # Backup server is only used when non-backup one is down. Thus we prevent possible deadlocks
+  # from OpenStack services writing to the database on different nodes at once.
+  n["backup"] = n["name"] != founder_name
+end
```
rhafer: How about just sorting the ha_server list and just add "backup" to all but the first one. This would avoid the search for the founder node.

jsuchome: Good point, we do not care if it really is cluster founder.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -142,3 +142,30 @@
...
+
+haproxy_loadbalancer "galera" do
+  address CrowbarDatabaseHelper.get_listen_address(node)
+  port 3306
+  mode "tcp"
+  options ["mysql-check"]
```
jsuchome: Hm, with "mysql-check", we'd need to raise the value of "max_connect_errors" or the clients could get blocked very soon with 
```
Host '192.168.243.81' is blocked because of many connection errors; unblock with 'mysqladmin flush-hosts'"
```

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: Done

PR #1124, opened 2017-08-03
Manila: get insecure values from databag

Itxaka:
the databag

### DIFF
--- a/chef/cookbooks/manila/recipes/common.rb
+++ b/chef/cookbooks/manila/recipes/common.rb
@@ -61,13 +61,10 @@ if neutron_servers.length > 0
     neutron_server,
     (neutron_server[:neutron][:ha][:server][:enabled] || false))
   neutron_server_port = neutron_server[:neutron][:api][:service_port]
-  neutron_insecure = neutron_protocol == "https" &&
-                     neutron_server[:neutron][:ssl][:insecure]
   neutron_service_user = neutron_server[:neutron][:service_user]
   neutron_service_password = neutron_server[:neutron][:service_password]
   Chef::Log.info("Neutron server at #{neutron_server_host}")
 else
-  neutron_insecure = false
   neutron_protocol = nil
   neutron_server_host = nil
   neutron_server_port = nil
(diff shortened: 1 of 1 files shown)


chef/cookbooks/manila/recipes/common.rb
```diff
@@ -79,39 +76,39 @@
   Chef::Log.warn("Neutron server not found")
 end
 
+neutron_config = Barclamp::Config.load("openstack", "neutron", node[:manila][:neutron_instance])
+neutron_insecure = neutron_config["insecure"] || false
+
```
nicolasbock: Too bad we can't get rid of the search in line 58 so that using the databag would speed things up. Is this change then for consistency instead of performance?

Itxaka: consistency. Plus I think the plan is to move more and more stuff to the databags, so potentially line 58 would disappear down the line and more and more stuff is gonna get saved/load to/from databags

chef/cookbooks/manila/recipes/common.rb
```diff
@@ -79,39 +76,39 @@
   Chef::Log.warn("Neutron server not found")
 end
 
+neutron_config = Barclamp::Config.load("openstack", "neutron", node[:manila][:neutron_instance])
+neutron_insecure = neutron_config["insecure"] || false
```
dirkmueller: isn't that supposed to be ```neutron_config[:ssl][:insecure]``` ? or how is that supposed to work? also why is this now outside the if condition?

chef/cookbooks/manila/recipes/common.rb
```diff
@@ -79,39 +76,39 @@
...
   cinder_admin_password = cinder[:cinder][:service_password]
 else
-  cinder_insecure = false
   cinder_admin_username = nil
   cinder_admin_password = nil
   Chef::Log.warn("cinder-controller not found")
 end
 
+cinder_config = Barclamp::Config.load("openstack", "cinder", node[:manila][:cinder_instance])
+cinder_insecure = cinder_config["insecure"] || false
```
dirkmueller: same here (missing [:ssl])

chef/cookbooks/manila/recipes/common.rb
```diff
@@ -79,39 +76,39 @@
...
   nova_admin_password = nova[:nova][:service_password]
 else
-  nova_insecure = false
   nova_admin_username = nil
   nova_admin_password = nil
   Chef::Log.warn("nova-controller not found")
 end
 
+nova_config = Barclamp::Config.load("openstack", "nova", node[:manila][:nova_instance])
+nova_insecure = nova_config["insecure"] || false
```
dirkmueller: same here?

chef/cookbooks/manila/recipes/common.rb
Itxaka: Not really, the databag contains the values directly, as we are only saving a small part of those configs, the databag is not a 1:1 copy of the attributes/templates.

This is what the databag contains:

```
root@crowbar:~ # knife data bag show crowbar-config openstack
default: 
  ceilometer: 
    insecure:  false
  cinder:     
    insecure:  false
  glance:     
    insecure:  false
  horizon:    
    insecure:  false
  keystone:   
    insecure:  false
  neutron:    
    insecure:  false
  nova:       
    insecure:  false
```

In this case, `Barclamp::Config.load("openstack", "neutron", node[:manila][:neutron_instance])` will return:

```
{insecure: false}
```

This values are set here: https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/models/openstack_service_object.rb#L41

And I believe the insecure value is the first one to be used due to the unnecessary searches that we do only to get that value.

For the seconds part, why is this outisde the if, this will work due to the `||` statement. We either have the databag, which indicates that neutron is deployed, or we dont, so the value defaults to false.

There is no need to set this inside the if/else anymore :)

Itxaka: I understand that this is confusing if you havent follow up the other databag patches, I expect to add this information to the weekly cloud email to have everyone aware of this new... feature?/fix?/approach?

Itxaka:
@dirkmueller is done in a migration in case this patch is applied to an existing installation which has the barclamp already deployed.

The way the databag is filled with the proper values is in the apply_role_post_chef_call method, which will only run after running the barclamp (i.e. apply method). In the case of develcloud8, this is ok, as this will get released all at once. But in the case of cloud7, if we want to provide this patches, we need to make sure that the databags are there, even if the proposal have been deployed already.

So the easiest way of doing this is through the migration. It will check if the proposal has been deployed, and if it has, it will fill up the databag with the *current* values of insecure, so the cookbooks can pick it up without requiring the user to go and apply every barclamp again.

Itxaka:
@dirkmueller btw, first time I heard about this `preapplychef` hook. Where is it? does it require to apply the barclamp if its already deployed to be executed?

Itxaka:
uhhh, how can the testbuild fail and mkcloud succeed? This was green a couple of hours ago ¬_¬

Itxaka:
reqrote it to be fixed according to the new approach at https://github.com/crowbar/crowbar-openstack/pull/1143/ so not to merge until that is merged

Itxaka:
Failure is expected as this uses classes not yet available in master

Itxaka:
[Use databag instead of node searches / manila](https://trello.com/c/CtfD18Ec/18-use-databag-instead-of-node-searches-manila)

Itxaka:
Need to rebuild the package so the changes are available to this PR

Itxaka:
@nicolasbock could I get a re-review? This is now based on https://github.com/crowbar/crowbar-openstack/pull/1143 which you also reviewed so you have the complete picture :D

PR #1125, opened 2017-08-03
keystone: fix ha race condition during default objects creation

stefannica:
During the deployment of the keystone barclamp on an
HA cluster, it was observed that the apache service was
being restarted on a controller node (due to a delayed
reload action) while on another controller node the
default users/tenants/domains/etc. were being created.
This caused the second node to fail deploying the barclamp,
because there was a small time window during which the
keystone WSGI entry points on one of the controller nodes
could not be accessed.

This commit fixes this problem by ensuring two things:
   - the default users/roles/domains/etc. are only created
   by the cluster founder node
   - an additional synchronization marker is used at the
     beginning of the section that creates the keystone
     objects to ensure that keystone services are properly
     configured and running on all controller nodes *before*
     the founder node creates them

NOTE: it would appear that a second synchronization marker
is also required *after* the keystone default objects are created,
to avoid running delayed actions (e.g. like restarting the apache
service) on any of the controller nodes while the founder
node creates them. However, this is not necessary, because
a wait/create synchronization section is already present and 
guarantees that all other nodes wait until the founder node
has completed creating these objects.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -536,6 +536,11 @@ if node[:keystone][:signing][:token_format] == "fernet"
   crowbar_pacemaker_sync_mark "create-keystone_fernet_rotate" if ha_enabled
 end
 
+# Wait for all nodes to reach this point so we know that all nodes will have
+# all the required services correctly configured and running before we create
+# the keystone resources
+crowbar_pacemaker_sync_mark "sync-keystone_before_register" if ha_enabled
+
 crowbar_pacemaker_sync_mark "wait-keystone_register" if ha_enabled
 
 keystone_insecure = node["keystone"]["api"]["protocol"] == "https" && node[:keystone][:ssl][:insecure]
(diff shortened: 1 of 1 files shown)


chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -630,6 +635,7 @@
     auth register_auth_hash
     tenant_name tenant
     action :add_tenant
+    only_if { ha_enabled && CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
aplanas: uhm are you sure that what you need is not `!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)`

stefannica:
@aplanas, @aspiers should I back-port this to SOC7 ? to SOC6 too ?

aplanas:
IMHO at least makes sense in SOC7

aspiers:
@stefannica Technically that's a decision for PM. I have no preference either way.

PR #1126, opened 2017-08-03
Nova: get insecure values from databag

Itxaka:
the databag

### DIFF
--- a/chef/cookbooks/nova/libraries/availability_zone.rb
+++ b/chef/cookbooks/nova/libraries/availability_zone.rb
@@ -18,7 +18,7 @@ module NovaAvailabilityZone
   def self.fetch_set_az_command_no_arg(node, cookbook_name)
     keystone_settings = KeystoneHelper.keystone_settings(node, cookbook_name)
 
-    nova_insecure = node[:nova][:ssl][:enabled] && node[:nova][:ssl][:insecure]
+    ssl_insecure = BarclampLibrary::Barclamp::Config.load("openstack", "nova")["insecure"] || false
 
     command = ["/usr/bin/crowbar-nova-set-availability-zone"]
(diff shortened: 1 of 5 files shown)


chef/cookbooks/nova/libraries/availability_zone.rb
```diff
@@ -18,7 +18,7 @@ module NovaAvailabilityZone
   def self.fetch_set_az_command_no_arg(node, cookbook_name)
     keystone_settings = KeystoneHelper.keystone_settings(node, cookbook_name)
 
-    nova_insecure = node[:nova][:ssl][:enabled] && node[:nova][:ssl][:insecure]
+    ssl_insecure = BarclampLibrary::Barclamp::Config.load("openstack", "nova")["insecure"] || false
```
nicolasbock: The insecure value is already in the node object, isn't it? Did you add the databag load here for consistency?

Itxaka: indeed. 
Plus when you pick from the databag is already does the whole `component[:ssl] || keystone_settings[:insecure]` which is nice :)

nicolasbock: True!

chef/cookbooks/nova/recipes/config.rb
```diff
@@ -131,10 +129,13 @@
     end
   end
 else
-  cinder_insecure = false
+  use_multipath = false
   keymgr_fixed_key = ""
 end
 
+cinder_config = Barclamp::Config.load("openstack", "cinder", node[:nova][:cinder_instance])
```
abel-navarro: still gives me shivers that we relay on cinder for the config although we agreed was the way to go.

PR #1131, opened 2017-08-03
Sahara: remove deprecated parameters

stephenbma:
The following sahara parameters are deprecated:

  - DEFAULT.verbose

In addition, new parameters to the neutron.conf template:
  - keystone_authtoken.service_token_roles_required
  - keystone_authtoken.service_token_roles
  - keystone_authtoken.memcached_server
  - keystone_authtoken.memcache_security_strategy
   - keystone_authtoken.memcache_secret_key

### DIFF
--- a/chef/cookbooks/sahara/recipes/common.rb
+++ b/chef/cookbooks/sahara/recipes/common.rb
@@ -51,6 +51,16 @@ nova_insecure = Barclamp::Config.load(
 
 use_ceilometer = !Barclamp::Config.load("openstack", "ceilometer").empty?
 
+memcached_servers = MemcachedHelper.get_memcached_servers(
+  if node[:sahara][:ha][:enabled]
+    CrowbarPacemakerHelper.cluster_nodes(node, "sahara-server")
+  else
+    [node]
+  end
+)
+
+memcached_instance("sahara") if node["roles"].include?("sahara-server")
+
 template node[:sahara][:config_file] do
   source "sahara.conf.erb"
   owner "root"
@@ -64,6 +74,7 @@ template node[:sahara][:config_file] do
     keystone_settings: KeystoneHelper.keystone_settings(node, :sahara),
     cinder_insecure: cinder_insecure,
     heat_insecure: heat_insecure,
+    memcached_servers: memcached_servers,
     neutron_insecure: neutron_insecure,
     nova_insecure: nova_insecure,
     use_ceilometer: use_ceilometer
--- a/chef/data_bags/crowbar/migrate/sahara/202_remove_deprecated_opts.rb
+++ b/chef/data_bags/crowbar/migrate/sahara/202_remove_deprecated_opts.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a.delete("verbose")
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["verbose"] = ta["verbose"]
+  return a, d
+end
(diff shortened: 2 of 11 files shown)


stephenbma:
Fixed the indentation issue in file 201_remove_deprecated_opts.rb.

crowbar_framework/app/views/barclamp/sahara/_edit_attributes.html.haml
```diff
@@ -14,5 +14,4 @@
     %fieldset
       %legend
         = t(".logging_header")
```
cmurphy: I think you can remove these two lines above and the logging_header key in the locale. the use_barbican_key_manager setting doesn't really have anything to do with logging AFAIK, so it was probably a mistake when it was added to put them in the same fieldset.

cmurphy:
I also see this in the sahara logs:

```
Using the in-process token cache is deprecated as of the 4.2.0 release and may be removed in the 5.0.0 release or the 'O' development cycle. The in-process cache causes inconsistent results and high memory usage. When the feature is removed the auth_token middleware will not cache tokens by default which may result in performance issues. It is recommended to use  memcache for the auth_token token cache by setting the memcached_servers option.
```

You can fix it like we did here: https://github.com/crowbar/crowbar-openstack/pull/1099

stephenbma:
The PR needs to be rebased.

chef/cookbooks/sahara/recipes/common.rb
```diff
@@ -51,6 +51,16 @@
...
 
+memcached_servers = MemcachedHelper.get_memcached_servers(
+  if node[:sahara][:ha][:enabled]
+    CrowbarPacemakerHelper.cluster_nodes(node, "sahara-api")
+  else
+    [node]
+  end
+)
+
+memcached_instance("sahara") if node["roles"].include?("sahara-api")
```
cmurphy: This role should be "sahara-server" not "sahara-api", same with line 56.

The way you can tell is from the role definition here: https://github.com/crowbar/crowbar-openstack/blob/9910a38b027459d7a088a50f1f91a223a4ebe866/chef/roles/sahara-server.rb#L1

Also on the admin node you can run `knife role list` and look for sahara roles.

chef/data_bags/crowbar/migrate/sahara/202_remove_deprecated_opts.rb
```diff
@@ -0,0 +1,14 @@
...
+    a["memcache_secret_key"] = service.random_password
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["verbose"] = ta["verbose"]
+  a.delete("memcache_secret_key")
+  return a, d
+end
```
cmurphy: I would rename this file, since it's not just removing opts it's adding them too. I think it would make sense for it to be two separate migrations.

crowbar_framework/config/locales/sahara/en.yml
```diff
@@ -27,7 +27,6 @@ en:
         glance_instance: 'Glance'
         cinder_instance: 'Cinder'
         logging_header: 'Logging'
```
cmurphy: You can remove this too

chef/data_bags/crowbar/migrate/sahara/202_remove_deprecated_opts.rb
stephenbma: Question: If I put the deprecation and the addition in two separate scripts.  Can the rev. number for each script file be identical?

cmurphy: No, they must be unique.

PR #1132, opened 2017-08-03
Barbican: keystone_authtoken service role and memcache usage

kantarun:
barbican: Add keystone_authtoken service role and memcache usage
    
Making change to address keystone middleware related deprecated warnings in barbican controller logs.
    
Use the service_token_roles_required and service_token_roles parameters in barbican service keystone auth config. Adding this will ensure that token validation to keystone is initiated by barbican service as service token needs to have configured service role. This will also take care of deprecation warning[1]
    
Add a memcached instance for the barbican service on controller and configure to use it for token caching. In-process token caching is deprecated and produces warnings[2]. As part of this change, need to add parameter to store randomly generated secret key which is used for cache encryption.
    
[2] https://docs.openstack.org/releasenotes/keystonemiddleware/mitaka.html


cmurphy:
The code looks fine and works for me, but please squash all the commits into one commit. If you like to use a branching/non-fast-forward-merging workflow when you're developing locally that's fine, but the merge commits in the PR make it hard to review and hard to deploy since it doesn't rebase cleanly against master, which I imagine might be why you were having trouble deploying your changes and also possibly why the testbuild job hasn't reported back.

See http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/ for a discussion on amending pull requests (written by our colleague @aspiers)

chef/data_bags/crowbar/template-barbican.json
```diff
@@ -33,14 +33,15 @@
...
+        "service_user": "barbican",
+        "memcache_secret_key": ""
      }
   },
   "deployment": {
     "barbican" : {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 101,
+      "schema-revision": 102,
```
cmurphy: Since this is for the new release series this should start at 200 (and the migration file should be renamed to 200_...

chef/data_bags/crowbar/migrate/barbican/200_add_memcache_secret_key.rb
```diff
@@ -0,0 +1,12 @@
+def upgrade(ta, td, a, d)
+  if a["memcache_secret_key"].nil? || a["memcache_secret_key"].empty?
```
dirkmueller: probably you want to use the symbol ```:memcache_secret_key``` instead - which is what you're using in barbican.conf.erb

chef/data_bags/crowbar/template-barbican.json
kantarun: Did not knew about this as did not come across this convention in docs and other reviews. Now, this change has been addressed in updated merged comments.

PR #1133, opened 2017-08-04
mysql: Use native SQL to create db_maker

rsalevsky:
This change uses the newly introduced database action's to create the
db_maker user and grant the required permissions. For doing so we have
to use the root user to connect to the database.


chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -128,23 +129,37 @@
...
-  mode "0600"
+database_user "create db_maker database user" do
+  connection db_settings[:connection]
+  username "db_maker"
+  password node[:database][:db_maker_password]
+  host "%"
+  provider db_settings[:user_provider]
   action :create
-  not_if { File.exist?("#{grants_key}") }
+end
```
jsuchome: I think we need 
`only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }` just like in other components

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -128,23 +129,37 @@
...
+    "CREATE USER",
+    "CREATE VIEW",
+    "SHOW VIEW",
+    "TRIGGER",
+    "EXECUTE",
+    "CREATE ROUTINE",
+    "ALTER ROUTINE"
+  ]
+  provider db_settings[:user_provider]
+  action :grant
```
jsuchome: Same here (`only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }`)

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -76,6 +76,14 @@ def self.database_settings(node, barclamp)
       else
         address = CrowbarDatabaseHelper.get_listen_address(database)
         backend_name = DatabaseLibrary::Database::Util.get_backend_name(database)
+        username = "db_maker"
+        password = database["database"][:db_maker_password]
+
+        if root && database["database"][:sql_engine] == "mysql"
+          address = "localhost"
```
rhafer: Why do you need "localhost" here? AFAIK we allow 'root'@'`nodes-admin-ip`' as well.

BTW, the resulting database_settings hash is cached throughout the chef-client run. And your code will cache it with the root@localhost setting. I don't think this works correctly.

rhafer: BTW, as we only the the root db connection in a single place (the mysql cookbook) it might just be easier to just update the connection information in the mysql cookbook instead for adding another parameter to `CrowbarOpenStackHelper::database_settings`

rsalevsky: Yes, this part needs some rework. But to answer the question why we need `localhost`, in my test env it always used the fqdn to connect and in this case `d52-54-77-77-01-01` doesn't match `d52-54-77-77-01-01.cloud.suse.de` and I'll get access denied.

jsuchome: Yes, the caching already caused the problem when asking for db_settings from keystone recipe

jsuchome:
[Fix granting the rights to db_maker user](https://trello.com/c/aoLOFF5W/9-fix-granting-the-rights-to-dbmaker-user)

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -128,23 +129,44 @@
...
+ha_enabled = node[:database][:ha][:enabled]
 
-grants_key = value_for_platform_family(
-  ["rhel", "suse", "fedora"] => "/etc/applied_grants",
-  "default" => "/etc/mysql/applied_grants"
-)
+db_settings = fetch_database_settings
+db_settings[:connection][:host] = "localhost"
+db_settings[:connection][:username] = "root"
+db_settings[:connection][:password] = node[:database][:mysql][:server_root_password]
```
rhafer: This is still suffering from the caching issue (you still overwrite the cached settings). To fix it you could `dup` the connnection hash. Something like:

```
db_settings[:connection][:host] = "localhost"
db_root_connection = db_settings[:connection].dup
db_root_connection[:username] = "root"
....
```
Should work. Of course you need to update the database_user calls to use the new hash ...

rsalevsky: Good catch, I did now `db_settings = fetch_database_settings.dup` which also solves the problem.

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -128,23 +129,44 @@
...
-  ["rhel", "suse", "fedora"] => "/etc/mysql_grants.sql",
-  "default" => "/etc/mysql/grants.sql"
-)
+ha_enabled = node[:database][:ha][:enabled]
 
-grants_key = value_for_platform_family(
-  ["rhel", "suse", "fedora"] => "/etc/applied_grants",
-  "default" => "/etc/mysql/applied_grants"
-)
+db_settings = fetch_database_settings.dup
```
rhafer: > Good catch, I did now db_settings = fetch_database_settings.dup which also solves the problem.

It doesn't. `dup` does a shallow copy not a deep copy. Since 'db_settings` is a nested hash you'd need a deep copy. (Or just shallow copy the internal `connection` "sub-hash)

> Beside this small change everything is tested in a non-HA deployment.

rsalevsky: Done

jsuchome:
And one more:

```
Mysql2::Error
-------------
Access denied; you need (at least one of) the RELOAD privilege(s) for this operation```

File: chef/cookbooks/mysql/recipes/server.rb [hunk 6]
```diff
@@ -128,23 +129,45 @@
...
+    "CREATE ROUTINE",
+    "CREATE TEMPORARY TABLES",
+    "CREATE USER",
+    "CREATE VIEW",
+    "EXECUTE",
+    "GRANT OPTION",
+    "LOCK TABLES",
+    "SHOW DATABASES",
+    "SHOW VIEW",
+    "TRIGGER"
```
rhafer: As @jsuchome  already mention we need at least "RELOAD" here as well.

With the previous code we gave db_maker the following privileges:

SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, RELOAD, INDEX, ALTER, SHOW DATABASES, CREATE TEMPORARY TABLES, LOCK TABLES, EXECUTE, CREATE USER, CREATE VIEW, SHOW VIEW, TRIGGER, EXECUTE, CREATE ROUTINE, ALTER ROUTINE, GRANT OPTION

And AFAIK the ```GRANT OPTION``` allows the user only to grant privileges that itself has. So I guess most of this is really needed. Otherwise  db_maker wouldn't be able to grant those privileges to anybody else. How about we first take that above list of privileges and try to size it down later (if that is even possible).

You might want to create a followup card for it.

rsalevsky: We have now, the default privileges ["SELECT", "INSERT", "UPDATE", "DELETE", "CREATE", "DROP", "INDEX", "ALTER"] + the current list. Which I hope is now the same.

rhafer: I created https://trello.com/b/EN16JVSt/cloud-8-databse-transition-squad-sprint-board for that now

jsuchome: Actually it's https://trello.com/c/EM0ClKxI/20-verify-that-the-openstack-database-users-have-the-correct-privileged-assigned-in-mariadb

jhesketh:
A little late, but decided to review anyway as learning the system. lgtm;

PR #1134, opened 2017-08-04
mysql: Add support for HA provisioning with Galera

rhafer:
Add a ha_galera recipe that will install Galera on a cluster. It will
create Pacemaker resources for bootstrapping and monitoring the cluster,
with other support sprinkled into the mysql::server recipe.

A current limitation of the configuration it creates is, that it relies
on the passwordless user "''@'localhost'" for monitoring the local
cluster node.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -1,0 +1,144 @@
+#
+# Cookbook Name:: mysql
+# Recipe:: ha_galera
+#
+# Copyright 2017, SUSE Linux GmbH.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+resource_agent = "ocf:heartbeat:galera"
+
+package "galera-3"
+
+unless node[:database][:galera_bootstrapped]
+  directory "/var/run/mysql/" do
+    owner "mysql"
+    group "root"
+    mode "0755"
+    action :create
+  end
+
+  execute "mysql_install_db" do
+    command "mysql_install_db"
+    action :run
+  end
+
+  crowbar_pacemaker_sync_mark "sync-database_after_install_db" do
+    revision node[:database]["crowbar-revision"]
+  end
+end
+
+# Wait for all nodes to reach this point so we know that all nodes will have
+# all the required packages installed before we create the pacemaker
+# resources
+crowbar_pacemaker_sync_mark "sync-database_before_ha" do
+  revision node[:database]["crowbar-revision"]
+end
+
+transaction_objects = []
+
+# Avoid races when creating pacemaker resources
+crowbar_pacemaker_sync_mark "wait-database_ha_resources" do
+  revision node[:database]["crowbar-revision"]
+end
+
+service_name = "galera"
+
+cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node, "database-server")
+nodes_names = cluster_nodes.map { |n| n[:hostname] }
+
+pacemaker_primitive service_name do
+  agent resource_agent
+  params({
+    "enable_creation" => true,
+    "wsrep_cluster_address" => "gcomm://" + nodes_names.join(","),
+    "check_user" => "''",
+    "socket" => "/var/run/mysql/mysql.sock"
+  })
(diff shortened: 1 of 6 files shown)


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,142 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+resource_agent = "ocf:heartbeat:galera"
+
+package "galera-3"
+
+unless node[:database][:galera_bootstrapped]
+  mysql_service_name = "mysql"
```
jsuchome: Does not seem to be used any more

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,142 @@
...
+
+cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node, "database-server")
+nodes_names = cluster_nodes.map { |n| n[:hostname] }
+
+pacemaker_primitive service_name do
+  agent resource_agent
+  params ({
+    "enable_creation" => true,
+    "wsrep_cluster_address" => "gcomm://" + nodes_names.join(","),
+    "check_user" => "''",
```
rhafer: As there is no good way to update a password before bootstrapping the cluster, we currently rely on the default ```''@'localhost'``` user, that doesn't have a password set.

In the long run I think I'd prefer to use some real user here. Which would require to update the pacemaker resource after the bootstrap completed and the user is created.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,142 @@
...
+  # note that this will also automatically start the resources
+  action :commit_new
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+crowbar_pacemaker_sync_mark "create-database_ha_resources" do
+  revision node[:database]["crowbar-revision"]
+end
+
+script 'wait galera bootstrap' do
```
jsuchome: Could you please add a comment why this block is needed?

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,142 @@
...
+    node.save
+  end
+  not_if {node[:database][:galera_bootstrapped]}
+end
+
+crowbar_pacemaker_sync_mark "sync-database_boostrapped" do
+  revision node[:database]["crowbar-revision"]
+end
+
+execute "assign-root-password-galera" do
```
jsuchome: Would it make sense to make sure this does not get executed on each run of chef-client?
(But I do not know how to nicely solve the case when the password is changed...)

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -122,10 +131,26 @@
...
+  include_recipe "mysql::ha_galera"
+else
+  log "HA support for mysql is disabled"
+end
+
+server_root_password = node[:mysql][:server_root_password]
+
+# set the root password on platforms
+# that don't support pre-seeding
+unless platform_family?("debian")
```
jsuchome: This was already removed, could you rebase?

rhafer: I rebased, but it seem I did something else wrong ... will remove it again.

chef/cookbooks/mysql/recipes/ha_galera.rb
rhafer: Actually it might be possible to remove the block from this recipe. And just rely on setting the password in the main recipe. (Still need to test that, though)

> But I do not know how to nicely solve the case when the password is changed

The current code doesn't handled that either, note the ```only_if``` clause. It only updates the password if the current one is not empty.

chef/cookbooks/mysql/recipes/ha_galera.rb
rhafer: Fixed

chef/cookbooks/mysql/recipes/ha_galera.rb
rhafer: done

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -0,0 +1,144 @@
...
+end
+
+crowbar_pacemaker_sync_mark "create-database_ha_resources" do
+  revision node[:database]["crowbar-revision"]
+end
+
+# The pacemaker galera resources agent will now bootstrap the cluster.
+# Which might take a while. Wait for that to complete by watching the
+# "wsrep_local_state_comment" status variable on all cluster nodes to reach
+# the "Synced" state, before continueing with the rest of the recipe.
```
jsuchome: Typo *continueing*

jsuchome:
Some of those hound complaints should be fixable...

jsuchome:
[Implement simple galera provisioning without HAProxy or VIP](https://trello.com/c/SU5hWOWX/8-implement-simple-galera-provisioning-without-haproxy-or-vip)

PR #1135, opened 2017-08-07
Ironic neutron decouple + HA fixes

skazi0:
This set of changes will enable deployment of ironic on non-neutron nodes. Also some old bugs with ironic deployment in HA clouds were fixed.

NOTE: these changes might possibly fix flapping status of CI jobs for ironic (looks like a race condition).
~~NOTE2: unfortunately the bug is still there after this change :o(~~
~~NOTE3: last commit is another attempt to fix this bug... looks promising~~

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -23,6 +23,8 @@ else
   nova_compute_ha_enabled = false
 end
 
+ironic_net = Barclamp::Inventory.get_network_definition(node, "ironic")
+
 # Disable rp_filter
 ruby_block "edit /etc/sysctl.conf for rp_filter" do
   block do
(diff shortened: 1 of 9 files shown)


skazi0:
Tested with following scenarios:
- HA + ironic and nova on separate nodes
- non-HA + ironic and nova on separate nodes
- non-HA + ironic and nova on controller

skazi0:
[Ironic Cloud Provo](https://trello.com/c/4lMUTWbA/394-ironic-cloud-provo)

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -23,6 +23,8 @@
   nova_compute_ha_enabled = false
 end
 
+ironic_net = node[:network][:networks]["ironic"]
```
rhafer: It might be better to just use node[:crowbar][:network]["ironic"] here. That is only present if the network is also enable for the node.

skazi0:
@rhafer I reorganized commits and fixed messages. For now I left the `ironic_net` to use network definition not checking if the network is actually enabled to the node as this was the original intention. I'm testing the updated version now.

skazi0:
rebased

PR #1136, opened 2017-08-07
migrations: fix databag migrations

Itxaka:
Looks like during the upgrade, the templates being passed
to the upgrade method contains all values in strings and we
were trying to access them by symbol instead.

Also fixes a nova migration numbering issue.


Itxaka:
Needs a package rebuild on trackupstream, its ongoing, will retrigger when done

PR #1139, opened 2017-08-08
testing: use knife to check syntax of cookbooks

Itxaka:
Use the installed knife command to check for cookbook syntax.
This will catch ruby issues on the cookbooks and also
template issues that we are not checking right now

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,7 @@ group :development do
   gem "sprockets-standalone", "~> 1.2.1"
   gem "sprockets", "~> 2.11.0"
   gem "rspec", "~> 3.1.0"
+  gem "chef", "~> 10.32.2"
 end
 
 unless ENV["PACKAGING"] && ENV["PACKAGING"] == "yes"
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -24,6 +24,7 @@ group :development do
   gem "sprockets-standalone", "~> 1.2.1"
   gem "sprockets", "~> 2.11.0"
   gem "rspec", "~> 3.1.0"
+  gem "chef", "~> 10.32.2"
```
nicolasbock: Shouldn't we add that to https://github.com/crowbar/crowbar-core/pull/1287 as well?

Itxaka: In crowbar core we take advantage that chef gem is already installed via crowbar_framework/Gemfile

In the other repos, its not included, Im guessing because they assume that crowbar-core is the base and will already have the chef gem installed, so no need to repeat it :)

nicolasbock: Ah, interesting. Thanks!

Itxaka:
sorry @nicolasbock I had to update the PR to include the older tests as well :)

Itxaka:
will require a rebase when https://github.com/crowbar/crowbar-openstack/pull/1180 is forwardported

Itxaka:
not moving forward

PR #1141, opened 2017-08-09
[4.0] ec2-api: allow a single ec2-api node or cluster

stefannica:
Fix the previous behaviour of allowing more than
one node or cluster to have the 'ec2-api' role
in the nova barclamp.

(backport of #1138)


stefannica:
The HA job failure is fixed by the recently merged PR #1142

PR #1144, opened 2017-08-09
[stable/4.0] mysql: Add support for HA provisioning with Galera

jsuchome:
This is a backport of https://github.com/crowbar/crowbar-openstack/pull/1134

Add a ha_galera recipe that will install Galera on a cluster. It will
create Pacemaker resources for bootstrapping and monitoring the cluster,
with other support sprinkled into the mysql::server recipe.

A current limitation of the configuration it creates is, that it relies
on the passwordless user "''@'localhost'" for monitoring the local
cluster node.

(cherry picked from commit 2cd79d39e755712d6b26ad58be5834f80c70eaed)

### DIFF
--- a/chef/cookbooks/database/libraries/crowbar.rb
+++ b/chef/cookbooks/database/libraries/crowbar.rb
@@ -10,7 +10,8 @@ module CrowbarDatabaseHelper
   end
 
   def self.get_listen_address(node)
-    if node[:database][:ha][:enabled]
+    # FIXME: temporarily do not expect there's any VIP for mysql database
+    if node[:database][:ha][:enabled] && node[:database][:sql_engine] != "mysql"
       vhostname = get_ha_vhostname(node)
       CrowbarPacemakerHelper.cluster_vip(node, "admin", vhostname)
     else
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -1,0 +1,144 @@
+#
+# Cookbook Name:: mysql
+# Recipe:: ha_galera
+#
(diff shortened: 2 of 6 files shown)


chef/cookbooks/database/libraries/crowbar.rb
```diff
@@ -10,7 +10,8 @@ def self.get_ha_vhostname(node)
   end
 
   def self.get_listen_address(node)
-    if node[:database][:ha][:enabled]
+    # FIXME: temporarily do not expect there's any VIP for mysql database
```
jsuchome: I'd rather merge it without this hack...

jsuchome:
[P1 (1) Implement galera provisioning without HAProxy or VIP](https://trello.com/c/SU5hWOWX/8-p1-1-implement-galera-provisioning-without-haproxy-or-vip)

chef/cookbooks/database/libraries/crowbar.rb
jsuchome: But I could live with it. It has to be removed for master as well and ported in next PR, so at least it would keep the commits the same...

rhafer:
It seems the galera resource agent on SP2 is missing at least: https://github.com/ClusterLabs/resource-agents/commit/a05eb8673bd1d5d3d41f2ed39df2650b19681d08

It fails to start the cluster with this: galera_start_0 on d52-54-77-77-01-03 'unknown error' (1): call=27, status=complete, exitreason='Unable to detect last known write sequence number'

rhafer:
> It seems the galera resource agent on SP2 is missing at least: ClusterLabs/resource-agents@a05eb86

filed: https://bugzilla.suse.com/show_bug.cgi?id=1053207 for that

PR #1145, opened 2017-08-10
mysql: Drop anonymous database user

rsalevsky:


jsuchome:

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -189,6 +189,15 @@
...
 
+database_user "drop anonymous database user" do
+  connection db_connection
+  username ""
+  host "*"
+  provider db_settings[:user_provider]
+  action :drop
+  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
```
nicolasbock: Would it make sense here to use `mysql_secure_installation` instead?

rsalevsky: No, `mysql_secure_installation` is only a bash script which does mysql query's which we can run easily in chef since we have the mysql2 adaptions.

nicolasbock: I see, thanks!

PR #1146, opened 2017-08-10
mysql: Debian cleanup

rsalevsky:
Not tested yet. Gating runs with mysql backend.


rsalevsky:
Tested the change and Gating failed during the Manila setup.

chef/cookbooks/mysql/recipes/client.rb
```diff
@@ -17,10 +17,10 @@
 # limitations under the License.
 #
 
-package "mysql-client" do
+package "mariadb-client" do
   package_name value_for_platform_family(
-    ["rhel", "suse", "fedora"] => "mysql",
-    "default" => "mysql-client"
+    ["rhel", "fedora"] => "mysql",
```
nicolasbock: Why are we removing 'suse' here?

rsalevsky: `suse` is now default ;)

nicolasbock: Well, then :smile:

PR #1147, opened 2017-08-10
nova: move leftover ec2-api attributes to ec2-api cookbook

stefannica:
This PR removes the old (and now unused) ec2-api service port
attributes from nova and move the current ec2-api service
port attributes from the nova to the ec2-api cookbook.
    
As a side-effect of this change, the tempest ec2-api
configuration, which was using the old and deprecated
port attributes, has been corrected.

NOTE: these changes are detached as a separate PR from the main bulk required to re-enable the ec2-api tempest test suite.


stefannica:
Abandoned in favor of #1151 (mostly because the main work in this PR of moving attributes from nova to ec2-api was incompatible with the haproxy_loadbalancer function logic and was removed).

PR #1148, opened 2017-08-10
ironic: Update for ocata

cmurphy:
This patch makes several updates for the ironic barclamp:

* Add service user credentials for the [glance], [neutron], [swift],
  [service_catalog], and [inspector] sections so that ironic can use 
  appropriate service users for each related service rather than getting
  credentials for all operations from [keystone_authtoken]. We still use 
  the ironic service user's credentials for the [inspector] and 
  [service_catalog] sections rather than creating new users for the 
  inspector service and API discovery. We must also include these auth
  options for the [swift] section even when swift is not being used in
  order to avoid a deprecation warning, but they of course do not have
  to work in that case. See the newton upgrade notes for more
  information about access credentials for services[1].
* Use the [keystone_authtoken]/service_token_roles and 
  [keystone_authtoken]/service_token_roles_required parameters that were
  introduced in Ocata and will soon become required[2] to ensure
  properly authenticated interservice communication.
* Add memcached settings for [keystone_authtoken] since the 
  keystonemiddleware in-process token cache has been deprecated since
  Newton[3].
* Add keystone v3 *_domain_name settings to [keystone_authtoken]
  section and update the auth_version to v3. 
* Explicitly set [deploy]/default_boot_option=netboot since the default
  is changing in Ocata[4].
* Change deprecated option [neutron]/cleaning_network_uuid to
  [neutron]/cleaning_network[4] and use the network name instead of uuid
  to enhance readability.
* Remove the "verbose" setting from the databag schema since it is
  deprecated in oslo.log[5] and it was never actually added to the 
  config template anyway.
* Stop forcing the glance barclamp to enable the v1 API when ironic is
  deployed since ironic can handle the glance v2 API now, and as of
  ocata defaults to using only v2[4].

[1] https://docs.openstack.org/releasenotes/ironic/newton.html#id13
[2] https://docs.openstack.org/releasenotes/keystonemiddleware/ocata.html
[3] https://docs.openstack.org/releasenotes/keystonemiddleware/newton.html
[4] https://docs.openstack.org/releasenotes/ironic/ocata.html
[5] https://docs.openstack.org/oslo.log/ocata/history.html#id33

### DIFF
--- a/chef/cookbooks/ironic/recipes/server.rb
+++ b/chef/cookbooks/ironic/recipes/server.rb
@@ -63,7 +63,7 @@ ironic_net_ip = Barclamp::Inventory.get_network_by_type(node, "ironic").address
 
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
-auth_version = "v2.0"
+auth_version = "v3"
 
 glance = node_search_with_cache("roles:glance-server").first
 raise "No glance-server found. Can not configure ironic." if glance.nil?
@@ -71,7 +71,9 @@ glance_settings = { protocol: glance[:glance][:api][:protocol],
                     host: CrowbarHelper.get_host_for_admin_url(
                       glance, glance[:glance][:ha][:enabled]
                     ),
-                    port: glance[:glance][:api][:bind_port] }
+                    port: glance[:glance][:api][:bind_port],
+                    service_user: glance[:glance][:service_user],
+                    service_password: glance[:glance][:service_password] }
 
 neutron = node_search_with_cache("roles:neutron-server").first
 raise "No neutron-server found. Can not configure ironic." if neutron.nil?
@@ -79,7 +81,9 @@ neutron_settings = { protocol: neutron[:neutron][:api][:protocol],
                      host: CrowbarHelper.get_host_for_admin_url(
                        neutron, neutron[:neutron][:ha][:server][:enabled]
                      ),
-                     port: neutron[:neutron][:api][:service_port] }
+                     port: neutron[:neutron][:api][:service_port],
+                     service_user: neutron[:neutron][:service_user],
+                     service_password: neutron[:neutron][:service_password] }
 
 api_port = node[:ironic][:api][:port]
 api_protocol = node[:ironic][:api][:protocol]
@@ -156,6 +160,12 @@ keystone_register "give ironic user admin role in service tenant" do
   action :add_access
 end
 
+ironic_net_name = "ironic"
+
+memcached_servers = MemcachedHelper.get_memcached_servers([node])
+
+memcached_instance("ironic")
+
 template node[:ironic][:config_file] do
   source "ironic.conf.erb"
   owner "root"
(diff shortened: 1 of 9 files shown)


chef/cookbooks/ironic/templates/default/ironic.conf.erb
```diff
@@ -26,11 +26,32 @@ glance_port=<%= @glance_settings[:port] %>
...
+auth_url=<%= @keystone_settings['internal_auth_url'] %>
+region_name=<%= @keystone_settings['endpoint_region'] %>
+username=<%= @swift_settings[:service_user] %>
+password=<%= @swift_settings[:service_password] %>
+project_name=<%= @keystone_settings['service_tenant'] %>
+project_domain_name=<%= @keystone_settings["admin_domain"]%>
+user_domain_name=<%= @keystone_settings["admin_domain"] %>
+<% else %>
+[swift]
+auth_type=none
```
cmurphy: This is wrong

skazi0:
@cmurphy forget my comment about waiting... the forward-port will not be clean cherry-picks anyway :(

cmurphy:
@skazi0 :(

I'm working on the other comments in any case.

cmurphy:
@skazi0 I incorporated your suggestions, thanks for those

chef/cookbooks/ironic/recipes/server.rb
```diff
@@ -169,8 +173,8 @@
 openstack_command << " --os-tenant-name #{keystone_settings["admin_tenant"]}"
 openstack_command << " --os-auth-url #{auth_url} #{insecure}"
 
-get_ironic_net_id = "#{openstack_command} network show ironic -f value -c id"
-ironic_net_id = Mixlib::ShellOut.new(get_ironic_net_id).run_command.stdout.chomp
+get_ironic_net = "#{openstack_command} network show ironic &>/dev/null"
```
skazi0: I think you can safely remove this part and set hardcoded 'ironic' network name. It is hardcoded in the part when the network is created.

cmurphy: done

cmurphy:
I can tell this is going to clash with #1163 at least on the net_id stuff so we should wait to merge this

chef/cookbooks/ironic/recipes/server.rb
```diff
@@ -169,8 +173,7 @@
 openstack_command << " --os-tenant-name #{keystone_settings["admin_tenant"]}"
 openstack_command << " --os-auth-url #{auth_url} #{insecure}"
 
-get_ironic_net_id = "#{openstack_command} network show ironic -f value -c id"
-ironic_net_id = Mixlib::ShellOut.new(get_ironic_net_id).run_command.stdout.chomp
+ironic_net_name = "ironic"
```
skazi0: Why not hardcode it directly in the template? It's not used for anything else but filling one param in the template.

skazi0:
#1163 merged, feel free to rebase and merge this.

chef/cookbooks/ironic/recipes/server.rb
cmurphy: My thinking was that in the event we eventually do want to change the name or make it configurable, it's easier to search for "ironic_net_name" than "ironic" and the plumbing will already be set up in the template.

cmurphy:
rebased

toabctl:
@skazi0 could you have another look please?

PR #1151, opened 2017-08-11
tempest: enable ec2-api testsuite

stefannica:
This PR updates the tempest cookbook to create
all the elements required to execute applicable
ec2-api test cases:
  - configure an image ID
  - create private network, subnet and router
  - create an EBS image by starting up a server
    and creating an image from it

The EBS image setup needs to be explicitly enabled by means of the (newly added) tempest.ec2_api.enable_ebs_image barclamp raw attribute.
    
The aws-cli package is required for this update, as well as the following upstream bugfix: https://review.openstack.org/#/c/493798/

EC2-API test statistics 
  - before:
```
Ran: 117 tests in 124.0000 sec.
 - Passed: 41
 - Skipped: 76
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 0
Sum of execute time for each test: 95.1280 sec.
```
  - after:
```
Ran: 177 tests in 1156.0000 sec.
 - Passed: 148
 - Skipped: 27
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 2
Sum of execute time for each test: 1912.7928 sec.
```
Reason for skipped test cases:
- 6 test cases skipped with 'vpnaas not available'
- 2 test cases skipped because they require an Ubuntu image
- 14 test cases skipped because they are test cases incompatible with Amazon (can be enabled by setting the aws.run_incompatible_tests configuration flag)
- 2 test cases skipped because the S3 image hasn't been configured (the effort is not justified for only 2 test cases)
- one test case (ec2api.tests.functional.api.test_instance_attributes.InstanceAttributeTest.test_instance_type_attribute) is disabled on purpose by not defining the alternative instance type, because it involves a VM migration which would only work in setups with 2 or more compute nodes

Reason for failed test cases:
- ec2api.tests.functional.api.test_images.ImageTest.test_check_image_operations_negative fails because it assumes the tempest test user does not have permission to modify the test image
- ec2api.tests.functional.scenario.test_instances.InstancesTest.test_metadata fails because neutron is not configured to use ec2-api-metadata instead of nova-api. Will be fixed by #1156.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -294,6 +294,11 @@ EOH
   })
 end
 
+ec2_apis = search(:node, "roles:ec2-api") || []
+use_ec2_api = ec2_apis.any?
+
+package "aws-cli" if use_ec2_api
+
 ruby_block "fetch ec2 credentials" do
   block do
     ec2_access = `#{openstackcli} ec2 credentials list -f value -c Access`.strip
@@ -302,6 +307,7 @@ ruby_block "fetch ec2 credentials" do
     node[:tempest][:ec2_access] = ec2_access
     node[:tempest][:ec2_secret] = ec2_secret
   end
+  only_if { use_ec2_api }
 end
 
 # FIXME: should avoid search with no environment in query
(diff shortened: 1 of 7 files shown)


chef/cookbooks/ec2-api/templates/default/ec2api.conf.erb
```diff
@@ -12,7 +12,7 @@ metadata_listen_port = <%= @bind_port_metadata %>
 s3_listen = <%= @bind_host %>
 s3_listen_port = <%= @bind_port_s3 %>
 transport_url = <%= @rabbit_settings[:url] %>
-full_vpc_support = false
+full_vpc_support = true
```
s-t-e-v-e-n-k: I'm concerned in changing this setting when the commit message doesn't mention it explicitly, especially since it enables it.

stefannica: This was only recently disabled by @dirkmueller in #803 based on the fact that it depends on Neutron VPNaaS (which we don't have). However, a lot of ec2-api VPC test cases are passing if I turn this on, while those that depend on VPNaaS are skipped, so it seems to me that this functionality isn't restricted to VPNaaS.

dirkmueller:
did you test this iwth a full tempest run (not only smoke)?

stefannica:
Turns out the CI failures are [well known](https://bugs.launchpad.net/ec2-api/+bug/1660888). In fact that ec2-api needs 'full_vpc_support = true' to enable neutron support, otherwise it will still assume it is talking to nova-network, which is deprecated.

stefannica:
@dirkmueller I did do a full tempest run for #1156, where I had to move the VPC enabling ec2-api configuration changes. You'll find the results there.

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -257,7 +257,7 @@
 
 glance #{insecure} image-list
 EOH
-  environment ({
+  environment ( lazy { {
```
cmurphy: The commit message doesn't mention why this is needed, was this not working before? I feel like "upload tempest test image" would affect a lot more than ec2-api if it was broken.

stefannica: When I started, I only needed a lazy dereference for the "create ec2 EBS image" bash block. Then I took a look at what @Itxaka din in #985 and I think all these other bash blocks need lazy reference as well, otherwise why would they work - they're not that different from the template resource, right ?
My initial plan was to extract them in a separate commit, maybe even a separate PR. Thanks for reminding me, @cmurphy. @Itxaka, could you please comment on this a bit ?

stefannica:
@cmurphy this needs #1156 to be merged and the aws-cli and python-s3transfer packages to be added to Devel:Cloud:8, otherwise the gating jobs will fail. I will keep it WIP until these prerequisites are met, but the code is otherwise ready.

chef/cookbooks/tempest/recipes/config.rb
Itxaka: Im running now, but I'll have a look at it tomorrow morning first thing.

Itxaka: Reading this, Im not sure if using `lazy` is needed tbh. The problem with the linked PR was that several resources were executed on the compilation phase, so any changes to the kesytone parameters (password, http<->https), will not be stored on the template parameters with their updated values, leading to the wrong vars being used. `lazy` is used there to delay the resolution to those parameters until they are written on the template.

For this PR, maybe the `OS_AUTH_URL` could be affected by the same, unfortunately I think the `environment` for a `bash` block do not work the same as the `variables` for a template (I think templates are all compiled during the compilation phase) so this may not be needed.

I doesn't hurt to have them, but you could probably test if this works without lazy by deploying, then changing the keystone barclamp from http to https and checking if the resource still uses the proper url for it. 

> The commit message doesn't mention why this is needed, was this not working before? I feel like "upload tempest test image" would affect a lot more than ec2-api if it was broken.

The problem may only arise when changing http to https or vice versa, and I think history has shown that we dont test for that anywhere, I mean we only merged the update_keystone_endpoint patch 5 months ago :D

stefannica:
rebased and removed the unneeded lazy dereference occurrences

stefannica:
pushed another commit that makes the EBS image setup configurable and disabled by default, in order to avoid making the tempest barclamp deployment the main detector for nova related issues.

chef/data_bags/crowbar/template-tempest.schema
```diff
@@ -55,6 +55,12 @@
               "mapping": {
                 "image_ref": { "type": "str", "required": true }
               }
+            },
+            "ec2_api": {
+              "type": "map",
+              "mapping": {
+                "enable_ebs_image": { "type": "bool", "required": true }
+              }
```
cmurphy: I think you're missing a migration for this

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -438,33 +434,50 @@
 echo $IMAGE_ID > #{ebs_image_id_file}
 echo "done."
 EOH
-  environment ( lazy { {
+  environment ({
```
cmurphy: Sorry to be nitpicky but since the lazy was introduced in the first commit of this PR can you fix it in that commit instead of this one?

chef/data_bags/crowbar/template-tempest.schema
stefannica: Hmm, I thought I added it. Silly no good non-telepathic git tools.

chef/cookbooks/tempest/recipes/config.rb
stefannica: I could do that, but then the previous commit would exhibit the same problem that made me add the lazy dereference in the first place - i.e. it would break the tempest barclamp. 
The reason I needed it in the first place was because the node[:tempest][:ec2_access] and node[:tempest][:ec2_secret] values were dereferenced too early and came up empty. The problem is still there with the last commit, which is why I use the lazy dereference for the "set up AWS CLI credentials" execute block I just introduced.

cmurphy: Oh I see, it was split into two blocks, got it.

PR #1154, opened 2017-08-14
Horizon: add SECRET_KEY as an attribute (bsc#1051439)

aplanas:
In a HA environment the secret key used in Horizon to store the
session data is not shared, producing form with different CFSR
tokens. This invalidate the session after a HA failover.

This patch also reload the apache service when the local_settings
file is updated, removing the race condition from bsc#1051439


aplanas:
[(HA) Unable to log in to openstack dashboard #1051439](https://trello.com/c/gwJoah7c/356-ha-unable-to-log-in-to-openstack-dashboard-1051439)

aplanas:
To test it manually:

* Setup an HA environment
* Deploy Horizon in a cluster
* Using the public IP for each controller, log in both nodes from the cluster (no CFSR error would happen)

PR #1155, opened 2017-08-15
Do not search for nodes with database-server role

jsuchome:
The role might not be saved yet. It should be safe to use the
default search (pacemaker-cluster-member) because cluster_members
helper will take care that result is in the same cluster as current
node (by checking pacemaker_config_environment).

This should fix [bsc#1053703](https://bugzilla.suse.com/show_bug.cgi?id=1053703)


jsuchome:
@s-t-e-v-e-n-k commit message adapted

PR #1156, opened 2017-08-15
neutron,ec2-api: use ec2-api-metadata in neutron-metadata-agent

stefannica:
When ec2-api is enabled, the ec2-api-metadata service
must be used instead of nova-metadata-agent when configuring
neutron.
The EC2 metadata service needs to act as a proxy between neutron
and the original nova metadata agent, adapting the metadata
information according to the EC2 API specifications.

This PR also re-enables VPC, for two main reasons:
  - without VPC enabled, ec2-api-metadata assumes it is running in a nova-network environment, which leads to failures (e.g. see [this upstream bug report](https://bugs.launchpad.net/ec2-api/+bug/1660888))
  - this was initially part of #1151, which aims to enable and pass as many applicable ec2-api tempest test cases as possible

Given the fact that VPC support was previously disabled precisely to correct some tempest test failures, a full tempest test run was executed on this PR to prove that no regression is incurred. The results show only 2 additional failures when compared to SOC7:

- new results (PR CI gating job):

```
      16:12:08 Ran: 2756 tests in 9265.0000 sec.
      16:12:08  - Passed: 2137
      16:12:08  - Skipped: 553
      16:12:08  - Expected Fail: 0
      16:12:08  - Unexpected Success: 0
      16:12:08  - Failed: 66
```
- old results (cloud 7 CI job - https://ci.suse.de/job/openstack-mkcloud/78718/console):

```
      06:52:37 Ran: 2340 tests in 8069.0000 sec.
      06:52:37  - Passed: 1729
      06:52:37  - Skipped: 547
      06:52:37  - Expected Fail: 0
      06:52:37  - Unexpected Success: 0
      06:52:37  - Failed: 64
```

The additional test failures are not related to the metadata service (they target the image and lbaas services).

### DIFF
--- a/chef/cookbooks/ec2-api/recipes/ec2api.rb
+++ b/chef/cookbooks/ec2-api/recipes/ec2api.rb
@@ -170,6 +170,19 @@ end
 
 crowbar_pacemaker_sync_mark "create-ec2_api_register" if ha_enabled
 
+nova_metadata_settings = {}
+nova = node_search_with_cache("roles:nova-controller").first
+unless nova.nil?
+  nova_ha_enabled = nova[:nova][:ha][:enabled]
+  nova_metadata_settings[:host] = CrowbarHelper.get_host_for_admin_url(nova, nova_ha_enabled)
+  nova_metadata_settings[:port] = nova[:nova][:ports][:metadata]
+  ssl_enabled = nova[:nova][:ssl][:enabled]
+  nova_metadata_settings[:protocol] = ssl_enabled ? "https" : "http"
+  ssl_insecure = nova[:nova][:ssl][:insecure]
+  nova_metadata_settings[:insecure] = ssl_enabled && ssl_insecure
+  nova_metadata_settings[:shared_secret] = nova[:nova][:neutron_metadata_proxy_shared_secret]
+end
+
 # ec2-api ssl
 if node[:nova]["ec2-api"][:ssl][:enabled]
   ssl_setup "setting up ssl for ec2-api" do
@@ -197,6 +210,7 @@ template node[:nova]["ec2-api"][:config_file] do
     bind_port_ec2api: bind_port_ec2api,
     bind_port_metadata: bind_port_metadata,
     bind_port_s3: bind_port_s3,
+    nova_metadata_settings: nova_metadata_settings
   )
 end
 
--- a/chef/cookbooks/ec2-api/templates/default/ec2api.conf.erb
+++ b/chef/cookbooks/ec2-api/templates/default/ec2api.conf.erb
@@ -24,6 +24,15 @@ ssl_ca_file = <%= node[:nova]["ec2-api"][:ssl][:ca_certs] %>
 s3_use_ssl = true
 <% end %>
 
+<% unless @nova_metadata_settings.empty? %>
+[metadata]
+nova_metadata_ip = <%= @nova_metadata_settings[:host] %>
+nova_metadata_port = <%= @nova_metadata_settings[:port] %>
+metadata_proxy_shared_secret = <%= @nova_metadata_settings[:shared_secret] %>
+nova_metadata_protocol = <%= @nova_metadata_settings[:protocol] %>
+nova_metadata_insecure = <%= @nova_metadata_settings[:insecure] %>
+<% end %>
+
 [database]
 connection = <%= @database_connection %>
 [oslo_concurrency]
--- a/chef/cookbooks/ec2-api/templates/default/ec2api.conf.erb
+++ b/chef/cookbooks/ec2-api/templates/default/ec2api.conf.erb
@@ -12,7 +12,7 @@ metadata_listen_port = <%= @bind_port_metadata %>
 s3_listen = <%= @bind_host %>
 s3_listen_port = <%= @bind_port_s3 %>
 transport_url = <%= @rabbit_settings[:url] %>
-full_vpc_support = false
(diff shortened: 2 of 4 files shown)


chef/cookbooks/ec2-api/recipes/ec2api.rb
```diff
@@ -180,6 +180,31 @@
 
 crowbar_pacemaker_sync_mark "create-ec2_api_register" if ha_enabled
 
+novas = Chef::Search::Query.new.search(:node, "roles:nova-controller")
```
cmurphy: It looks like there's a handy function for this that might be faster (I'm guessing, based on its name): https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/utils/libraries/search.rb#L52

Could that be used here?

chef/cookbooks/ec2-api/templates/default/ec2api.conf.erb
```diff
@@ -24,6 +24,13 @@ ssl_ca_file = <%= node[:nova]["ec2-api"][:ssl][:ca_certs] %>
 s3_use_ssl = true
 <% end %>
 
+[metadata]
+nova_metadata_ip = <%= @nova_metadata_host %>
+nova_metadata_port = <%= @nova_metadata_port %>
```
cmurphy: So what happens when these end up empty?

stefannica:
@cmurphy I know what you mean, there is very little info about how ec2 services should be configured. I had to revert to devstack scripts and even the code itself to figure this out. Had this misconfiguration not been revealed through one of the tempest test failures, we would have never known that something was wrong until the customers would start complaining.

chef/cookbooks/ec2-api/recipes/ec2api.rb
stefannica: Excellent suggestion, I used it and it worked, thanks.

chef/cookbooks/ec2-api/templates/default/ec2api.conf.erb
stefannica: The short answer is they will never end up empty. The ec2-api cookbook and nova cookbook are part of the same barclamp and the nova-controller role is mandatory.
Even so, I updated the code to address even this extreme (im)possibility.

stefannica:
Looks like this is breaking some of the non-ec2-api tempest test cases. I may need to fix those to accept the fact that ec2-api-metadata might be present.

chef/cookbooks/neutron/templates/default/metadata_agent.ini.erb
```diff
@@ -1,8 +1,11 @@
 [DEFAULT]
-nova_metadata_ip = <%= @nova_metadata_host %>
-metadata_proxy_shared_secret = <%= @metadata_proxy_shared_secret %>
-nova_metadata_protocol = <%= @nova_metadata_protocol %>
-nova_metadata_insecure = <%= @nova_metadata_insecure ? "True" : "False" %>
+<% unless @metadata_settings.empty? %>
+nova_metadata_ip = <%= @metadata_settings[:host] %>
+metadata_proxy_shared_secret = <%= @metadata_settings[:shared_secret] %>
+nova_metadata_protocol = <%= @metadata_settings[:protocol] %>
+nova_metadata_insecure = <%= @metadata_settings[:insecure] ? "True" : "False" %>
```
s-t-e-v-e-n-k: Couldn't we just use @metadata[:insecure] here anyway? It will always be a boolean based on how I'm reading the code.

stefannica:
@cmurphy, @s-t-e-v-e-n-k, I pushed another commit that re-enables VPC support, which fixes the remaining ec2-api metadata problems. Could you re-review please ?

chef/cookbooks/ec2-api/recipes/ec2api.rb
```diff
@@ -170,6 +170,19 @@
 
 crowbar_pacemaker_sync_mark "create-ec2_api_register" if ha_enabled
 
+metadata_settings = {}
+nova = node_search_with_cache("roles:nova-controller").first
+unless nova.nil?
+  ha_enabled = nova[:nova][:ha][:enabled]
```
cmurphy: You should rename this variable to distinguish it from the one meaning `node[:nova]["ec2-api"][:ha][:enabled]`

chef/cookbooks/ec2-api/recipes/ec2api.rb
```diff
@@ -170,6 +170,19 @@
...
+  metadata_settings[:host] = CrowbarHelper.get_host_for_admin_url(nova, ha_enabled)
+  metadata_settings[:port] = nova[:nova][:ports][:metadata]
+  ssl_enabled = nova[:nova][:ssl][:enabled]
+  metadata_settings[:protocol] = ssl_enabled ? "https" : "http"
+  ssl_insecure = nova[:nova][:ssl][:insecure]
+  metadata_settings[:insecure] = ssl_enabled && ssl_insecure
+  metadata_settings[:shared_secret] = nova[:nova][:neutron_metadata_proxy_shared_secret]
```
cmurphy: I would rename this `nova_metadata_settings`, and the same in the neutron cookbook, since it's otherwise not clear that this metadata is specific to nova and not the present cookbook (ec2-api or neutron)

stefannica:
Thanks @cmurphy, those were good suggestions. I integrated them into the code.

PR #1157, opened 2017-08-15
database: Move connection string creation to a helper

rsalevsky:
Not tested yet and needs to be done for all barclamps.

### DIFF
--- a/chef/cookbooks/cinder/recipes/common.rb
+++ b/chef/cookbooks/cinder/recipes/common.rb
@@ -56,18 +56,16 @@ include_recipe "database::client"
 include_recipe "#{db_settings[:backend_name]}::client"
 include_recipe "#{db_settings[:backend_name]}::python-client"
 
-db_password = ""
-if node.roles.include? "cinder-controller"
-  db_password = node[:cinder][:db][:password]
-else
+db_auth = node[:cinder][:db].dup
+unless node.roles.include? "cinder-controller"
   # pickup password to database from cinder-controller node
   node_controllers = node_search_with_cache("roles:cinder-controller")
   if node_controllers.length > 0
-    db_password = node_controllers[0][:cinder][:db][:password]
+    db_auth[:password] = node_controllers[0][:cinder][:db][:password]
   end
 end
 
-sql_connection = "#{db_settings[:url_scheme]}://#{node[:cinder][:db][:user]}:#{db_password}@#{db_settings[:address]}/#{node[:cinder][:db][:database]}"
+sql_connection = fetch_database_connection_string(db_auth)
 
 dirty = false
 
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -24,6 +24,11 @@ class Chef
       CrowbarOpenStackHelper.database_settings(node, barclamp)
     end
 
+    def fetch_database_connection_string(db_auth, barclamp = @cookbook_name)
+      db_settings = CrowbarOpenStackHelper.database_settings(node, barclamp)
+      CrowbarOpenStackHelper.database_connection_string(db_settings, db_auth)
+    end
+
     def fetch_rabbitmq_settings(barclamp=@cookbook_name)
       CrowbarOpenStackHelper.rabbitmq_settings(node, barclamp)
     end
@@ -39,6 +44,11 @@ class Chef
     class Template
       def fetch_database_settings(barclamp=@cookbook_name)
         CrowbarOpenStackHelper.database_settings(node, barclamp)
+      end
+
+      def fetch_database_connection_string(db_auth, barclamp = @cookbook_name)
+        db_settings = CrowbarOpenStackHelper.database_settings(node, barclamp)
+        CrowbarOpenStackHelper.database_connection_string(db_settings, db_auth)
       end
 
       def fetch_rabbitmq_settings(barclamp=@cookbook_name)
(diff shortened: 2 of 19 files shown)


chef/cookbooks/nova/recipes/config.rb
```diff
@@ -46,25 +46,23 @@
   include_recipe "#{db_settings[:backend_name]}::client"
   include_recipe "#{db_settings[:backend_name]}::python-client"
```
rsalevsky: I'm wondering if we can get rid of them.

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -98,6 +106,20 @@ def self.database_settings(node, barclamp)
     @database_settings[instance]
   end
 
+  def self.database_connection_string(user, password, database)
```
jsuchome: Do you think it is possible to just pass barclamp name and then fetch these data from the `node[barclamp]` here?

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -98,6 +106,20 @@ def self.database_settings(node, barclamp)
...
+    db_settings = fetch_database_settings
```
jsuchome: Hm, this requires arguments we do not have, right?

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
rsalevsky: Just passing the barclamp name is not enough since we can have multiple databases for one barclamp e.g. for nova we have nova, placement and api. But I think it is possible to pass for example `node[:nova][:db]` and do the rest in the helper.

stefannica:
Do you plan to update other barclamps to use the helper or is nova enough for this PR ?

jsuchome:
AFAIK the original bug we're trying to fix here is in cinder, right? Does it help, btw?

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -98,6 +106,20 @@ def self.database_settings(node, barclamp)
...
+
+    db_conn_scheme = db_settings[:url_scheme]
+    db_charset = ""
+
+    if db_conn_scheme == "mysql"
+      db_conn_scheme = "mysql+pymysql"
+      db_charset = "?charset=utf8"
+    end
+
+    "#{db_conn_scheme}://#{user}:#{password}@#{db_settings[:address]}/#{database}#{charset}"
```
jsuchome: `db_charset`, not `charset`

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
rsalevsky: Not sure, we will just use the default values and so far I saw no issues.

jsuchome: It fails for me as it does not know `fetch_database_settings`

Itxaka:
Is this related to a bug? Is there any more context for the error that is being fixed here? I want to review it properly (And it looks good) but not sure what it fixes or how :)

jsuchome:
I think this is the bug https://bugzilla.suse.com/show_bug.cgi?id=1053827

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -24,6 +24,11 @@ def fetch_database_settings(barclamp=@cookbook_name)
       CrowbarOpenStackHelper.database_settings(node, barclamp)
     end
 
+    def fetch_database_connection_string(db_auth, barclamp = @barclamp)
```
stefannica: Not sure about this, but if fetch_database_settings is using `@cookbook_name`, maybe this new method should use the same convention instead of `@barclamp`.

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -41,6 +46,11 @@ def fetch_database_settings(barclamp=@cookbook_name)
         CrowbarOpenStackHelper.database_settings(node, barclamp)
       end
 
+      def fetch_database_connection_string(db_auth, barclamp = @barclamp)
```
stefannica: Same question here: shouldn't default value be `@cookbook_name` instead of `@barclamp` ?

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -98,6 +108,21 @@ def self.database_settings(node, barclamp)
...
+    db_conn_scheme = db_settings[:url_scheme]
+    db_charset = ""
+
+    if db_conn_scheme == "mysql"
+      db_conn_scheme = "mysql+pymysql"
+      db_charset = "?charset=utf8"
+    end
+
+    "#{db_conn_scheme}://" \
+    "#{db_auth_hash[:user]}:#{db_auth_hash[:password]}@#{db_settings[:address]}/" \
```
stefannica: s/db_auth_hash/db_auth/

chef/cookbooks/cinder/recipes/common.rb
```diff
@@ -67,7 +67,7 @@
   end
 end
 
-sql_connection = "#{db_settings[:url_scheme]}://#{node[:cinder][:db][:user]}:#{db_password}@#{db_settings[:address]}/#{node[:cinder][:db][:database]}"
+sql_connection = fetch_database_connection_string(node[:cinder][:db])
```
stefannica: You missed the block above which decides which password to use (same thing as for ceilometer).

chef/cookbooks/ec2-api/recipes/ec2api.rb
```diff
@@ -40,19 +40,9 @@
...
-end
-
 crowbar_pacemaker_sync_mark "wait-ec2_api_database" if ha_enabled
 
-database_connection = "#{db_conn_scheme}://" \
-  "#{node[:nova]["ec2-api"][:db][:user]}" \
-  ":#{node[:nova]["ec2-api"][:db][:password]}" \
-  "@#{db_settings[:address]}" \
-  "/#{node[:nova]["ec2-api"][:db][:database]}"
+database_connection = fetch_database_connection_string(node[:nova]["ec2-api"][:db])
```
stefannica: The ec2-api cookbook is special because it doesn't have its own barclamp yet. This is also the exception to my previous statement about using `@cookbook_name` instead of `@barclamp` as default parameter values for the new helper methods. I.e. see the `db_settings = fetch_database_settings "nova"` line above.

jsuchome:
@rsalevsky could you please update the commit message? This is not only about moving the connection string creation, but (even more) about creating correct connection string value

PR #1160, opened 2017-08-16
Removed deprecated options from trove ocata

stephenbma:
Made the following changes to the Trove cookbooks for Ocata:

1. Removed the deprecated option verbose.
2. Added the [keystone_authtoken]/service_token_roles and [keystone_authtoken]/service token_roles_rqeuired options.  They are introduced in ocata and will be required in the future.  See [1].
3. Used memcached settings in the [keystone_authtoken] section due to the deprecation of the keystone middleware in-process token cache. [2] 

[2] https://docs.openstack.org/releasenotes/keystonemiddleware/newton.html


chef/cookbooks/trove/recipes/api.rb
```diff
@@ -143,6 +143,13 @@
 
 crowbar_pacemaker_sync_mark "create-trove_register" if ha_enabled
 
+memcached_servers = MemcachedHelper.get_memcached_servers(
+  if node[:trove][:ha][:enabled]
+    CrowbarPacemakerHelper.cluster_node(node, "trove-server")
+  else
+    [node]
+  end
+)
```
cmurphy: You should also declared a memcached_instance, similar to here: https://github.com/crowbar/crowbar-openstack/blob/eb28b5b964be0458cc0507896b1e5e262abe20d8/chef/cookbooks/nova/recipes/config.rb#L94

The reason is if the trove API server is deployed on its own machine we want it to use a local memcached instance.

chef/cookbooks/trove/attributes/default.rb
```diff
@@ -36,3 +35,5 @@
 default[:trove][:db][:password] = nil
 default[:trove][:db][:database] = "trove"
 default[:trove][:db][:user] = "trove"
+
+default[:trove][:ha][:enabled] = false
```
cmurphy: You don't need to add this here

chef/cookbooks/trove/recipes/api.rb
```diff
@@ -143,6 +143,13 @@
 
 crowbar_pacemaker_sync_mark "create-trove_register" if ha_enabled
 
+memcached_servers = MemcachedHelper.get_memcached_servers(
+  if node[:trove][:ha][:enabled]
+    CrowbarPacemakerHelper.cluster_node(node, "trove-server")
+  else
```
cmurphy: We don't yet support HA for trove so you can skip all this logic

chef/cookbooks/trove/recipes/api.rb
stephenbma: Will fix.

chef/cookbooks/trove/recipes/api.rb
stephenbma: Will fix

chef/cookbooks/trove/attributes/default.rb
stephenbma: Will fix

PR #1161, opened 2017-08-17
Use the same user for haproxy and RA monitoring.

jsuchome:


jsuchome:

PR #1168, opened 2017-08-17
Finetune haproxy settings for galera.

jsuchome:
1. lower the number of unsuccessful checks needed for declaring server DOWN (default is 5)
2. lower the interval checking after first failure is found (default is 2 seconds)

Requires https://github.com/crowbar/crowbar-ha/pull/220


jsuchome:
[P3 (3) Confiugure HAProxy for galera](https://trello.com/c/BQJLG0lL/7-p3-3-confiugure-haproxy-for-galera)

jhesketh:
Change looks good to me, but I have not tested it.

PR #1170, opened 2017-08-18
ceilometer: replaced auth_token in process cache with memcached in ocata cookbook

charleswang007:
The following Ceilometer configuration options deprecated in Ocata (or earlier) are removed or replaced in this PR:

- [keystone_authtoken]/service_token_roles_required  is set to True to enable fetching expired tokens based on valid service tokens and [keystone_authtoken]/service_token_roles is set to the admin role

The deprecated keystonemiddleware.auth_token in-process token cache used by ceilometer services is replaced with a memcached server. When HA is used, several memcached servers are configured and used, one for each ceilometer-server node. The keystone token data stored in memcached is authenticated and encrypted using a generated secret key.

References:

- Ocata release notes for Ceilometer: https://docs.openstack.org/releasenotes/ceilometer/ocata.html
- Mitaka release notes for keystonemiddleware: https://docs.openstack.org/releasenotes/keystonemiddleware/mitaka
- memcached protection options: https://docs.openstack.org/keystonemiddleware/latest/middlewarearchitecture.html#memcache-protection

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/common.rb
+++ b/chef/cookbooks/ceilometer/recipes/common.rb
@@ -1,5 +1,14 @@
 include_recipe "apache2"
 
+is_controller = node["roles"].include?("ceilometer-server")
+ha_enabled = node[:ceilometer][:ha][:server][:enabled]
+
+memcached_servers = MemcachedHelper.get_memcached_servers(
+  ha_enabled ? CrowbarPacemakerHelper.cluster_nodes(node, "ceilometer-server") : [node]
+)
+
+memcached_instance("ceilometer-server") if is_controller
+
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
 if node[:ceilometer][:use_mongodb]
@@ -68,6 +77,7 @@ template node[:ceilometer][:config_file] do
       debug: node[:ceilometer][:debug],
       rabbit_settings: fetch_rabbitmq_settings,
       keystone_settings: keystone_settings,
+      memcached_servers: memcached_servers,
       bind_host: bind_host,
       bind_port: bind_port,
       metering_secret: node[:ceilometer][:metering_secret],
--- a/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
+++ b/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
@@ -31,6 +31,11 @@ project_name = <%= @keystone_settings['service_tenant'] %>
 project_domain_name = <%= @keystone_settings["admin_domain"]%>
 user_domain_name = <%= @keystone_settings["admin_domain"] %>
 auth_type = password
+memcached_servers = <%= @memcached_servers.join(',') %>
+memcache_security_strategy = ENCRYPT
+memcache_secret_key = <%= node[:ceilometer][:memcache_secret_key] %>
+service_token_roles_required = true
+service_token_roles = admin
 
 [notification]
 workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
--- a/chef/data_bags/crowbar/template-ceilometer.json
+++ b/chef/data_bags/crowbar/template-ceilometer.json
@@ -15,6 +15,7 @@
       "keystone_instance": "none",
       "service_user": "ceilometer",
       "service_password": "",
+      "memcache_secret_key": "",
       "api": {
         "protocol": "http",
         "port": 8777
@@ -42,7 +43,7 @@
     "ceilometer": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 200,
+      "schema-revision": 201,
       "element_states": {
         "ceilometer-server": [ "readying", "ready", "applying" ],
         "ceilometer-central": [ "readying", "ready", "applying" ],
--- a/chef/cookbooks/ceilometer/metadata.rb
+++ b/chef/cookbooks/ceilometer/metadata.rb
@@ -7,6 +7,7 @@ version "0.1"
 
 depends "nagios"
 depends "keystone"
+depends "memcached"
 depends "database"
 depends "crowbar-openstack"
 depends "crowbar-pacemaker"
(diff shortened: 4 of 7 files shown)


chef/data_bags/crowbar/template-ceilometer.json
```diff
@@ -43,7 +44,7 @@
     "ceilometer": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 104,
+      "schema-revision": 201,
```
nicolasbock: Please start numbering migrations at 200.

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
```diff
@@ -32,6 +32,9 @@ project_name = <%= @keystone_settings['service_tenant'] %>
 project_domain_name = <%= @keystone_settings["admin_domain"]%>
 user_domain_name = <%= @keystone_settings["admin_domain"] %>
 auth_type = password
+memcached_servers = <%= @memcached_servers.join(',') %>
+memcache_security_strategy = ENCRYPT
+memcache_secret_key = <%= node[:aodh][:memcache_secret_key] %>
```
stefannica: I think I see some copy/paste trouble here: you're using the ceilometer cookbook, not aodh :)

chef/data_bags/crowbar/template-ceilometer.json
stefannica: Maybe @charleswang007's idea was to avoid merge conflicts with #1169 by hoping to merge them in sequence.

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
```diff
@@ -32,6 +32,9 @@ project_name = <%= @keystone_settings['service_tenant'] %>
...
```
stefannica: @charleswang007, what about the service_token_role_required and service_token_roles options for keystonemiddleware ? See here for more info: https://docs.openstack.org/releasenotes/keystonemiddleware/ocata.html

stefannica:
@charleswang007 it's up to you, but you should know that it's good practice to group several commits together in a single git branch and under a single pull request, especially if they are small and are related to each-other somehow or if they each tell a part of some bigger story, as is the case with this PR and #1169.

chef/cookbooks/ceilometer/recipes/server.rb
```diff
@@ -353,4 +360,24 @@
...
+    memcached_servers: memcached_servers,
+    bind_host: bind_host,
+    bind_port: bind_port,
+    database_connection: db_connection,
+    node_hostname: node["hostname"],
+    ceilometer_ssl: node[:ceilometer][:ssl],
+    evaluation_interval: node[:ceilometer][:evaluation_interval]
+  )
+  notifies :reload, resources(service: "apache2")
+end
```
cmurphy: This config file template is already defined in the common recipe. You can add the memcache stuff there, and condition the memcached_instance on whether the node is a controller, the way the nova recipe does:

https://github.com/crowbar/crowbar-openstack/blob/eb28b5b964be0458cc0507896b1e5e262abe20d8/chef/cookbooks/nova/recipes/config.rb#L22
https://github.com/crowbar/crowbar-openstack/blob/eb28b5b964be0458cc0507896b1e5e262abe20d8/chef/cookbooks/nova/recipes/config.rb#L94

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -1,5 +1,13 @@
 include_recipe "apache2"
 
+is_controller = node["roles"].include?("ceilometer-server")
+
+memcached_servers = MemcachedHelper.get_memcached_servers(
+  ha_enabled ? CrowbarPacemakerHelper.cluster_nodes(node, "ceilometer-server") : [node]
```
cmurphy: ha_enabled isn't defined yet in this recipe, you can look it up like this:

https://github.com/crowbar/crowbar-openstack/blob/a20fc751a0bc35e620255126ca77a4c133277924/chef/cookbooks/ceilometer/recipes/central.rb#L25

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -1,5 +1,14 @@
 include_recipe "apache2"
 
+is_controller = node["roles"].include?("ceilometer-server")
+ha_enabled = node[:ceilometer][:ha][:central][:enabled]
```
stefannica: The ceilometer-server and ceilometer-central roles can both be assigned to HA clusters, so they both have separate HA settings. The ceilometer-server HA setting is not node[:ceilometer][:ha][:central], but rather node[:ceilometer][:ha][:server].

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -1,5 +1,14 @@
...
+
+memcached_servers = MemcachedHelper.get_memcached_servers(
+  ha_enabled ? CrowbarPacemakerHelper.cluster_nodes(node, "ceilometer-server") : [node]
+)
+
+memcached_instance("ceilometer-server") if is_controller
```
stefannica: At first, I thought we might need to do this for ceilometer-central nodes/clusters too, but after reading more on the [ceilometer architecture](https://docs.openstack.org/ceilometer/ocata/architecture.html), I have to agree with @cmurphy that 
1. I don't understand the ceilometer architecture either and 
2. the celometer-central service does not provide a REST API, so it doesn't use memcached for keystonemiddleware

chef/cookbooks/ceilometer/recipes/common.rb
cmurphy: Good catch

chef/cookbooks/ceilometer/recipes/common.rb
cmurphy: I think this makes sense but I'll see if I can confirm this

PR #1153, opened 2017-08-14
MariaDB config rebase and cleanup

rsalevsky:
Thinks to do:

- [ ] Slow query logging which works
- [ ] Move tuning related options to separate log and check if they still make sense

PR #1171, opened 2017-08-18
Set State Snapshot Transfer Settings

jhesketh:
Use xtrabackup-v2 for transferring the state to new nodes joining the cluster.
This requires access to a root mysql user to do the transfer. Set up a new one
called "sstuser".

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -41,9 +41,95 @@ unless node[:database][:galera_bootstrapped]
   end
 end
 
+unless node[:database][:galera_bootstrapped]
+  if CrowbarPacemakerHelper.is_cluster_founder?(node)
+    # To bootstrap for the first time, start galera on one node
+    # to set up the seed sst user.
+
+    template "temporary bootstrap /etc/my.cnf.d/galera.cnf" do
+      path "/etc/my.cnf.d/galera.cnf"
+      source "galera.cnf.erb"
+      owner "root"
+      group "mysql"
+      mode "0640"
+      variables(
+        cluster_addresses: "gcomm://",
+        sstuser: "root",
+        sstuser_password: ""
+      )
+    end
+
+    case node[:platform_family]
+    when "rhel", "fedora"
+      mysql_service_name = "mysqld"
+    else
+      mysql_service_name = "mysql"
+    end
+
+    # use the initial root:'' credentials to set up the new user. The
+    # unauthenticated root user is later removed in server.rb after the
+    # bootstraping. Once the cluster has started other nodes will pick up on
+    # the sstuser and we are able to use these credentails.
+    db_settings = fetch_database_settings
+    db_connection = db_settings[:connection].dup
+    db_connection[:host] = "localhost"
+    db_connection[:username] = "root"
+    db_connection[:password] = ""
+
+    service "mysql-temp start" do
+      service_name mysql_service_name
+      supports status: true, restart: true, reload: true
+      action :start
+    end
+
+    database_user "create state snapshot transfer user" do
+      connection db_connection
+      username "sstuser"
+      password node[:database][:mysql][:sstuser_password]
+      host "localhost"
+      provider db_settings[:user_provider]
+      action :create
+    end
+
+    database_user "grant sstuser root privileges" do
+      connection db_connection
+      username "sstuser"
+      password node[:database][:mysql][:sstuser_password]
+      host "localhost"
+      provider db_settings[:user_provider]
+      action :grant
+    end
+
+    service "mysql-temp stop" do
+      service_name mysql_service_name
+      supports status: true, restart: true, reload: true
+      action :stop
+    end
+  end
+end
+
+service_name = "galera"
+
+cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node)
+nodes_names = cluster_nodes.map { |n| n[:hostname] }
+
+cluster_addresses = "gcomm://" + nodes_names.join(",")
+
+template "/etc/my.cnf.d/galera.cnf" do
+  source "galera.cnf.erb"
+  owner "root"
+  group "mysql"
+  mode "0640"
+  variables(
+    cluster_addresses: cluster_addresses,
+    sstuser: "sstuser",
+    sstuser_password: node[:database][:mysql][:sstuser_password]
+  )
+end
+
 # Wait for all nodes to reach this point so we know that all nodes will have
-# all the required packages installed before we create the pacemaker
-# resources
+# all the required packages and configurations installed before we create the
+# pacemaker resources
 crowbar_pacemaker_sync_mark "sync-database_before_ha" do
   revision node[:database]["crowbar-revision"]
 end
@@ -55,16 +141,11 @@ crowbar_pacemaker_sync_mark "wait-database_ha_resources" do
   revision node[:database]["crowbar-revision"]
 end
 
-service_name = "galera"
-
-cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node)
-nodes_names = cluster_nodes.map { |n| n[:hostname] }
-
 pacemaker_primitive service_name do
   agent resource_agent
   params({
     "enable_creation" => true,
-    "wsrep_cluster_address" => "gcomm://" + nodes_names.join(","),
+    "wsrep_cluster_address" => cluster_addresses,
     "check_user" => "monitoring",
     "socket" => "/var/run/mysql/mysql.sock",
     "datadir" => node[:database][:mysql][:datadir]
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -19,7 +19,7 @@
 
 resource_agent = "ocf:heartbeat:galera"
 
-["galera-3-wsrep-provider", "mariadb-tools", "xtrabackup"].each do |p|
+["galera-3-wsrep-provider", "mariadb-tools", "xtrabackup", "socat"].each do |p|
   package p
 end
 
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -35,10 +35,6 @@ unless node[:database][:galera_bootstrapped]
     command "mysql_install_db"
     action :run
   end
-
-  crowbar_pacemaker_sync_mark "sync-database_after_install_db" do
-    revision node[:database]["crowbar-revision"]
-  end
 end
 
 unless node[:database][:galera_bootstrapped]
--- a/chef/cookbooks/mysql/recipes/server.rb
+++ b/chef/cookbooks/mysql/recipes/server.rb
@@ -66,10 +66,6 @@ service mysql start
 EOC
 end
 
-cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node)
-nodes_names = cluster_nodes.map { |n| n[:hostname] }
-cluster_addresses = "gcomm://" + nodes_names.join(",")
-
 template "/etc/my.cnf.d/openstack.cnf" do
   source "my.cnf.erb"
   owner "root"
@@ -101,19 +97,6 @@ template "/etc/my.cnf.d/tuning.cnf" do
     max_heap_table_size: node[:database][:mysql][:max_heap_table_size]
   )
   notifies :restart, "service[mysql]", :immediately
-end
-
-if node[:database][:ha][:enabled]
-  template "/etc/my.cnf.d/galera.cnf" do
-    source "galera.cnf.erb"
-    owner "root"
-    group "mysql"
-    mode "0640"
-    variables(
-      cluster_addresses: cluster_addresses
-    )
-    notifies :restart, "service[mysql]", :immediately
-  end
 end
 
 unless Chef::Config[:solo]
--- a/chef/cookbooks/mysql/recipes/server.rb
+++ b/chef/cookbooks/mysql/recipes/server.rb
@@ -44,7 +44,19 @@ end
 
 service "mysql" do
   service_name mysql_service_name
-  supports status: true, restart: true, reload: true
+  if ha_enabled
+    supports status: true,
+             restart: true,
+             reload: true,
+             restart_crm_resource: true,
+             pacemaker_resource: "galera",
+             crm_resource_stop_cmd: "force-demote",
+             crm_resource_start_cmd: "force-promote"
+  else
+    supports status: true,
+             restart: true,
+             reload: true
+  end
   action :enable
   provider Chef::Provider::CrowbarPacemakerService if ha_enabled
 end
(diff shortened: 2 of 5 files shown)


jhesketh:
Note to self: need to set the permissions correctly for sstuser.

jsuchome:
[P4 (5) Sanitize and enhance mariadb configuration done by cookbook](https://trello.com/c/PcPTyEX7/1-p4-5-sanitize-and-enhance-mariadb-configuration-done-by-cookbook)

rhafer:
@jhesketh Am I right that this is based off of https://github.com/crowbar/crowbar-openstack/pull/1153  and you're interested in reviews for the last commit?

jsuchome:
Please rebase, @rsalevsky's PR has been merged.

crowbar_framework/app/models/database_service.rb
```diff
@@ -174,6 +174,7 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
 
     if ( sql_engine == "mysql" )
       role.default_attributes["database"]["mysql"]["server_root_password"] = (old_role && old_role.default_attributes["database"]["mysql"]["server_root_password"]) || random_password
+      role.default_attributes["database"]["mysql"]["sstuser_password"] = (old_role && old_role.default_attributes["database"]["mysql"]["sstuser_password"]) || random_password
```
jhesketh: This is shorter than neighbouring lines.. Do we not care about hound results? If so, can we tune it not to warn on this type of error?

jsuchome: I think we ignore the warnings in these kind of situations - when long lines are used in the area anyway and/or if it does not get better with shortening. So I would treat it this way, probably.

But in general we're trying to follow hounds recommendations, even in this case.

(This is not a rule written anywhere, just my POV)

jsuchome:
@rhafer that's my concern as well...

jsuchome:
[galera config: Find the best value for SST](https://trello.com/c/27Q5XW7K/28-galera-config-find-the-best-value-for-sst)

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -167,6 +179,16 @@
...
 
+  database_user "grant sstuser root privileges" do
+    connection db_connection
+    username "sstuser"
+    password node[:database][:mysql][:sstuser_password]
+    host "%"
+    provider db_settings[:user_provider]
+    action :grant
+    only_if { ha_enabled && CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
```
rsalevsky: Where are the privileges?

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -85,7 +85,9 @@
     group "mysql"
     mode "0640"
     variables(
-      cluster_addresses: cluster_addresses
+      cluster_addresses: cluster_addresses,
+      sstuser: "sstuser",
```
rsalevsky: I would just call it sst.

jhesketh:
This works for me bringing up a new cluster correctly. WIP removed as should be ready for review.

However, I'm currently debugging adding a new node to the cluster which appears to be failing in my setup, but my current hunch is that it is unrelated.

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -183,6 +188,42 @@
...
+      source "galera.cnf.erb"
+      owner "root"
+      group "mysql"
+      mode "0640"
+      variables(
+        cluster_addresses: cluster_addresses,
+        sstuser: "sstuser",
+        sstuser_password: node[:database][:mysql][:sstuser_password]
+      )
+      notifies :restart, "service[mysql]", :immediately
```
jsuchome: Don't we need `only_if { node[:database][:galera_bootstrapped] }` here?

jhesketh: No, we want this to apply to all the nodes so that any of them can act as a donor for future nodes.

jsuchome: But `galera_bootstrapped` is set for all nodes. I assumed you wanted to change my.cnf after you are sure bootstrapping is done

jhesketh: Right, but this section is under " unless node[:database][:database_bootstrapped]"

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -79,15 +79,20 @@
...
-  template "/etc/my.cnf.d/galera.cnf" do
-    source "galera.cnf.erb"
-    owner "root"
-    group "mysql"
-    mode "0640"
-    variables(
-      cluster_addresses: cluster_addresses
-    )
-    notifies :restart, "service[mysql]", :immediately
+  unless node[:database][:galera_bootstrapped]
```
jsuchome: Does this mean you are not saving any my.cnf template during bootstrapping? It's probably fine if we do not need it, but it might be enough just to adapt the variables based on the value of `galera_bootstrapped`

jsuchome: Sorry, that was wrong

chef/cookbooks/mysql/recipes/server.rb
jhesketh: good point, I can fix that

jsuchome:
Seems to work correctly

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -79,15 +79,20 @@
...
+      source "galera.cnf.erb"
+      owner "root"
+      group "mysql"
+      mode "0640"
+      variables(
+        cluster_addresses: cluster_addresses,
+        sstuser: "root",
+        sstuser_password: ""
+      )
+      notifies :restart, "service[mysql]", :immediately
```
rhafer: Hm, does this really work? (I mean will it restart correctly?) 

Remember, we're not using the "lsb:" or "systemd:" resource agents in pacemaker for mariadb (but the "galera" MS resource). I somehow thing that calling "systemd restart mysql" will start mariadb with a different set of command line options than pacemaker did.

But I still need to test that.

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -79,15 +79,20 @@
...
-    owner "root"
-    group "mysql"
-    mode "0640"
-    variables(
-      cluster_addresses: cluster_addresses
-    )
-    notifies :restart, "service[mysql]", :immediately
+  unless node[:database][:galera_bootstrapped]
+    # For bootstrapping sst, use root with no password
+    template "/etc/my.cnf.d/galera.cnf" do
```
rhafer: As this is pretty specific to HA setups, I think this belongs to the ha_galera recipe.

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -183,6 +188,49 @@
     action :drop
     only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
+
+  if node[:database][:ha][:enabled]
```
rhafer: Same here. I think this belongs to the ha_galera recipe.

chef/cookbooks/mysql/recipes/server.rb
jhesketh: This was already here, but I'm happy to move it.

chef/cookbooks/mysql/recipes/server.rb
jhesketh: So this does work in my testing, however I couldn't tell you why. We should investigate if this is correct. However, this is what was already here (this change doesn't affect it).

chef/cookbooks/mysql/recipes/server.rb
jhesketh: So there is some logic grouping here with the above users being created, but I take your point and happy to move it.

chef/cookbooks/mysql/recipes/server.rb
rhafer: Hm, true this ended here as part of https://github.com/crowbar/crowbar-openstack/pull/1153. Didn't notice that, sorry.

Still I think it should move to the HA recipe. (Maybe as part of another commit)

chef/cookbooks/mysql/recipes/server.rb
rhafer: Hm, true the ha_proxy user is also created here. I guess we should keep it here. Sorry for the noise.

chef/cookbooks/mysql/recipes/server.rb
rhafer: Actually it just "works" by accident. (or because "systemctl status mysql" is smart enough to notice that it didn't start mysqld itself):

```
[2017-08-23T14:50:08+00:00] INFO: template[/etc/my.cnf.d/galera.cnf] updated content
[2017-08-23T14:50:08+00:00] INFO: template[/etc/my.cnf.d/galera.cnf] owner changed to 0
[2017-08-23T14:50:08+00:00] INFO: template[/etc/my.cnf.d/galera.cnf] group changed to 485
[2017-08-23T14:50:08+00:00] INFO: template[/etc/my.cnf.d/galera.cnf] mode changed to 640
[2017-08-23T14:50:08+00:00] INFO: template[/etc/my.cnf.d/galera.cnf] sending restart action to service[mysql] (immediate)
[2017-08-23T14:50:08+00:00] INFO: Processing service[mysql] action restart (mysql::server line 45)
[2017-08-23T14:50:08+00:00] INFO: Ignoring restart action for mysql service since not running on this node (d52-54-77-77-01-01)
´´´

But this also means that mysql is not restarted after the 2nd galera.cnf update further down in the recipe. So the updated sst config is not really active. (I guess this becomes is an issue if one node is stopped for long enough to require a full sst on restart, right?)

Note: restarting a pacemaker managed server on a specific node is tricky (if possible at all) when not using the lsb or systemd resource agents. I don't yet have a good idea how to solve this special case. I see two possible solutions:

1. explicitly call "crm resource restart galera" on the founder node after all (!) nodes updated the galera.cnf file, this is somewhat ugly as it requires yet another syncmark + a wait loop to make sure the cluster is fully up again. But this might be acceptable for the first bootstrap. (Though it doesn't solve the restart issue after later configuration changes)

2. rework the way we bootstrap (again) to first run on a single node, create the sst user there and only after the bootstrap the full cluster.

Comment (jhesketh) [inline chef/cookbooks/mysql/recipes/server.rb]: Because the configuration is for the node to be a donor we do need to somehow reload it incase another node needs to get the full state or is coming online. (If the configuration was for itself when it was coming online it wouldn't matter, but this is not the case).

I'll take a look at the options you put forward and give it some more thought tomorrow.

File: chef/cookbooks/mysql/recipes/server.rb (hunk 7 above)
Comment (jhesketh) [inline chef/cookbooks/mysql/recipes/server.rb]: So I went to move it, but I don't think it make sense at the moment. Primarily because we have to wait for the users to be set up before switching the configuration back, so the newer configuration (below) can't easily be placed in ha_galera.rb and therefore it makes more sense to keep them in a similar location (in my opinion).

(This may all change once we solve the restart issue anyway)

**Comment by jhesketh** [comment]
This should be ready for review in the sense that it configures SST correctly. It does, however require Ralf's fix to the galera-3 dependencies: https://build.opensuse.org/request/show/518768.

Also my testing of adding a new node to the cluster fails. I don't think this is due to the SST configuration but rather how it is being added into pacemaker (I may be wrong)

File: chef/cookbooks/mysql/recipes/ha_galera.rb [hunk 9]
```diff
@@ -46,10 +42,70 @@
...
+    end
+
+    case node[:platform_family]
+    when "rhel", "fedora"
+      mysql_service_name = "mysqld"
+    else
+      mysql_service_name = "mysql"
+    end
+
+    db_settings = fetch_database_settings
```
jsuchome: I think this part would deserve a comment explaining why are we using this (non)-credentials...

jhesketh: Ack, done.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -35,15 +35,96 @@
...
+end
 
-  crowbar_pacemaker_sync_mark "sync-database_after_install_db" do
-    revision node[:database]["crowbar-revision"]
+unless node[:database][:galera_bootstrapped]
+  if CrowbarPacemakerHelper.is_cluster_founder?(node)
+    # To bootstrap for the first time, start galera on one node
+    # to set up the seed sst user.
+
+    template "/etc/my.cnf.d/galera.cnf" do
```
rhafer: Could you please give this resource a different name than the main one on line 113. Something like:

```
template "temporary bootstrap galera.cnf" do
  path "/etc/my.cnf.d/galera.cnf"
  source "galera.cnf.erb"
  ...
end
```

That should avoid those `WARN: Cloning resource attributes for template[/etc/my.cnf.d/galera.cnf] from prior resource (CHEF-3694)` warning during the chef-client run on the founder node.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -35,15 +35,96 @@
...
+    # unauthenticated root user is later removed in server.rb after the
+    # bootstraping. Once the cluster has started other nodes will pick up on
+    # the sstuser and we are able to use these credentails.
+    db_settings = fetch_database_settings
+    db_connection = db_settings[:connection].dup
+    db_connection[:host] = "localhost"
+    db_connection[:username] = "root"
+    db_connection[:password] = ""
+
+    service "mysql-temp" do
```
rhafer: Same here. Please give the resource for the temporary start a different name than the one that stop mysql again. chef doesn't really like multiple resources with the same name. We should avoid that if possible.

jsuchome:
I agree with some squashing, but do we really need single commit? IMHO the restart actions and package updates could live as separate commits

jsuchome:
In general, I wanted to also add that it seems good to merge (with implementing latest Ralf's ideas)

jhesketh:
I agree with the feedback. I have implemented the changes and squashed the commits into what seem sensible divisions of code. Happy to squash more or less.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -41,9 +41,94 @@
   end
 end
 
+unless node[:database][:galera_bootstrapped]
+  if CrowbarPacemakerHelper.is_cluster_founder?(node)
+    # To bootstrap for the first time, start galera on one node
+    # to set up the seed sst user.
+
+    template "temporary bootstrap /etc/my.cnf.d/galera.cnf" do
+      source "galera.cnf.erb"
```
rhafer: Does this really work? I think you need to add a `path` setting here so that the resource knows where to put the processed template.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -35,15 +35,96 @@
...
+end
 
-  crowbar_pacemaker_sync_mark "sync-database_after_install_db" do
-    revision node[:database]["crowbar-revision"]
+unless node[:database][:galera_bootstrapped]
+  if CrowbarPacemakerHelper.is_cluster_founder?(node)
+    # To bootstrap for the first time, start galera on one node
+    # to set up the seed sst user.
+
+    template "temporary bootstrap /etc/my.cnf.d/galera.cnf" do
```
jsuchome: Oh. I think you need to specify `path` property if the template name cannot serve as a path to file.

jhesketh:
Thanks for the fix, I've squashed it into the appropriate commit.

PR #1172, opened 2017-08-18
neutron: enable dns extension

dirkmueller:
This is a prerequisite for DNS as a Service integration,
but also allows assigning a dns_name attribute to ports,
see

https://docs.openstack.org/newton/networking-guide/config-dns-int.html


nicolasbock:
Will we make the DNS domain configurable in a follow on PR?

dirkmueller:
@nicolasbock dns_domain is configurable via the expert settings in the proposal already.

nicolasbock:
Thanks @dirkmueller.

PR #1173, opened 2017-08-18
sahara: fix sahara sync mark name

Itxaka:
Name has spaces on it, while the rest of the other sync marks dont have
any spaces. This also causes problems with the existing task of trying
to move the sync marks into the pacemaker attributes, as the attributes
do not accept spaces, and we are using the name in the sync mark to set
it as the attribute.

Blocks: https://github.com/crowbar/crowbar-ha/pull/217


Itxaka:

PR #1176, opened 2017-08-18
[4.0] mariadb: Drop anonymous database user

rsalevsky:
The mysql_secure_install script drops all anonymous database users to
prevent unauthorized logins. Let's do the same in our cookbooks.

(cherry picked from commit 7c4e10f2430021cbe708001f3b906e20c8e74e74)

Backport from https://github.com/crowbar/crowbar-openstack/pull/1145


jsuchome:

PR #1179, opened 2017-08-21
Forwardport neutron-ha-tool service

Itxaka:
Turn neutron-ha-tool into a service

As an operator I want to see that neutron-ha-tool is a service monitored by pacemaker.
Currently NHT (Neutron HA-TOOL) is a one-shot script that's "managed" by pacemaker. This is a problem, because the model for a daemon and a one-time script differ. We want to modify NHT so that it is a service rather than a one-shot script.
https://suse-jira.dyndns.org/browse/SCRD-117
https://trello.com/c/qeBCsC0d/9-turn-neutron-ha-tool-into-a-service


Itxaka:
[Forward-port neutron-ha-tool servicification](https://trello.com/c/XHzUu7Ug/53-forward-port-neutron-ha-tool-servicification)

Itxaka:
rebased!

PR #1181, opened 2017-08-21
Cookbooks: sync all nodes before setting wait database sync mark

Itxaka:
As we are not setting a sync mark before the database creation/migration
of nova, this could lead to a potential timeout in which the non-founder
nodes are faster than the founder and start waiting on the wait mark
earlier than the founder, which leads to the non-founder nodes starting
their timeout countdown earlier than the founder, potentially triggering
a timeout for no reason as the database creation/migration is not taking
120 seconds in the founder.

This makes it so all nodes are synced before starting the timeout
countdown for the database creation/migration.

### DIFF
--- a/chef/cookbooks/nova/recipes/database.rb
+++ b/chef/cookbooks/nova/recipes/database.rb
@@ -25,6 +25,10 @@ ha_enabled = node[:nova][:ha][:enabled]
 
 db_settings = fetch_database_settings
 
+# Wait for all nodes to reach this point so we avoid any timeouts due to the
+# non-founders being faster than the founder and not syncing properly with it
+crowbar_pacemaker_sync_mark "sync-nova_before_database" if ha_enabled
+
 crowbar_pacemaker_sync_mark "wait-nova_database" do
   # the db sync is very slow for nova
   timeout 120


Itxaka:

Itxaka:
@stefannica Not sure, what are those HA failures? Do we have a card to track that?

I dont think it will fix anything tbh, this is mostly for https://github.com/crowbar/crowbar-ha/pull/217 as with that approach to sync marks, non-founders are much faster than founders, so this gets triggered. Timing it, the database recipe takes about 63-68 seconds, so we still have a buffer of 50~ seconds, and I dont think on the current approach to sync marks makes the non-founders be that fast in regards to the founder.

but if it does fix stuff, Im taking all credit for sure, even if it was your similar patch on keystone the one that gave me the idea :P

stefannica:
@Itxaka it just occurred to me: the PR/commit titles are wrong. It should say nova instead of keystone.

Itxaka:
thanks @stefannica, I was just re-reading the commit and thinking that there was something wrong but I could not find what exactly. What an oversight :laughing:

Itxaka:
I had a small talk with @stefannica about this, and it seems like this is not the right approach. I mean, it kind of works, but only because its merely introducing an extra timeout by using the sync mark instead of addressing the real problem, that the founder and non-founders can drift apart in their execution, so by the time the founder sets the mark, the non-founders have already timed out.

Im gonna close this and send a different PR for a different approach, instead of micro-managing each sync mark and allowing the nodes to drift apart and fix every timeout, Im gonna introduce a sync recipe that should be run at the start of each cookbook (if HA enabled) so the nodes can sync at each recipe start. That should leave us with a more uniform run times for founder and non-founders and should execute the resources much close to each other, hopefully avoiding anymore timeouts.

Run time should not be affected as we are constrained to always wait for the founder node, which is the one that takes more time, so it should be ok (although ugly as it has to be a cross-repo fix :/ )

aspiers:
@Itxaka This PR looked good to me - even if we enhance the overall approach to syncing, is there any downside to also merging this extra database sync?  I can't think of one.

> Im gonna close this and send a different PR for a different approach, instead of micro-managing each sync mark and allowing the nodes to drift apart and fix every timeout, Im gonna introduce a sync recipe that should be run at the start of each cookbook (if HA enabled) so the nodes can sync at each recipe start. That should leave us with a more uniform run times for founder and non-founders and should execute the resources much close to each other, hopefully avoiding anymore timeouts.

Yeah, I thought about doing that a long time ago to address [bsc#935462 (missing synchronization means chef-client runs on cluster nodes drift apart)](https://bugzilla.suse.com/show_bug.cgi?id=935462).  I think that would probably be fine, but please read the history of that bug and also the workaround that @vuntz came up with in https://github.com/crowbar/crowbar-ha/pull/60 to make sure you understand all the nuances.  In particular it's important to realise that the sync marks only get reset (and hence have an effect) when applying a proposal.  For periodic chef-client runs, they will not.

Also take a look at [this tool I wrote](https://github.com/aspiers/SUSE-dist/blob/master/bin/lnav-chef-sync-marks) which makes it a lot easier to follow the timings of the sync marks across all cluster nodes.

Itxaka:
> Also take a look at this tool I wrote which makes it a lot easier to follow the timings of the sync marks across all cluster nodes.

WHY I DID NOT KNOW ABOUT THIS BEFORE CHECKING THE SYNC MARKS MANUALLY!

Awesome resource @aspiers, storing it on bookmarks!

> Yeah, I thought about doing that a long time ago to address bsc#935462 (missing synchronization means chef-client runs on cluster nodes drift apart). I think that would probably be fine, but please read the history of that bug and also the workaround that @vuntz came up with in crowbar/crowbar-ha#60 to make sure you understand all the nuances. In particular it's important to realise that the sync marks only get reset (and hence have an effect) when applying a proposal. For periodic chef-client runs, they will not.

Thanks for the pointers, they will be really helpful.

> @Itxaka This PR looked good to me - even if we enhance the overall approach to syncing, is there any downside to also merging this extra database sync? I can't think of one.

Should probably have to add them to all the database recipes, so I thougth into a more generic approach for this instead.

PR #1183, opened 2017-08-21
cookbooks: sync all nodes before setting wait database sync mark

Itxaka:
As we are not setting a sync mark before the database creation/migration
of heat, this could lead to a potential timeout in which the non-founder
nodes are faster than the founder and start waiting on the wait mark
earlier than the founder, which leads to the non-founder nodes starting
their timeout countdown earlier than the founder, potentially triggering
a timeout.

This makes it so all nodes are synced before starting the timeout
countdown for the database wait mark.


Itxaka:

Itxaka:

Itxaka:
Looks like there is more than one barclamp affected, I will push the rest to https://github.com/crowbar/crowbar-openstack/pull/1181 instead

PR #1185, opened 2017-08-22
database: Fix monitor op for galera pacemaker resource

rhafer:
Enable monitor also for nodes that are currently slaves. The galera
resource agent does regular checks for missing galera members and
latest commit in that case. Without this there might be issues
after a full reboot of the cluster.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -68,8 +68,7 @@ pacemaker_primitive service_name do
   })
   op({
     "monitor" => {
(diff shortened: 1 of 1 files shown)


rhafer:
[P9 (5) Fix problems with restarting galera cluster](https://trello.com/c/xyoZFWnT/10-p9-5-fix-problems-with-restarting-galera-cluster)

rhafer:
This PR is wrong. As it breaks monitoring for the Master nodes. Also my initial assumption about bootstrapping issues with a rebooted cluster seem to be wrong. With a fixed galera resource agent bootstrapping after reboot seems to work fine.

PR #1186, opened 2017-08-22
nova: Clean up config template

cmurphy:
Currently, the nova.conf that is rendered by chef contains a lot of
empty newlines. This is cosmetically displeasing but also hinders
debugging because of the need to scroll further to see all declared
configuration options. Moreover, the markup in this template was
inconsistent with itself, so new additions had no good example to copy
from.

Chef uses Erubnis, which is not actually the same as ERB. One of the
main differences is that in Erubnis, `<% %>`, `<%- %>`, `<% -%>`, and
`<%- -%>` are all equivalent, i.e. the `-` is completely unnecessary.
Where the `-` makes a difference is when closing a `<%=` tag, especially
if the expression within the tags is conditional. There is not really
any documentation about this, so this evaluation is based on trial and
error and stackoverflow[1].

This change changes the markup to consistently use conditional
beginnings and ends on their own lines, using `<% %>`, with the
expression in between, rather than in one line within `<%= %>` tags or
in one line between two separate tags, which also causes unnecessary
newlines to be generated. This way there's no need to guess whether to
close the `<%=` with a `-%>` or a `%>`. Also removes use of `<%-` and
`-%>` since those are meaningless here.

It is too much work to clean up every template like this but we can now
reference the nova config template for future additions.

[1] https://stackoverflow.com/questions/7996695/what-is-the-difference-between-and-in-erb-in-rails

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -2,11 +2,13 @@
 ssl_only = <%= @vncproxy_ssl_enabled ? "True" : "False" %>
 cert = <%= @vncproxy_cert_file %>
 key = <%= @vncproxy_key_file %>
-<% if @libvirt_type.eql?('zvm') -%>
+<% if @libvirt_type.eql?('zvm') %>
 instance_name_template=zvm%05x
 <% end %>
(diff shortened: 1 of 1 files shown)


rsalevsky:
Very nice! Thanks for doing this.

PR #1188, opened 2017-08-22
ceilometer: combined commit to replace auth_token in process cache with memcached in ocata release

charleswang007:
The following Ceilometer configuration options deprecated in Ocata (or earlier) are removed or replaced in this PR:

- [keystone_authtoken]/service_token_roles_required is set to True to enable fetching expired tokens based on valid service tokens and [keystone_authtoken]/service_token_roles is set to the admin role

The deprecated keystonemiddleware.auth_token in-process token cache used by aodh services is replaced with a memcached server. When HA is used, several memcached servers are configured and used, one for each aodh-server node. The keystone token data stored in memcached is authenticated and encrypted using a generated secret key.

References:

- Ocata release notes for Ceilometer: https://docs.openstack.org/releasenotes/ceilometer/ocata.html
- Mitaka release notes for keystonemiddleware: https://docs.openstack.org/releasenotes/keystonemiddleware/mitaka
- Memcached protection options: https://docs.openstack.org/keystonemiddleware/latest/middlewarearchitecture.html#memcache-protection

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/common.rb
+++ b/chef/cookbooks/ceilometer/recipes/common.rb
@@ -1,5 +1,13 @@
 include_recipe "apache2"
 
+is_controller = node["roles"].include?("ceilometer-controller") 
+
+memcached_servers = MemcachedHelper.get_memcached_servers(
+  ha_enabled ? CrowbarPacemakerHelper.cluster_nodes(node, "ceilometer-server") : [node]
+)
+
+memcached_instance("ceilometer-server") if is_controller
+
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
 if node[:ceilometer][:use_mongodb]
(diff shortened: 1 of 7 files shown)


chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -1,5 +1,13 @@
 include_recipe "apache2"
 
+is_controller = node["roles"].include?("ceilometer-controller")
```
cmurphy: This should be ceilometer-server, not ceilometer-controller.

You can tell by checking the role definition here: https://github.com/crowbar/crowbar-openstack/blob/a20fc751a0bc35e620255126ca77a4c133277924/chef/roles/ceilometer-server.rb#L1

Or also by running `knife role list` on the crowbar admin node and looking for ceilometer-related roles.

PR #1189, opened 2017-08-22
mariadb: Enable slow query logging

rsalevsky:
Before the config cleanup some archs had slow_query_logging enabled.
Let's reenable it for all archs.

### DIFF
--- a/chef/cookbooks/mysql/recipes/server.rb
+++ b/chef/cookbooks/mysql/recipes/server.rb
@@ -78,6 +78,17 @@ template "/etc/my.cnf.d/openstack.cnf" do
   notifies :restart, "service[mysql]", :immediately
 end
 
+template "/etc/my.cnf.d/logging.cnf" do
+  source "logging.cnf.erb"
+  owner "root"
+  group "mysql"
+  mode "0640"
+  variables(
+    slow_query_logging_enabled: node[:database][:mysql][:slow_query]
+  )
+  notifies :restart, "service[mysql]", :immediately
+end
+
 if node[:database][:ha][:enabled]
   template "/etc/my.cnf.d/galera.cnf" do
     source "galera.cnf.erb"
--- a/chef/cookbooks/mysql/templates/default/logging.cnf.erb
+++ b/chef/cookbooks/mysql/templates/default/logging.cnf.erb
@@ -1,0 +1,7 @@
+[mysqld]
+log_error=/var/log/mysql/mysql_error.log
+
+<% if @slow_query_logging_enabled -%>
+slow_query_log=1
+slow_query_log_file=/var/log/mysql/mysql_slow.log
+<% end -%>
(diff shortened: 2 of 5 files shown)


jsuchome:
What is the meaning of slow query and why do we want it?

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -66,6 +66,11 @@
 EOC
 end
 
+# FIXME: The configs installed by the package are currently not overwritable
```
jsuchome: So is it a packaging bug? Should we file a bug report so it gets fixed?

nicolasbock:
@jsuchome The slow query log logs queries that are slow. You can tune what is considered slow with the `long_query_time` setting. The default is 10 seconds though (see https://mariadb.com/kb/en/the-mariadb-library/slow-query-log-overview/). Nice feature!

rsalevsky:
@jsuchome Added the explanation to the commit message.

chef/cookbooks/mysql/recipes/server.rb
rsalevsky: Done

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -66,6 +66,11 @@
 EOC
 end
 
+# FIXME: The configs installed by the package are currently not overwritable (bsc#1055268)
+file "/etc/my.cnf.d/error_log.cnf" do
+  action :delete
```
rhafer: Hm, this means with every mariadb maintenance update we will get the file back, until chef-client runs again and deletes it. As chef-client won't restart the server after deleting the file (but rpm will do), we might endup running with the wrong configuration. I think I'd prefer to just ship a template for "/etc/my.cnf.d/error_log.cnf" for now. 

Or alternatively just add the "slow_query_log=1" settings to separate file "slow_query_log.cnf" and leave "/etc/my.cnf.d/error_log.cnf" in place. I think that should work as well.

rsalevsky:
@rhafer I removed the part completely as the logging config gets loaded after the error_log config. So it's not a problem for us.

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -78,6 +78,17 @@
...
 
+template "/etc/my.cnf.d/logging.cnf" do
+  source "logging.cnf"
+  owner "root"
+  group "mysql"
+  mode "0640"
+  variables(
+    slow_query_logging_enabled: node[:database][:mysql][:slow_query]
+  )
+  notifies :restart, "service[mysql]", :immediately
```
rhafer: Yet another immediate restart. This means in the worst case mysql will be restarted 3 times during a chef-client. But I am fine to merge this for now an address that later.

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -78,6 +78,17 @@
   notifies :restart, "service[mysql]", :immediately
 end
 
+template "/etc/my.cnf.d/logging.cnf" do
+  source "logging.cnf"
```
jsuchome: `"logging.cnf.erb"`

PR #1192, opened 2017-08-23
mariadb: Install the galera-3-lib package

rsalevsky:
The galera-3 got splited into a base package and a lib package.


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -19,7 +19,7 @@
 
 resource_agent = "ocf:heartbeat:galera"
 
-package "galera-3"
+package "galera-3-lib"
```
nicolasbock: We only need the 'lib' package? Aren't there binaries or services in the galera-3 package we might need as well?

jsuchome:
Was the same packaging done for SOC7 as well?

chef/cookbooks/mysql/recipes/ha_galera.rb
rhafer: @nicolasbock Current we really only need that library. (Which contains the replication provider code). The other stuff in that package (mainly garbd) is not used in our configuration.

rsalevsky: Yes, the other parts are in the MariaDB package.

rsalevsky:
@jsuchome Yes, this needs a backport.

chef/cookbooks/mysql/recipes/ha_galera.rb
nicolasbock: Thanks!

rsalevsky:
There was some disagreement between the maintainers, now we have the final name.

PR #1193, opened 2017-08-23
(NOT FULLY TESTED) nova: Subscribe to placement config (bsc#1055188)

cmurphy:
Without this change, an HA deployment will change the placement config file to
point to the cluster endpoint, but if this happens after the compute
service is refreshed (triggered by a change in the main config file), it
will not receive the changes and will fail to be able to access the
placement API. The nova compute service needs to be refreshed when this
file changes because it uses the placement user credentials contained in
it to look up, through keystone, the placement API endpoint. This change
ensures that nova services subscribe to the placement config file in
addition to the main config file.


cmurphy:
The problem seems to be somewhere else.

PR #1194, opened 2017-08-23
nova: Randomize placement password

cmurphy:
Without this change, the nova placement user has a default password of
"placement", set in attributes.rb, and it is impossible to change it.
This is a massive security flaw for obvious reasons. This change adds
the ability to change the placement service password via the raw view of
the nova barclamp and also ensures it defaults to a random one.


cmurphy:
> Don't you also need to use this attribute in the nova::config recipe?
> Wait, it's used, but there is an entry in the default attributes. We could remove that.

Yeah that is kind of the problem, there is a default non-empty password defined and we're using that in nova::config:

https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/nova/recipes/config.rb#L329

I decided not to remove it from the default attributes file since other passwords are already defined there, this just ensures that crowbar will ignore them.

PR #1195, opened 2017-08-23
nova: Use internal placement url (bsc#1055188)

cmurphy:
By default, the nova compute service tries to access the public endpoint
of the placement API. The "public" network defined in network.json isn't
accessible to the compute nodes. Without this patch, the nova-compute
service is unable to connect to the placement API and so is unable to be
scheduled. This patch forces the service to request the internal
placement endpoint, which it can reach.


cmurphy:
Woops, yes, two thoughts collided there

PR #1197, opened 2017-08-24
nova: Subscribe to placement config (bsc#1055188)

cmurphy:
Without this change, changing the placement configuration after the
initial deployment will have no effect on any of the nova services. This
is a problem if we want to change the way the compute service connects
to the placement API after the initial deployment. This patch ensures
that nova services subscribe to both configuration files. The placement
config file is really just an extension of the main config file, so it
makes sense for the services to subscribe to both.

### DIFF
--- a/chef/cookbooks/nova/definitions/nova_package.rb
+++ b/chef/cookbooks/nova/definitions/nova_package.rb
@@ -55,7 +55,8 @@ define :nova_package, enable: true, use_pacemaker_provider: false, restart_crm_r
       end
     end
 
-    subscribes :restart, resources(template: node[:nova][:config_file])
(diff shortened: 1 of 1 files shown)


cmurphy:
Supercedes #1193 which I can't reopen because I reused that branch for something else

toabctl:
@cmurphy isn't it just the ```nova-placement``` service that would need to be restarted if the file changes? Also ```nova-placement``` is running under apache. This change would have no effect I think.

cmurphy:
@toabctl no, the nova-compute service must be restarted too. The nova-compute service makes requests to the placement API service to report its available resources. It uses the [placement] section of the nova config file to connect to keystone to look up the placement endpoint.

PR #1198, opened 2017-08-24
MariaDB config tuning

rsalevsky:
This PR splits the tuning related config options in a extra config and makes them available via crowbar (for tuning purposes).

This settings expect at least 1GB of total memory and are optimized for this.

### DIFF
--- a/chef/data_bags/crowbar/template-database.json
+++ b/chef/data_bags/crowbar/template-database.json
@@ -6,7 +6,11 @@
       "sql_engine": "postgresql",
       "mysql": {
         "datadir": "/var/lib/mysql",
-        "slow_query_logging": true
+        "slow_query_logging": true,
+        "innodb_buffer_pool_size": 256,
+        "max_connections": 800,
+        "tmp_table_size": 64,
+        "max_heap_table_size": 64
       },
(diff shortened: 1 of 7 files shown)


jsuchome:
@rsalevsky not exactly related to these changes, but more to the process. What about moving further cleanups to next related card (P13 cookbook update)?

chef/data_bags/crowbar/template-database.json
```diff
@@ -6,7 +6,11 @@
       "sql_engine": "postgresql",
       "mysql": {
         "datadir": "/var/lib/mysql",
-        "slow_query_logging": true
+        "slow_query_logging": true,
+        "innodb_buffer_pool_size": 256,
+        "max_connections": 800,
```
s-t-e-v-e-n-k: Why 800? The default value for Postgres is 1000 (and is visible as a default option when creating/editing the database barclamp)

rsalevsky: For MariaDB it is not recommended to use such a high max_connection number unless required; as higher the number is as more memory is used. In most of our use cases 800 should be enough if not the operator can increase the number to find the right balance.

PR #1200, opened 2017-08-25
mariadb: Rename slow_query to slow_query_logging

rsalevsky:
I just realized that slow query is not specific enough and does not
provide any context. This commit renames slow_query too
slow_query_logging and changes the migrations.

### DIFF
--- a/chef/cookbooks/mysql/recipes/server.rb
+++ b/chef/cookbooks/mysql/recipes/server.rb
@@ -84,7 +84,7 @@ template "/etc/my.cnf.d/logging.cnf" do
   group "mysql"
   mode "0640"
   variables(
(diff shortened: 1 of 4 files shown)


jsuchome:
@nicolasbock I think that because no package with previous change was released yet, we're fine with just a rename...

nicolasbock:
Ok, that sounds reasonable @jsuchome. But wouldn't it be possible that a customer had received a PTF which included the first migration which would then require a second migration?

rsalevsky:
@nicolasbock The change was not backported to Cloud 7 (yet) so there is no way a customer has a PTF, also there was no Cloud 8 Milestone with this change.

nicolasbock:
Thanks @rsalevsky, now it makes sense. I hadn't considered that this is Cloud 8.

PR #1201, opened 2017-08-28
mariadb: Pull in additional packages required for SST

rhafer:
"mariadb-tools" and "xtrabackup" contain the tooling for "xtrabackup"
and "rsync" based SST (State Snapshot Transfer) to work.


jsuchome:
[P11 galera config: Find the best value for SST](https://trello.com/c/27Q5XW7K/28-p11-galera-config-find-the-best-value-for-sst)

PR #1204, opened 2017-08-29
keystone: Stop syncing signing keys (SCRD-781)

cmurphy:
PKI tokens have been removed from keystone, and the pki_setup command
that used to generate the initial signing key pair has been removed from
the keystone package[1]. Since there is no longer an initial signing key
pair at the point where the keystone barclamp is supposed to try to
synchronize them, the keystone barclamp fails on pike with "No such file
or directory". If a key pair is needed to HTTPS, the barclamp will
generate it, but it is no longer needed for token functionality, so we
can just remove the synchronization steps.

[1] https://build.opensuse.org/package/rdiff/Cloud:OpenStack:Master/openstack-keystone?linkrev=base&rev=1001


cmurphy:
Updated, I think the sync marks can go away too

PR #1205, opened 2017-08-29
Do not deploy monasca-server to the node running database cluster

jsuchome:
Node with monasca-server deploys its own mariadb instances.
Let's not mix it with OpenStack database (now using mariadb as well).

### DIFF
--- a/crowbar_framework/app/models/monasca_service.rb
+++ b/crowbar_framework/app/models/monasca_service.rb
@@ -91,8 +91,13 @@ class MonascaService < OpenstackServiceObject
     base = super
 
     nodes = NodeObject.all
+    non_db_nodes = nodes.reject do |n|
+      # Do not deploy monasca-server to the node running database cluster (already running mariadb)
+      n.roles.include?("database-server") && n[:database][:sql_engine] == "mysql"
+    end
 
-    monasca_server = select_nodes_for_role(nodes, "monasca-server", "monitoring") || []
+    monasca_server = select_nodes_for_role(non_db_nodes, "monasca-server", "monitoring") || []
+
     log_agent_nodes = select_nodes_for_role(nodes, "monasca-log-agent", "compute") || []
     agent_nodes = select_nodes_for_role(nodes, "monasca-agent") || []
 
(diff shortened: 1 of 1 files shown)


rsalevsky:
@jsuchome Can you explain why we should not mix it? Can't they share the database?

jsuchome:
@rsalevsky that might be possible yes, although I have no idea if it is recommended.

But as I assume this mariadb instance is set up by ansible playbook, someone else should step in.

toabctl:
> But as I assume this mariadb instance is set up by ansible playbook, someone else should step in.

Yes. mariadb is installed via ansible. I think we can disable the installation in ansible when we have mariadb available. But for now I'm fine with adding this constraint

jsuchome:
@toabctl would it be enough to disable the installation from the playbook? We'd need at least provide correct credentials (and I have no idea what else does the playbook do regarding the DB).

crowbar_framework/app/models/monasca_service.rb
```diff
@@ -91,8 +91,13 @@ def create_proposal
     base = super
 
     nodes = NodeObject.all
+    non_db_nodes = nodes.reject do |n|
+      # Do not deploy monasca-server to the node running database cluster (already running mariadb)
+      n.roles.include?("database-server")
+    end
+
+    monasca_server = select_nodes_for_role(non_db_nodes, "monasca-server", "monitoring") || []
```
toabctl: Can we check that mariadb is used as database? It's totally fine to have monasca-master together with database-server if the database is postgres. And if we want to backport this patch, only checking for the database-server might break things.

toabctl:
> would it be enough to disable the installation from the playbook? We'd need at least provide correct credentials (and I have no idea what else does the playbook do regarding the DB).

It would be possible afaics. But for now I would go with this solution.

crowbar_framework/app/models/monasca_service.rb
jsuchome: With Cloud8, mariadb is the only option, postgres willbe dropped

jsuchome: @toabctl done.

jsuchome:
[P7 Update mariadb package in SOC7](https://trello.com/c/o5EjZr3I/29-p7-update-mariadb-package-in-soc7)

crowbar_framework/app/models/monasca_service.rb
```diff
@@ -135,6 +140,12 @@ def validate_proposal_after_save(proposal)
     validate_one_for_role proposal, "monasca-server"
     nodes = proposal["deployment"][@bc_name]["elements"]
     nodes["monasca-server"].each do |node|
+      n = NodeObject.find_node_by_name(node)
+      if n.roles.include?("database-server") && n[:database][:sql_engine] == "mysql"
+        validation_error(
+          "monasca-server role cannot be deployed to the node with database-server role."
```
toabctl: Should mention mysql. the database-server role itself is ok

PR #1209, opened 2017-08-30
mysql: Provide datadir value to the pacemaker resource

jsuchome:
We allow users to change the default value of mysql data directory,
so we have to make sure galera resource agent knows about it.


jsuchome:
[P4 (5) Sanitize and enhance mariadb configuration done by cookbook](https://trello.com/c/PcPTyEX7/1-p4-5-sanitize-and-enhance-mariadb-configuration-done-by-cookbook)

PR #1211, opened 2017-08-30
horizon: Reload horizon after migration instead of restart

Itxaka:
Apache reload (graceful) will signal the apache process, that will
restart the child process (WSGI).  This can cause a race, were an
OpenStack client try to contact to one of the WSGI services, and
the old WSGI process can answer.  This process will fail because
the WSGI socket is not anymore in the file system.

The aproach here is to use a dummy service for 'horizon', what
will SIGUSR1 the wsgi:horizon childs from apache, signaling the
reload of the configuration file which was introduced on commit 02120688d5d3bf35ba12d29ec68be3eba19253f9

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -389,7 +389,7 @@ execute "python manage.py migrate" do
   group node[:apache][:group]
   action :nothing
   subscribes :run, "template[#{local_settings}]", :immediately
-  notifies :restart, resources(service: "apache2"), :immediately
+  notifies :reload, resources(service: "horizon"), :immediately
 end
 
 # Force-disable multidomain support when the default keystoneapi version is too


aplanas:
lets see the CI : )

Itxaka:
@nicolasbock I just lifted the commit message from the original commit that fixed this into different locations.
What it means its that the common `apache reload` action is the one that can cause this, so a new reload was introduced by @aplanas in https://github.com/crowbar/crowbar-openstack/commit/02120688d5d3bf35ba12d29ec68be3eba19253f9 which manually sends SIGUSR1 to the horizon childs from apache.

Maybe I should also needed to steal that part of the commit message to make it clearer :D

nicolasbock:
Ah, I had forgotten about @aplanas 's change. Thanks for reminding me! I think it would help already if you reference that commit (https://github.com/crowbar/crowbar-openstack/commit/02120688d5d3bf35ba12d29ec68be3eba19253f9) in your commit message.

Itxaka:
@nicolasbock indeed, it was a mistake to not include it. Updated commit message with the extended explanation+commit link

PR #1212, opened 2017-08-30
[4.0] monasca: Do not deploy monasca-server to the node running mariadb

jsuchome:
Node with monasca-server deploys its own mariadb instances.
Let's not mix it with OpenStack database that can use mariadb as well.

If we'd want to use shared database for both usages, ansible based
monasca-installer has to be adapted.

(cherry picked from commit 417d0578a8ae4a7de3b53faee9146bec0f31a840)

Backport of https://github.com/crowbar/crowbar-openstack/pull/1205

### DIFF
--- a/crowbar_framework/app/models/monasca_service.rb
+++ b/crowbar_framework/app/models/monasca_service.rb
@@ -91,8 +91,13 @@ class MonascaService < OpenstackServiceObject
     base = super
 
     nodes = NodeObject.all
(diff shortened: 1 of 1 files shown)


jsuchome:
[P7 Update mariadb package in SOC7](https://trello.com/c/o5EjZr3I/29-p7-update-mariadb-package-in-soc7)

PR #1214, opened 2017-08-30
Local cache for keystoneHelper

sjamgade:
_find_id makes many redundant calls to keystone for listing users,
tenants... during one chef-client, but would return only what is
required right now.

This patch changes that behaviour, instead of throwing away all the
additional data, _find_id would cache the data in KeystoneHelper allowing
the listing (user, tenants...) to be reused, bypassing the need to
actually create a socket, connect and so.

### DIFF
--- a/chef/cookbooks/keystone/libraries/helpers.rb
+++ b/chef/cookbooks/keystone/libraries/helpers.rb
@@ -46,6 +46,7 @@ module KeystoneHelper
       end
       @keystone_settings = nil
       @keystone_node = nil
+      cache_reset
       @keystone_settings_cache_time = current_node[:ohai_time]
     end
 
@@ -122,5 +123,17 @@ module KeystoneHelper
 
     Chef::Log.info("Keystone server found at #{@keystone_node[instance].name}")
     return @keystone_node[instance]
+  end
+
+  def self.cache
+    @cache
+  end
+
+  def self.cache_update(update)
+    @cache = @cache.merge(update)
+  end
+
+  def self.cache_reset
+    @cache = {}
   end
 end
--- a/chef/cookbooks/keystone/providers/register.rb
+++ b/chef/cookbooks/keystone/providers/register.rb
@@ -419,24 +419,35 @@ def _build_connection(new_resource)
 end
 
 def _find_id(http, headers, item_name, path, dir, key = "name", ret = "id")
-  # Construct the path
-  my_item_id = nil
+  # this can break your code, if you are asking for name (ret),
+  # find_id will have to be modified to not search the cache.
+  # the cache stores only the "ret" that it was first querried with
+  my_item_id = _find_id_in_cache(item_name, path)
   error = false
-  resp = http.request_get(path, headers)
-  if resp.is_a?(Net::HTTPOK)
-    data = JSON.parse(resp.read_body)
-    data = data[dir]
+  unless my_item_id
+    resp = http.request_get(path, headers)
+    if resp.is_a?(Net::HTTPOK)
+      data = JSON.parse(resp.read_body)
+      data = data[dir]
+      data2hash = {}
 
-    data.each do |item|
-      my_item_id = item[ret] if item[key] == item_name
-      break if my_item_id
+      data.each do |item|
+        my_item_id = item[ret] if item[key] == item_name
+        data2hash[[path, item[key]]] = item[ret]
+      end
+      KeystoneHelper.cache_update(data2hash) if my_item_id
+    else
+      log_message = "Find #{path}: #{item_name}: Unknown response from Keystone Server"
+      _log_error(resp, log_message)
+      error = true
     end
-  else
-    log_message = "Find #{path}: #{item_name}: Unknown response from Keystone Server"
-    _log_error(resp, log_message)
-    error = true
   end
   [my_item_id, error]
+end
+
+def _find_id_in_cache(rsc_name, rpath)
+  cache = KeystoneHelper.cache
+  cache[[rpath, rsc_name]]
 end
 
 def _build_service_object(svc_name, svc_type, svc_desc)


chef/cookbooks/keystone/providers/register.rb
```diff
@@ -46,7 +46,8 @@
   dir = "OS-KSADM:services"
 
   # Lets verify that the service does not exist yet
-  item_id, error = _find_id(http, headers, new_resource.service_name, path, dir)
+  item_id, error = _find_id_in_cache(new_resource.service_name, path, dir) ||
+                   _find_id(http, headers, new_resource.service_name, path, dir)
```
nicolasbock: Can you hide this `||` logic inside the helper method? In other words, call only the `_find_id_in_cache` method and that method checks whether the requested information is already cached or not and then calls `_find_id` if necessary.

sjamgade:
This optimization achieves a speed of minimum 10 secs over a single node deployment. I could come up with an upper bound. The it varied from 10 to 30 secs.  (I did not actually measure but observed it from the time logged after every client run)

The cache uses the `[spath, dir]` as key, and not just `spath` just to be future safe. Other wise in all the usecase a `spaht` and `dir` were in 1:1 correspondence.

The implementation could be simplified by calling `_find_id_from_cache` from `_find_id` instead of calling it independently. But I have delirately done it otherwise to highlight that `http` object need not exist in cases when there is cache-hit and thus avoiding the call to `_build_connection` (which actually does get the `auth_token`  in headers). There is a cache-hit in almost all `add_*` actions.

chef/cookbooks/keystone/providers/register.rb
sjamgade: Can you please see my comment.
I plan to handle this in a separate PR.

nicolasbock:
If I understand you correctly then you are planning on another PR in which you avoid the `_build_connection` call in cases where there is a cache already?

sjamgade:
yes

sjamgade:
thanks @Itxaka for catching that one. 
I have added that parameter for that.

sjamgade:
[Local cache for keystoneHelper](https://trello.com/c/AYbUFg1a/442-local-cache-for-keystonehelper)

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -504,6 +523,19 @@ def _find_id(http, headers, svc_name, spath, dir, key = "name", ret = "id")
 end
 
 private
+
+def _find_id_in_cache(svc_name, spath, dir, key = "name", ret = "id")
```
cmurphy: Don't use the names svc/svc_name/spath, it's a mistake in _find_id that the variables are named that way since this is generic to any keystone resource, not just services

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -504,6 +523,19 @@ def _find_id(http, headers, svc_name, spath, dir, key = "name", ret = "id")
...
+  caches = KeystoneHelper.caches()
+  if caches and caches[[spath, dir]]
```
cmurphy: An empty hash is truthy so .caches() is always going to return a truthy value, so you can either remove the check here or explicitly check whether it is empty, not just falsy.

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -504,6 +523,19 @@ def _find_id(http, headers, svc_name, spath, dir, key = "name", ret = "id")
...
+    caches[[spath, dir]].each do |svc|
+      return svc[ret] if svc[key] == svc_name
```
cmurphy: You are already keying on [spath, dir], why not key on [spath, dir, key] and avoid having to loop?

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -120,4 +121,14 @@ def self.keystone_settings(current_node, cookbook_name)
...
   end
+
+  def self.caches(d = nil, renew=false)
+    if renew
+        @caches = Hash.new
+    else
+        @caches ||= Hash.new
+    end
+    @caches = @caches.merge(d) if d
+    @caches
```
cmurphy: Are you sure these don't need to be @@caches? Since this is being used as a class method and not an object method?

sjamgade: yes I have tested this, and just for brevity i have renamed the variable.

chef/cookbooks/keystone/providers/register.rb
sjamgade: okay. I think the newer ones make more sense now

chef/cookbooks/keystone/providers/register.rb
sjamgade: the point is
 `["/v2.0/services","services"]  => [{name:keystone, id:....}, {name:glance, id:....},....]`
So now I have all the the data of all the `services (dir)` at `/v2.0/services (spath)`, then i have `id (ret)` of `glance` (when deploying `glance`) without asking `keystone` again.

The cache lives only through one chef client run and is automatically invalidated in case of a miss for a particular `[spath, dir]`

aplanas: I still consider the invalidation comment relevant. A POST / PUT / DELETE will change / create / delete some data that can already be in the cache, but that for the action is now not valid.

sjamgade:
Merging the two `find_id` requires substantial refactoring. That will increase the size of the change.
I like to think of them as two separate logical changes. Intermediate change does not really break or degrade.

The lifespan of the cache is only across chef-client run, so cache is not really broken, unless the admin goes ahead and alters any of these entities during the chef-client runs

chef/cookbooks/keystone/providers/register.rb
sjamgade: I did look for any cases in the `crowbar-openstack` codebase where we were altering any keystone resource without using `KeystoneHelper` class. I could not find any.

Also since our cookbooks are supposed to _pseudo re-entrant_, altering any keystone resource without checking if it needs modifications, feels unnecessary, So if it all there is a place where we are doing this (and is outside the scope of KeystoneHelper) I would be a happy to patch that to reset the cache also.

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -504,6 +523,19 @@ def _find_id(http, headers, svc_name, spath, dir, key = "name", ret = "id")
 end
 
 private
+
+def _find_id_in_cache(rsc_name, rpath, dir, key = "name", ret = "id")
+  caches = KeystoneHelper.caches()
+  if !caches.empty? and caches[[rpath, dir]]
+    caches[[rpath, dir]].each do |svc|
+      return svc[ret] if svc[key] == rsc_name
```
cmurphy: Instead of looping in the read here I think you could push the workload to the write on line 512 so that the cache key includes the resource name and the read is immediate. In the case of users, for example, there could be thousands, so looping over every one on every cache lookup would be really intensive.

sjamgade:
Though the use of a finer cache `key` makes the access the `value` easier but it diverts away from the api provided by `_find_id`, so if a certain `add_name` function wants to `name` for a `user=crowbar` and `add_user` wants `id` for a `user=crowbar` we will have a cache collision.


By keeping the parameter of `_find_id_in_cache` and `_find_id` same, the use of cache becomes transparent to caller of `_find_id`

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -604,18 +602,22 @@ def endpoint_needs_update(interface, endpoints, new_resource)
 end
 
 def _get_service_id(http, headers, svc_name)
+  _find_id_in_cache(svc_name, "/v3/services") ||
```
aspiers: Ditto here, there's a bunch of duplicated code which suggests you can roll the cache lookup into `_find_id` or at least into a new method which combines them both.  [DRY!](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -419,26 +419,42 @@ def _build_connection(new_resource)
...
 
 def _find_id(http, headers, item_name, path, dir, key = "name", ret = "id")
-  # Construct the path
-  my_item_id = nil
+  # this can break your code, if you are asking for name (ret),
+  # find_id will have to be modified to not search the cache.
+  # the cache stores only the "ret" that it was first querried with
+  my_item_id = _find_id_in_cache(item_name, path)
   error = false
   resp = http.request_get(path, headers)
```
cmurphy: So this looks for the item in the cache, and then does the GET irrespective of whether it had a cache hit?

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -419,26 +419,42 @@ def _build_connection(new_resource)
 end
 
 def _find_id(http, headers, item_name, path, dir, key = "name", ret = "id")
-  # Construct the path
-  my_item_id = nil
+  # this can break your code, if you are asking for name (ret),
```
cmurphy: "this can break your code" - that sounds scary but I'm unclear from the comment what the problem is or how we're supposed to circumvent it. Is this TODO work for the future or instructions for developers on how to use _find_id?

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -419,26 +419,42 @@ def _build_connection(new_resource)
...
-    error = true
   end
   [my_item_id, error]
 end
 
+private
+
+def _find_id_in_cache(rsc_name, rpath)
+  cache = KeystoneHelper.cache
+  return cache[[rpath, rsc_name]] unless cache.empty?
```
cmurphy: I'm not sure the `unless cache.empty?` is needed, since {}['somekey'] will return nil without erroring out

chef/cookbooks/keystone/providers/register.rb
sjamgade: I tried to explain it [in this comment ](https://github.com/crowbar/crowbar-openstack/pull/1214#issuecomment-329098727)

Its not exactly a TODO but more like a comment about the possible cache collision.

sjamgade:
the logs from two subsequent proposal commits, 
after patch 
`[2017-10-05T12:48:46+00:00] INFO: Chef Run complete in 120.142086048 seconds`
and before patch:
`[2017-10-05T13:01:45+00:00] INFO: Chef Run complete in 135.69424046 seconds`

sjamgade:
damn-it, I cant even type !
s/lastest/latest

PR #1217, opened 2017-08-31
cinder: volume snapshots for NFS (bsc#1053414)

aplanas:
Add a new parameter in the UI to allow volume snapshots in NFS. This
feature must also disable the parameter `nas_secure_file_operations`.

This is configuration have some security implications[1] and a
warning text is added as a help hint in the UI.

[1] https://docs.openstack.org/security-guide/block-storage/checklist.html#check-block-07-is-nas-operating-in-a-secure-environment

### DIFF
--- a/chef/data_bags/crowbar/template-cinder.json
+++ b/chef/data_bags/crowbar/template-cinder.json
@@ -75,7 +75,8 @@
           "iscsi_ip": ""
         },
         "nfs": {
-          "nfs_shares": ""
+          "nfs_shares": "",
+          "nfs_snapshot": false
         },
         "rbd": {
           "use_crowbar": true,
@@ -176,7 +177,7 @@
     "cinder": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 205,
+      "schema-revision": 206,
       "element_states": {
           "cinder-controller": [ "readying", "ready", "applying" ],
           "cinder-volume": [ "readying", "ready", "applying" ]
--- a/chef/cookbooks/cinder/templates/default/cinder.conf.erb
+++ b/chef/cookbooks/cinder/templates/default/cinder.conf.erb
@@ -180,6 +180,10 @@ nfs_shares_config=/etc/cinder/nfs_shares-<%= backend_id %>
   <% if volume['backend_driver'] == 'nfs' -%>
 nfs_shares_config=/etc/cinder/nfs_shares-<%= backend_id %>
 volume_driver = cinder.volume.drivers.nfs.NfsDriver
+    <% if volume['nfs']['nfs_snapshot'] -%>
+nfs_snapshot_support = true
+nas_secure_file_operations = false
+    <% end -%>
   <% end -%>
 
   <% if volume['backend_driver'] == 'nimble' -%><% end -%>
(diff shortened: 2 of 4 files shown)


crowbar_framework/config/locales/cinder/en.yml
```diff
@@ -62,6 +62,7 @@ en:
           nfs_volume_driver: 'NFS'
           nfs:
             nfs_shares_config_hint: 'Format one entry per line in the form hostname:/vol/path -o optional-nfs-mount-options'
+            nfs_snapshot_hint: 'Security caution. This feature will is not recommended in security deployments.'
```
nicolasbock: Just some wording: drop the 'will'.

Itxaka: Code looks good to me, but I dont understand this. what is a `security deployment`. Is it a secure environment?

nicolasbock: Good point, I missed that earlier :smile:. How about `This feature is not recommended because of its significant security implications.`

aplanas: Thanks, done!

aplanas:
Do not need a backport, this fix is only for master (SOC8)

chef/data_bags/crowbar/template-cinder.json
```diff
@@ -163,7 +164,7 @@
     "cinder": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 202,
+      "schema-revision": 203,
```
dirkmueller: needs to be 204 now.

crowbar_framework/config/locales/cinder/en.yml
```diff
@@ -62,6 +62,7 @@ en:
           nfs_volume_driver: 'NFS'
           nfs:
             nfs_shares_config_hint: 'Format one entry per line in the form hostname:/vol/path -o optional-nfs-mount-options'
+            nfs_snapshot_hint: 'Security caution. This feature is not recommended because of its significant security implications.'
```
dirkmueller: maybe don't expose it in the ui then? if its in the UI we need to add documentation.

aplanas: There is a pending bug to documentation after this merge (it is as task in the card).

What do you refer to it, the comment or the field? The field needs to be in the UI, because adding it in the raw is not trivial, but if you talk about the hint, I would prefer to be also in the UI, that is more probable to be read than the documentation.

aplanas:
* Rebased and moved to migration 204

crowbar_framework/config/locales/cinder/en.yml
dirkmueller: @aplanas I think this looks terrible from a first perspective. if we don't recommend enabling it then don't expose it in the ui. if we expose it in the UI then we need to explain what the risk is. "significant security implications" isn't very telling.

aplanas: Uhm, indeed is not a good text, maybe this?:

"Enable this option if the NAS is in a secure environment"

nicolasbock: Maybe: "Note that the network traffic between server and client is not encrypted and this option should only be enabled if the NAS is on a secured network."

aplanas: This sound much better, but the issue AFAIU is not about encryption, but the recognition of the root user.

nicolasbock: Oh, sorry, I misunderstood the security implications... :smile:

"Note that using NFS has security implications that should be considered prior to deployment."

aplanas:
@dirkmueller can you re-review?

vuntz:
@aplanas is this something to backport to stable/4.0?

vuntz:
Ah, probably not, I think the feature only became available in Ocata...

aplanas:
@vuntz I think that not, is not a feature available in Newton

PR #1218, opened 2017-08-31
Ceilometer: use nova-api for discovery hyp=xen, (bsc#1056142)

sjamgade:
Instead of using libvirt to get metadata about a domain, use nova in
case of xen as it does not support call to DomainGetMetadata

[relevant ceilometer change](https://review.openstack.org/#/c/397654/)

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/common.rb
+++ b/chef/cookbooks/ceilometer/recipes/common.rb
@@ -38,7 +38,11 @@ is_swift_proxy = node.roles.include?("ceilometer-swift-proxy-middleware") && nod
 # Find hypervisor inspector
 hypervisor_inspector = nil
 libvirt_type = nil
+discovery_method = nil
 if is_compute_agent
+  if ["xen"].include?(node[:nova][:libvirt_type])
+    discovery_method = "workload_partitioning"
+  end
   if node.roles.include?("nova-compute-vmware")
     hypervisor_inspector = "vmware"
   else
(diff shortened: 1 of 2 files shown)


sjamgade:
[1056142 ceilometer-polling: libvirt: Domain Config error : this function is not supported by the connection driver: virDomainGetMetadata SUSE OpenStack Cloud 8 Ceilometer](https://trello.com/c/aAY1zRAX/419-1056142-ceilometer-polling-libvirt-domain-config-error-this-function-is-not-supported-by-the-connection-driver-virdomaingetmetad)

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -87,6 +90,7 @@
       libvirt_type: libvirt_type,
       metering_time_to_live: metering_time_to_live,
       event_time_to_live: event_time_to_live,
+      discovery_method: discovery_method
```
toabctl: Isn't ```discovery_method``` uninitialized in the non-xen case?

sjamgade: yes that goes to default ie trying through `libvirt`,
But I am not sure what all hypervisors do not support discovery through libvirt.

aplanas: @sjamgade but tom is right: this will produce and error in ruby because the variable is not defined. You need to set a default value for not-xen deployments

sjamgade: ah, yes.
Handled.

PR #1219, opened 2017-08-31
cinder: volume snapshots for NFS (bsc#1053414)

aplanas:
Add a new parameter in the UI to allow volume snapshots in NFS. This
feature must also disable the parameter `nas_secure_file_operations`.

This is configuration have some security implications[1] and a
warning text is added as a help hint in the UI.

[1] https://docs.openstack.org/security-guide/block-storage/checklist.html#check-block-07-is-nas-operating-in-a-secure-environment


aplanas:
Newton driver do not allow snapshots in NFS, so closing.

PR #1220, opened 2017-09-01
tempest: Set domain specifc drivers identitiy feature flag

toabctl:
The option "domain_specific_drivers" must be enabled when testing an environment
that is configured to use domain-specific identity drivers.
Otherwise the tempest test
tempest.api.identity.admin.v3.test_groups.GroupsV3TestJSON.test_list_groups
fails.


toabctl:
Backport for stable/4.0 is https://github.com/crowbar/crowbar-openstack/pull/1221

PR #1222, opened 2017-09-01
mariadb: Set mariadb as the default database

rsalevsky:
In Cloud 8 we will use maridb as the default and only supported database
for OpenStack. This change sets the database to mysql (mariadb).


rsalevsky:
My idea is to switch all non-HA CI jobs to MariaDB to get some more testing.

rsalevsky:
https://github.com/SUSE-Cloud/automation/pull/2166 needs to be merged first

rsalevsky:
Lets see how it behaves in Jenkins over the weekend.

PR #1223, opened 2017-09-01
Stop exposing passwords in the process table

cmurphy:
Using --os-password in command line calls exposes the password in the
process table and may therefor potentially expose it to unprivileged
users. This patch converts all shell-outs to use the environment
variable version of their credential settings, which is not leaked to
the process table.

We can omit passing '--os-auth-type password' since it is always implied
by the use of '--os-password' or OS_PASSWORD.

For openstackclient commands, if we want to specify a particular
endpoint type, we should use --os-interface/OS_INTERFACE, not
--os-endpoint/OS_ENDPOINT.

Using OS_PROJECT_NAME instead of OS_TENANT_NAME is acceptable in all
cases except for the nova availability zone setter, which is a custom
script.

### DIFF
--- a/chef/cookbooks/glance/recipes/setup.rb
+++ b/chef/cookbooks/glance/recipes/setup.rb
@@ -11,12 +11,15 @@ end
 
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
-glance_args = "--os-username '#{keystone_settings["admin_user"]}'"
-glance_args = "#{glance_args} --os-password '#{keystone_settings["admin_password"]}'"
-glance_args = "#{glance_args} --os-tenant-name '#{keystone_settings["admin_tenant"]}'"
-glance_args = "#{glance_args} --os-auth-url '#{keystone_settings["internal_auth_url"]}'"
-glance_args = "#{glance_args} --os-endpoint-type internalURL"
-glance_args = "#{glance_args} --os-region-name '#{keystone_settings["endpoint_region"]}'"
+env = {
+  "OS_USERNAME" => keystone_settings["admin_user"],
+  "OS_PASSWORD" => keystone_settings["admin_password"],
+  "OS_PROJECT_NAME" => keystone_settings["admin_tenant"],
+  "OS_AUTH_URL" => keystone_settings["internal_auth_url"],
+  "OS_ENDPOINT_TYPE" => "internalURL",
+  "OS_REGION_NAME" => keystone_settings["endpoint_region"]
+}
+env_to_s = env.to_s[1..-2].delete(',', '').gsub('=>', '=').gsub('"', "'")
 
 #
 # Download and install AMIs
@@ -54,7 +57,7 @@ ruby_block "load glance images" do
       Chef::Log.info("Extracting #{name} into #{tmpdir}")
       ::Kernel.system("tar -zxf \"#{rawdir}/#{name}\" -C \"#{tmpdir}/\"")
       if wait
-        ::Kernel.system("glance #{glance_args} image-list")
+        ::Kernel.system(env, "glance image-list")
         sleep 15
         wait = false
       end
@@ -65,7 +68,7 @@ ruby_block "load glance images" do
         ["initrd", "loader"],
         ["image", ".img"]].each do |part|
         next unless image_part = Dir.glob("#{tmpdir}/*#{part[1]}").first
-        cmd = "glance #{glance_args} image-create --name #{basename}-#{part[0]} --public"
+        cmd = "#{env_to_s} glance image-create --name #{basename}-#{part[0]} --public"
         case part[0]
         when "kernel" then cmd << " --disk-format aki --container-format aki"
         when "initrd" then cmd << " --disk-format ari --container-format ari"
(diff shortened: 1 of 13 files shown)


nicolasbock:
Great idea!

rsalevsky:
Should we backport this change?

cmurphy:
@rsalevsky yes probably

chef/cookbooks/glance/recipes/setup.rb
```diff
@@ -11,12 +11,12 @@
...
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
-glance_args = "--os-username '#{keystone_settings["admin_user"]}'"
-glance_args = "#{glance_args} --os-password '#{keystone_settings["admin_password"]}'"
-glance_args = "#{glance_args} --os-tenant-name '#{keystone_settings["admin_tenant"]}'"
-glance_args = "#{glance_args} --os-auth-url '#{keystone_settings["internal_auth_url"]}'"
-glance_args = "#{glance_args} --os-endpoint-type internalURL"
-glance_args = "#{glance_args} --os-region-name '#{keystone_settings["endpoint_region"]}'"
+env = "OS_USERNAME=#{keystone_settings["admin_user"]} "
+env << "OS_PASSWORD=#{keystone_settings["admin_password"]} "
```
dirkmueller: you removed the (insecure) quoting from the command line use before here, can you add it back?

chef/cookbooks/glance/recipes/setup.rb
```diff
@@ -54,7 +54,7 @@
       Chef::Log.info("Extracting #{name} into #{tmpdir}")
       ::Kernel.system("tar -zxf \"#{rawdir}/#{name}\" -C \"#{tmpdir}/\"")
       if wait
-        ::Kernel.system("glance #{glance_args} image-list")
+        ::Kernel.system("#{env} glance image-list")
```
dirkmueller: should be according to https://ruby-doc.org/core-2.2.0/Kernel.html#method-i-system:

```ruby
::Kernel.system(env, "glance image-list")
```
(not sure which type this env needs to be, likely a hash). then the quoting comment from just above is obsolete, as it doesn't get passed through the shell then anymore.

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -361,14 +361,13 @@
...
---os-tenant-name #{keystone_settings["admin_tenant"]} \
---os-auth-url=#{url} \
---os-region-name='#{keystone_settings["endpoint_region"]}' \
---os-identity-api-version=3 \
-#{insecure} \
+    env = "OS_USERNAME=#{keystone_settings["admin_user"]} "
+    env << "OS_PASSWORD=#{keystone_settings["admin_password"]} "
+    env << "OS_PROJECT_NAME=#{keystone_settings["admin_tenant"]} "
+    env << "OS_AUTH_URL=#{url} "
+    env << "OS_REGION_NAME=#{keystone_settings["endpoint_region"]} "
```
dirkmueller: quoting lost here..

chef/cookbooks/ironic/libraries/helpers.rb
```diff
@@ -23,11 +23,11 @@ def auth_url(keystone_settings)
 
     def openstack_command(keystone_settings)
       insecure = keystone_settings["insecure"] ? "--insecure" : ""
-      "openstack --os-username #{keystone_settings["admin_user"]}"\
-      " --os-auth-type password --os-identity-api-version 3"\
-      " --os-password #{keystone_settings["admin_password"]}"\
-      " --os-tenant-name #{keystone_settings["admin_tenant"]}"\
-      " --os-auth-url #{auth_url(keystone_settings)} #{insecure}"
+      env = "OS_USERNAME=#{keystone_settings["admin_user"]} "
+      env << "OS_PASSWORD=#{keystone_settings["admin_password"]} "
```
dirkmueller: quoting needed here..

chef/cookbooks/magnum/recipes/post_install.rb
```diff
@@ -41,15 +41,14 @@
...
 
-# create basic arguments for openstack client
-openstack_args = "--os-username #{keystone_settings["service_user"]}"
-openstack_args += " --os-auth-type password --os-identity-api-version 3"
-openstack_args += " --os-password #{keystone_settings["service_password"]}"
-openstack_args += " --os-tenant-name #{keystone_settings["service_tenant"]}"
-openstack_args += " --os-auth-url #{keystone_settings["internal_auth_url"]}"
-openstack_args += " --os-endpoint internalURL"
+env = "OS_USERNAME=#{keystone_settings["service_user"]} "
+env << "OS_PASSWORD=#{keystone_settings["service_password"]} "
```
dirkmueller: quoting?

chef/cookbooks/swift/recipes/dispersion.rb
```diff
@@ -87,10 +87,16 @@
 end
 
 dispersion_cmd="swift-dispersion-populate"
+
+env = "OS_USERNAME=#{service_user} "
+env << "OS_PASSWORD=#{service_password} "
```
dirkmueller: quoting lost here..

chef/cookbooks/glance/recipes/setup.rb
cmurphy: Done, but also added a string version since this is re-used in a backtick command a few lines down.

cmurphy:
Backport: https://github.com/crowbar/crowbar-openstack/pull/1253

PR #1226, opened 2017-09-04
[4.0] database: Replace MySQL UI string with MariaDB

jsuchome:
We're not using MySQL, but MariaDB as an option for OpenStack
database.

Instead of replacing all relevant mysql strings (including variable
and cookbook names), we go the easy way and only change the part
that is visible to user in crowbar UI.

(cherry picked from commit 49c0782483b8ff81eaf2b689068a4426c7e79c7e)

Backport of https://github.com/crowbar/crowbar-openstack/pull/1225

### DIFF
--- a/crowbar_framework/app/helpers/barclamp/database_helper.rb
+++ b/crowbar_framework/app/helpers/barclamp/database_helper.rb
@@ -20,7 +20,7 @@ module Barclamp
     def engines_for_database(selected)
       options_for_select(
         [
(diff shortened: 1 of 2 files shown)


jsuchome:
[Replace MySQL string with mariadb in the crowbar UI](https://trello.com/c/aJ43hlqO/39-replace-mysql-string-with-mariadb-in-the-crowbar-ui)

PR #1233, opened 2017-09-05
nova: add rate limiting options to raw template

Itxaka:
Adds rate limiting options to the raw template for the different
services that depend from nova. Defaults to 0 as the values, which means
no rate limiting

Depends-on: https://github.com/crowbar/crowbar-ha/pull/228

### DIFF
--- a/chef/data_bags/crowbar/template-nova.schema
+++ b/chef/data_bags/crowbar/template-nova.schema
@@ -246,6 +246,15 @@
                   }
                 }
               }
+            },
+            "ha_rate_limit": {
+              "type": "map", "required": true, "mapping": {
+                "nova-api": { "type": "int", "required": true },
+                "nova-placement-api": { "type": "int", "required": true },
+                "nova-metadata": { "type": "int", "required": true },
+                "nova-novncproxy": { "type": "int", "required": true },
+                "nova-serialproxy": { "type": "int", "required": true }
+              }
             }
           }
         }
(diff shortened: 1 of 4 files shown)


Itxaka:
[Nova OpenStack API rate Limit](https://trello.com/c/GsZ6iHh9/40-nova-openstack-api-rate-limit)

toabctl:
@Itxaka this only works in the HA case. Should we name the rate_limit var ```haproxy_rate_limit``` to make it clear that this is a HA-only thing?

Itxaka:
@toabctl thats a good point. Maybe we should do that yes, so people dont set it on non-ha environments thinking that its gonna do something.

Let me rework this, sorry @cmurphy @dirkmueller

Itxaka:
Looks like even with spaces Im missing something in the schema file
```
Failed with exception: file: crowbar/template-nova.schema, line 251: flow mapping is not closed by '}'.
```

chef/data_bags/crowbar/template-nova.schema
```diff
@@ -246,6 +246,15 @@
                   }
                 }
               }
+            },
+            "ha_rate_limit": {
+              "type": "map": "required": true, "mapping": {
```
Itxaka: huuur, durrrrrr, using `:` instead of a comma its a good way of making this fail.

vuntz:
For the record, I'm re-triggering the ha job to also run with the patch from https://github.com/crowbar/crowbar-ha/pull/228 (which is required).

dirkmueller:
assuming this wasn't actually "do not merge yet" as it got merged by @rsalevsky

Itxaka:
@dirkmueller the label was there to wait for the parent patch to be merged and I forgot to remove it, so it should be ok.

PR #1234, opened 2017-09-05
mysql: Add a timeout to galera bootstrapping

jsuchome:
If bootstrapping fails, we have to raise some error, not wait in the endless loop.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -189,15 +189,28 @@ end
 # Which might take a while. Wait for that to complete by watching the
 # "wsrep_local_state_comment" status variable on all cluster nodes to reach
 # the "Synced" state, before continuing with the rest of the recipe.
-script "wait galera bootstrap" do
-  interpreter "bash"
-  code <<-EOH
-    sync_state=""
-    while [[ $sync_state != "Synced" ]]; do
-      sleep 1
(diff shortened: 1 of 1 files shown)


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -188,15 +188,27 @@
...
-  EOH
+ruby_block "wait galera bootstrap" do
+  seconds = 300
+  block do
+    require "timeout"
+    begin
+      cmd = "mysql -u '' -N -B -e \"SHOW STATUS WHERE Variable_name='wsrep_local_state_comment';\""
+      sync_state = ""
+      Timeout.timeout(seconds) do
+        while sync_state !~ /Synced/
```
rhafer: In general I am not a big fan of using pattern matching where it's easy to get a more accurate result with just a few basic string ops (`get_state.stdout.strip.split("\t") ...`, or the cut pipe that was used before), but I guess I can live with that.

jsuchome: Yes, I've had that pipe here before.
The only reason I changed it was hound complaining about line (with the command) being too long... :-)

s-t-e-v-e-n-k: I personally still like the shell version -- fixing the hound complaint is as easy as moving the SHOW SQL into a variable, and adding a counter and calling exit 1 if $count -gt 300?

jsuchome: Why? 
Here we got standard timeout library with easy error reporting.

Anyway, I'll remove the regexp to please you guys.

PR #1187, opened 2017-08-22
Rabbit Active-Active

matelakat:
Implement Active-Active RabbitMQ configuration

This removes the need for shared storage and also enables active/active HA as well as some load balancing.

This is the re-based version of #1104

PR #1235, opened 2017-09-05
Fetch HA resource name for rabbitmq from rabbitmq settings

dirkmueller:
This is in preparation to the HA support in RabbitMQ using
durable queues, which requires the pacemaker resource name
to change away from rabbitmq to ms-rabbitmq.

this is a split from #1187 in order to make that one smaller.


chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -77,6 +77,9 @@ lock_path = /var/run/neutron
 driver = neutron.openstack.common.notifier.rpc_notifier
 
 [oslo_messaging_rabbit]
+amqp_durable_queues = <%= @rabbit_settings[:durable_queues] %>
+rabbit_ha_queues = <%= @rabbit_settings[:ha_queues] %>
+<% end -%>
```
vuntz: Bad rebase.

chef/cookbooks/barbican/templates/default/barbican.conf.erb
```diff
@@ -3,10 +3,13 @@
 debug=<%= node[:barbican][:debug] %>
 host_href = <%= @host_href %>
 sql_connection = <%= @database_connection %>
+<% if @rabbit_settings[:cluster] -%>
+transport_url = <%= @rabbit_settings[:url] %>
```
toabctl: Don't we need the ```transport_url``` also in non-cluster mode? I wonder why that was not set previously...

chef/cookbooks/barbican/templates/default/barbican.conf.erb
```diff
@@ -3,10 +3,13 @@
...
+<% end -%>
 
 [oslo_messaging_rabbit]
-
-amqp_durable_queues = False
+amqp_durable_queues = <%= @rabbit_settings[:durable_queues] %>
```
toabctl: Should this be in a ```<% if @rabbit_settings[:cluster] -%>``` block?

chef/cookbooks/keystone/templates/default/keystone.conf.erb
```diff
@@ -84,11 +84,16 @@ auth_pool_size = <%= node[:keystone][:ldap][:auth_pool_size] %>
 auth_pool_connection_lifetime = <%= node[:keystone][:ldap][:auth_pool_connection_lifetime] %>
 
 [oslo_messaging_rabbit]
+<% if @rabbit_settings[:cluster] -%>
+amqp_durable_queues = <%= @rabbit_settings[:durable_queues] %>
+rabbit_ha_queues = <%= @rabbit_settings[:ha_queues] %>
+<% else -%>
```
toabctl: Why an ```else``` here? All other ```[oslo_messaging_rabbit]``` don't have this.

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -384,6 +384,7 @@
     lazy {
       {
         debug: node[:heat][:debug],
+        verbose: node[:heat][:verbose],
```
toabctl: Unrelated to this commit

toabctl:
Commit 0d72909b677d37ad7ecd7caaa8f2d573189762f8 can be squashed into the previous one

chef/cookbooks/barbican/templates/default/barbican.conf.erb
dirkmueller: yeah, that looks like a mistake indeed. @jgrassler @sayalilunkad any comments?

matelakat: So the original idea behind introducing this `<% if @rabbit_settings[:cluster] -%>` stanza everywhere is coming from the times when we wanted this to be an update to an already deployed product, because in those cases configuration file changes would cause re-start of services. Given that we are aiming for only new installations for C8 this extra logic can be skipped.

chef/cookbooks/barbican/templates/default/barbican.conf.erb
matelakat: It really depends what do we decide - if we don't care about update as we only support new installations, I think getting rid of all `<% if @rabbit_settings[:cluster] -%>` lines make sense. If we want to make backport easier, we should have `<% if @rabbit_settings[:cluster] -%>` everywhere as you pointed out.

chef/cookbooks/neutron/templates/default/neutron.conf.erb
matelakat: what do you mean on this?

matelakat:
@dirkmueller I would squash 04c91c027892b3a5801adf280613a260f31b35fd and 0d72909b677d37ad7ecd7caaa8f2d573189762f8
Triggering a re-build

chef/cookbooks/trove/libraries/helpers.rb
```diff
@@ -15,21 +15,6 @@
 #
 
 module TroveHelper
-  def self.get_rabbitmq_trove_url(node, rabbitmq_servers)
-    # get rabbitmq-server information
-    # NOTE: Trove uses it's own user/pass/vhost instead of the default one
```
toabctl: @dirkmueller have you seens this comment? Just using the default rabbit_settings seems not to work for Trove.

matelakat: That's a good point

vuntz: @toabctl it's fine because `@rabbit_settings[:trove_url]` is used in the templates

matelakat:
If we squash those 2 commits it makes the cluster variable not needed as well.

matelakat:
[RabbitMQ Active-Active setup (No HAProxy)](https://trello.com/c/aaR7h4q1/8-rabbitmq-active-active-setup-no-haproxy)

PR #1238, opened 2017-09-06
[4.0] Fetch HA resource name for rabbitmq from rabbitmq settings

matelakat:
This is in preparation to the HA support in RabbitMQ using
durable queues, which requires the pacemaker resource name
to change away from rabbitmq to ms-rabbitmq.

This is the backport of #1235


matelakat:
We would still need to think about how the updates will go - so if someone installs this change it won't cause re-starts across all of her services.

matelakat:
[RabbitMQ Active-Active backport](https://trello.com/c/186xQCwn/73-rabbitmq-active-active-backport)

matelakat:
The last commit is addressing backport issues

matelakat:
> @matelakat The first 5 commits change nothing for a user. It is just preparation work. I don't see why we need the last commit you added. So please remove the last commit.

@toabctl As the commit message says, we need it so **nothing** changes in the rendered files. Imagine a scenario where people have a running cloud. They decide to install an update. Given that the update is  making modifications to all of the configuration files - even if that's only a new-line character, it will cause the services to be reloaded/restarted. The purpose of the last commit is to prevent this unwanted restart.

matelakat:
I need to re-do this as the commit references have changed since on the original PR

matelakat:
Here is the output of `git diff origin/stable/4.0 -- $(git diff origin/stable/4.0 --name-only  | grep -e 'erb$')`

http://pastebin.suse.de/26562

PR #1239, opened 2017-09-06
nova: remove/rename config options deprecated in Pike

stefannica:
The following nova configuration options deprecated in Pike
are removed or renamed:

 - [DEFAULT]/default_floating_pool has been deprecated for neutron
users in favor of [neutron]/default_floating_pool [1]
 - the following [ssl] options are moved to the [glance] group:
ca_file, cert_file and key_file and renamed to cafile, certfile
and keyfile respectively (these options were only used by the
Nova code that interacts with the Glance client) [1]
 - api_insecure option from [glance] group is renamed to insecure [1]
 - the [DEFAULT]/firewall_driver is deprecated as its default
value has been changed to nova.virt.firewall.NoopFirewallDriver
to coincide with the default value of [DEFAULT]/use_neutron=True [1]
 - the --version optional parameter accepted by the
'nova-manage api_db sync' and 'nova-manage db sync' commands
is now a positional argument [1]
 - Nova is now configured to use the v3 version of the Cinder API.
Using the v2 version is deprecated [1]
 - the RamFilter entry is being removed from the enabled_filters
value, since the placement service is used to verify RAM resources
when using FilterScheduler [1]
 - the following nova-network [DEFAULT] group options are deprecated:
default_floating_pool, firewall_driver, dhcpbridge, linuxnet_interface_driver [1]
 - the [oslo_messaging_rabbit]/rabbit_use_ssl option is deprecated
in favor of the ssl option under the same group [2]

More information:
 - [1] https://docs.openstack.org/releasenotes/nova/pike.html
 - [2] https://review.openstack.org/#/c/438455/

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -43,13 +43,9 @@ compute_driver = libvirt.LibvirtDriver
 <% if @libvirt_type.eql?('xen') %>
 use_cow_images = false
 <% end %>
-firewall_driver = nova.virt.firewall.NoopFirewallDriver
 <% if @ssl_cert_required %>
 ssl_ca_file = <%= @ssl_ca_file %>
 <% end %>
-default_floating_pool = floating
-dhcpbridge = <%= @dhcpbridge %>
-linuxnet_interface_driver = ""
 <% if @force_config_drive || @libvirt_type.eql?('zvm') %>
 flat_injected = true
 <% end %>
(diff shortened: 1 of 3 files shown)


chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -261,11 +263,6 @@ serialproxy_host = 0.0.0.0
 serialproxy_port = 6083
 
 [ssl]
```
cmurphy: Does this section header need to stick around?

stefannica: no :)

stefannica:
rebased

PR #1240, opened 2017-09-06
keystone: Update for pike (SCRD-781)

cmurphy:
Remove the [signing] section which is only applicable to the now-defunct PKI token format, and remove the [resource]/driver setting.

### DIFF
--- a/chef/data_bags/crowbar/migrate/keystone/202_remove_signing.rb
+++ b/chef/data_bags/crowbar/migrate/keystone/202_remove_signing.rb
@@ -1,0 +1,16 @@
+def upgrade(ta, td, a, d)
+  a["token_format"] = a["signing"]["token_format"]
+  a.delete("signing")
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["signing"] = {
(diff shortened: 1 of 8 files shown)


chef/data_bags/crowbar/template-keystone.json
```diff
@@ -178,7 +173,7 @@
     "keystone": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 200,
+      "schema-revision": 202,
```
cmurphy: This presumes #1001 goes in first

chef/data_bags/crowbar/migrate/keystone/202_remove_signing.rb
```diff
@@ -0,0 +1,16 @@
+def upgrade(ta, td, a, d)
+  a["token_format"] = a["signing"]["token_format"]
+  a.delete("signing")
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["signing"] = {
```
dirkmueller: I believe you could just do 

```ruby
a["signing"] = ta["signing"]
a["signing"]["token_format"] = a["token_format"]
```

PR #1241, opened 2017-09-07
keystone: Increase timeout for keystone_register actions

jsuchome:
In various tests, actions happening inside keystone_register syncmark
block tend to take more than 40 secods. Depending on the setup,
it could be even little bit over 1 minute which is the default value
when the waiting nodes time out.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -454,7 +454,11 @@ end
 # the keystone resources
 crowbar_pacemaker_sync_mark "sync-keystone_before_register" if ha_enabled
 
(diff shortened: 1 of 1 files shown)


jsuchome:
@rhafer yesterday I've seen it succeed with mariadb and fail with postgres, so it's not that much depending on the DB backend

jsuchome:
@cmurphy nothing that much specific, sorry. I refer to the manual mkcloud runs I've run this week.

These are the actions taking about a minute: http://pastebin.nue.suse.com/26578/src

jsuchome:
Also here is the same problem I think https://ci.suse.de/job/openstack-mkcloud/82459/

rhafer:
@cmurphy I think this can largely be attributed to overloaded test workers :/

jsuchome:
rebased

PR #1243, opened 2017-09-07
nova: respect image_cache_manager_interval set in proposal (bsc#1057233)

dirkmueller:
The parameter was made configuable but it wasn't actually written
to the config file.

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -18,7 +18,7 @@ metadata_listen = <%= @metadata_bind_address %>
 metadata_listen_port = <%= @bind_port_metadata %>
 metadata_workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
 instance_usage_audit_period = hour
-image_cache_manager_interval = -1
+image_cache_manager_interval = <%= node[:nova][:image_cache_manager_interval] %>
 <% if @libvirt_type.eql?('kvm') %>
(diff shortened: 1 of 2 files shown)


sayalilunkad:
We also need to set the default value. I would suggest changing the default to 0 instead of -1 so it is not disabled by default.

cmurphy:
Letting it default to 0 conflicts with the reason it was added in the first place https://github.com/crowbar/crowbar-openstack/pull/861

sayalilunkad:
@cmurphy yes but @bmwiedemann  and I discussed this and we think it should enable it by default and when we want to migrate then set it to -1. This seems like a better approach for this setting. And since it was only a internal bug it should be ok to change this.

dirkmueller:
@sayalilunkad the point was that it needs to be set that way way before migration iirc to avoid the migration breaking at a later point in time. at least thats how it used to be before.

sayalilunkad:
@dirkmueller yes that was the reason to disable it but it sees to be not so convenient for the operators. Either ways I guess we will need to set a default. I'm fine with -1 or 0.

bmwiedemann:
I'd vote for a default of 0 (enabling cache manager), because the old https://bugzilla.suse.com/show_bug.cgi?id=1015324 was supposed to only happen with shared NFS backing storage
and only when an extended network outage happened (24+h) and thus was hard to trigger.

bmwiedemann:
@dirkmueller I added a commit with a lengthy explanation to properly fix that regression.

PR #1244, opened 2017-09-07
keystone: Set an origin flag on apache2 restart

Itxaka:
When we do a apache2 restart action, as the apache2 service is declared
on the apache2 cookbook, we have no idea of what triggered the restart
for it.

This makes it a problem to differentiate between the different cookbooks
calling the restart for the service, if we want to introduce a feature
to avoid restarts based on origin/barclamps.

This PR introduces a ruby_block that sets a temporal key on the node
when apache2 is restarted, to flag where the restart was triggered from.

The run_state values are ephemeral so we dont modify the node attributes
at all.

Only affected cookbook is keystone for the moment, as the other cookbooks do a reload instead of a restart.

Unblocks: This unblocks https://github.com/crowbar/crowbar-core/pull/1307
Required-By: https://github.com/crowbar/crowbar-core/pull/1307

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -63,6 +63,15 @@ memcached_servers = MemcachedHelper.get_memcached_servers(
 
 memcached_instance "keystone"
 
+# resource to set a flag when apache2 is restarted so we now which cookbook was the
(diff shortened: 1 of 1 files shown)


Itxaka:
[Neutron: Avoid Chef re-starting services](https://trello.com/c/BwtWivrg/67-neutron-avoid-chef-re-starting-services)

PR #1245, opened 2017-09-07
neutron: Fix metadata server for ocata again

dirkmueller:
The recent PR #1229 introduced config options that don't work
in Ocata. Revert to an Ocata compatible code.


dirkmueller:
@stefannica yeah, but it helps if we don't merge patches that don't pass CI (like last time)

PR #1248, opened 2017-09-08
ceilometer: fix libvirt_type & host_ip null error

charleswang007:
mkcloud deployed with cloudsource=pikecloud8 has below error in controller:

/var/log/ceilometer/ceilometer-collector.log (cotyledon._service_manager [-] ServiceManager raised exception during new_worker hooks: **ConfigFileValueError: Value for option libvirt_type is not valid: Valid values are [kvm, lxc, qemu, uml, xen], but found ''**)
/var/log/ceilometer/ceilometer-polling-central.log (cotyledon._service_manager [-] ServiceManager raised exception during new_worker hooks: **ConfigFileValueError: Value for option libvirt_type is not valid: Valid values are [kvm, lxc, qemu, uml, xen], but found ''**)
/var/log/ceilometer/ceilometer-polling-central.log (cotyledon._service_manager **ConfigFileValueError: Value for option host_ip is not valid:  is not a valid host address**)

Local tested the fix. **_ConfigFileValueError_** is now gone in ceilometer logs.

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/common.rb
+++ b/chef/cookbooks/ceilometer/recipes/common.rb
@@ -38,12 +38,15 @@ is_swift_proxy = node.roles.include?("ceilometer-swift-proxy-middleware") && nod
 # Find hypervisor inspector
 hypervisor_inspector = nil
 libvirt_type = nil
-discovery_method = nil
+instance_discovery_method = "libvirt_metadata"
 if is_compute_agent
   if ["xen"].include?(node[:nova][:libvirt_type])
-    discovery_method = "workload_partitioning"
+    instance_discovery_method = "workload_partitioning"
+    hypervisor_inspector = "libvirt"
+    libvirt_type = node[:nova][:libvirt_type]
   end
-  if node.roles.include?("nova-compute-vmware")
+  if ["vmware"].include?(node[:nova][:libvirt_type])
+    instance_discovery_method = "workload_partitioning"
     hypervisor_inspector = "vmware"
   else
     hypervisor_inspector = "libvirt"
@@ -91,7 +94,8 @@ template node[:ceilometer][:config_file] do
       libvirt_type: libvirt_type,
       metering_time_to_live: metering_time_to_live,
       event_time_to_live: event_time_to_live,
-      discovery_method: discovery_method
+      instance_discovery_method: instance_discovery_method,
+      is_compute_agent: is_compute_agent
     )
     if is_compute_agent
       notifies :restart, "service[nova-compute]"
--- a/chef/cookbooks/ceilometer/recipes/common.rb
+++ b/chef/cookbooks/ceilometer/recipes/common.rb
@@ -47,7 +47,7 @@ if is_compute_agent
   end
   if ["vmware"].include?(node[:nova][:libvirt_type])
     instance_discovery_method = "workload_partitioning"
-    hypervisor_inspector = "vmware"
+    hypervisor_inspector = "vsphere"
   else
     hypervisor_inspector = "libvirt"
     libvirt_type = node[:nova][:libvirt_type]
--- a/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
+++ b/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
@@ -61,7 +61,7 @@ password = <%= @keystone_settings['service_password'] %>
 [service_types]
 neutron_lbaas_version = v2
 
-<% if @hypervisor_inspector == "vmware" -%>
+<% if @hypervisor_inspector == "vsphere" -%>
 [vmware]
 host_ip = <%= node[:nova][:vcenter][:host] rescue "" %>
 host_port = <%= node[:nova][:vcenter][:port] rescue "443" %>
--- a/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
+++ b/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
@@ -1,7 +1,9 @@
 [DEFAULT]
 shuffle_time_before_polling_task = 15
+<% if @is_compute_agent -%>
 hypervisor_inspector = <%= @hypervisor_inspector %>
 libvirt_type = <%= @libvirt_type %>
+<% end -%>
 host = <%= @node_hostname %>
 debug = <%= @debug ? "true" : "false" %>
 log_dir = /var/log/ceilometer
@@ -59,16 +61,18 @@ password = <%= @keystone_settings['service_password'] %>
 [service_types]
 neutron_lbaas_version = v2
 
+<% if @hypervisor_inspector == "vmware" -%>
 [vmware]
 host_ip = <%= node[:nova][:vcenter][:host] rescue "" %>
 host_port = <%= node[:nova][:vcenter][:port] rescue "443" %>
 host_username = <%= node[:nova][:vcenter][:user] rescue "" %>
 host_password = <%= node[:nova][:vcenter][:password] rescue "" %>
-<% if @hypervisor_inspector == "vmware" && !node[:nova][:vcenter][:ca_file].empty? -%>
+<% if !node[:nova][:vcenter][:ca_file].empty? -%>
 ca_file = <%= node[:nova][:vcenter][:ca_file] -%>
 <% end -%>
-<% if @hypervisor_inspector == "vmware" && node[:nova][:vcenter][:insecure] -%>
+<% if node[:nova][:vcenter][:insecure] -%>
 insecure = <%= node[:nova][:vcenter][:insecure] %>
+<% end -%>
 <% end -%>
 
 [oslo_messaging_rabbit]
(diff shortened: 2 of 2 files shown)


chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -37,14 +37,16 @@
...
+libvirt_type = "kvm"
 discovery_method = nil
 if is_compute_agent
   if ["xen"].include?(node[:nova][:libvirt_type])
     discovery_method = "workload_partitioning"
+    libvirt_type = "xen"
   end
   if node.roles.include?("nova-compute-vmware")
     hypervisor_inspector = "vmware"
+    libvirt_type = node[:nova][:libvirt_type]
```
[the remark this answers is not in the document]
nicolasbock: Please listen to hound :smile:

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -37,14 +37,16 @@
 
 # Find hypervisor inspector
 hypervisor_inspector = nil
-libvirt_type = nil
+libvirt_type = "kvm"
```
stefannica: The `hypervisor_inspector`, `libvirt_type` and `instance_discovery_method` options only apply to ceilometer-polling services running in the compute namespace. `hypervisor_inspector` identifies one of the supported compute inspectors (libvirt, hyperv, vmware, xenapi or powervm). `libvirt_type` is only used with the libvirt inspector and identifies the libvirt domain type (kvm', 'lxc', 'qemu', 'uml' or 'xen'). `instance_discovery_method` controls the method used to discover new instances (libvirt_metadata - only for libvirt-based computes, naive, or workload_partitioning).

You need to configure these options properly depending on the `node[:nova][:libvirt_type]` value:
- if the chef recipe is not running on a "compute agent" node (i.e is_compute_agent is False), then neither option should appear in the configuration file (see how this is done if discovery_method is nil in the ceilometer.conf.erb file).
- otherwise, the options should be set to their correct values corresponding to the type of compute node that the recipe is running on:
  - for a qemu, kvm or xen compute node, the libvirt inspector is used (hypervisor_inspector=libvirt, libvirt_type=node[:nova])
  - for a vmware compute node, the vsphere inspector is used (current code value is wrong) (hypervisor_inspector=vsphere), while the libvirt_type option no longer applies and shouldn't be configured at all
  - for all other supported compute nodes (ironic, zvm), I have no idea how this should be configured, so we should keep the current behavior unchanged

Note that the `instance_discovery_method` option default value (libvirt_metadata) is only correct if the libvirt inspector is also used, otherwise the 'workload_partitioning' value should be used.

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
```diff
@@ -60,7 +60,7 @@ password = <%= @keystone_settings['service_password'] %>
 neutron_lbaas_version = v2
 
 [vmware]
-host_ip = <%= node[:nova][:vcenter][:host] rescue "" %>
```
stefannica: Likewise here, this whole block should only be added to the configuration file if hypervisor_inspector==vsphere, as covered above. This will solve problems with empty options that aren't actually needed.

chef/cookbooks/ceilometer/recipes/common.rb
stefannica: Forgot to mention that the compute node type can be identified either by the node[:nova][:libvirt_type] value of the node roles (e.g. node.roles.include?("nova-compute-vmware")). The two methods are equivalent.

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -38,14 +38,17 @@
 # Find hypervisor inspector
 hypervisor_inspector = nil
 libvirt_type = nil
-discovery_method = nil
+instance_discovery_method = "workload_partitioning"
 if is_compute_agent
-  if ["xen"].include?(node[:nova][:libvirt_type])
-    discovery_method = "workload_partitioning"
```
stefannica: You have accidentally removed the xen case, where instance_discovery_method must be "workload_partitioning". In fact, I think it's easier if you just initialize instance_discovery_method to libvirt_metadata instead, and just address the corner cases of xen and vmware separately:

```
if ["xen", "vmware"].include?(node[:nova][:libvirt_type])
    instance_discovery_method = "workload_partitioning"
end

```

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
```diff
@@ -79,7 +79,11 @@ rabbit_use_ssl = <%= @rabbit_settings[:use_ssl] %>
 kombu_ssl_ca_certs = <%= @rabbit_settings[:client_ca_certs] %>
 <% end -%>
 
-<% if @discovery_method -%>
 [compute]
+<% if @discovery_method -%>
 instance_discovery_method = <%= @discovery_method %>
 <% end -%>
+<% if @is_compute_agent -%>
```
stefannica: Having these options conditionally added to the configuration file depending on `is_compute_agent` is indeed a great idea, but:
- if you decide to reference a new variable in the template, you need to pass it to the template resource (e.g. the same way the libvirt_type variable is passed [here](https://github.com/charleswang007/crowbar-openstack/blob/b1e69db243420044a89a61c82599a006de971e90/chef/cookbooks/ceilometer/recipes/common.rb#L94))
- the hypervisor_inspector and libvirt_type configuration options belong to the DEFAULT group, not to the compute group, even if they are consumed only by the compute namespace polling agents. You can find the documentation on ceilometer configuration options [here](https://docs.openstack.org/ceilometer/pike/configuration/).

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
```diff
@@ -60,14 +58,16 @@ password = <%= @keystone_settings['service_password'] %>
 neutron_lbaas_version = v2
 
 [vmware]
+<% if @hypervisor_inspector == "vsphere" -%>
```
stefannica: I think you should also include the [vmware] group in this conditional block, otherwise it will look rather silly on non-vmware compute nodes.

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
```diff
@@ -60,14 +58,16 @@ password = <%= @keystone_settings['service_password'] %>
...
 host_ip = <%= node[:nova][:vcenter][:host] rescue "" %>
 host_port = <%= node[:nova][:vcenter][:port] rescue "443" %>
 host_username = <%= node[:nova][:vcenter][:user] rescue "" %>
 host_password = <%= node[:nova][:vcenter][:password] rescue "" %>
-<% if @hypervisor_inspector == "vmware" && !node[:nova][:vcenter][:ca_file].empty? -%>
```
stefannica: Likewise here, you can extend the conditional block to include the rest of the vmware configuration options - ca_file and insecure. This will also help you simplify the conditions associated with them.

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
```diff
@@ -79,7 +79,11 @@ rabbit_use_ssl = <%= @rabbit_settings[:use_ssl] %>
 kombu_ssl_ca_certs = <%= @rabbit_settings[:client_ca_certs] %>
 <% end -%>
 
-<% if @discovery_method -%>
 [compute]
+<% if @discovery_method -%>
```
stefannica: Two things are problematic with this change:
 - you already renamed the input discovery_method parameter to instance_discovery_method [here](https://github.com/charleswang007/crowbar-openstack/blob/b1e69db243420044a89a61c82599a006de971e90/chef/cookbooks/ceilometer/recipes/common.rb#L97), so you need to be consistent in your choices
 - you already chose a non-nil default value for instance_discovery_method variable [here](https://github.com/charleswang007/crowbar-openstack/blob/b1e69db243420044a89a61c82599a006de971e90/chef/cookbooks/ceilometer/recipes/common.rb#L41), which makes this condition useless. Again, you need to be consistent in your choices. I would use `<% if @is_compute_agent -%>` here instead.

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -38,13 +38,16 @@
...
+    instance_discovery_method = "workload_partitioning"
+    hypervisor_inspector = "libvirt"
+    libvirt_type = node[:nova][:libvirt_type]
   end
-  if node.roles.include?("nova-compute-vmware")
-    hypervisor_inspector = "vmware"
+  if ["vmware"].include?(node[:nova][:libvirt_type])
+    instance_discovery_method = "workload_partitioning"
+    hypervisor_inspector = "vsphere"
```
stefannica: Could you please separate the "vmware" to "vsphere" change in a separate commit, so we can back-port it to SOC7 ? Ideally, the "vsphere" commit would come first, then the rest of the changes in a second commit.
You can use `git reset HEAD~1` followed by `git add -p` in some cases to speed this up.

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -38,13 +38,16 @@
 # Find hypervisor inspector
 hypervisor_inspector = nil
 libvirt_type = nil
-discovery_method = nil
+instance_discovery_method = "libvirt_metadata"
 if is_compute_agent
   if ["xen"].include?(node[:nova][:libvirt_type])
-    discovery_method = "workload_partitioning"
+    instance_discovery_method = "workload_partitioning"
+    hypervisor_inspector = "libvirt"
```
stefannica: You can safely remove this line and the next one, given that they repeat in the else clause below anyway.

PR #1250, opened 2017-09-08
mariadb: Increase timeout for the promote operation

rhafer:
Promoting the galera resource to master might require a full State
Transfer (SST) which can be time consuming. Increase the default timeout
for that to 300 seconds.

Partial-Bug: #1057806


rhafer:
Haven't tested this yet. And it might make sense to turn this into chef attributes at least.

PR #1254, opened 2017-09-08
keystone: Make password hash algorithm configurable

toabctl:
Since Pike, the new default password hash algorithm is
"bcrypt"[1]. This is more secure but also significantly slower (factor 10
in my tests) when eg. trying to get a new token via the v3 API.
To be able to increase performance (if wanted), make the hashing
algorithm and the hashing rounds for the selected algorithm
configurable. Reducing the value for "password_hash_rounds" increases
the performance.

[1]
https://docs.openstack.org/releasenotes/keystone/pike.html#new-features

### DIFF
--- a/crowbar_framework/app/models/keystone_service.rb
+++ b/crowbar_framework/app/models/keystone_service.rb
@@ -87,6 +87,25 @@ class KeystoneService < OpenstackServiceObject
       validation_error I18n.t("barclamp.#{@bc_name}.validation.enable_keystone_api")
     end
 
+    # validate the password hash configuration
+    if proposal["attributes"][@bc_name]["identity"]["password_hash_rounds"]
+      password_hash_rounds = proposal["attributes"][@bc_name]["identity"]["password_hash_rounds"]
+      case proposal["attributes"][@bc_name]["identity"]["password_hash_algorithm"]
+      when "bcrypt"
+        if password_hash_rounds < 4 || password_hash_rounds > 31
+          validation_error I18n.t("barclamp.#{@bc_name}.validation.bcrypt_password_hash_rounds")
+        end
+      when "scrypt"
+        if password_hash_rounds < 1 || password_hash_rounds > 31
+          validation_error I18n.t("barclamp.#{@bc_name}.validation.scrypt_password_hash_rounds")
+        end
+      when "pbkdf_sha512"
+        if password_hash_rounds < 1 || password_hash_rounds > (1 << 32) - 1
+          validation_error I18n.t("barclamp.#{@bc_name}.validation.pbkdf_sha512_password_hash_rounds")
+        end
+      end
+    end
+
     keystone_timeout = proposal["attributes"]["keystone"]["token_expiration"]
     horizon_proposal = Proposal.find_by(barclamp: "horizon", name: "default")
     unless horizon_proposal.nil?
--- a/crowbar_framework/config/locales/keystone/en.yml
+++ b/crowbar_framework/config/locales/keystone/en.yml
@@ -47,3 +47,6 @@ en:
         api_version: 'API version %{api_version} not recognized.'
         enable_keystone_api: 'Keystone API version 3 needs to be enabled in order to use domain specific drivers.'
         timeout: 'Setting the Horizon timeout (%{horizon_timeout} minutes) longer than the Keystone token expiration timeout (%{keystone_timeout} minutes) is not supported.'
+        bcrypt_password_hash_rounds: "Password hash rounds for bcrypt must be between 4 and 31"
+        scrypt_password_hash_rounds: "Password hash rounds for scrypt must be between 1 and 31"
+        pbkdf_sha512_password_hash_rounds: "Password hash rounds for pbkdfsha512 must be between 1 and 4294967295"
--- a/chef/cookbooks/keystone/templates/default/keystone.conf.erb
+++ b/chef/cookbooks/keystone/templates/default/keystone.conf.erb
@@ -20,6 +20,10 @@ idle_timeout = <%= @sql_idle_timeout %>
 max_active_keys = <%= @max_active_keys %>
 
 [identity]
+password_hash_algorithm = <%= node[:keystone][:identity][:password_hash_algorithm] %>
+<% if node[:keystone][:identity].key?("password_hash_rounds") -%>
+password_hash_rounds = <%= node[:keystone][:identity][:password_hash_rounds] %>
+<% end -%>
 domain_specific_drivers_enabled = <%= node[:keystone][:domain_specific_drivers] ? "True" : "False" %>
 
 <% if node[:keystone][:domain_specific_drivers] -%>
--- a/chef/data_bags/crowbar/migrate/keystone/203_add_password_hash_algorithm.rb
+++ b/chef/data_bags/crowbar/migrate/keystone/203_add_password_hash_algorithm.rb
@@ -1,0 +1,12 @@
+def upgrade(ta, td, a, d)
+  a["identity"]["password_hash_algorithm"] = ta["identity"]["password_hash_algorithm"]
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["identity"].delete("password_hash_algorithm")
+  if a["identity"].key?("password_hash_rounds")
+    a["identity"].delete("password_hash_rounds")
+  end
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-keystone.json
+++ b/chef/data_bags/crowbar/template-keystone.json
@@ -54,7 +54,8 @@
         "password": "crowbar"
       },
       "identity": {
-        "driver": "sql"
+        "driver": "sql",
+        "password_hash_algorithm": "bcrypt"
       },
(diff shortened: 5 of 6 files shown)


toabctl:
Crowbar code not yet tested. For testing the performance, I used https://gist.github.com/toabctl/0dba262b9422c3d92e044dc83048db5a together with:
```
PYTHONPATH=. python -m timeit -n 100 -r 2 -s "from keystone_token_get import do_get" "do_get()"
```

With the Ocata code, this results in:
```
100 loops, best of 2: 61.3 msec per loop
```

With the Pike code (brcypt):
```
100 loops, best of 2: 467 msec per loop
```

And with Pike, (brypt, but ```password_hash_rounds=4```):
```
100 loops, best of 2: 54.8 msec per loop
```

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -87,6 +87,25 @@ def validate_proposal_after_save(proposal)
       validation_error I18n.t("barclamp.#{@bc_name}.validation.enable_keystone_api")
     end
 
+    # validate the password hash configuration
+    if proposal["attributes"][@bc_name]["identity"]["password_hash_rounds"]
+      password_hash_rounds = proposal["attributes"][@bc_name]["identity"]["password_hash_rounds"]
+      case proposal["attributes"][@bc_name]["identity"]["password_hash_algorithm"]
+      when "bcrypt"
+        if password_hash_rounds < 4 || password_hash_rounds > 31
+          validation_error I18n.t("barclamp.#{@bc_name}.validation.bcrypt_password_hash_rounds")
```
nicolasbock: Where are the hash rounds documented?

toabctl: In the sample config. The code for that can be found here: https://github.com/openstack/keystone/blob/master/keystone/conf/identity.py#L120

nicolasbock: Thanks!

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -87,6 +87,25 @@ def validate_proposal_after_save(proposal)
       validation_error I18n.t("barclamp.#{@bc_name}.validation.enable_keystone_api")
     end
 
+    # validate the password hash configuration
+    if proposal["attributes"][@bc_name]["identity"]["password_hash_rounds"]
+      password_hash_rounds = proposal["attributes"][@bc_name]["identity"]["password_hash_rounds"]
+      case proposal["attributes"][@bc_name]["identity"]["password_hash_algorithm"]
+      when "bcrypt"
+        if password_hash_rounds < 4 || password_hash_rounds > 31
```
cmurphy: The documentation says

> The default for bcrypt is 12, must be between 4 and 31, inclusive

So I think this needs to be > 32.

It's pulled from the passlib docs which is why it's worded so differently for bcrypt and scrypt even though they have the same upper bound.

nicolasbock: I interpret this wording to mean `[4, 31]` which means the check here is correct.

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -87,6 +87,25 @@ def validate_proposal_after_save(proposal)
...
+          validation_error I18n.t("barclamp.#{@bc_name}.validation.bcrypt_password_hash_rounds")
+        end
+      when "scrypt"
+        if password_hash_rounds < 1 || password_hash_rounds > 32
+          validation_error I18n.t("barclamp.#{@bc_name}.validation.scrypt_password_hash_rounds")
+        end
+      when "pbkdf_sha512"
+        if password_hash_rounds < 1 || password_hash_rounds > 4294967296
```
nicolasbock: Since

> The default for pbkdf_sha512 is 60000, must be within `range(1,1<<32)`

This check is incorrect as `range(1, 1 << 32)` is `[1, 4294967296)` or `[1, 4294967295]`. Instead of using such a large integer, rewriting the check as `password_hash_rounds > (1 << 32) - 1` would be a easier to read.

crowbar_framework/app/models/keystone_service.rb
```diff
@@ -87,6 +87,25 @@ def validate_proposal_after_save(proposal)
...
+    # validate the password hash configuration
+    if proposal["attributes"][@bc_name]["identity"]["password_hash_rounds"]
+      password_hash_rounds = proposal["attributes"][@bc_name]["identity"]["password_hash_rounds"]
+      case proposal["attributes"][@bc_name]["identity"]["password_hash_algorithm"]
+      when "bcrypt"
+        if password_hash_rounds < 4 || password_hash_rounds > 31
+          validation_error I18n.t("barclamp.#{@bc_name}.validation.bcrypt_password_hash_rounds")
+        end
+      when "scrypt"
+        if password_hash_rounds < 1 || password_hash_rounds > 32
```
nicolasbock: Thanks @cmurphy for pointing out the difference in wording:

> The default for scrypt is 16, must be within `range(1,32)`.

Since `range(1, 32)` is `[1, 32)` or equivalently `[1, 31]` the check here is incorrect. It should be `password_hash_rounds > 31`.

crowbar_framework/app/models/keystone_service.rb
cmurphy: I had to go back and forth on this in my head but you're right, if the user gives password_hash_rounds=31 then 31 is not > 31 so it should pass the validation. The one below is incorrect. Good call.

nicolasbock: It's definitely worded in a very confusing way :smile:

crowbar_framework/app/models/keystone_service.rb
toabctl: Done

crowbar_framework/app/models/keystone_service.rb
toabctl: Done

crowbar_framework/app/models/keystone_service.rb
toabctl: Done

crowbar_framework/config/locales/keystone/en.yml
```diff
@@ -47,3 +47,6 @@ en:
         api_version: 'API version %{api_version} not recognized.'
         enable_keystone_api: 'Keystone API version 3 needs to be enabled in order to use domain specific drivers.'
         timeout: 'Setting the Horizon timeout (%{horizon_timeout} minutes) longer than the Keystone token expiration timeout (%{keystone_timeout} minutes) is not supported.'
+        bcrypt_password_hash_rounds: "Password hash rounds for bcrypt must be between 4 and 31"
+        crypt_password_hash_rounds: "Password hash rounds for scrypt must be between 1 and 32"
+        pbkdf_sha512_password_hash_rounds: "Password hash rounds for bcrypt must be between 1 and 4294967296"
```
nicolasbock: The UI messages need to be updated as well.

toabctl: Done

crowbar_framework/config/locales/keystone/en.yml
```diff
@@ -47,3 +47,6 @@ en:
         api_version: 'API version %{api_version} not recognized.'
         enable_keystone_api: 'Keystone API version 3 needs to be enabled in order to use domain specific drivers.'
         timeout: 'Setting the Horizon timeout (%{horizon_timeout} minutes) longer than the Keystone token expiration timeout (%{keystone_timeout} minutes) is not supported.'
+        bcrypt_password_hash_rounds: "Password hash rounds for bcrypt must be between 4 and 31"
+        crypt_password_hash_rounds: "Password hash rounds for scrypt must be between 1 and 32"
+        pbkdf_sha512_password_hash_rounds: "Password hash rounds for pbkdfsha512 must be between 1 and 4294967296"
```
nicolasbock: I don't see the updates? Is GitHub a little slow today?

toabctl: I do see the update

crowbar_framework/config/locales/keystone/en.yml
```diff
@@ -47,3 +47,6 @@ en:
         api_version: 'API version %{api_version} not recognized.'
         enable_keystone_api: 'Keystone API version 3 needs to be enabled in order to use domain specific drivers.'
         timeout: 'Setting the Horizon timeout (%{horizon_timeout} minutes) longer than the Keystone token expiration timeout (%{keystone_timeout} minutes) is not supported.'
+        bcrypt_password_hash_rounds: "Password hash rounds for bcrypt must be between 4 and 31"
+        scrypt_password_hash_rounds: "Password hash rounds for scrypt must be between 1 and 32"
+        pbkdf_sha512_password_hash_rounds: "Password hash rounds for pbkdfsha512 must be between 1 and 4294967296"
```
nicolasbock: Sorry, yes, I see an update but not the one I was looking for :wink: You need to update the ranges :smile:

toabctl: Done

PR #1257, opened 2017-09-11
mysql: Added SSL configuration for client-server traffic

jsuchome:
This does not include replication traffic within Galera Cluster.

**Notes**:
 - Generating certificates had to be enhanced, because:
    - our `ssl_setup` helper did not generate CA's and that seems to be required for this database setup
    - python connection (for OpenStack services seems to require either cn=hostname for server certificate (_and it is actually server IP, not hostname in our setup_!) or subjectAltName values that match the server hostname(s) .... see  `match_hostname` in `/usr/lib64/python2.7/ssl.py` ... So I'm adding cluster hostname as CN

**Missing**: 
 - Client side configuration:
    -  REQUIRE SSL for OpenStack database users to allow only ssl connections - currently done for keystone only
    - ~~database connection string needs location of CA~~ 2nd commit
    - ca-cert needs to be copied to clients (if they are outside of the cluster)

 - UI
 - ~~SST setup (not required by spec)~~
 - ~~Galera related configuration (probably nothing else is needed, but might require further work during certs generation)~~

**Documentation**:
 - http://galeracluster.com/documentation-webpages/ssl.html
 - https://www.cyberciti.biz/faq/howto-install-configure-mariadb-galera-master-cluster-ubuntulinux/ (contains other bits I've used for testing)
 - https://docs.openstack.org/security-guide/databases/database-access-control.html mentiones the connection string update


chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -92,6 +92,7 @@ def action_grant
           grant_sql += " ON #{db_name}.#{tbl_name}"
           grant_sql += " TO '#{username}'@'#{host}'"
           grant_sql += " IDENTIFIED BY '#{password}'"
+          grant_sql += " REQUIRE SSL" if new_resource.require_ssl
```
rsalevsky: What happens for existing users if we switch from non-SSL to SSL?

chef/cookbooks/mysql/templates/default/my.cnf.erb
```diff
@@ -10,10 +10,19 @@ user                   = mysql
...
 # Instead of skip-networking the default is now to listen only on
 # localhost which is more compatible and is not less secure.
 bind-address            = <%= node['mysql']['bind_address'] %>
 
 [mysqldump]
 # FIXME: Remove after MariaDB 10.2.X switch (new default is 16777216)
 max_allowed_packet      = <%= node['mysql']['tunable']['max_allowed_packet'] %>
+
+<% if node['mysql']['ssl']['enabled'] %>
+ssl-ca = <%= node['mysql']['ssl']['ca_certs'] %>
```
rsalevsky: Why is this required?

chef/cookbooks/mysql/templates/default/my.cnf.erb
```diff
@@ -10,10 +10,19 @@ user                   = mysql
 datadir                = <%= node['mysql']['datadir'] %>
 tmpdir                 = <%= node['mysql']['tmpdir'] %>
 
+<% if node['mysql']['ssl']['enabled'] %>
+ssl-cert = <%= node['mysql']['ssl']['certfile'] %>
```
rsalevsky: I would probably split this into a SSL config to follow the path to split the config in single logical pieces.

rsalevsky:
I can't comment inline but if I understand the config correct then we have to add the `wsrep_provider_options` to work with SSL.

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
jsuchome: I assume they will require SSL since the switch. In other words, they will not be able to connect non-SSL way

chef/cookbooks/mysql/templates/default/my.cnf.erb
jsuchome: It's not necessary for server part itself, but when I was playing around with clients, I did not find a better way (as written in PR description)

chef/cookbooks/mysql/templates/default/my.cnf.erb
jsuchome: Makes sense

jsuchome:
@rsalevsky  `wsrep_provider_options` is for replication traffic if I get it correct, and it is not required AFAIK

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
rsalevsky: I guess we should test this, my assumption is that they will work in the SSL and non-SSL way as they don't require SSL.

jsuchome: Isn't the grant action executed for existing users as well?

chef/cookbooks/mysql/templates/default/my.cnf.erb
jsuchome: Ha! `[client]` section is missing here

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
rsalevsky: You are right, the privileges should get updated (if the code works as expected).

chef/cookbooks/mysql/templates/default/my.cnf.erb
jsuchome: Done

rhafer:
> REQUIRE SSL for OpenStack database users to allow only ssl connections - currently done for keystone only

@jsuchome  Shouldn't we at least do it for the user created by the mysql cookbook itself (i.e. "db_maker" and "monitoring")

jsuchome:
REQUIRE SSL - makes sense for db_maker, monitoring is being reworked by @jhesketh in another PR.
But we will need it for all OpenStack users, I've just did this one for experiments.

rhafer:
> REQUIRE SSL - makes sense for db_maker [..]
> But we will need it for all OpenStack users, I've just did this one for experiments.

Sure, I was just a bit confuse not to see it enabled for the "db_maker" from the beginning.

Another thing I find a bit confusing:

> our ssl_setup helper did not generate CA's and that seems to be required for this database setup

Do you have more details about this requirement? I assume that's because ssl can only be enabled for the openstack service(s) by adding some "ssl_" parameter to the connection string? SQLalchemy seems to behave really weird in that regard. :(
If you want to avoid that ca creation part you could use "ssl_key=some_value" and leave out the ssl_ca values. They main issue with requiring the ca seems to be that we need to distribute the ca to all database clients (i.e. openstack controllers). How do we do that if they're on a separate cluster?

jsuchome:
> Do you have more details about this requirement? 

@rhafer not exactly, it's based on my experiments.

It's possible to use client and server certificates,  https://docs.openstack.org/security-guide/databases/database-access-control.html suggests how to adapt the connection strings. But I have no idea how to set the correct rights to e.g. client keys if all openstack services should have access to them.

Also, HOS percona setup is also using this kind of setup (ca only for client part).

> They main issue with requiring the ca seems to be that we need to distribute the ca to all database
> clients  ...

Yes, I've already adapted PR description with another TODO item.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -137,6 +137,29 @@
   revision node[:database]["crowbar-revision"]
 end
 
+# Sync generated ssl certificates from the founder
+if node[:mysql][:ssl][:enabled] &&
```
s-t-e-v-e-n-k: This does not go far enough. If we have a cluster of 3 machines for the database, and a separate controller cluster for the OpenStack controllers (which is certainly plausible), we don't sync the CA files to the other controllers and they will not be able to connect.

chef/data_bags/crowbar/template-database.json
```diff
@@ -10,7 +10,14 @@
         "innodb_buffer_pool_size": 256,
         "max_connections": 800,
         "tmp_table_size": 64,
-        "max_heap_table_size": 64
+        "max_heap_table_size": 64,
+        "ssl": {
+          "enabled": false,
```
s-t-e-v-e-n-k: Do we also want to export this option onto the barclamp page so users can avoid using raw mode to set it?

chef/cookbooks/database/libraries/crowbar.rb
```diff
@@ -15,11 +15,13 @@ def self.get_ha_vhostname(node)
   end
 
   def self.get_listen_address(node)
+    # For SSL we prefer a cluster hostname (for certificate validation)
+    use_ssl = node[:database][:sql_engine] == "mysql" && node[:database][:mysql][:ssl][:enabled]
     if node[:database][:ha][:enabled]
       vhostname = get_ha_vhostname(node)
-      CrowbarPacemakerHelper.cluster_vip(node, "admin", vhostname)
+      use_ssl ? vhostname : CrowbarPacemakerHelper.cluster_vip(node, "admin", vhostname)
```
s-t-e-v-e-n-k: I wonder if we want to use the hostname all the time, regardless of SSL.

chef/data_bags/crowbar/template-database.schema
```diff
@@ -25,7 +25,16 @@
...
                 "tmp_table_size": { "type": "int", "required": true },
-                "max_heap_table_size": { "type": "int", "required": true }
+                "max_heap_table_size": { "type": "int", "required": true },
+                "ssl": {
+                  "type": "map", "required": true, "mapping": {
+                    "enabled": { "type": "bool", "required": true },
+                    "generate_certs": { "type": "bool", "required": true },
+                    "certfile": { "type" : "str", "required" : true },
+                    "keyfile": { "type" : "str", "required" : true },
+                    "ca_certs": { "type" : "str", "required" : true }
```
s-t-e-v-e-n-k: Do we also need an insecure flag, when using self-signed certs?

chef/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb
```diff
@@ -11,12 +11,47 @@ def load_current_resource
...
+
+            package "openssl"
+
+            dir = ::File.dirname(@current_resource.ca_certs)
+            ::FileUtils.mkdir_p(dir) unless ::File.exist?(dir)
+
+            # Generate CA key
+            ca_key_file = "#{dir}/ca_key.pem"
+            `openssl genrsa -out #{ca_key_file} 4096`
+            if $?.exitstatus != 0
```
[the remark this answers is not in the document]
s-t-e-v-e-n-k: Can we silence this warning? It keeps coming up, and everyone ignores it.

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: Yes, I know, I have the need for copying CA's mentioned in Missing part of PR description.

chef/data_bags/crowbar/template-database.json
jsuchome: Yep (another bit mentioned in description, as "UI missing")

chef/data_bags/crowbar/template-database.schema
jsuchome: I did not find if any service actually requires insecure flag for that case. I think they do not seem to care.

chef/cookbooks/database/libraries/crowbar.rb
jsuchome: Good point. But this is in line of what we do for OpenStack services ssl setup - hostname only in case of SSL, IP otherwise.

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: But anyway it would need to be at different place (not in database barclamp but in the barclamp relevant for such controller node(s).

rhafer:
@jsuchome Sorry, this got somewhat longer than what I planned for :)

> It's possible to use client and server certificates, https://docs.openstack.org/security-guide/databases/database-access-control.html suggests how to adapt the connection strings. But I have no idea how to set the correct rights to e.g. client keys if all openstack services should have access to them.

That's ok, I don't think we want to use client certificates for mariadb. We're not doing that anywhere in SOC. I was mainly wondering why we need to create (and distribute) a CA when using self-signed certificates. And the answer to that seems to be the weird configuration mechanism that sqlalchemy/pymysql uses for setting up ssl. I.e. you can only enable ssl by setting one, or more, of the ssl_ options (there is no generic "use_ssl" switch or similar). These are the ssl_ options know by sqlalchemy/pymsql:

* ssl_ca
* ssl_key
* ssl_cert
* ssl_capath
* ssl_cipher

Setting any of these to a non-empty value will cause the client to setup an ssl connection. If we use "ssl_ca" that will also implicitly enable server certificate verification, which in general is a good thing, however we are not doing that anywhere else when we using the self-signed certificates created by our chef recipes (that's why we have that --insecure setting everywhere for the openstack service). And it will also mean that need to worry about those certificates (and the ca) becoming invalid.

The equivalent of doing what we do for the openstack services (i.e. non server cert verification) could be achieved by not settting "ssl_ca" on the connection string, but e.g. by using "ssl_key" with some arbitrary value. Just setting "ssl_key" without settting "ssl_cert" has no real effect on the ssl configuraton in sqlalchemy/pymysql, other than enabling ssl with keeping server cert verification disabled. For reference:

SQLAlchemy's ssl code for the pymysql connector:
https://bitbucket.org/zzzeek/sqlalchemy/src/d8cddf61f9de1570257c575a149738ecec6a9d91/lib/sqlalchemy/dialects/mysql/mysqldb.py?at=master#mysqldb.py-152

PyMySQL's ssl initialization:
https://github.com/PyMySQL/PyMySQL/blob/master/pymysql/connections.py#L709

Don't get me wrong, in general I think creating correct certification and enabling verification is the right thing to do. But we need to thing through the consequences of it.

> Also, HOS percona setup is also using this kind of setup (ca only for client part).

Do you know how they create the certifcates? E.g. what lifetime. And e.g. how the ca is distributed there?

> But In general, I'm not saying this is the only correct setup, it's just the only one that I was able to make work.

What the correct setup is really depends on the requirements. If server-cert verification is wanted for self-signed certificates, the configuration in this PR seems correct. We "just" need to solve how the CA is going to be distributed. If we just want the "--insecure" variant by default as for everything else we'd need to change some things. Maybe it's worth to get some input from TPM here.

jsuchome:
@rhafer this is what @jodavis-suse wrote me:
_HP had worked out an Ephemeral CA scheme. Basically, since "noone uses CRLs", they set the certificate expiration date to something pretty short (like a day) and update the certificates frequently (like every half a day)._

jsuchome:
@rhafer also, my understanding is that this whole 'generate certificates' part we're offering is mostly for testing and POC's. As it is insecure itself, we do not expect customers are gonna use it in production.

rhafer:
> also, my understanding is that this whole 'generate certificates' part we're offering is mostly for testing and POC's. As it is insecure itself, we do not expect customers are gonna use it in production.

In that case I'd say the "ssl_key=value_doesnt_matter" approach is a lot simpler to implement than the CA distribution.

chef/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb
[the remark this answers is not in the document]
rsalevsky: @s-t-e-v-e-n-k Why not following it? I would also prefer `$CHILD_STATUS`.

jsuchome:
closed in favor of https://github.com/crowbar/crowbar-openstack/pull/1280

PR #1261, opened 2017-09-12
nova: enable nested virt on Intel

bmwiedemann:
because it defaults to off
but a lot of people rely on nested virt being available

While in https://fate.suse.com/320082 the virtualisation team
declined to promote nested virt to fully supported status for SLE12,
we are using this since 2012 in all kinds of places without problems.
So it remains in tech-preview.

### DIFF
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -111,7 +111,12 @@ case node[:nova][:libvirt_type]
             # load modules only when appropriate kernel is present
             execute "loading kvm modules" do
               command <<-EOF
-                  grep -qw vmx /proc/cpuinfo && /sbin/modprobe kvm-intel
+                  if grep -qw vmx /proc/cpuinfo ; then
+                    ! grep -q nested /etc/modprobe.d/80-kvm-intel.conf &&
+                      echo "options kvm_intel nested=1" > /etc/modprobe.d/80-kvm-intel.conf &&
+                      /sbin/modprobe -r kvm-intel
+                    /sbin/modprobe kvm-intel
+                  fi
                   grep -qw svm /proc/cpuinfo && /sbin/modprobe kvm-amd
                   grep -q POWER /proc/cpuinfo && /sbin/modprobe kvm
                   /sbin/modprobe vhost-net


chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -111,7 +111,12 @@
             # load modules only when appropriate kernel is present
             execute "loading kvm modules" do
               command <<-EOF
-                  grep -qw vmx /proc/cpuinfo && /sbin/modprobe kvm-intel
+                  if grep -qw vmx /proc/cpuinfo ; then
+                    ! grep -q nested /etc/modprobe.d/80-kvm-intel.conf &&
+                      echo "options kvm_intel nested=1" > /etc/modprobe.d/80-kvm-intel.conf &&
```
nicolasbock: This overwrites the conf file and would remove anything that's in it already. What do you think of `>>` instead?

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -111,7 +111,12 @@
...
+                      /sbin/modprobe -r kvm_intel
+                    /sbin/modprobe kvm-intel
```
nicolasbock: Why load the module twice?

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -111,7 +111,12 @@
             # load modules only when appropriate kernel is present
             execute "loading kvm modules" do
               command <<-EOF
-                  grep -qw vmx /proc/cpuinfo && /sbin/modprobe kvm-intel
+                  if grep -qw vmx /proc/cpuinfo ; then
+                    ! grep -q nested /etc/modprobe.d/80-kvm-intel.conf &&
+                      echo "options kvm_intel nested=1" > /etc/modprobe.d/80-kvm-intel.conf &&
+                      /sbin/modprobe -r kvm_intel
```
nicolasbock: nit: you are using an `_` in the module name and one line further a `-`. Personally I prefer the `-` but as long as the naming is consistent I don't mind either form.

chef/cookbooks/nova/recipes/compute.rb
bmwiedemann: modprobe -r unloads it. so that the line after loads it with the new option applied.

chef/cookbooks/nova/recipes/compute.rb
bmwiedemann: actually, the module itself is inconsistently named kvm-intel.ko in the filesystem, but kvm_intel in `lsmod` when loaded. I'll update the modprobe line, though.

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -111,7 +111,12 @@
             # load modules only when appropriate kernel is present
             execute "loading kvm modules" do
               command <<-EOF
-                  grep -qw vmx /proc/cpuinfo && /sbin/modprobe kvm-intel
+                  if grep -qw vmx /proc/cpuinfo ; then
+                    ! grep -q nested /etc/modprobe.d/80-kvm-intel.conf &&
+                      echo "options kvm_intel nested=1" > /etc/modprobe.d/80-kvm-intel.conf &&
+                      /sbin/modprobe -r kvm-intel
```
aplanas: `rmmod`?

chef/cookbooks/nova/recipes/compute.rb
nicolasbock: Thanks.

chef/cookbooks/nova/recipes/compute.rb
nicolasbock: Thanks, for some reason I was thinking of 'reload' but you are right, it's 'remove'.

nicolasbock:
It looks like nested virtualization is not going to be supported for SP2/3 (Feature #320082). We shouldn't enable it by our barclamp in that case.

aplanas:
@nicolasbock netsted virtualization was never supported AFAIK, but I do not see this rejected for SLE15. But at least needs to be mentioned in the documentation as 'no supported' in SP3 if this reach SOC7

nicolasbock:
That's true @aplanas. As long as we are mentioning it, it should be fine. Thanks!

bmwiedemann:
It has tech-preview status in SLE12 for some years. I updated the commit message to include
```
    While in https://fate.suse.com/320082 the virtualisation team
    declined to promote nested virt to fully supported status for SLE12,
    we are using this since 2012 in all kinds of places without problems.
```

nicolasbock:
Thanks for the update @bmwiedemann!

bmwiedemann:
>> but a lot of people rely on nested virt being available

> Who?

in SUSE basically everyone who already uses kvm
- SOC (e.g. cleanvm CI)
- CAP (Alex Barclay explicitly asked for it)
and I guess susestudio, OBS, openQA teams could all have use-cases for it.

OTOH - what is the disadvantage of having it enabled?

vuntz:
> OTOH - what is the disadvantage of having it enabled?

Having it enabled by default is setting the expectation that it's fully supported.

dirkmueller:
not for cloud 7, lets continue the discussion in the cloud 8 PR #1261

PR #1265, opened 2017-09-13
mysql: Make sure galera resources are started on controller nodes only

jsuchome:
Adding the location constraint should ensure that the service
is not running at remote cluster nodes.


jsuchome:
Still does not seem to work, btw

jsuchome:
Squashed one commit from @rhafer

PR #1266, opened 2017-09-13
keystone: Switch memcache backend to oslo_cache.memcache_pool

rhafer:
Per the oslo.cache documentation it's recommended to use the pooling
backend for eventlet based and threaded services. As our setup for the
keystone vhost in apache uses threads by default, let's use the pooling
backend. This seems to significantly reduce the number of open
connections to memcached.

Partial-Bug: bsc#1057822

Note: An alternative fix might be to disable threads for the WSGIDaemonProcess. (I don't really have an idea how much good that does anyway, give pythons GIL)

### DIFF
--- a/chef/cookbooks/keystone/templates/default/keystone.conf.erb
+++ b/chef/cookbooks/keystone/templates/default/keystone.conf.erb
@@ -8,7 +8,7 @@ log_dir = /var/log/keystone
 driver = <%= node[:keystone][:assignment][:driver] %>
 
 [cache]
-backend = dogpile.cache.memcached
+backend = oslo_cache.memcache_pool
 enabled = True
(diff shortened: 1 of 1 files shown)


rhafer:
> When you write that this change 'seems to significantly reduce the number of open connections', do you have some numbers in terms of how many connections there are before this change and how many after?

In the worst case (I haven't see that during my tests yet) with the new backend and a three node cluster each keystone process should have 30 connections open to memcached. (The pool is by default limited to 10 connections per destination server).

With the old backend I've seen it using more then twice of that. (I somehow suspect the old backend is leaking descriptors in multithreaded envs, but I don't have any prove for that :-( )

nicolasbock:
Thanks @rhafer, that's quite an improvement. :+1:

PR #1276, opened 2017-09-15
heat: update cookbook for pike

aagate:
rabblt_use_ssl and kombu_ssl_ca_certs oslo messaging
config options were deprecated in Ocata.
They were replaced by ssl and ssl_ca_file instead.

Reference: https://review.openstack.org/#/c/438455/


toabctl:
@aagate fyi, we usually add people to the "reviewers" list when we request a review.

PR #1277, opened 2017-09-15
Use httpchk in haproxy for galera

jhesketh:
By using a script provided in the galera-clustercheck package we are
able to ensure a node is in sync with the galera cluster rather than
the previously superficial check of just ensuring mysql is running.

Set up the correct user required by the package and start the
systemd service.

This requires a change to crowbar-ha in PR #242 to allow setting of
server-default across the haproxy.


abel-navarro:
in my understanding *mysql-check* does better check than *http* one. And both are better than the *tcp* one. Could you explain why removing *mysql-check* while there's support for Galera? [1]

[1] https://severalnines.com/resources/tutorials/mysql-load-balancing-haproxy-tutorial

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -253,11 +253,33 @@
   n["fastinter"] = 1000
 end
 
+# Set up monitoring user and service
+db_settings = fetch_database_settings
```
jsuchome: We're fetching this already at some other place of this recipe, aren't we?

chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -153,16 +153,6 @@
     only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
   end
 
-  database_user "create haproxy and galera monitoring user" do
-    connection db_connection
-    username "monitoring"
```
jsuchome: We're still using 'monitoring' user for galera monitoring (see resource agent setting in ha_galera.rb)

jsuchome:
@abel-navarro but here we're going for `httpchk` and extra script, see also https://etherpad.nue.suse.com/p/SCRD-1208 for the reasearch

(But your comments are welcomed!)

chef/cookbooks/mysql/recipes/ha_galera.rb
jhesketh: Not exactly. We set up empty credentials while we create the sstuser but only on the founder node. We'll need to do this for each node.

chef/cookbooks/mysql/recipes/server.rb
jhesketh: Hmm okay. I did it this way as the clustercheck script defaults to the running user to authenticate with mysql which happens to be a new user: galera-clustercheck. So we have a couple of options here:
 - We create a my.cnf with the monitoring (or similar) that we use
 - or we change the monitoring user for galera to also be galera-clustercheck

jsuchome: Yeah, I think we probably could use the new user for monitoring, just wanted to point out the inconsistency.

jhesketh:
@abel-navarro, the page you linked to actually comments that mysql-check isn't fully complete "Take note that this does not check database presence nor database consistency". However it does reference mysqlchk which is a part of their ClusterControl product. mysqlchk is just a script that they run via either httpchk or tcp-check in the same way we are. We're just getting our check script from the galera-clustercheck package.

abel-navarro:
@jhesketh @jsuchome thanks for the explanation. I couldn't infer that from the PR description. Now I can see why it's needed.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -253,11 +263,18 @@
   n["fastinter"] = 1000
 end
 
+# Start galera-clustercheck which serves the cluster state as http return codes
+# on port 8000
+service "galera-clustercheck" do
+  action :start
```
jsuchome: What about `:enable` action?

jsuchome:
@jhesketh so when you've introduced `default-server`, don't you wanna use it for other values we set for galera  as well?

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -253,11 +263,18 @@
   n["fastinter"] = 1000
 end
 
+# Start galera-clustercheck which serves the cluster state as http return codes
+# on port 8000
```
jsuchome: Is it correct to use port 8000? @rsalevsky @rhafer aren't we using it for something else?

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -96,6 +96,16 @@
       action :grant
     end
 
+    # Also create a monitoring user for haproxy and galera to use
+    database_user "create haproxy and galera monitoring user" do
+      connection db_connection
+      username "galera-clustercheck"
```
jsuchome: Where is this user actually used? Should it be part of `galera-clustercheck` config file?

chef/cookbooks/mysql/recipes/ha_galera.rb
rhafer: Shouldn't this be managed by pacemaker anyways? To avoid problems with start ordering and to have it monitored and automatically restarted in some way. (I think it kind of belongs into the g-haproxy resource group, but that might be tricky to setup)

jhesketh: Ah, so we have to enable and then start? (I'm not very familiar with chef still).

@rhafer: There shouldn't be any problems with start ordering as the script won't begin to connect to the database until hit with an http request. However it's an interesting idea having pacemaker run it. It's likely a good idea, but I'm not sure how much we gain or what the effort required is.

chef/cookbooks/mysql/recipes/ha_galera.rb
jhesketh: This is the default for the galera-clustercheck package. We can tune it if needed. I'll check to see if other services are using that port later, but if we're worried we can change it regardless?

chef/cookbooks/mysql/recipes/ha_galera.rb
jhesketh: The script uses MySQLdb which defaults to the currently running user. The systemd service runs as galera-clustercheck so that is what MySQLdb authenticates to mysql with (and no password). We can change all their either by configuring systemd, editing the script, or supplying a my.ini file to the script. I just went with what seemed the easiest, but I'm open to any of the above.

jhesketh:
@jsuchome: regarding default-server, yes we can move some values into there. I'll look at that as a follow up commit (but will put it on this branch/pr).

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: I see, so this should be fine.
On similar note, are we fine with the alera-clustercheck config file as it is provided? AFAICS it points to non-existing my.cnf file (which might not be a problem)

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: What is needed is imho just calling `pacemaker_group g-haproxy` with one member, which would be pacemaker primitive build for `galera-clustercheck` service. See
https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/recipes/haproxy.rb#L100

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -142,7 +152,7 @@
   params({
     "enable_creation" => true,
     "wsrep_cluster_address" => cluster_addresses,
-    "check_user" => "monitoring",
+    "check_user" => "galera-clustercheck",
```
rsalevsky: Why do we have to rename the user? I like the old name a bit more.

jhesketh: It's explained a little bit above (in this PR), but clearly this needs a comment in the code.

Basically it's because the script authenticates with mysql as the user who is running the script. The way it is currently packaged is to be ran by a user galera-clustercheck. Given we are discussing taking more ownership and possible modifications of the package we could modify this further. Rather than having 2 monitoring users it was easier to have pacemaker also use this for it's monitoring.

chef/cookbooks/mysql/recipes/ha_galera.rb
jhesketh: Right, we don't need a my.cnf as far as I can tell, but we have the option of turning the service in the future.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -255,11 +255,29 @@
...
+template "/etc/galera-python-clustercheck/my.cnf" do
+  source "galera-python-clustercheck-my.cnf.erb"
+  owner "galera-python-clustercheck"
+  group "mysql"
+  mode "0640"
+end
+
+# Start galera-clustercheck which serves the cluster state as http return codes
+# on port 8000
+pacemaker_group "g-#{service_name}" do
```
jsuchome: This should be just adapting existing haproxy group ("g-haproxy" if I am not mistaken)

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -255,11 +255,29 @@
...
+  members ["galera-python-clustercheck", service_name]
```
jsuchome: Here we need new pacemker primitive (for your service) as a member

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -255,11 +255,38 @@
...
+# Start galera-clustercheck which serves the cluster state as http return codes
+# on port 8000
+
+pacemaker_primitive "galera-python-clustercheck" do
+  agent "systemd:galera-python-clustercheck"
+  op node[:mysql][:ha][:op]
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects.push("pacemaker_primitive[galera-python-clustercheck]")
```
jsuchome: So, the `transaction_objects` are not used at this place any more (the transaction was executed earlier, see `pacemaker_transaction`.

It's possible we're gonna need a new transaction here that will just group those new actions (primitive creation and haproxy group update). I'd suggest to go without the transaction for now (for first round of testing) and possibly adding it later.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -19,7 +19,7 @@
 
 resource_agent = "ocf:heartbeat:galera"
 
-["galera-3-wsrep-provider", "mariadb-tools", "xtrabackup", "socat"].each do |p|
+["galera-3-wsrep-provider", "mariadb-tools", "xtrabackup", "socat", "galera-python-clustercheck"].each do |p|
```
rsalevsky: As already discussed in IRC I don't believe that `galera-python-clustercheck` is the right name for this package as alternative I suggest `galery-pyclustercheck`.

jhesketh: I'm open to changing it, but it has been accepted into server:database (the submitreq was opened before the IRC discussion): https://build.opensuse.org/request/show/530396

I'll hold off doing a linkpac into cloud:openstack until we've finalised this in case we need to change it.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -255,11 +255,37 @@
...
+
+pacemaker_primitive "galera-python-clustercheck" do
+  agent "systemd:galera-python-clustercheck"
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+transaction_objects.push("pacemaker_primitive[galera-python-clustercheck]")
+
+pacemaker_group "g-haproxy" do
+  members ["galera-python-clustercheck"]
```
jsuchome: Ah, so this seems to be wrong. It removes previous members of the group

jhesketh: Yep, I was just testing this. I think we need to make the group extendable in the haproxy recipe (see new commit).

jsuchome:
As an alternative to https://github.com/crowbar/crowbar-ha/pull/256 , we could explicitely check for current members (this might be seen as uglier): 
```
haproxy_cmd = Mixlib::ShellOut.new("crm configure show g-haproxy").run_command
haproxy_members = haproxy_cmd.stdout.chop
haproxy_members = haproxy_members.split(" g-haproxy ").last

pacemaker_group "g-haproxy" do
  members haproxy_members.split + ["galera-python-clustercheck"]
  action :update
end
```

jsuchome:
@jhesketh @s-t-e-v-e-n-k I think this might do the job (instead of that `g-haproxy` related code):
```
transaction_objects = []
service_name = "galera-python-clustercheck"

pacemaker_primitive service_name do
  agent "systemd:#{service_name}"
  action :update
end

transaction_objects.push("pacemaker_primitive[#{service_name}]")

clone_name = "cl-#{service_name}"
pacemaker_clone clone_name do
  rsc service_name
  meta CrowbarPacemakerHelper.clone_meta(node, remote: false)
  action :update
end

transaction_objects.push("pacemaker_clone[#{clone_name}]")

clone_location_name = openstack_pacemaker_controller_only_location_for clone_name
transaction_objects << "pacemaker_location[#{clone_location_name}]"

pacemaker_transaction "clustercheck" do
  cib_objects transaction_objects
  action :commit_new
end

```

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: I'm fine with `galera-python-clustercheck` (but there are problems with its python-MySQL-python dependency that needs to be addressed)

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -236,6 +236,33 @@
...
+  owner "galera-python-clustercheck"
+  group "mysql"
+  mode "0640"
+end
+
+# Start galera-clustercheck which serves the cluster state as http return codes
+# on port 8000
+
+pacemaker_primitive "galera-python-clustercheck" do
+  agent "systemd:galera-python-clustercheck"
```
s-t-e-v-e-n-k: I wonder if this is too complicated? With the new logic that just landed to restart systemd resources when they fail, why don't we just start the service on all cluster members and have systemd care if the script dies?

chef/cookbooks/mysql/recipes/ha_galera.rb
jhesketh: What are the problems? Do you mean with the other packages we'll need to pull in?

jsuchome: galera-python-clustercheck requires python-MySQL-python but that conflicts with python-mysql required by other recipes (e.g. keystone)

chef/cookbooks/mysql/recipes/ha_galera.rb
jhesketh: @jsuchome: could you please comment on this if you have time. Happy with either solution personally.

jhesketh:
Thanks jsuchome for the clone code. I had found openstack_pacemaker_controller_clone_for_transaction which does a large portion of this, but I think since this isn't an openstack service your solution is neater.

chef/cookbooks/mysql/recipes/ha_galera.rb
jhesketh: I think an earlier version of the package may have, but the version in server:database depends on python-mysql: https://build.opensuse.org/package/view_file/server:database/galera-python-clustercheck/galera-python-clustercheck.spec?expand=1

Unless it's a dependency of python-twisted or I've missed something.

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: Right, that we could do. But the clone variant works as well and might be preferred as it is in line with other pacemaker related code

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: It's possible I tested with older package version; in such case, good for us

stefannica:
FYI, the cloud-mkcloud8-job-ha-linuxbridge-x86_64 Jenkins job might also be fixed by this. See, for instance [this recent failed build](https://ci.suse.de/job/openstack-mkcloud/89066/console).

jhesketh:
Thanks Ralf, incoming fix.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -238,6 +238,59 @@
...
 end
 
+# Configuration files for galera-python-clustercheck
+template "/etc/galera-python-clustercheck/galera-python-clustercheck.conf" do
+  source "galera-python-clustercheck.conf.erb"
+  owner "galera-python-clustercheck"
+  group "mysql"
+  mode "0640"
+  variables(
+    cluster_address: node["hostname"]
```
rhafer: I think I'd prefer to use the IP address instead of the hostname here. You can get that via: `Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address`

Keeps us on the save side in case there is some hickup with the resolver.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -238,6 +238,59 @@
...
+  )
+end
+
+template "/etc/galera-python-clustercheck/my.cnf" do
+  source "galera-python-clustercheck-my.cnf.erb"
+  owner "galera-python-clustercheck"
+  group "mysql"
+  mode "0640"
+  variables(
+    cluster_address: node["hostname"]
```
rhafer: Same here for hostname vs. IP address.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -238,6 +247,59 @@
...
 end
 
+# Configuration files for galera-python-clustercheck
+template "/etc/galera-python-clustercheck/galera-python-clustercheck.conf" do
+  source "galera-python-clustercheck.conf.erb"
+  owner "galera-python-clustercheck"
+  group "mysql"
+  mode "0640"
+  variables(
+    cluster_address: Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
```
jsuchome: Why are you calling this `cluster_address` when it actually is node address? (same for the other file)

jhesketh: This was a mistake between revisions. Pushing up a fix.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -261,7 +323,8 @@
   address CrowbarPacemakerHelper.cluster_vip(node, "admin")
   port 3306
   mode "tcp"
-  options ["mysql-check user monitoring"]
+  options ["httpchk"]
+  default_server "port 8000"
```
jsuchome: @rhafer just found out that this port is conflicting with another (I think some heat service)

jhesketh: Darn. Thanks for catching though, should be a trivial fix. Do you have any suggestions for ports?

jsuchome: Citing Ralf again, "maybe something from 55XX range? We've been using these for the services that use haproxy as frontend (550x for keystone, 551x for glance ...)"

I think 5555 is free

jhesketh: okay thanks, I'll update it now but I don't have time to test it

jsuchome: It was actually catched by the (HA) gating job, so getting that green would be a good indication

PR #1281, opened 2017-09-19
[4.0] nova: add HA rate limiting options to raw template

Itxaka:
Adds HA rate limiting options to the raw template for the different
services that depend from nova. Defaults to 0 as the values, which means
no rate limiting

(cherry picked from commit ffcc557fcc85888b95d058de7e72d315ba0008f6)

Backport of https://github.com/crowbar/crowbar-openstack/pull/1233

### DIFF
--- a/chef/cookbooks/nova/recipes/controller_ha.rb
+++ b/chef/cookbooks/nova/recipes/controller_ha.rb
@@ -29,6 +29,7 @@ haproxy_loadbalancer "nova-api" do
   port node[:nova][:ports][:api]
   use_ssl node[:nova][:ssl][:enabled]
   servers CrowbarPacemakerHelper.haproxy_servers_for_service(node, "nova", "nova-controller", "api")
+  rate_limit node[:nova][:ha_rate_limit]["nova-api"]
   action :nothing
 end.run_action(:create)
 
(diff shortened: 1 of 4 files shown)


vuntz:
@Itxaka seeing that we don't have the attribute for `nova-placement-api` here, it means we'll need a migration script in master to add it (and the current migration as in https://github.com/crowbar/crowbar-openstack/pull/1233/files#diff-640c8361247ad9016d10dc4223a89cb6 won't work for this). Can you do that?

aspiers:
Same comment as other PR - please mention that this is a backport of #1233 in the PR description.

vuntz:
HA job is failing because we need to update the crowbar-ha package after the other patch was merged.

Itxaka:
@vuntz will a modification to master's migration for this would be enough? Like modifying https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/migrate/nova/205_add_rate_limit.rb to have an extra
```
a["ha_rate_limit"]["nova-placement-api"] = ta["ha_rate_limit"]["nova-placement-api"] unless a["ha_rate_limit"].key? "nova-placement-api"
```

Itxaka:
crowbar-ha package for cloud7 rebuilt!

vuntz:
> @vuntz will a modification to master's migration for this would be enough?

Yes, what you put should be fine.

PR #1282, opened 2017-09-19
tempest: readd img_dir and ami/ari/aki_img_file options

stefannica:
The img_dir, ami_img_file, ari_img_file and aki_img_file
options from the scenario group are still being used by
some test cases:

  scenario.test_minimum_basic.TestMinimumBasicScenario.*
  scenario.test_encrypted_cinder_volumes.TestEncryptedCinderVolumes.*

, which fail unless these configuration options are supplied.

These options were removed during the Ocata update in #1167.

More info:
 - tempest configuration doc: https://github.com/openstack/tempest/blob/17.0.0/doc/source/configuration.rst#images

### DIFF
--- a/chef/cookbooks/tempest/templates/default/tempest.conf.erb
+++ b/chef/cookbooks/tempest/templates/default/tempest.conf.erb
@@ -142,7 +142,11 @@ instance_type = <%= @heat_flavor_ref %>
 lock_path = /var/run/tempest
 
 [scenario]
+img_dir = <%= @tempest_path %>/etc/cirros/
 img_file = cirros-<%= @cirros_version %>-<%= @cirros_arch %>-disk.img
+ami_img_file = cirros-<%= @cirros_version %>-<%= @cirros_arch %>-blank.img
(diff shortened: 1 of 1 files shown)


stefannica:
@cmurphy the failure is:
`No such file or directory: '/var/lib/openstack/etc/cirros//cirros-0.3.1-x86_64-vmlinuz'`
This happens because the default values do not reflect the reality.

toabctl:
@stefannica why ```readd```? Were the options removed? If so, in which commit and why?

stefannica:
@toabctl I updated the PR description to reflect the original PR.

toabctl:
> @toabctl I updated the PR description to reflect the original PR.

@stefannica can you add that to the commit msg please?

stefannica:
> > @toabctl I updated the PR description to reflect the original PR.
> 
> @stefannica can you add that to the commit msg please?

@toabctl done

cmurphy:
@stefannica maybe we should be fixing the defaults upstream then?

PR #1283, opened 2017-09-19
[4.0] rabbitmq: do not create the erlang cookie file on cluster deployment

Itxaka:
There should be no need to create the erlang cookie file for rabbit on
cluster deployments as the pacemaker resource agent already creates
it for us.

(cherry picked from commit ffe2732e464667f1fb3907cf9e4d2554b63f809b)


Itxaka:
[RabbitMQ active/active cleanup](https://trello.com/c/90sXOb0U/105-rabbitmq-active-active-cleanup)

Itxaka:
Doesnt seem to be safe without investigating further, closing for the moment.

PR #1284, opened 2017-09-19
glance: load glance metadefs into the database

stefannica:
Glance metadata definitions are missing from the database,
which causes the following fulltempest test cases to fail:

  tempest.api.image.v2.test_images_metadefs_namespace_properties.MetadataNamespacePropertiesTest.test_basic_meta_def_namespace_property
  tempest.api.image.v2.test_images_metadefs_resource_types.MetadataResourceTypesTest.test_basic_meta_def_resource_type_association

The metadef can be loaded into the database right after the
db sync step.

More info:
 - official openstack docs: https://docs.openstack.org/image-guide/image-metadata.html

### DIFF
--- a/chef/cookbooks/glance/recipes/registry.rb
+++ b/chef/cookbooks/glance/recipes/registry.rb
@@ -42,17 +42,26 @@ template node[:glance][:registry][:config_file] do
 end
 
 ha_enabled = node[:glance][:ha][:enabled]
+is_founder = CrowbarPacemakerHelper.is_cluster_founder?(node)
 
-crowbar_pacemaker_sync_mark "wait-glance_db_sync" if ha_enabled
+crowbar_pacemaker_sync_mark "wait-glance_database" if ha_enabled
 
 execute "glance-manage db sync" do
   user node[:glance][:user]
   group node[:glance][:group]
-  # We know the glance-api.conf file is not updated yet, so forcefully ignore it
-  command "glance-manage --config-file \"#{node[:glance][:registry][:config_file]}\" db sync"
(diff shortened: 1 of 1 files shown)


chef/cookbooks/glance/recipes/registry.rb
```diff
@@ -55,6 +55,16 @@
   only_if { !node[:glance][:db_synced] && (!node[:glance][:ha][:enabled] || CrowbarPacemakerHelper.is_cluster_founder?(node)) }
 end
 
+execute "glance-manage db_load_metadefs" do
+  user node[:glance][:user]
+  group node[:glance][:group]
+  # We know the glance-api.conf file is not updated yet, so forcefully ignore it
+  command "glance-manage --config-file \"#{node[:glance][:registry][:config_file]}\" db_load_metadefs"
```
dirkmueller: shouldn't we pass in config-dir instead of config-file here?

stefannica: This seems to be the way puppet does it: https://review.openstack.org/#/c/447480/

stefannica: Actually, there is a better explanation for this: a long time ago [there was a problem](https://github.com/crowbar/barclamp-glance/pull/227) with glance-manage loading the incorrect database configuration from a wrong configuration file. However, now I think neither --config-file nor --config-dir are needed anymore, since glance-manage is automatically loading everything.

stefannica:
@Itxaka , @nicolasbock could you re-review please ?

PR #1286, opened 2017-09-20
[4.0] all: Use new pacemaker option to stop managing stateless a/a services

vuntz:
When this option is used, only the following resources stay managed by
pacemaker (assuming the relevant roles are added to the cluster):

 - rabbitmq
 - database
 - haproxy
 - keystone-fernet-rotate
 - neutron-ha-tool
 - ceilometer-central
 - cinder-volume

All other services basically work in the same way as without HA, with a
simple systemd service.

This reduces the pacemaker complexity, and therefore the impact of
pacemaker, as well as side-effects that it may cause when
mis-used/mis-configured. But this also removes self-healing (which will
be added back with Restart= in systemd services in some followup work),
and the ordering constraints that are cluster-wide.

Note that the default value of the option in the pacemaker barclamp
implies no change at all for now; we will consider changing the default
once we implement Restart= in systemd services if it is found out that
this has no negative impact.

Backport of https://github.com/crowbar/crowbar-openstack/pull/1256
Depends on https://github.com/crowbar/crowbar-ha/pull/248


aspiers:
#toomanyreviewers (not a criticism of anyone here, just collecting more [evidence](https://github.com/search?q=toomanyreviewers&type=Issues) for [the need to optimise distribution of review work](http://mailman.suse.de/mlarch/SuSE/cloud-devel/2017/cloud-devel.2017.08/msg00207.html))

PR #1287, opened 2017-09-20
all: Make systemd restart services on failures

vuntz:
This offers some automatic recovery, which is nice. It turns out it's
also recommended by systemd documentation, in case a service is
long-running, which is the case here.

It's worth mentioning that in the past, the periodic chef-client has
been providing this, to some extent. But this means the recovery could
take up to 15 minutes.

This is of course disabled when a service is managed by Pacemaker, as we
do not want to have systemd interfere with Pacemaker in that case.

### DIFF
--- a/chef/cookbooks/barbican/recipes/worker.rb
+++ b/chef/cookbooks/barbican/recipes/worker.rb
@@ -28,3 +28,6 @@ service "openstack-barbican-worker" do
   action [:enable, :start]
   provider Chef::Provider::CrowbarPacemakerService if use_crowbar_pacemaker_service
 end
+utils_systemd_service_restart "openstack-barbican-worker" do
+  action use_crowbar_pacemaker_service ? :disable : :enable
+end
--- a/chef/cookbooks/postgresql/recipes/server_redhat.rb
+++ b/chef/cookbooks/postgresql/recipes/server_redhat.rb
@@ -88,10 +88,10 @@ service "postgresql" do
   service_name node["postgresql"]["server"]["service_name"]
   supports restart: true, status: true, reload: true, restart_crm_resource: true
   action [:enable, :start]
-  provider Chef::Provider::CrowbarPacemakerService if node[:database][:ha][:enabled]
+  provider Chef::Provider::CrowbarPacemakerService if ha_enabled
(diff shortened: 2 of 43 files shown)


chef/cookbooks/barbican/recipes/worker.rb
```diff
@@ -28,3 +28,6 @@
   action [:enable, :start]
   provider Chef::Provider::CrowbarPacemakerService if use_crowbar_pacemaker_service
 end
+utils_systemd_service_restart "openstack-barbican-worker" do
```
Itxaka: So hound will complain on the most absurd thing but will ignore the ugliness of not having a new line between blocks. I see how it is, hound ¬_¬

chef/cookbooks/postgresql/recipes/server_redhat.rb
```diff
@@ -90,6 +90,9 @@
   action [:enable, :start]
   provider Chef::Provider::CrowbarPacemakerService if node[:database][:ha][:enabled]
 end
+utils_systemd_service_restart "postgresql" do
+  action node[:database][:ha][:enabled] ? :disable : :enable
```
Itxaka: you can reuse the `ha_enabled` attribute here

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -75,3 +75,6 @@
   action [:enable, :start]
   provider Chef::Provider::CrowbarPacemakerService if node[:rabbitmq][:ha][:enabled]
 end
+utils_systemd_service_restart "rabbitmq-server" do
+  action node[:rabbitmq][:ha][:enabled] ? :disable : :enable
```
Itxaka: same as above

chef/cookbooks/postgresql/recipes/server_redhat.rb
vuntz: done

chef/cookbooks/rabbitmq/recipes/default.rb
vuntz: done

PR #1290, opened 2017-09-20
rabbitmq: Increase timeouts for start/promote actions (bsc#1059532)

Itxaka:
When starting rabbitmq, especially if we are recovering from fencing,
the start could take a while as it needs to check for the tables
to be correct and/or reset the mnesia dir depending on the outcome, of
that check so the whole process could take much longer than the
default 60 seconds.

We also increase the promote timeout as it also needs to do some
internal checking for the master score and it could easily go over the
default timeout.

### DIFF
--- a/chef/cookbooks/rabbitmq/attributes/default.rb
+++ b/chef/cookbooks/rabbitmq/attributes/default.rb
@@ -45,3 +45,6 @@ default[:rabbitmq][:erlang_cookie_path] = "/var/lib/rabbitmq/.erlang.cookie"
 # ha
 default[:rabbitmq][:ha][:enabled] = false
 default[:rabbitmq][:ha][:storage][:mode] = nil
+default[:rabbitmq][:ha][:op][:start][:timeout] = "300s"
+default[:rabbitmq][:ha][:op][:promote][:timeout] = "180s"
+default[:rabbitmq][:ha][:op][:monitor][:interval] = "10s"
--- a/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
+++ b/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
@@ -16,9 +16,6 @@
 pid_file = "/var/run/rabbitmq/pid"
 
 agent_name = "ocf:rabbitmq:rabbitmq-server-ha"
-rabbitmq_op = {}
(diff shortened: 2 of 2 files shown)


Itxaka:
[Bug 1059532 - RabbitMQ failed to start after cluster failover](https://trello.com/c/7KtD7Db7/115-bug-1059532-rabbitmq-failed-to-start-after-cluster-failover)

rhafer:
@gosipyan Is there any chance you could test this on your setup?

Itxaka:
Not enough, we need to actually pass those values to the primitive!

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -16,7 +16,7 @@
 pid_file = "/var/run/rabbitmq/pid"
 
 agent_name = "ocf:rabbitmq:rabbitmq-server-ha"
-rabbitmq_op = {}
+rabbitmq_op = node[:rabbitmq][:ha][:op] || {}
```
vuntz: Careful here: I think you may need to copy the structure, as the few lines afterwards may actually modify the node attributes otherwise.

rhafer: @Itxaka Why don't you just add the monitor values to the chef-attributes as well?

Itxaka: @both of you are right, let me redo this!

chef/cookbooks/rabbitmq/attributes/default.rb
```diff
@@ -45,3 +45,6 @@
 # ha
 default[:rabbitmq][:ha][:enabled] = false
 default[:rabbitmq][:ha][:storage][:mode] = nil
+default[:rabbitmq][:ha][:op][:start][:timeout] = 300
+default[:rabbitmq][:ha][:op][:promote][:timeout] = 180
+default[:rabbitmq][:ha][:op][:monitor][:interval] = "10s"
```
rhafer: I'd suggest you use the same variant everywhere. So either `"300s"` and `"180s"` or `10` without the `s`. Looking at some other cookbooks it seems we're preferring the string variant.

PR #1292, opened 2017-09-20
nova: use the proper vars for serialproxy

Itxaka:
We were leaving the default hardcoded values for serialproxy which could
collide with haproxy if deployed on HA as both will try to use the same
ports.


Itxaka:
[Nova - serialproxy uses hardcoded values](https://trello.com/c/DHDiOm5P/120-nova-serialproxy-uses-hardcoded-values)

PR #1301, opened 2017-09-22
Revert ha service restarts

cmurphy:
Originally, crowbar could not restart the postgresql or rabbitmq resource agents. We "fixed" this so that crowbar could properly manage it, but this is really not a safe thing to do for HA services. This PR reverts those commits so that we go back to the original behavior. We don't make either postgresql or rabbitmq highly configurable so neither the original changes nor this revert will have a significant impact on users.

See also this discussion: https://github.com/crowbar/crowbar-openstack/pull/974#pullrequestreview-53856237


cmurphy:
Backports at #1332 and #1333

PR #1307, opened 2017-09-22
database: Check if correct privileges are present before granting 

jsuchome:
Previous check for privileges always forced new grant action for users that required specific rights (like db_user)

### DIFF
--- a/chef/cookbooks/database/libraries/provider_database_mysql_user.rb
+++ b/chef/cookbooks/database/libraries/provider_database_mysql_user.rb
@@ -99,12 +99,13 @@ class Chef
           Chef::Log.info("Granting privileges for '#{new_resource.username}@#{new_resource.host}'")
 
           password = client.escape(new_resource.password)
+          requires = new_resource.require_ssl ? "SSL" : "NONE"
 
           grant_sql = "GRANT #{new_resource.privileges.join(",")}"
           grant_sql += " ON #{db_name}.#{tbl_name}"
           grant_sql += " TO '#{username}'@'#{host}'"
           grant_sql += " IDENTIFIED BY '#{password}'"
-          grant_sql += " REQUIRE SSL" if new_resource.require_ssl
+          grant_sql += " REQUIRE #{requires}"
           client.query(grant_sql)
           client.query("FLUSH PRIVILEGES")
         ensure
--- a/chef/cookbooks/database/libraries/provider_database_mysql_user.rb
+++ b/chef/cookbooks/database/libraries/provider_database_mysql_user.rb
@@ -63,29 +63,41 @@ class Chef
           db_name = new_resource.database_name || "*"
           tbl_name = new_resource.table || "*"
 
-          # test
-          incorrect_privileges = false
-          test_sql = client.prepare("SELECT * from mysql.db WHERE User = ? AND Host = ? AND Db = ?")
-          results = test_sql.execute(
-            new_resource.username,
-            new_resource.host,
-            db_name
-          )
-          incorrect_privileges = true if results.size.zero?
-          # These should all by 'Y'
-          results.each do |result|
-            new_resource.privileges.each do |privileges|
-              key = "#{privileges.capitalize}_priv"
-              incorrect_privileges = true if privileges[key] != "Y"
+          username = client.escape(new_resource.username)
+          host = client.escape(new_resource.host)
+
+          # test the presence of privileges
+          privs = {}
+          if new_resource.privileges == [:all]
+            privs = { "ALL" => false }
+          else
+            new_resource.privileges.each { |p| privs[p] = false }
+          end
+
+          ssl_set = false
+
+          client.query("SHOW GRANTS FOR '#{username}'@'#{host}'").each do |row|
+            row.each do |grant, val|
+              # example of vals:
+              # GRANT USAGE ON *.* TO 'nova'@'%' IDENTIFIED BY PASSWORD '..' REQUIRE SSL
+              # GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, INDEX, ALTER ON `nova`.* TO 'nova'@'%'
+              ssl_set = true if val.include?("REQUIRE SSL")
+              # make sure to check privileges for selected database
+              next unless db_name == "*" || val.include?("\`#{db_name}\`")
+              privs.keys.each do |priv|
+                privs[priv] = true if val.include? priv
+              end
             end
           end
 
+          correct_privileges = privs.reject { |priv, present| present }.empty? &&
+            (ssl_set == new_resource.require_ssl)
+
+          return if correct_privileges
+
           # grant
-          return unless incorrect_privileges
           Chef::Log.info("Granting privileges for '#{new_resource.username}@#{new_resource.host}'")
 
-          username = client.escape(new_resource.username)
-          host = client.escape(new_resource.host)
           password = client.escape(new_resource.password)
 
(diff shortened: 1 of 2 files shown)


jsuchome:
Note: it seems to produce some failures with `sst` user set up while bootstrapping galera...

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -80,6 +80,17 @@ def action_grant
             end
           end
 
+          if new_resource.require_ssl
+            ssl_required = false
+            grants = client.query("show grants for #{new_resource.username}")
```
rsalevsky: What is the output of show grants? Can we combine the incorrect_privileges and this check?

jsuchome: Unfortunatelly the previos check for incorrect_privileges does not list SSL part.

But you are suggesting we might use show grants for both, right? Maybe. It looks like this:
```
{"Grants for keystone@%"=>"GRANT USAGE ON *.* TO 'keystone'@'%' IDENTIFIED BY PASSWORD '*8849BFB74A78E172855C23FFDE2B1FB335AD2D65' REQUIRE SSL"}
{"Grants for keystone@%"=>"GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, DROP, INDEX, ALTER ON `keystone`.* TO 'keystone'@'%'"}
```

jsuchome: Hm, actually I think current check (the original for  incorrect_privileges) is already broken. 

```
> SELECT * from mysql.db WHERE User = "keystone" AND Host = "%" AND Db = "keystone";
+------+----------+----------+-------------+-------------+-------------+-------------+-------------+-----------+------------+-----------------+------------+------------+-----------------------+------------------+------------------+----------------+---------------------+--------------------+--------------+------------+--------------+
| Host | Db       | User     | Select_priv | Insert_priv | Update_priv | Delete_priv | Create_priv | Drop_priv | Grant_priv | References_priv | Index_priv | Alter_priv | Create_tmp_table_priv | Lock_tables_priv | Create_view_priv | Show_view_priv | Create_routine_priv | Alter_routine_priv | Execute_priv | Event_priv | Trigger_priv |
+------+----------+----------+-------------+-------------+-------------+-------------+-------------+-----------+------------+-----------------+------------+------------+-----------------------+------------------+------------------+----------------+---------------------+--------------------+--------------+------------+--------------+
| %    | keystone | keystone | Y           | Y           | Y           | Y           | Y           | Y         | N          | N               | Y          | Y          | N                     | N                | N                | N              | N                   | N                  | N            | N          | N            |
+------+----------+----------+-------------+-------------+-------------+-------------+-------------+-----------+------------+-----------------+------------+------------+-----------------------+------------------+------------------+----------------+---------------------+--------------------+--------------+------------+--------------+
```

This is not changed by our default `grant` action, And the check sets `incorrect_privileges` if any of those privileges is set to 'N': https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/database/libraries/provider_database_mysql_user.rb#L75

rsalevsky: It also just adds privileges but doesn't remove them. In my opinion we should combine both checks and also take care of removing privileges.

jsuchome: Well, removing that SSL is not an easy action, I fear

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -63,25 +63,22 @@ def action_grant
...
-          results.each do |result|
-            new_resource.privileges.each do |privileges|
-              key = "#{privileges.capitalize}_priv"
-              incorrect_privileges = true if privileges[key] != "Y"
+          # test the presence
+          privs = {}
+          new_resource.privileges.each { |p| privs[p] = false }
+          privs["REQUIRE SSL"] = false if new_resource.require_ssl
+
+          client.query("show grants for #{new_resource.username}").do |row|
```
rsalevsky: Please keep the prepare and execute if possible or escape the username. Additional it would be nice to use the capital syntax to be equal with the other statements.

jsuchome: I can add quotes and capital syntax, but I was not able to do it with prepare/execute. Glad to change it of you know correct way.

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -63,25 +63,22 @@ def action_grant
...
-            new_resource.privileges.each do |privileges|
-              key = "#{privileges.capitalize}_priv"
-              incorrect_privileges = true if privileges[key] != "Y"
+          # test the presence
+          privs = {}
+          new_resource.privileges.each { |p| privs[p] = false }
+          privs["REQUIRE SSL"] = false if new_resource.require_ssl
+
+          client.query("SHOW GRANTS FOR \"#{new_resource.username}\"").do |row|
+            row.each do |grant, val|
```
rsalevsky: Another approach for this would be:

```ruby
privileges = []
grants.each { |row| grant.push(row["Grants for #{new_resource.username}@%"].split) }
privileges.flatten!

privs.keys.each do |priv|
  privs[priv] = true if val.include? priv
end
```

It would save one loop but I'm not sure if it is the cleaner method.

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -63,25 +63,22 @@ def action_grant
...
-          results.each do |result|
-            new_resource.privileges.each do |privileges|
-              key = "#{privileges.capitalize}_priv"
-              incorrect_privileges = true if privileges[key] != "Y"
+          # test the presence
+          privs = {}
+          new_resource.privileges.each { |p| privs[p] = false }
+          privs["REQUIRE SSL"] = false if new_resource.require_ssl
+
+          client.query("SHOW GRANTS FOR \"#{new_resource.username}\"").do |row|
```
rsalevsky: `client.query("SHOW GRANTS FOR #{client.escape(new_resource.username)}")`

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -63,25 +63,22 @@ def action_grant
...
+            row.each do |grant, val|
+              privs.keys.each do |priv|
+                privs[priv] = true if val.include? priv
+              end
```
rsalevsky: This would we an alternate approach to save one loop:

```ruby  
client.query("SHOW GRANTS FOR \"#{new_resource.username}\"").do |row|
  privs.keys.each do |priv|
    privs[priv] = true if ["Grants for #{user}@%"].include? priv
    end
end
```

jsuchome: There's something missing in `["Grants for #{user}@%"].include?` ...

jsuchome:
In the end, it might be easier to just grant always without complicated checking...

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
rsalevsky: Oh, true it should be `row["Grants for #{user}@%"].include?`.

jsuchome:
@rsalevsky what about now?

PR #1309, opened 2017-09-22
database: Rename all mysql folders to mariadb

rsalevsky:
This commit changes /var/lib/mysql(tmp) to /var/lib/mariadb(tmp). The
goal of this commit is make the mariadb integration more aligned with
the 10.2 release which depricates the use of the mysql string.


jsuchome:
default certificate locations - `/etc/mysql/ssl/`

PR #1310, opened 2017-09-22
database: Remove require ssl from user

rsalevsky:
In case we disable SSL for the database we want to remove also the
require for the user to allow non-SSL connections again.


rsalevsky:
This needs some testing.

jsuchome:
Looks correct itself, but I think we should combine it with https://github.com/crowbar/crowbar-openstack/pull/1307

rsalevsky:
@jsuchome Feel free to cherry-pick it and close this PR.

jsuchome:
Part of https://github.com/crowbar/crowbar-openstack/pull/1307 now

PR #1312, opened 2017-09-25
magnum: Use credential env to setup domain role

sayalilunkad:
In a1accbaf we switched out the credentials in the command flags for
environment variables, but we forgot to use the environment in one of
the magnum commands. This was causing the magnum domain admin to not
have a role in its own domain and causing our tests to fail when trying
to create a cluster template. This patch adds the environment setting to
the last mixlib call so that it works the same as the other shellouts.

(cherry picked from commit e58608a66c6faad7936dda329fbcfb54dcef1abc)


sayalilunkad:
The gating failure is moves past the failure we get without this patch which which can be found in the failing jobs here https://ci.suse.de/view/Cloud/view/Cloud7/job/cloud-mkcloud7-job-magnum-x86_64/
So this solves the problem at hand and can be merged.

sayalilunkad:
@dirkmueller can you please merge this?

PR #1314, opened 2017-09-25
mysql: Use [:database][:mysql] attributes rather than just [:mysql]

jsuchome:
Use correct attribute hierarchy that is located under [:database] [bsc#1058876](https://bugzilla.suse.com/show_bug.cgi?id=1058876).

Not tested. Not sure if it is enough.

Does not touch postgresql parts (we should drop postgresql anyway...)


rhafer:
> Does not touch postgresql parts (we should drop postgresql anyway...)

I think we should try to keep postgresql working for now if possible.

rhafer:
I am closing this in favor of https://github.com/crowbar/crowbar-core/pull/1355 and https://github.com/crowbar/crowbar-openstack/pull/1352 for now. We can reopen it if decide to go for the renaming approach.

PR #1320, opened 2017-09-26
neutron-l3-ha-service: don't capture SIGTERM

abel-navarro:
The service captures SIGTERM when migrating routers from dead agents.
This is a process that can take long time and can prevent Pacemaker to
properly shutdown the service leading to node fencing.

This change delegates the SIGTERM handling to neutron-ha-tool, which
does so in a per router basis instead of for the whole batch.

Reference: https://bugzilla.suse.com/show_bug.cgi?id=1059530

### DIFF
--- a/chef/cookbooks/neutron/files/default/neutron-l3-ha-service.rb
+++ b/chef/cookbooks/neutron/files/default/neutron-l3-ha-service.rb
@@ -30,8 +30,6 @@ class HAToolLog
 end
 
 def main
-  terminator = Terminator.new
-  terminator.register_term_signal_handler
   service_options = ServiceOptions.load ARGV[0]
   ha_functions = HAFunctions.new(service_options)
   error_counter = ErrorCounter.new service_options.max_errors_tolerated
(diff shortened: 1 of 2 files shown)


abel-navarro:
[Make neutron-l3-ha-service respond to TERM signal as quickly as possible](https://trello.com/c/O2sWBMIc/142-make-neutron-l3-ha-service-respond-to-term-signal-as-quickly-as-possible)

abel-navarro:
Related: https://github.com/SUSE-Cloud/cookbook-openstack-network/pull/35

abel-navarro:
After being enlighted by @matelakat , this change is actually not needed. The original behavior was already handling the situation where a SIGTERM is received and both the service and the tool receive that signal. With this change we just prevented the service to handle the signal, not the tool and that was causing problems. The service will stop as soon as the tool stops, so no need for this at all.

matelakat:
[investigate - neutron-l3-ha-service: will TERM signal reach the python script](https://trello.com/c/E64VmRtZ/144-investigate-neutron-l3-ha-service-will-term-signal-reach-the-python-script)

PR #1323, opened 2017-09-26
neutron: Fixes for ACI integration - updates for Newton and later

mmnelemane:
This commit fixes minor issues encountered while testing ACI
integration with SUSE OpenStack Cloud 7. The ACI changes were
so far only being tested for Cloud 6 on the Cisco Lab.
Also recent changes in the crowbar framework caused failures while
applying the barclamp by enabling APIC as the ML2 backend.
The fixes include:
 - config setting for of_interface
 - modified path for linking ml2_conf_cisco_apic.ini
These changes were tested with Cloud7 on the Cisco ACI lab for L2 and L3
functionality and also shared with the Sales Engineering team.
So the commit needs to be merged with Cloud7 so PTFs can be
generated if required by customers wanting to upgrade from
Cloud6 to Cloud7. The changes are not tested against Cloud8

### DIFF
--- a/chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
+++ b/chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
@@ -70,6 +70,9 @@ if node.roles.include?("nova-compute-kvm")
   include_recipe "neutron::common_config"
 
   # Agent configurations for Cisco APIC driver
+  # The ACI setup for OpenStack releases before Pike use "of_interface" options
+  # set to "ovs-ofctl". This option has been deprecated in Pike and removed
+  # from this config file for Pike. It is still included in Newton (Cloud7)
   agent_config_path = "/etc/neutron/plugins/ml2/openvswitch_agent.ini"
   template agent_config_path do
     cookbook "neutron"
@@ -80,8 +83,10 @@ if node.roles.include?("nova-compute-kvm")
     variables(
       ml2_type_drivers: ml2_type_drivers,
       tunnel_types: "",
+      enable_tunneling: false,
       use_l2pop: false,
       dvr_enabled: false,
+      ovsdb_interface: neutron[:neutron][:ovs][:ovsdb_interface],
       bridge_mappings: ""
     )
   end
--- a/chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
@@ -3,10 +3,11 @@
 <% unless @tunnel_types.empty? -%>
 tunnel_types = <%= @tunnel_types.join(",") %>
 <% end -%>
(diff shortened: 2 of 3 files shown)


chef/cookbooks/neutron/templates/default/openvswitch_agent.ini.erb
```diff
@@ -3,10 +3,11 @@
 <% unless @tunnel_types.empty? -%>
 tunnel_types = <%= @tunnel_types.join(",") %>
 <% end -%>
-<% if @use_l2pop -%>
-l2_population = True
+<% if @ml2_type_drivers.include?("cisco_apic_ml2") || @ml2_type_drivers.include?("apic_gbp") -%>
+enable_tunneling = <%= @enable_tunneling %>
```
dirkmueller: hmm, whats the rationale of this change? iirc tunneling only works with l2_pop enabled. is this fixing something for the case l2_pop disabled?

mmnelemane: @dirkmueller : W.r.t l2pop, I have only removed repeated condition statements and grouped (l2_population and arp_responder) under one condition statement. I understood that l2_pop works only when with tunneling and not the other way (i.e) Tunneling can still be enabled without l2_pop except it will have an impact on the tunneling performance. So IMO enable_tunneling can be independently used irrespective of l2_pop otherwise whats the reason for having separate options. The other change done here is to always disable tunneling in case of ACI usage, since all L2 functions are handled internally by the APIC and doesn't bother about l2_pop.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -80,8 +80,11 @@
     variables(
       ml2_type_drivers: ml2_type_drivers,
       tunnel_types: "",
+      enable_tunneling: false,
       use_l2pop: false,
       dvr_enabled: false,
+      of_interface: "ovs-ofctl",
```
rhafer: I don't see this used anywhere in the template. Am I missing something?

mmnelemane: @rhafer: Thanks for noticing that. I missed it. Corrected now.

mmnelemane:
@rhafer: The changes are copied to "master" just for the process. I intended the changes go into Cloud7.  Its tested only with Cloud7.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
rhafer: @mmnelemane I believe the way you fixed it is not quite correct. AFAIK the "of_interface" settings has been removed from Pike (which "master" is currently based on). I guess you'll unfortunately need differrent patches for master and stable/4.0 here :(

PR #1325, opened 2017-09-27
mysql: tune innodb log size / writeback

dirkmueller:
We're already raising innodb_buffer_pool_size, and log_file_size should
be scaled accordingly (typical recommendation is 25% of the buffer pool
size which is what we're doing here). Also tune writeback to O_DIRECT
as this is what upstream is doing:

https://git.openstack.org/cgit/openstack/openstack-ansible-galera_server/tree/templates/my.cnf.j2#n80


dirkmueller:
@jsuchome probably. we could also leave it out of that commit for now.. it looks like it isn't configurable upstream

jsuchome:
Well, since we have those `.d/` configuration directories, users can always add a new file replacing the option that is not chef-configurable. So it's not really needed.

PR #1327, opened 2017-09-27
nova: reduce excessive node searches on compute role nodes

dirkmueller:
This was only a marginal security feature to begin with, as
you could just copy the other instance ssh key with one hop
more. Created a followup card to make this more secure.

(cherry picked from commit ae1345b916848af7bade0e170c05b0159ba61d55)

Backport of #648 648

### DIFF
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -300,42 +300,30 @@ elsif !node[:nova]["use_shared_instance_storage"]
   end
 end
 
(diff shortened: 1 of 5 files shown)


dirkmueller:
[avoid excessive searches on compute recipe](https://trello.com/c/MJpY601t/146-avoid-excessive-searches-on-compute-recipe)

matelakat:
Gate failure: 
```
apache2: migration-threshold=3 fail-count=1 last-failure='Wed Sep 27 16:42:26 2017'
Error (55): Cluster resources' failures detected
```

re-triggering gate.

PR #1330, opened 2017-09-28
neutron-ha-tool: remove non-used parameters

matelakat:
Given that ha-tool is no longer set up through a custom resource agent,
we no longer need the settings for that resource.


matelakat:
[neutron-l3-ha-service cleanup](https://trello.com/c/izcHFkSp/149-neutron-l3-ha-service-cleanup)

PR #1331, opened 2017-09-28
neutron-l3-ha-service: tweak timeouts

abel-navarro:
We found out that default timeouts for neutron-l3-ha-service were bigger
than Pacemaker default timeout (60s). This lead to a situation where we
tried to gracefully shutdown the service, but since we allowed that much
time, Pacemaker kicked in and fenced the node.

Since Pacemaker will call systemd service to stop the l3-service + tool,
we are reducing the time systemd takes to send a kill signal. Now
systemd will kill the l3-service + tool before Pacemaker tries to fence
the node.

Pacemaker timeout is configured to 60s. We want systemd to send the kill
signal at 55s. The systemd timeout is calculated as the maximum
l3-service :kill timeout + 5s. That's why we are changing the :kill
timeouts to 50s.

### DIFF
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -247,9 +247,9 @@ default[:neutron][:ha][:ports][:server] = 5530
 default[:neutron][:ha][:neutron_l3_ha_resource][:op][:monitor][:interval] = "10s"
 
 default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:terminate] = 300
-default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:kill] = 120
+default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:kill] = 50
 default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:router_migration][:terminate] = 1800
-default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:router_migration][:kill] = 120
+default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:router_migration][:kill] = 50
 default[:neutron][:ha][:neutron_l3_ha_service][:hatool][:program] = "/usr/bin/neutron-ha-tool"
 default[:neutron][:ha][:neutron_l3_ha_service][:hatool][:env] = {}
 default[:neutron][:ha][:neutron_l3_ha_service][:seconds_to_sleep_between_checks] = 10


abel-navarro:
[investigate - set pacemaker timeout for neutron-ha-service](https://trello.com/c/VAF5yLH0/145-investigate-set-pacemaker-timeout-for-neutron-ha-service)

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -246,11 +246,11 @@
 
 default[:neutron][:ha][:neutron_l3_ha_resource][:op][:monitor][:interval] = "10s"
 
-default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:terminate] = 300
-default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:kill] = 120
-default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:router_migration][:terminate] = 1800
-default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:router_migration][:kill] = 120
+default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:terminate] = 45
+default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:kill] = 5
```
matelakat: Actually, only kill timeout is considered when calculating the systemd timeout, so I would set it as 50, and leave the rest un-changed.

The logic used to calculate the max_kill timeout is:
```ruby
  def self.max_kill_timeout(timeout_records)
    timeouts = []
    timeout_records.each do |component, timeout_record|
      timeout_record.each do |operation, timeout|
        timeouts << timeout if operation.to_s == "kill"
      end
    end
    timeouts.max
  end
```

Basically kill timeout is the time the service leaves for the tool to terminate gracefully after the term signal has been sent.

Also please leave a comment here that says we are lowering it to `50` because pacemaker has a fixed timeout of `60` so we need to make sure systemd sends the kill signal before pacemaker times out.

By adjusting the service timeouts we'll end up with:

 - pacemaker `60` <- fixed
 - service `55` max_kill_timeout + 5

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -246,11 +246,11 @@
 
 default[:neutron][:ha][:neutron_l3_ha_resource][:op][:monitor][:interval] = "10s"
 
-default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:terminate] = 300
-default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:kill] = 120
-default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:router_migration][:terminate] = 1800
-default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:router_migration][:kill] = 120
+default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:terminate] = 45
```
matelakat: This terminate has nothing to do with our scenario when we are stopped by pacemaker. This time is used by the service. So when service calls the ha-tool to ask for the status of the cloud, it will give `terminate` seconds to it to finish. If it still hasn't finished, the service will send a TERM signal to the tool, and then the tool will have `kill` seconds to respond to that TERM signal before the service will send the KILL signal.

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -246,11 +246,11 @@
...
+default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:status][:kill] = 5
+default[:neutron][:ha][:neutron_l3_ha_service][:timeouts][:router_migration][:terminate] = 45
```
matelakat: We really don't want to change this value. This is the time the service leaves for the tool to migrate ALL routers away from dodgy agents. Once this time has passed, and the tool is still working, service will send it the TERM signal, and wait `kill` seconds to send the KILL signal and end the tool's run.

matelakat:
pacemaker-systemd-service-tool

abel-navarro:
@nicolasbock after being enlighted by @matelakat we are indeed keeping the old timeouts except the `:kill` ones, which are used to calculate systemd timeouts. I updated the PR description as well as the commit. I hope it answer your question.

nicolasbock:
That's an awesome graph @matelakat !

Itxaka:
@matelakat -1 from not using http://asciiflow.com/ for the flow.

aspiers:
Indeed those sequence diagrams are great! What tool did you use @matelakat ?

matelakat:
> Indeed those sequence diagrams are great! What tool did you use @matelakat ?

I used the standalone version of umlet. http://umlet.com/

PR #1333, opened 2017-09-28
[4.0] Revert ha service restarts

cmurphy:
Originally, crowbar could not restart the postgresql or rabbitmq resource agents. We "fixed" this so that crowbar could properly manage it, but this is really not a safe thing to do for HA services. This PR reverts those commits so that we go back to the original behavior. We don't make either postgresql or rabbitmq highly configurable so neither the original changes nor this revert will have a significant impact on users.

Backport of #1301


vuntz:
@Itxaka fyi, this is the bit that impacts https://github.com/crowbar/crowbar-openstack/pull/1354

PR #1335, opened 2017-09-28
horizon: Explicit set REST_API_REQUIRED_SETTINGS (bsc#1046616)

toabctl:
SOC6 updated local_settings.py via Crowbar. Because of that, the file will
not be updated during a RPM package update (due to "config(noreplace)").
So explicitly set REST_API_REQUIRED_SETTINGS to get the correct settings.

Note: This fix is not needed for SOC8 because SOC7 will put the config
file into local_settings.d/ as a snippet. So local_settings.py will be
updated (if no user edited the file manually) during a SOC7->SOC8 update.

### DIFF
--- a/chef/cookbooks/horizon/templates/default/local_settings.py.erb
+++ b/chef/cookbooks/horizon/templates/default/local_settings.py.erb
@@ -57,6 +57,13 @@ OPENSTACK_HYPERVISOR_FEATURES = {
     'requires_keypair': False,
 }
 
+# explicitly set this (bsc#1046616)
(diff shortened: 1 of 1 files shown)


toabctl:
[P2: 1046616 SOC7 Since upgrade from SOC6 to 7 Horizon has error when you try to upload new images](https://trello.com/c/WBC5zYdA/446-p2-1046616-soc7-since-upgrade-from-soc6-to-7-horizon-has-error-when-you-try-to-upload-new-images)

PR #1337, opened 2017-09-29
glance: fix loading glance metadefs to database

stefannica:
Fixes an error from the previous PR #1284 , which introduced
loading glance metadefs into the glance database: the 'glance-manage
db_load_metadefs' block is never executed because the ruby_block
that ensures single execution is triggered too quickly.

### DIFF
--- a/chef/cookbooks/glance/recipes/registry.rb
+++ b/chef/cookbooks/glance/recipes/registry.rb
@@ -74,7 +74,7 @@ ruby_block "mark node for glance db_sync" do
     node.save
   end
   action :nothing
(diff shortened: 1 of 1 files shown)


stefannica:
On a somewhat related note, isn't [this nova db execute block](https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/nova/recipes/database.rb#L169) in the same situation ?

nicolasbock:
> On a somewhat related note, isn't this nova db execute block in the same situation ?

If I understand this fix correctly then it would certainly seem so.

PR #1340, opened 2017-10-02
[4.0] memcached: Make config non-HA-aware (bsc#1038223)

cmurphy:
Without this patch, the keystone and nova barclamps set their cache
servers to all of the memcached servers in the cluster in
lexicographical order. This is not actually an optimal way to configure
memcached servers since if part of the cluster is down, the memcached
servers living on it will be inaccessible. The python-memcached backend
is not tied to pacemaker and has no way of knowing that the server is
down, so it attempts to connect to each server serially, not attempting
the next one until the first times out. The effect is that any query to
the OpenStack service will take a very long time. This patch fixes the
issue by only using the local memcached server for keystonemiddleware
instead of using all in the cluster. This means every controller in the
cluster will use only its own memcached server, similar to how it would
work if it was using an in-process cache.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -209,17 +209,8 @@ crowbar_pacemaker_sync_mark "create-keystone_database" if ha_enabled
 
 sql_connection = fetch_database_connection_string(node[:keystone][:db])
 
-if ha_enabled
-  memcached_nodes = CrowbarPacemakerHelper.cluster_nodes(node, "keystone-server")
-  memcached_servers = memcached_nodes.map do |n|
-    node_admin_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address
-    "#{node_admin_ip}:#{n[:memcached][:port]}"
-  end
-else
-  node_admin_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
-  memcached_servers = ["#{node_admin_ip}:#{node[:memcached][:port]}"]
-end
-memcached_servers.sort!
+node_admin_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+memcached_servers = ["#{node_admin_ip}:#{node[:memcached][:port]}"]
 
 # we have to calculate max_active_keys for fernet token provider
 # http://docs.openstack.org/admin-guide/identity-fernet-token-faq.html# \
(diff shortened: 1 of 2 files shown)


cmurphy:
Cloud8 version is here: https://github.com/crowbar/crowbar-openstack/pull/1341 (not cherry-picked)

cmurphy:
@nicolasbock what performance gains did we get from using multiple cache instances? I can't find information on how configuring memcached in a cluster improves performance, and in fact from https://github.com/memcached/memcached/wiki/Performance#maximum-number-of-nodes-in-a-cluster there is the potential for it to impede performance.

The issue is actually in python-memcached, not oslo.cache: https://github.com/linsomniac/python-memcached/blob/1.58/memcache.py#L444-L448
python-memcached is common to both the dogpile.cache.memcached and oslo_cache.memcache_pool backends of oslo.cache, so I don't think there's any way to tune oslo.cache for this. I guess it's worth noting that I reproduced this with cache backend oslo_cache.memcache_pool but Dirk reported the bug when we were still using dogpile.cache.memcache as the cache backend.

nicolasbock:
> what performance gains did we get from using multiple cache instances?

In a multiprocess keystone setup in which all keystone instances access the same pool of memcached servers I would expect a potential increase in keystone performance because of the shared cache. While you are correct to point out that a `large number of servers can slow down a client` we don't know at what number this happens. And of course it would help if we had some actual benchmarks to show that we do end up with a performance gain. At this point I am just speculating. :smile:

It sounds like the memcached design does not consider failed nodes though (I couldn't find anything in their documentation) and waiting for a connection to time out definitely decreases performance. Since you found that we can't tune anything to change that behavior I agree that it's better to simply use the local memcached only.

Thanks for the additional details!

stefannica:
@cmurphy there were two memcached_servers.join in the nova config file

cmurphy:
@stefannica thanks, fixed

cmurphy:
I'm pretty sure the failure here must be caused by this change

cmurphy:
@nicolasbock after doing a lot of reading I understand better what you were saying about performance gains - if we have two separate caches then each controller has to cache everything itself, doubling or tripling the number of writes we have to do. Not ideal.

The HA job is failing here because when the ceph cookbook tries to make a role assignment, it makes two requests, one to GET the role assignments for the ceph user and one to PUT the new role assignment. It was lucky that it was fairly consistent about which controller ended up receiving the requests. One GET for role assignments would go to one controller and produce a cache hit containing just the role _member_, which comes from the ceph user having a default tenant set, and which is not the intended assignment of 'admin'. It would then issue a PUT to try to correct the role assignment and fail with an HTTP 409 because it had already created this role assignment in an earlier chef run, just the request had gone to a different controller and therefore was only cached on that controller.

I think this particular issue could be corrected by using the keystone v3 API for role assignments (which we already do in master) which wouldn't consider a default project to be a role assignment and would therefore have a cache miss and seek the role assignments from the database. But this illustrates the potential for a sort of split-brain problem that is not really acceptable, in addition to the performance hit.

I commented on the bug that I think the problem that prompted this is not really a problem any more since we switched to the memcache_pool backend, so closing this.

PR #1341, opened 2017-10-02
memcached: Make config non-HA-aware (bsc#1038223)

cmurphy:
Without this patch, when deployed in an HA configuration, all of the
barclamps set their cache servers to all of the memcached servers in the
cluster in lexicographical order. This is not actually an optimal way to
configure memcached servers since if part of the cluster is down, the
memcached servers living on it will be inaccessible. python-memcached is
not tied to pacemaker and has no way of knowing that, so it passes the
entire list to python-memcached which attempts to connect to each server
serially, not attempting the next one until the first times out. The
effect is that any query to the OpenStack service will take a very long
time if the first memcached server in the list is down.

This patch fixes the issue by only using the local memcached server
instead of using all in the cluster. This also adjusts the
get_memcached_servers helper method to only accept one node as input
since, knowing what we now know, we're unlikely to need more than one.
The get_memcached_servers method was implemented while updating the
barclamps to prevent deprecation warnings emitted by keystonemiddleware
in Ocata[1] and was mimicking old behavior used to set the cache servers
for keystone and nova.

### DIFF
--- a/chef/cookbooks/aodh/recipes/aodh.rb
+++ b/chef/cookbooks/aodh/recipes/aodh.rb
@@ -147,9 +147,7 @@ end
 node.normal[:apache][:listen_ports_crowbar] ||= {}
 node.normal[:apache][:listen_ports_crowbar][:aodh] = { plain: [bind_port] }
 
-memcached_servers = MemcachedHelper.get_memcached_servers(
-  ha_enabled ? CrowbarPacemakerHelper.cluster_nodes(node, "aodh-server") : [node]
-)
+memcached_servers = MemcachedHelper.get_memcached_servers([node])
 
 memcached_instance("aodh-server")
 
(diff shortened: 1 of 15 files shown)


cmurphy:
Cloud7 version is here: https://github.com/crowbar/crowbar-openstack/pull/1340 (not cherry-picked)

dirkmueller:
NoMethodError: undefined method `get_memcached_servers' for MemcachedHelper:Module

we need some other patch elsewhere? 

where is the code that sets up memchached replication? basically when starting the memcached we need to tell it which one is the replication master and how to reach it.

cmurphy:
@dirkmueller I had forgotten that the swift barclamp was still using that method name.

I decided that since swift is special I would rather leave it alone, so I changed the method name back to plural and reverted the variable name changes, so this patch is a lot smaller now.

cmurphy:
Closing for the reasons given here: https://github.com/crowbar/crowbar-openstack/pull/1340#issuecomment-336167666

PR #1344, opened 2017-10-04
nova: enable default filters explicitly along with additonal filters …

aagate:
…for tempest tests

Tracking down the following tempest test failure:
```
tempest.api.compute.admin.test_servers_on_multinodes.ServersOnMultiNodesTest.test_create_servers_on_sam
e_host
```
Since SameHostFilter is not enabled in nova.conf, even though the tempest test
makes create server call with hints set to:
```
{ same_host: "..." }
```
The same host hint is ignored and nova scheduler schedules the instance creation on any
compute host. (See [1] to for more on filters)

This change enables default set of filters by setting them explicitly
in the nova.conf property file and also enables
SameHostFilter and DifferentHostFilter for tempest tests.

[1] https://docs.openstack.org/nova/latest/user/filter-scheduler.html


stefannica:
Kicked off another CI HA job. This time it should pass, as most problems have been addressed during the jenkins green day.

PR #1346, opened 2017-10-04
Extend DB sync timeouts  (bsc#1060421)

dirkmueller:
It looks like it takes just above 3 minutes to complete using mysql/galera, so we need to raise limits.

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -258,7 +258,7 @@ ha_enabled = node[:neutron][:ha][:server][:enabled]
 crowbar_pacemaker_sync_mark "wait sync mark for neutron db sync" do
   mark "neutron_db_sync"
   action :wait
-  timeout 120
+  timeout 300
   only_if { ha_enabled }
 end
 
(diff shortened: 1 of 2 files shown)


dirkmueller:
Note: this was raised as a Milestone 3 release blocker, so we need to get this in relatively short and respin Milestone3.

aplanas:
The real reason of the timeout is that galera / mariadb is doing one fsync / fdatasync before each close, and qa2 and qa3 have very slow SSDs.

We compared the performance between qa[23] and qa4 and we found a two order of magnitude difference in the f[data]sync syscall, using a small C program and galera itself. So maybe an alternative solution is configure innodb to do not do fsync so frequently in qa[23] (via mkcloud)

dirkmueller:
@aplanas https://github.com/crowbar/crowbar-openstack/pull/1325 should help with that problem. It is somewhat not understandable to me why we haven't done that before.

aplanas:
@dirkmueller I am not sure that #1325 will help. O_DIRECT will do not decrease the number of `fsync`. Maybe `O_DIRECT_NO_FSYNC` or the risky `nosync` will do (via mkcloud, of course, and only for qa[23])

PR #1347, opened 2017-10-04
tempest: using guid for flavor ids to avoid any conflicts    

sjamgade:
tempest: using guid for flavor ids to avoid any conflicts  for a long long time.                                        
Resolves conflict with the ids in nova::flavor

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -272,9 +272,10 @@ EOH
   not_if { File.exist?(machine_id_file) }
 end
 
-flavor_ref = "6"
-alt_flavor_ref = "7"
-heat_flavor_ref = "8"
+# pre-generated guid, to avoid conflicts
+flavor_ref = "2ace2042-a736-4114-af67-61fef1b80d5a"
+alt_flavor_ref = "bf9d7697-a348-4e1a-b983-753698e33596"
+heat_flavor_ref = "a9d20c85-9ef6-41f2-a833-0367f973bca2"
 
 bash "create_yet_another_tiny_flavor" do
   code <<-EOH


Itxaka (approved):

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -272,9 +272,10 @@
   not_if { File.exist?(machine_id_file) }
 end
 
-flavor_ref = "6"
-alt_flavor_ref = "7"
-heat_flavor_ref = "8"
+# pre-generated guid, to avoid conflicts
```
dirkmueller: hmm.. I would have expected to just deduplicate the ids (by e,.g. adding a 10 or so), as generating a fixed GUID is going to create potentially a conflict. can we change the code to live without hardcoded ids? I'm not sure what we're actually using them for, maybe the name reference is good enough?

Itxaka: why not use on-the-fly guid creation instead? 
As long as we need it only in this recipe then we can get awya with it
```
require 'securerandom'
flavor_ref = SecureRandom.uuid
alt_flavor_ref = SecureRandom.uuid
heat_flavor_ref = SecureRandom.uuid
```

sjamgade: I do not want to of import a new package/module, given that we consume a decent memory overhead with chef-client.

I think it will be really very difficult to regenerate a Globally-Unique-Identifier

sjamgade:
i got [this](https://stackoverflow.com/questions/39771/is-a-guid-unique-100-of-the-time)
if you want I can still change it

chef/cookbooks/tempest/recipes/config.rb
nicolasbock: You could also let `nova` generate the UUID for you and `nova flavor-create tempest-stuff auto ...`.

sjamgade: that was I my first though too, but the this `id` needs to be extracted back and stored in tempest conf file.

PR #1348, opened 2017-10-04
nova: improve flavors creation performance

sjamgade:
- reusing the output of flavor-list
- subscribes and not_if are independent
  as a result the block will always try to create the flavors even if
they exist

This saves about 15-20 secs on non-shared hardware and about 1min on a
loaded CI

### DIFF
--- a/chef/cookbooks/nova/recipes/flavors.rb
+++ b/chef/cookbooks/nova/recipes/flavors.rb
@@ -79,22 +79,28 @@ env = "OS_USERNAME='#{keystone_settings["service_user"]}' "
 env << "OS_PASSWORD='#{keystone_settings["service_password"]}' "
 env << "OS_PROJECT_NAME='#{keystone_settings["service_tenant"]}' "
 env << "OS_AUTH_URL='#{keystone_settings["internal_auth_url"]}' "
-env << "OS_REGION_NAME='#{keystone_settings["endpoint_region"]}'"
+env << "OS_REGION_NAME='#{keystone_settings["endpoint_region"]}' "
+env << "OS_IDENTITY_API_VERSION=#{keystone_settings["api_version"]}"
 novacmd = "#{env} nova"
+openstack = "#{env} openstack"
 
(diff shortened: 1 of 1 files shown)


Itxaka:
Looks great and indeed its a big time saver @sjamgade 

Im thinking out loud here but wouldnt it be better to move from the `nova` command to the `openstack` one so its much easier to extract the info due to the new formats and colums present in the new command:

```
root@d52-54-77-77-01-01:~ # openstack flavor list -c ID -c Name -f value
1 m1.tiny
10 m1.trusted.medium
100 manila-service-image-flavor
11 m1.trusted.large
12 m1.trusted.xlarge
2 m1.small
3 m1.medium
4 m1.large
5 m1.xlarge
6 tempest-stuff
7 tempest-stuff-2
8 m1.trusted.tiny
9 m1.trusted.small
```

nicolasbock:
I agree with @Itxaka. Using the `openstack` command also has the advantage that it's well, the `openstack` command and not some services CLI comand. We should be migrating to it whenever we can.

sjamgade:
ha failures are unrelated.

chef/cookbooks/nova/recipes/flavors.rb
```diff
@@ -79,30 +79,35 @@
 env << "OS_PASSWORD='#{keystone_settings["service_password"]}' "
 env << "OS_PROJECT_NAME='#{keystone_settings["service_tenant"]}' "
 env << "OS_AUTH_URL='#{keystone_settings["internal_auth_url"]}' "
-env << "OS_REGION_NAME='#{keystone_settings["endpoint_region"]}'"
+env << "OS_REGION_NAME='#{keystone_settings["endpoint_region"]}' "
+env << "OS_IDENTITY_API_VERSION=#{keystone_settings["api_version"]}"
 novacmd = "#{env} nova"
+openstack = "#{env} openstack"
```
Itxaka: nit: `openstackcmd` ?

PR #1349, opened 2017-10-04
monasca: enable libvirt monitoring

jgrassler:
This pull request enables libvirt monitoring for `monasca-agent`. I haven't tested it, yet (first I need to get monasca-setup working on `stable/ocata`...), so very much WIP for now...

### DIFF
--- a/crowbar_framework/config/locales/monasca/en.yml
+++ b/crowbar_framework/config/locales/monasca/en.yml
@@ -35,6 +35,7 @@ en:
           insecure: 'Do you want insecure connection?'
           ca_file: 'Sets the path to the ca certs file if using certificates. Required only if insecure is set to False (ca_file)'
           log_level: 'Log level'
+          monitor_libvirt: 'Whether to monitor the libvirt process on compute nodes'
           statsd_port: 'Monasca Statsd port'
           check_frequency: 'Time to wait between collection runs (check_frequency)'
           num_collector_threads: 'Number of Collector Threads to run (num_collector_threads)'
--- a/chef/cookbooks/monasca/attributes/default.rb
+++ b/chef/cookbooks/monasca/attributes/default.rb
@@ -27,6 +27,7 @@ default[:monasca][:api][:bind_port] = 8070
 default[:monasca][:log_api][:bind_port] = 5607
 
 default[:monasca][:kibana][:bind_port] = 5601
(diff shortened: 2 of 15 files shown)


crowbar_framework/config/locales/monasca/en.yml
```diff
@@ -35,6 +35,7 @@ en:
           insecure: 'Do you want insecure connection?'
           ca_file: 'Sets the path to the ca certs file if using certificates. Required only if insecure is set to False (ca_file)'
           log_level: 'Log level'
+          monitor_libvirt: 'Whether to monitor the libvirt process on compute nodes (defaults to True)'
```
dirkmueller: Enable monitoring of libvirtd on compute nodes

IMHO remove the "default" part, we don't have that for any other exposed option.

jgrassler:
Of course I'll squash and clean up the commit message once I'm done. First of all I'll need to get this running, though.

On a more general note, this is definitely not ready for comment, yet. I just opened the pull request so I could have something to link to already (maybe we should introduce a "do not comment, yet" label for this purpose.

toabctl:
> On a more general note, this is definitely not ready for comment, yet. I just opened the pull request so I could have something to link to already (maybe we should introduce a "do not comment, yet" label for this purpose.

Just don't create the PR if you don't want comments :-)

jgrassler:
| Just don't create the PR if you don't want comments :-)

Yeah...then there will be people who will complain about a link to a plain branch not being a PR. Damned if you do, damned if you don't :-)

jgrassler:
Ok, ready for review as far as I'm concerned. Please do not merge, yet. I haven't tested this with all the Hound related changes in place, yet, so I still want to put the current state of the request through its paces on a fresh cloud.

jgrassler:
Yep. Addressing Hound comments did introduce a bug. Thanks, Hound. Let me quickly fix it...

jgrassler:
Ok, tested and works (and Hound is happy, too). So feel free to merge right away if you end up +1ing.

jgrassler:
Ok, I added two fixup commits now: without these the `node[:monasca][:agent][:monitor_libvirt]` setting would have been ignored when configuring the agent side. Also, before these the controller side HTTP and databese checks would also have gotten enabled on the compute nodes.

jgrassler:
Sure. Commits squashed.

PR #1353, opened 2017-10-06
Extend timeouts for db syncs

dirkmueller:
This should help with initial setup of database in galera/mysql case


stefannica:
The HA failure is not related (see [this pacemaker bug](https://bugzilla.suse.com/show_bug.cgi?id=1059187) already fixed in soc8)

PR #1264, opened 2017-09-13
revert commit 0dd6cc373e97832e0738bd2c8fe537eb4782cb7f 

Itxaka:
revert commit https://github.com/crowbar/crowbar-openstack/commit/0dd6cc373e97832e0738bd2c8fe537eb4782cb7f as the underlying problem has been fixed

PR #1354, opened 2017-10-06
[4.0] Backport rabbit cleanup

Itxaka:
Revert a commit that its not necessary anymore ( rabbitmq: dont let the template changes restart if in cluster mode )

Fixes: rabbitmq: use proper stop/start cmd for cluster

Backport of https://github.com/crowbar/crowbar-openstack/pull/1264

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -21,8 +21,6 @@
 ha_enabled = node[:rabbitmq][:ha][:enabled]
 # we only do cluster if we do HA
 cluster_enabled = node[:rabbitmq][:cluster] && ha_enabled
-# dont let the changes to the templates restart the rabbitmq in cluster mode
-service_action = cluster_enabled ? :nothing : :restart
 quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
 
 cluster_partition_handling = if cluster_enabled
@@ -94,7 +92,7 @@ template "/etc/rabbitmq/rabbitmq-env.conf" do
   owner "root"
   group "root"
   mode 0644
-  notifies service_action, "service[rabbitmq-server]"
+  notifies :restart, "service[rabbitmq-server]"
 end
 
 `systemd-detect-virt -v -q`
@@ -111,7 +109,7 @@ template "/etc/rabbitmq/rabbitmq.config" do
     cluster_partition_handling: cluster_partition_handling,
     hipe_compile: hipe_compile
   )
-  notifies service_action, "service[rabbitmq-server]"
+  notifies :restart, "service[rabbitmq-server]"
 end
 
 # create a file with definitions to load on start, to be 100% sure we always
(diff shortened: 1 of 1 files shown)


chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -66,12 +66,16 @@
   environment "HOME" => "/root/"
   code "#{rabbitmq_plugins} #{rabbitmq_plugins_param} enable rabbitmq_management > /dev/null"
   not_if "#{rabbitmq_plugins} list -E | grep rabbitmq_management -q", environment: {"HOME" => "/root/"}
-  notifies service_action, "service[rabbitmq-server]"
+  notifies :restart, "service[rabbitmq-server]"
 end
 
 service "rabbitmq-server" do
-  supports restart: true, start: true, stop: true, status: true, \
-           restart_crm_resource: true, pacemaker_resource_name: "rabbitmq"
```
vuntz: Hmm, don't we need to possibly keep these attributes? They got dropped in master with https://github.com/Itxaka/crowbar-openstack/commit/2600ec22e483bd937663d1eaf194009ccebfa800 but I somehow think they will be needed for this change to really work.

Itxaka: Im checking it

Itxaka: According to the PR you linked, then we dont need that anymore 👍

vuntz: Well, I'm not sure ;-) That PR was created when there was no native clustering, I think. Especially the `pacemaker_resource_name: "rabbitmq"` bit may matter: I don't understand how pacemaker can promote/demote without this.

Itxaka: I think Pacemaker already knows the name of the service due to https://github.com/crowbar/crowbar-ha/blob/stable/4.0/chef/cookbooks/crowbar-pacemaker/providers/service.rb#L181

vuntz: Sure. My point is that `rabbtimq-server` != `rabbitmq` (service name != pacemaker resource name)

Itxaka: Wait, so if https://github.com/crowbar/crowbar-openstack/pull/1333 has also been merged to 4.0 then this should work the same as it works on master now right @vuntz ?

rsalevsky:
Needs rebase

chef/cookbooks/rabbitmq/recipes/default.rb
Itxaka: So this is actually ok, because even with the fix on master we do not allow postgres or rabbit to be restarted if they are on HA as this can probably break the cluster in so many ways.

The only thing is, for rabbit this works by having a mismatch on the service name VS the actual resource name so the restarts are "trigerred" but never executed.

This is how it works on master but that should be fixed. What about merging this then doing the fix on master+backport so we have the branches properly synced?

vuntz: > What about merging this then doing the fix on master+backport so we have the branches properly synced?

deal

PR #1355, opened 2017-10-06
[4.0] Fix neutron ha service logging

abel-navarro:
backport from https://github.com/crowbar/crowbar-openstack/pull/1343

The neutron-l3-ha-service will log to:
`/var/log/neutron/neutron-l3-ha-service.log`
With a daily rotation, to avoid polluting messages


abel-navarro:
[Bug 1059460 neutron-l3-ha-service logging fix](https://trello.com/c/GfalPbOW/116-bug-1059460-neutron-l3-ha-service-logging-fix)

PR #1356, opened 2017-10-09
neutron: fix neutron_default_networks HA timeout error

stefannica:
Fixes the error introduced in #1338 that causes a consistent timeout error in the neutron barclamp:

RuntimeError: crowbar-pacemaker_sync_mark[wait sync mark for neutron default networks] (neutron::server line 451) had an error: RuntimeError: Cluster founder didn't set neutron_default_networks!\n"}

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -448,9 +448,9 @@ if node[:neutron][:create_default_networks]
   # take a long time. Otherwise, the founder node will be delayed and
   # might produce pacemaker sync timeouts in other recipes where such
   # a big delay isn't expected (e.g. the network_agents recipe).
-  crowbar_pacemaker_sync_mark "wait sync mark for neutron default networks" do
+  crowbar_pacemaker_sync_mark "sync mark for neutron default networks" do
     mark "neutron_default_networks"
-    action :wait
+    action :sync
     timeout 180
     only_if { ha_enabled }
   end


stefannica:
[RuntimeError: Cluster founder didn't set neutron_default_networks](https://trello.com/c/asp0Lf88/222-runtimeerror-cluster-founder-didnt-set-neutrondefaultnetworks)

toabctl:
And the CI reported that problem already in https://github.com/crowbar/crowbar-openstack/pull/1338 . ;We should not merge stuff that is broken...

rsalevsky:
@toabctl Yep my mistake, sorry. The issue is that the HA gating is not really reliable in the last few weeks and tend to be always red.

toabctl:
> Yep my mistake, sorry. The issue is that the HA gating is not really reliable in the last few weeks and tend to be always red.

np. I did the same mistake a couple of times. just want to remind that we should look at the CI results even if a job is not reliable.

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -450,7 +450,7 @@
   # a big delay isn't expected (e.g. the network_agents recipe).
   crowbar_pacemaker_sync_mark "wait sync mark for neutron default networks" do
```
rhafer: Would you mind removing the "wait" from the resource's name as well?

I think it's a bit confusing as ```crowbar_pacemaker_sync_mark``` usually sets the default action depending on the name of the resource. Names starting with ```wait-``` get the ```:wait``` action. Likewise for names starting with ```sync-```. See here: https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/providers/sync_mark.rb#L21 for details.

stefannica:
@rhafer done, thanks

PR #1357, opened 2017-10-09
neutron, nova: Revert use of Restart= for ovs and nfs

vuntz:
These services are oneshot, and this is not compatible.

This is not visible in our CI, due to the fact that the code path is
only used for some exotic configuration.


toabctl:
Backport for stable-4.0 is https://github.com/crowbar/crowbar-openstack/pull/1359

PR #1359, opened 2017-10-09
neutron, nova: Revert use of Restart= for ovs and nfs

toabctl:
These services are oneshot, and this is not compatible.

This is not visible in our CI, due to the fact that the code path is
only used for some exotic configuration.

(cherry picked from commit 28cf45c4519f43bab7a004fa3c7e16006c9d7250)


toabctl:
Merged https://github.com/crowbar/crowbar-openstack/pull/1358

PR #1360, opened 2017-10-09
ceilometer: add separate polling configuration file

aagate:
In Ocata use of pipeline.yaml for polling by ceilometer
polling agent was deprecated. Instead it is recommended
to add a separate polling section in ceilometer config
file and set cfg_file to a new file called polling.yaml
(See [1] [2] Ocata and Pike release notes)

This is preferred method because it allows to define
a set of meters that can be polled which is different
from meters that are collected via notifications.

The structure of the polling.yaml is similar to
pipeline.yaml but only requires fields necessary for
polling that is interval (polling interval) and
list of patterns for source meters.

[1] Ceilometer Ocata Release Notes:
    https://docs.openstack.org/releasenotes/ceilometer/ocata.html

[2] Ceilometer Pike Release Notes:
    https://docs.openstack.org/releasenotes/ceilometer/pike.html

### DIFF
--- a/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
+++ b/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
@@ -105,3 +105,6 @@ service_region_name= <%= @keystone_settings['endpoint_region'] %>
 
 [publisher_notifier]
 metering_topic=metering
+
+[polling]
+cfg_file=/etc/ceilometer/pipeline.yaml


rsalevsky:
Needs a rebase.

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -124,3 +124,18 @@
     notifies :restart, "service[swift-proxy]"
   end
 end
+
+template "/etc/ceilometer/polling.yaml" do
```
sjamgade: this block will always overwrite changes made by customer.

Since customers are likely the adapt this files to their requirement, we should not overwrite existing file. 
So I would recommend writing this file only if it doesnot exist.

aagate: Good point (getting overwritten by chef) will be true for all config files (pipeline.yaml, event_pipeline.yaml, polling.yaml etc) that are managed by chef.  Will make that change for all. Thanks!

One other alternative will be to expose these list of meters for each of the yaml property files via crowbar proposal mechanism, but that will be a lot more work.

sjamgade:
or 
may be add a `boolean` accessible in raw mode to overwrite the file or not.

PR #1361, opened 2017-10-10
keystone: Fix updating admin password

cmurphy:
If the keystone admin password is updated through the keystone barclamp,
the change happens at converge time and the chef node object is also
updated at converge time. This does not help other recipes that have
already evaluated the auth hash at compile time. This patch ensures the
update stores the entire credentials hash on the node object, and then
uses the lazy syntax to evaluate that value at converge time so that an
update to the password during the keystone proposal run will propagate
to the other proposals.

### DIFF
--- a/chef/cookbooks/aodh/recipes/aodh.rb
+++ b/chef/cookbooks/aodh/recipes/aodh.rb
@@ -67,10 +67,6 @@ end
 
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
-register_auth_hash = { user: keystone_settings["admin_user"],
-                       password: keystone_settings["admin_password"],
-                       project: keystone_settings["admin_project"] }
-
 my_admin_host = CrowbarHelper.get_host_for_admin_url(node, ha_enabled)
 my_public_host = CrowbarHelper.get_host_for_public_url(
   node, node[:aodh][:api][:protocol] == "https", ha_enabled)
@@ -82,7 +78,7 @@ keystone_register "register aodh user" do
   insecure keystone_settings["insecure"]
   host keystone_settings["internal_url_host"]
   port keystone_settings["admin_port"]
-  auth register_auth_hash
+  auth lazy { node[:keystone][:admin][:credentials] }
   user_name keystone_settings["service_user"]
   user_password keystone_settings["service_password"]
   project_name keystone_settings["service_tenant"]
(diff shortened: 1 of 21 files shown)


cmurphy:
To address https://trello.com/c/iH2KAeIn

cmurphy:
This is not tested yet but @Itxaka do you think my reasoning makes sense?

stefannica:
@cmurphy any idea why this is only happening if HA is enabled ?

cmurphy:
Never mind, this doesn't work at all :(

cmurphy:
I think I was on the right track - this now works for keystone + swift. I still don't have a good answer for why this is only a problem on HA scenarios.

Itxaka:
> If the keystone admin password is updated through the keystone barclamp,
the change happens at converge time 

This is not really true, the action for changing the password is being done on the compilation phase: https://docs.chef.io/resource_common.html#run-in-compile-phase

```
keystone_register "update admin password" do
      protocol node[:keystone][:api][:protocol]
      insecure keystone_insecure
      host my_admin_host
      port node[:keystone][:api][:admin_port]
      auth register_auth_hash
      user_name node[:keystone][:admin][:username]
      user_password updated_password
      tenant_name node[:keystone][:admin][:tenant]
      action :nothing
    end.run_action(:add_user)
```

The reason this may not work on HA...no idea tbh, I kind of remember testing it and it working but I may have not deployed swift (nor the tests may have deployed it??) but it may be the same reason as why it didnt work with tempest and @cmurphy is on the rigth track using `lazy` on the auth_hash/credentials so they dont get "cached"

cmurphy:
> This is not really true, the action for changing the password is being done on the compilation phase: https://docs.chef.io/resource_common.html#run-in-compile-phase

Oh, you're right - so I wonder if maybe it's just being run in the wrong order, even though it was happening in the same phase?

cmurphy:
Running https://github.com/SUSE-Cloud/automation/pull/2242 with this PR and hacloud=1, will update the commit message after that if all goes well.

Itxaka:
> Oh, you're right - so I wonder if maybe it's just being run in the wrong order, even though it was happening in the same phase?

I may as well be, the only reason it should work rigth now is due to the recipes ordering, because keystone comes first in the recipe run list. So yeah, maybe thats the issue with HA :/

nicolasbock:
If it's the ordering because several nodes execute chef-client than this would need a sync mark, wouldn't it?

cmurphy:
@nicolasbock but sync marks happen during the converge phase don't they? And I think all this is taking place during the compile phase?

rsalevsky:
Needs rebase.

stefannica:
This has been replaced by #1365 and #1389

PR #1364, opened 2017-10-11
crowbar-pacemaker: clone restarts by default

abel-navarro:
When STONITH is enabled in a cluster, the default behavior for the
primitives is to fence the node on failure. This is a desirable behavior
for stateful services because it can lead to a corrupt state. On
stateless services (those that we can clone) fencing may be a too
intrusive action. We are setting the default action for clones to
restart instead of [implicit] fencing.

### DIFF
--- a/chef/cookbooks/crowbar-openstack/definitions/openstack_pacemaker_controller_clone_for_transaction.rb
+++ b/chef/cookbooks/crowbar-openstack/definitions/openstack_pacemaker_controller_clone_for_transaction.rb
@@ -26,6 +26,9 @@ define :openstack_pacemaker_controller_clone_for_transaction,
   raise "No agent specified for #{primitive_name}!" if agent.nil?
 
   fake_params = {}
+  unless op["monitor"].nil?
+    op["monitor"] = op["monitor"].merge("on-fail" => "restart")
+  end
 
   pacemaker_primitive primitive_name do
     agent agent


abel-navarro:
[Rework on-fail for stop actions setting in pacemaker](https://trello.com/c/Jkr9NDFX/112-rework-on-fail-for-stop-actions-setting-in-pacemaker)

chef/cookbooks/crowbar-openstack/definitions/openstack_pacemaker_controller_clone_for_transaction.rb
```diff
@@ -26,10 +26,11 @@
   raise "No agent specified for #{primitive_name}!" if agent.nil?
 
   fake_params = {}
+  op_onfail = {'monitor' => { 'interval' => '10s', 'on-fail' => 'restart' } }
 
   pacemaker_primitive primitive_name do
     agent agent
-    op op
+    op op.empty?? op_onfail : op
```
vuntz: Don't we have most (all?) cookbooks specify a monitor interval? In which case, `op_onfail` would never be used.

vuntz: Btw, we could also use your change to make sure we have a common monitor interval, that is defined only here.

abel-navarro: @vuntz you are right, it is defined in all cookbooks. I was mistakenly looking at a controller information stored in chef, but turned out only the founder node has that attribute set, the other has all blanks, hence my patch. I'll rework it so it takes that into account.

About setting that default monitor interval, should then remove it from the openstack cookbooks? Or contrary-wise, add the `on_fail=restart` to those cookbooks?

abel-navarro: well, mysql service has a [monitor interval of 20s](https://github.com/crowbar/crowbar-openstack/blob/939c78ac6961dc1b9f19141f8c50fcfb115df638/chef/cookbooks/mysql/attributes/server.rb#L38). I guess a common approach is not possible in this case.

abel-navarro:
As agreed, we will make this option configurable. This PR will be obsoleted by the new approach. Link to the new PR as soon as it's ready.

PR #1365, opened 2017-10-11
keystone: store old admin password in node attr for update

stefannica:
Use a node attribute to store a backup copy of the admin password, to be used when the admin password is changed.
This replaces the previous method of managing the admin password change through a dedicated UI `updated_password` attribute and allows the user to change the password by simply modifying the regular `password` UI attribute.
    
The advantage of having a backup copy of the admin password in the node attributes is that the new password value is immediately available after a password change, even during the chef compilation phase. The implications of this:
 - the new password value can immediately be used by all recipes during the chef compilation phase, which removes the need of having to deal with a password change scenario in every openstack cookbook using keystone for authentication
 - reconfiguring keystone with the new password doesn't need to happen in the compilation phase anymore. Doing it in the convergence phase, at the same time as everything else, removes this unnecessary exception from the keystone cookbook and aligns everything together better

### DIFF
--- a/chef/data_bags/crowbar/migrate/keystone/204_remove_updated_password.rb
+++ b/chef/data_bags/crowbar/migrate/keystone/204_remove_updated_password.rb
@@ -1,0 +1,14 @@
+def upgrade(ta, td, a, d)
+  a["admin"].delete("updated_password")
+  nodes = NodeObject.find("roles:keystone-server")
+  nodes.each do |node|
+    node[:keystone][:admin][:old_password] = node[:keystone][:admin][:password]
+    node.save
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["admin"]["updated_password"] = ta["admin"]["updated_password"]
+  return a, d
+end
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -475,6 +475,39 @@ end
 
 keystone_insecure = node["keystone"]["api"]["protocol"] == "https" && node[:keystone][:ssl][:insecure]
 
+register_auth_hash = { user: node[:keystone][:admin][:username],
+                       password: node[:keystone][:admin][:password],
+                       project: node[:keystone][:admin][:project] }
+
+old_password = node[:keystone][:admin][:old_password]
+old_register_auth_hash = register_auth_hash.clone
+old_register_auth_hash[:password] = old_password
+
+keystone_register "update admin password" do
+  protocol node[:keystone][:api][:protocol]
+  insecure keystone_insecure
+  host my_admin_host
+  port node[:keystone][:api][:admin_port]
+  auth old_register_auth_hash
+  user_name node[:keystone][:admin][:username]
+  user_password node[:keystone][:admin][:password]
+  project_name node[:keystone][:admin][:project]
+  action :add_user
+  only_if do
+    node[:keystone][:bootstrap] &&
+      (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)) &&
+      old_password && !old_password.empty? &&
+      old_password != node[:keystone][:admin][:password]
+  end
+end
+
+ruby_block "backup current admin password on node attributes" do
+  block do
+    node.set[:keystone][:admin][:old_password] = node[:keystone][:admin][:password]
+    node.save
+  end
+end
+
 # Creates admin user, admin role and admin project
 execute "keystone-manage bootstrap" do
   command "keystone-manage bootstrap \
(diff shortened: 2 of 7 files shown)


stefannica:
Moved the keystone admin password update in the convergence phase. Preliminary testing on a non-HA setup looks promising.

cmurphy:
Can you propose a change to the automation repo to modify the old password change test for this and reenable the test?

stefannica:
sure @cmurphy, the PR re-enabling the keystone password update is here: https://github.com/SUSE-Cloud/automation/pull/2243

chef/data_bags/crowbar/migrate/keystone/204_remove_updated_password.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  a["admin"].delete("updated_password")
+  a["admin"]["old_password"] = ta["admin"]["password"]
```
cmurphy: old_password is not defined in the schema so the schema migration fails

Itxaka: Im not sure why you would need and old_password attribute here as its a ndoe attribute, not a crowbar one.

stefannica: @Itxaka I need old_password to be immediately initialize to the current admin password value during schema migration. This is based on the fact that there is no chef-run triggered during migration. If the user updateshe crowbar-openstack package and then immediately tries to change the keystone admin password, the old password value is lost.
However, I didn't realize that there is a schema validation made on attributes created during migration, so I thought I could 'trick' crowbar into initializing the old_password node attr for me here. I have to try something else.

stefannica: Curious though that this migration works fine for me on my current local setup.

cmurphy: It's valid to define it in the .schema as required => false, and then just not put it in the .json. Then it wouldn't appear to the user in the raw view.

stefannica: @cmurphy so the only reason it's complaining with a `key 'old_password:' is undefined` error is because it's missing from the .schema ?

cmurphy: Yes

Itxaka: In that case, you can iterate over the nodes and save it manually in the migrations, no need for an extra attribute in the schema :)

Something like https://github.com/crowbar/crowbar-core/blob/master/chef/data_bags/crowbar/migrate/dns/003_add_element_run_list_order.rb :)

stefannica:
So this is what I tried, but it's not setting `old_password` during migration on my (probably too dirty) local setup. I'll build a new one to test on:
```
  nodes = NodeObject.find("roles:keystone-server")
  nodes.each do |node|
    node[:keystone][:admin][:old_password] = node[:keystone][:admin][:password]
    node.save
  end
```

cmurphy:
> So this is what I tried, but it's not setting old_password during migration on my (probably too dirty) local setup.

Just a guess, maybe you didn't run the downgrade so it never re-ran the upgrade?

stefannica:
> > So this is what I tried, but it's not setting old_password during migration on my (probably too dirty) local setup.
> Just a guess, maybe you didn't run the downgrade so it never re-ran the upgrade?
> 
Well @cmurphy, according to @rsalevsky the downgrade scripts are currently completely disabled in crowbar.

stefannica:
@cmurphy anyway, I retried this on a new setup and it works as expected

cmurphy:
I went and deployed this and it did not work as expected. Relevant logs: http://paste.opensuse.org/95262553

I confirmed that the schema is at the right revision and that the migration correctly added the old_password field to the node.

stefannica:
@cmurphy Maybe there's something different in the way we deploy the barclamp. I just applied it again on a fresh setup and it works fine. See: http://pastebin.nue.suse.com/17855

The way that I tested it:
 - started a (non-HA) 2 node mkcloud setup on my local machine
 - build a new crowbar-openstack package with the changes in my local branch
 - install the updated crowbar-openstack package
 - change the keystone admin password using the crowbar UI and apply

I like to test it this way because it better reflects the way used by customers to apply PTFs, updates, etc.

cmurphy:
I tested this in a five-node HA environment. I don't build the package but I sync all the files, reinstall the barclamps, upload the cookbooks, and restart crowbar. Testing this in an HA environment is critical because we only uncovered this bug in the HA jobs.

stefannica:
@cmurphy I did the same thing on an HA setup, with slightly different results (see http://pastebin.nue.suse.com/17887/src), but ultimately didn't encountered any errors. I will re-test it on the HA gate setup, when the CI job is complete, but so far it looks like there is something different in the way we test crowbar updates.

stefannica:
@cmurphy wrt the 89307 build, the password update test worked, but the neutron-l3-ha-service CRM resource had a minor hiccup, most likely during the password change keystone outage window, from which the cluster then later on recovered. But because of this single CRM failure, the build failed.

stefannica:
I addressed the CRM failure in #1389. This PR can be merged.

PR #1366, opened 2017-10-12
mariadb: Add expire_logs_days config option

rsalevsky:
This commit makes the expire_logs_days option configurable via crowbar.
More information about the option can be found at
https://mariadb.com/kb/en/library/replication-and-binary-log-server-system-variables/#expire_logs_days.

### DIFF
--- a/chef/cookbooks/mysql/templates/default/logging.cnf.erb
+++ b/chef/cookbooks/mysql/templates/default/logging.cnf.erb
@@ -2,6 +2,6 @@
 log_error=/var/log/mysql/mysql_error.log
 
 <% if @slow_query_logging_enabled -%>
-slow_query_log=1
-slow_query_log_file=/var/log/mysql/mysql_slow.log
+slow_query_log = 1
+slow_query_log_file = /var/log/mysql/mysql_slow.log
 <% end -%>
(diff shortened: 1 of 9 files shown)


chef/cookbooks/mysql/templates/default/logging.cnf.erb
```diff
@@ -2,6 +2,8 @@
 log_error=/var/log/mysql/mysql_error.log
 
 <% if @slow_query_logging_enabled -%>
-slow_query_log=1
-slow_query_log_file=/var/log/mysql/mysql_slow.log
+slow_query_log = 1
+slow_query_log_file = /var/log/mysql/mysql_slow.log
 <% end -%>
+
+expire_logs_days = <%= @expire_logs_days %>
```
rhafer: Note ```expire_logs_days``` is not really a setting for logging in the sense of debug-logging or syslog. It is about the binary log that tracks all the changes made to the database. (They are used for replication and can be useful for backup/recovery).

So I don't believe this setting belongs to the logging.cnf file. I'd put it into galera.cnf (as that's where we set the binlog format as well)

crowbar_framework/config/locales/database/en.yml
```diff
@@ -26,6 +26,7 @@ en:
         mysql:
           datadir: 'Datadir'
           max_connections: 'Maximum Number of simultaneous Connections'
+          expire_logs_days: 'Number of days after which the log can be automatically removed'
```
rhafer: s/log/binary logs/ just to make clear that this is not about "usual" log files

PR #1367, opened 2017-10-12
keystone: Simplify password update code

nicolasbock:
The `ruby_block` to update the node attributes after a password update
is not necessary. It is run at compile phase which can be achieved
without the `ruby_block` also.


nicolasbock:
Thanks @Itxaka for the comments. I am also always a little suspicious with Chef. I never quite know what it will end up doing :smile:.

PR #1372, opened 2017-10-16
memcached: increase max connections limit

Itxaka:
In a big deployment, the default connections limit for memcached is too
low (1024) so its easy to reach it.
Increase the max connections limit for memcached to 4096 which should be
enough to deal with big deployments.

### DIFF
--- a/chef/cookbooks/memcached/attributes/default.rb
+++ b/chef/cookbooks/memcached/attributes/default.rb
@@ -2,6 +2,7 @@ default[:memcached][:memory] = 64
 default[:memcached][:port] = 11211
 default[:memcached][:listen] = "0.0.0.0"
 default[:memcached][:daemonize] = true
+default[:memcached][:max_connections] = 4096
 
 case node[:platform_family]
 when "suse"
--- a/chef/cookbooks/memcached/recipes/default.rb
+++ b/chef/cookbooks/memcached/recipes/default.rb
@@ -17,6 +17,8 @@
 # limitations under the License.
 #
 
+service_name = "memcached"
+
 case node[:platform_family]
 when "suse", "rhel"
   package "memcached" do
@@ -30,11 +32,11 @@ when "debian"
   end
 end
 
-service "memcached" do
+service service_name do
   action :nothing
   supports status: true, start: true, stop: true, restart: true
 end
-utils_systemd_service_restart "memcached"
+utils_systemd_service_restart service_name
 
 template "/etc/memcached.conf" do
   case node[:platform_family]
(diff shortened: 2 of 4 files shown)


Itxaka:
[Increase Max open files for memcached](https://trello.com/c/qCfMzD5n/170-increase-max-open-files-for-memcached)

Itxaka:
[Increase Max open files for memcached](https://trello.com/c/qCfMzD5n/170-increase-max-open-files-for-memcached)

chef/cookbooks/memcached/attributes/default.rb
```diff
@@ -2,6 +2,7 @@
 default[:memcached][:port] = 11211
 default[:memcached][:listen] = "0.0.0.0"
 default[:memcached][:daemonize] = true
+default[:memcached][:resource_limits] = 4096
```
nicolasbock: I would pick a more descriptive name for this option. Why not call it simply `connections`?

Itxaka: Mainly to have it in sync with the other `resource_limits` being "used" like in https://github.com/crowbar/crowbar-openstack/pull/974

Itxaka: althougth as this is only for connections, it may need a different name indeed

Itxaka:
I have not tested this, we do not need the sysctl changes,  not sure what I was thinking.

Itxaka:
@nicolasbock thanks for the review as this was totally wrong from the start. I jumped between PR and mixed them. Now the attribute name is changed and there is no sysctl limits anywhere as they are not needed.

Itxaka:
Im not really sure if this would work as expected without increasing the system limit for open files as it would still get the default of 1024 🤔

tpatzig:
@Itxaka it works by just increasing the connection limit. memcached is using this value to increase 'Max open files' for the memcached process as well:

```
root@d9c-dc-71-4a-27-40:~ # ulimit -n
1024
root@d9c-dc-71-4a-27-40:~ # ps axu | grep memcached
memcach+ 25193  0.0  0.0 325128  2156 ?        Ssl  08:17   0:00 /usr/sbin/memcached -m 64 -l 10.244.85.18 -p 11211 -c 4096
root     25213  0.0  0.0   9284  1616 pts/0    S+   08:17   0:00 grep --color=auto memcached
root@d9c-dc-71-4a-27-40:~ # cat /proc/25193/limits
Limit                     Soft Limit           Hard Limit           Units
Max cpu time              unlimited            unlimited            seconds
Max file size             unlimited            unlimited            bytes
Max data size             unlimited            unlimited            bytes
Max stack size            8388608              unlimited            bytes
Max core file size        0                    unlimited            bytes
Max resident set          unlimited            unlimited            bytes
Max processes             2063417              2063417              processes
Max open files            4096                 4096                 files
Max locked memory         65536                65536                bytes
Max address space         unlimited            unlimited            bytes
Max file locks            unlimited            unlimited            locks
Max pending signals       2063417              2063417              signals
Max msgqueue size         819200               819200               bytes
Max nice priority         0                    0
Max realtime priority     0                    0
Max realtime timeout      unlimited            unlimited            us
```

But nevertheless I think having global `nofile` limits increased to 65536 on the control and network nodes is a good idea anyhow.

PR #1375, opened 2017-10-17
mariadb: Drop unneeded root users (bsc#106062)

rhafer:
We only use root access via the unix domain socket so we can remove any
other root user.
Additional fix for https://bugzilla.suse.com/show_bug.cgi?id=1060628

### DIFF
--- a/chef/cookbooks/mysql/recipes/server.rb
+++ b/chef/cookbooks/mysql/recipes/server.rb
@@ -228,6 +228,18 @@ unless node[:database][:database_bootstrapped]
       only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
     end
   end
+
+  # Drop unneeded root users, we only use root access via unix domain socket
+  ["127.0.0.1", "::1", node[:hostname]].each do |hostname|
+    database_user "drop unneeded root database user at #{hostname}" do
+      connection db_connection
+      username "root"
+      host hostname
+      provider db_settings[:user_provider]
+      action :drop
+      only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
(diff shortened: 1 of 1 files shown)


chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -228,6 +228,18 @@
       only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
     end
   end
+
+  # Drop unneeded root users, we only use root access via unix domain socket
+  ["127.0.0.1", "::1", node[:hostname]].each do |hostname|
```
s-t-e-v-e-n-k: I think we should loop through all cluster members in fact if we're only running this on the cluster founder.

rhafer: Hm yes, that might be a good idea. OTOH I think we're always using the founder node for bootstrap so `root@founder` should be the only user with a real hostname in there if I am not mistaken.

Need to take a closer look.

rhafer: @s-t-e-v-e-n-k I just verified once more. We always bootstrap on the founder node. Which means that we will only have those `root@founder`, `root@::1` and `root@127.0.0.1` which we want to delete. Are you ok with that or would you prefer to have the other cluster members added to the list.

There is also `root@localhost` but we want to keep that, as it is required for local access via unix domain socket.

s-t-e-v-e-n-k: I think I've been conflating multiple runs I've been doing, with root@<founder> changing, depending on which node was pushed forward as the cluster founder.

andreasstieger:
This references bsc#106062, should be bsc#1060628

PR #1377, opened 2017-10-18
barbican: reorder config creation and initial db sync 

rhafer:
By creating the config file before running the initial database sync be
can remove the addtional db connection parameter from the "barbican db
upgrade" call. Which brings the recipe more in sync with how we do
things in the other barclamps.

It also avoid a shell escaping issues with db_connection uri that
contain special character like '&' that would cause issues on the
shell. (e.g. for ssl enabled MariaDB we add some ssl parameters to it
via "&ssl...=")

This should fix the current failures in the ssl CI jobs e.g.: https://ci.suse.de/job/openstack-mkcloud/88868/parsed_console/

### DIFF
--- a/chef/cookbooks/barbican/recipes/common.rb
+++ b/chef/cookbooks/barbican/recipes/common.rb
@@ -32,6 +32,33 @@ public_host = CrowbarHelper.get_host_for_public_url(node,
 
 database_connection = fetch_database_connection_string(node[:barbican][:db])
 
+memcached_servers = MemcachedHelper.get_memcached_servers(
+  if node[:barbican][:ha][:enabled]
+    CrowbarPacemakerHelper.cluster_nodes(node, "barbican-controller")
+  else
+    [node]
+  end
+)
+
+memcached_instance("barbican") if node["roles"].include?("barbican-controller")
+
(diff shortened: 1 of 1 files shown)


rhafer:
> Just a suggestion, if you have the time: if you generate the barbican configuration before the "db upgrade" command, this would enable you to remove the -d #{database_connection} argument completely.

Yes, that was also something I thought about. Then (shame on me ;) ) I went for the quick solution as I don't have much insight into barbican and not much time to spend on it.

@sayalilunkad, you did the initial implementation it seems. Is there any special reason why the config file is only written after the db sync happened?

sayalilunkad:
@rhafer As far as I can remember there was no specific reason for that and I think it should be ok to write it before the db_sync.

stefannica:
@rhafer I too think this was just overlooked and that you can simply correct it by moving some blocks of code around without worrying about implications :smile:

rhafer:
> I too think this was just overlooked and that you can simply correct it by moving some blocks of code around without worrying about implications 😄

@stefannica Ok, I did that now. It survived a quick test on my system. (even with HA)

PR #1378, opened 2017-10-18
Add HA rate limiting options to raw template to neutron and cinder

ilausuch:
Adds HA rate limiting options to the raw template for the different
services that depend from neutron and cinder. Defaults to 0 as the values, which means
no rate limiting


Itxaka:
[OpenStack Rate Limit Neutron, Cinder](https://trello.com/c/1rNJXMh4/66-openstack-rate-limit-neutron-cinder)

PR #1379, opened 2017-10-18
nova: get pub key from file instead of stdin

abel-navarro:
We found in some situations that when generating a public ssh key from a
private key using stdin, the ssh-keygen command would fail. In those
situations it helps to read the private from the file it was just
writen in the previous code block.


abel-navarro:
[Nova: generating public key fails with incorrect passphrase](https://trello.com/c/LWDyA5CH/176-nova-generating-public-key-fails-with-incorrect-passphrase)

PR #1385, opened 2017-10-19
nova: skip unchanged nodes chef-run

Itxaka:
By using skip_batch_for_node? we can avoid running chef on nodes that do
not change role and if there is no attribute change for the barclamp.

Does not skip remote HA nodes, nodes that have changed roles, all nodes if
attributes have changed and when deploying the barclamp for the first
time.

Stolen from https://github.com/vuntz/crowbar-openstack/commit/5cf4e69dadfd1ac804b75b4a8bd2a0f4b2248277
Requires https://github.com/crowbar/crowbar-core/pull/1374

### DIFF
--- a/crowbar_framework/app/models/nova_service.rb
+++ b/crowbar_framework/app/models/nova_service.rb
@@ -507,7 +507,82 @@ class NovaService < OpenstackServiceObject
     super
   end
 
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node, old_role, new_role)
+    # if old_role is nil, then we are applying the barclamp for the first time, so no skip
+    return false if old_role.nil?
+
+    # if the node changed roles, then we need to apply, so no skip
+    return false if node_changed_roles?(node, old_role, new_role)
+
+    # if attributes have changed, we need to apply, so no skip
+    return false if node_changed_attributes?(node, old_role, new_role)
+
+    # if we use remote HA, let's be safe, so no skip
+    return false if node_is_remote_ha?(node, new_role)
+
+    # if the node is a controller, then we only need to apply if we move from
+    # non-HA to HA (or vice-versa), since the config didn't change
+    if node_is_nova_controller?(node, new_role)
+      return false if node_changed_ha?(node, old_role, new_role)
+    end
+
+    # by this point its safe to assume that we can skip the node as nothing has changed on it
+    # same attributes, same roles so skip it
+    @logger.info("#{@bc_name} skip_batch_for_node? skipping: #{node_name}")
+    true
+  end
+
   private
+
+  def node_changed_attributes?(node, old_role, new_role)
+    old_role_attributes = old_role.default_attributes[@bc_name].deep_dup
+    # we need to remove the HA keys from the old_role for nova/ec2-api,
+    # as they get added afterwards to the new role during apply_role_pre_chef_call
+    # so if we dont remove them, the comparision is always gonna fail
+    old_role_attributes.delete("ha")
+    old_role_attributes["ec2-api"].delete("ha")
+    if old_role_attributes != new_role.default_attributes[@bc_name]
+      logger.debug("Nova skip_batch_for_node?: not skipping #{node} (attribute change)")
+      return true
+    end
+
+    false
+  end
+
+  def node_is_remote_ha?(node, new_role)
+    new_role.override_attributes[@bc_name]["elements"].each do |role_name, elements|
+      next unless role_name =~ /^nova-compute-/
+      elements.each do |element|
+        if is_remotes?(element)
+          @logger.debug("Nova skip_batch_for_node?: not skipping #{node} (compute HA)")
+          return true
+        end
+      end
+    end
+
+    false
+  end
+
+  def node_changed_ha?(node, old_role, new_role)
+    old_elements = old_role.override_attributes[@bc_name]["elements"]
+    new_elements = new_role.override_attributes[@bc_name]["elements"]
+
+    if old_elements["nova-controller"] == new_elements["nova-controller"]
+      @logger.debug("Nova skip_batch_for_node?: skipping #{node} (no controller change)")
+      return false
+    end
+
+    true
+  end
+
+  # find out if this node is a controller
+  def node_is_nova_controller?(node, role)
+    return false if role.nil?
+
+    _, controller_nodes, = role_expand_elements(role, "nova-controller")
+    controller_nodes.include?(node)
+  end
 
   def hyperv_available?
     return File.exist?("/opt/dell/chef/cookbooks/hyperv")


Itxaka:
[Skip batches when applying nova barclamp](https://trello.com/c/rYkjyvHO/161-skip-batches-when-applying-nova-barclamp)

crowbar_framework/app/models/nova_service.rb
```diff
@@ -507,8 +507,88 @@ def validate_proposal_after_save(proposal)
     super
   end
 
+  # try to know if we can skip a node from running chef-client
+  def skip_batch_for_node?(batch, node, old_role, new_role)
```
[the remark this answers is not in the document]
Itxaka: ¯\\\_(ツ)\_/¯

crowbar_framework/app/models/nova_service.rb
```diff
@@ -507,8 +507,88 @@ def validate_proposal_after_save(proposal)
     super
   end
 
+  # try to know if we can skip a node from running chef-client
+  def skip_unchanged_node?(node, old_role, new_role)
```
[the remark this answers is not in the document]
aspiers: Actually I agree with Hound here.  When introducing new code, we should not introduce new technical debt.  Breaking it into small methods is the single most effective technique for ensuring that it will be correct, easy to understand, test, and maintain in the long run.  Although not huge by `#apply_role` standards, this method is still way too big for my tastes.  For example I can't get any sense of how it works just by quickly glancing at it.  If it was just a few lines which called out to other methods (with meaningful names) then it would be easy to do that.

Yes, I know you didn't write it, but I already tried and failed to persuade @vuntz of the importance of small methods, so maybe I'll have better luck with you ;-)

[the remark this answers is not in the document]
Itxaka: @aspiers will you be ok with a follow up PR for this? it would be great if we could merge this so SAP can take advantage of it and they are starting deployment of the big cloud this week (today or tomorrow I think), so it would be much easier for us to merge and improve than to delay this further.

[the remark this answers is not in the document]
Itxaka: @aspiers updated with a method per thing to check to make it more clear as its easier to do it like this than to make it a patch follow up especially if we need to also do the backport+follow up patch for backport.

crowbar_framework/app/models/nova_service.rb
```diff
@@ -507,8 +507,112 @@ def validate_proposal_after_save(proposal)
...
+    # if we use remote HA, let's be safe, and don't skip anything
+    return false if node_is_remote_ha?(node, new_role)
+
+    # if the node changed roles, then we need to apply
+    return false if node_changed_roles?(node, old_role, new_role)
+
+    # if the node is only a compute node, or not relevant to nova at all
+    # (shouldn't happen), then we don't do anything since the nova proposal
+    # didn't change and we have the same role
+    unless node_controller?(node, new_role)
```
skazi0: I would rather call it e.g. `node_is_nova_controller?` or it's not clear if you're checking for "node's controller" and if you care about other controllers or just `nova-controller`.

crowbar_framework/app/models/nova_service.rb
```diff
@@ -507,8 +507,112 @@ def validate_proposal_after_save(proposal)
...
+    @logger.debug("Nova skip_batch_for_node? entering: #{node}")
+
+    # if old_role is nil, then we are applying the barclamp for the first time
+    if old_role.nil?
+      logger.debug("Nova skip_batch_for_node?: not skipping #{node} (new role)")
+      return false
+    end
+
+    # if attributes have changed, we need to run
+    return false if node_attributes_changed?(node, old_role, new_role)
```
skazi0: To match the other methods this should probably be `node_changed_attributes?`.

Itxaka: indeed!

crowbar_framework/app/models/nova_service.rb
Itxaka: +1

crowbar_framework/app/models/nova_service.rb
```diff
@@ -507,8 +507,112 @@ def validate_proposal_after_save(proposal)
...
+    # (shouldn't happen), then we don't do anything since the nova proposal
+    # didn't change and we have the same role
+    unless node_is_nova_controller?(node, new_role)
+      @logger.debug("Nova skip_batch_for_node?: skipping #{node} (compute only)")
+      return true
+    end
+
+    # if the node is a controller, then we only need to apply if we move from
+    # non-HA to HA (or vice-versa), since the config didn't change
+    return true unless node_changed_ha?(node, old_role, new_role)
```
bmwiedemann: Could also use `return false if` to be similar to the lines 515 to 527 (and update the default below to return true)

Itxaka: We default to false by default at the end of the method if nothing has been triggered(it should not happen but better safe than sorry) so there is only 2 cases in which if we reach them we can safely-ish skip the node. Changing the default return value to true could have some unwanted consequences, so I dont feel safe enough to do so.

I could probably rework the `node_changed_ha` function to return the flipped value but Im struggling to find a decent name for it `node_not_changed_ha` sounds bad and goes against the rest of the other method names :/

Thougths?

crowbar_framework/app/models/nova_service.rb
```diff
@@ -507,8 +507,112 @@ def validate_proposal_after_save(proposal)
...
+    end
+
+    nil
+  end
+
+  # find out if this node is a controller
+  def node_is_nova_controller?(node, role)
+    return false if role.nil?
+
+    _, controller_nodes, = role_expand_elements(role, "nova-controller")
```
aspiers: Is there supposed to be a trailing comma here? Although I guess it probably does no harm?

crowbar_framework/app/models/nova_service.rb
aspiers: Although I initially had the same thought as @bmwiedemann here, I eventually decided that I think it's good how it is. I agree that a default fallback returning `false` is a good idea in terms of safety.

crowbar_framework/app/models/nova_service.rb
Itxaka: yup! Seems that if you just want to discard the rest of the return values hound complains that you should use a trailing coma instead of a `_`

```
Style/TrailingUnderscoreVariable: Do not use trailing _s in parallel assignment. Prefer _, controller_nodes, = role_expand_elements(role, "nova-controller").
```

vuntz:
> Thanks! This is SOOO much easier to review than before

Nooooo, it's horrible!!!!! ;-) (therefore I will let someone else press the merge button)

Itxaka:
awaiting for the fixes tot he base function https://github.com/crowbar/crowbar-core/pull/1395 before merging

Itxaka:
Somehow its not working as expected as it always trigger the force-run due to an attribute change. Looking at the differences, there is always the following in the `new_role.default_attributes[@bc_name]` (array.to_a for easy comparison between 2 hashes)

```
[["ec2-api", {"db"=>{"password"=>"xxxxxxxx", "user"=>"ec2", "database"=>"ec2"}, "ssl"=>{"enabled"=>false, "certfile"=>"/etc/ec2api/ssl/certs/signing_cert.pem", "keyfile"=>"/etc/ec2api/ssl/certs/signing_key.pem", "generate_certs"=>false, "insecure"=>false, "cert_required"=>false, "ca_certs"=>"/etc/ec2api/ssl/certs/ca.pem"}, "ha"=>{"enabled"=>false}}], ["ha", {"enabled"=>false}]]
```

Looking at the ec2-api parameters now...

Itxaka:
umm, this may be becuase some of those attributes are set at `apply_role_pre_chef_call` which is called afterwards, but I would expect those to appear on the old role as well...

crowbar_framework/app/models/nova_service.rb
```diff
@@ -507,8 +507,112 @@ def validate_proposal_after_save(proposal)
...
+    return true unless node_changed_ha?(node, old_role, new_role)
+
+    @logger.debug("Nova skip_batch_for_node? exiting: #{node}")
+    false
+  end
+
   private
 
+  def node_changed_attributes?(node, old_role, new_role)
+    if old_role.override_attributes[@bc_name] != new_role.override_attributes[@bc_name]
```
vuntz: I guess I confused you when I explained things ;-)

You want `default_attributes` here. What we want to change is https://github.com/Itxaka/crowbar-openstack/blob/1d4a4e504d4f4541506c45e8abb21ad17b7a0fd6/crowbar_framework/app/models/nova_service.rb#L285-L287

Itxaka:
@vuntz I think this is ready as the fixes to make this work are already merge but we may want to wait for https://github.com/crowbar/crowbar-core/pull/1408/ to get in so we can clean parts of this? Or maybe @ilausuch can just cherry-pick this change and add it to the same PR so we merge all at the same time? Or that may cause less reviews as it would be a big PR.

Thougths?

vuntz:
> @vuntz I think this is ready as the fixes to make this work are already merge but we may want to wait for crowbar/crowbar-core#1408 to get in so we can clean parts of this?

If you review that PR, then we can just merge it ;-)

crowbar_framework/app/models/nova_service.rb
```diff
@@ -507,8 +507,117 @@ def validate_proposal_after_save(proposal)
...
+  end
+
   private
 
+  def node_changed_attributes?(node, old_role, new_role)
+    old_role_attributes = old_role.default_attributes[@bc_name]
+    # we need to remove the HA keys from the old_role for nova/ec2-api, as they get added afterwards to the
+    # new role during apply_role_pre_chef_call so if we dont remove them the comparision is always gonna fail
+    old_role_attributes.delete("ha")
+    old_role_attributes["ec2-api"].delete("ha")
```
vuntz: Eek, please copy the attributes and don't modify the original ones.

crowbar_framework/app/models/nova_service.rb
```diff
@@ -507,8 +507,73 @@ def validate_proposal_after_save(proposal)
...
+
+    # if attributes have changed, we need to apply, so no skip
+    return false if node_changed_attributes?(node, old_role, new_role)
+
+    # if we use remote HA, let's be safe, so no skip
+    return false if node_is_remote_ha?(node, new_role)
+
+    # if the node is a controller, then we only need to apply if we move from
+    # non-HA to HA (or vice-versa), since the config didn't change
+    return false if node_changed_ha?(node, old_role, new_role)
```
vuntz: How do we know that the node is a controller here?

Itxaka: woops, I went overboard with the cleaning

PR #1387, opened 2017-10-20
pacemaker: created openstack_pacemaker_primitive

abel-navarro:
Depends on: https://github.com/crowbar/crowbar-ha/pull/262

This change is meant to provide a default value for pacemaker ops. In
particular we are interested in providing a default for 'op on-fail'
actions.

With Pacemaker we are not able to configure such behavior. The approach
here is to provide a wrapper definition around pacemaker_primitive that
can read that default value from the Pacemaker barclamp. For that reason
this change depends on a pull request in the crowbar-ha repository
addressing this.

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/central_ha.rb
+++ b/chef/cookbooks/ceilometer/recipes/central_ha.rb
@@ -25,7 +25,7 @@ rabbit_settings = fetch_rabbitmq_settings
 transaction_objects = []
 
 service_name = "ceilometer-central"
-pacemaker_primitive service_name do
+openstack_pacemaker_primitive service_name do
   agent node[:ceilometer][:ha][:central][:agent]
   op node[:ceilometer][:ha][:central][:op]
   action :update
--- a/chef/cookbooks/crowbar-openstack/definitions/openstack_pacemaker_primitive.rb
+++ b/chef/cookbooks/crowbar-openstack/definitions/openstack_pacemaker_primitive.rb
@@ -1,0 +1,41 @@
+#
+# Copyright 2017, SUSE
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+define :openstack_pacemaker_primitive,
+    agent: nil,
+    params: {},
+    op: {},
+    action: [] do
+  primitive_name = params[:name]
+  agent = params[:agent]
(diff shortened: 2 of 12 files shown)


abel-navarro:
Dependency: https://github.com/crowbar/crowbar-ha/pull/262

abel-navarro:
Dependency: https://github.com/crowbar/crowbar-ha/pull/262

abel-navarro:
[Rework on-fail for stop actions setting in pacemaker](https://trello.com/c/Jkr9NDFX/112-rework-on-fail-for-stop-actions-setting-in-pacemaker)

chef/cookbooks/ceilometer/recipes/central_ha.rb
```diff
@@ -25,13 +25,13 @@
...
 service_name = "ceilometer-central"
-pacemaker_primitive service_name do
+openstack_pacemaker_primitive service_name do
   agent node[:ceilometer][:ha][:central][:agent]
   op node[:ceilometer][:ha][:central][:op]
   action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
-transaction_objects << "pacemaker_primitive[#{service_name}]"
+transaction_objects << "openstack_pacemaker_primitive[#{service_name}]"
```
vuntz: I'm not sure that `pacemaker_transaction` can deal with `openstack_pacemaker_primitive`. Are you sure it can?

abel-navarro: it indeed can't use that resource. I'm trying to fix that right now.

aspiers:
@abel-navarro In [the absence of GitHub supporting PR dependencies](https://github.com/isaacs/github/issues/95%39), I recommend mentioning dependencies in the PR description at the top - if it's only mentioned in a comment then it easily gets buried and missed :-)

chef/cookbooks/crowbar-openstack/definitions/openstack_pacemaker_primitive.rb
```diff
@@ -0,0 +1,41 @@
...
+    agent: nil,
+    params: {},
+    op: {},
+    action: [] do
+  primitive_name = params[:name]
+  agent = params[:agent]
+  op = params[:op]
+  action = params[:action]
+
+  fake_params = {}
```
aspiers: Why fake instead of `params[:params]`?  Surely it's not right to just ditch the parameters passed in?

abel-navarro: I'm still having problems passing params to this `openstacl_pacemaker_primitive` definition. Chef doesn't like that and need help solving it. That's why this PR is marked as WIP.

abel-navarro:
About the dependency among PRs, I will update the description to add it in the beginning.

abel-navarro:
When diving deep into this implementation proposal we found that using a definition to wrap `pacemaker_primitive` to provide a default value was causing problems with conflicting parameters. The workaround to make this work (changing parameter names) wasn't solving the problem. Any other approach would similarly fail.

This change is being abandoned in favor of a solution extending the original `pacemaker_primitive` resource instead of wrapping it with a definition.

PR #1388, opened 2017-10-20
[4.0] keystone: store old admin password in node attr for update

stefannica:
(back-port of #1365 )

Use a node attribute to store a backup copy of the admin password, to be used when the admin password is changed.
This replaces the previous method of managing the admin password change through a dedicated UI updated_password attribute and allows the user to change the password by simply modifying the regular password UI attribute.

The advantage of having a backup copy of the admin password in the node attributes is that the new password value is immediately available after a password change, even during the chef compilation phase. The implications of this:
 - the new password value can immediately be used by all recipes during the chef compilation phase, which removes the need of having to deal with a password change scenario in every openstack cookbook using keystone for authentication
 - reconfiguring keystone with the new password doesn't need to happen in the compilation phase anymore. Doing it in the convergence phase, at the same time as everything else, removes this unnecessary exception from the keystone cookbook and aligns everything together better


stefannica:
Fixed a small typo during merge

stefannica:
(rebased)

PR #1389, opened 2017-10-20
neutron: use service account for neutron-l3-ha service

stefannica:
Update the neutron barclamp to use the neutron service user credentials for the neutron-l3-ha/neutron-ha-tool service.

The neutron-l3-ha service was previously configured to use the admin user, which:
 1. is not necessary - the neutron service user is more  appropriate
 2. is problematic during a keystone admin password update, because it takes some time for the new admin password to reach the neutron barclamps in the chef-run, which means the neutron-l3-ha service may be triggered and run with the old admin password value and ultimately fail.

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -35,7 +35,7 @@ if use_l3_agent
     owner "root"
     group "root"
     mode "0600"
-    content keystone_settings["admin_password"]
+    content keystone_settings["service_password"]
     # Our Chef is apparently too old for this :-/
     #sensitive true
     action :create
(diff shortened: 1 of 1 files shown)


chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -70,8 +70,8 @@
                                                                    insecure_flag do |env|
       env["OS_AUTH_URL"] = os_auth_url_v2
       env["OS_REGION_NAME"] = keystone_settings["endpoint_region"]
-      env["OS_TENANT_NAME"] = keystone_settings["admin_tenant"]
-      env["OS_USERNAME"] = keystone_settings["admin_user"]
+      env["OS_TENANT_NAME"] = keystone_settings["service_tenant"]
```
cmurphy: Since this line is already being changed anyway, could you make this "service_project" instead of "service_tenant"? They are actually aliases for the same thing and "project" is the newer name for the thing

https://github.com/crowbar/crowbar-openstack/blob/25906a22980d0da656569303c159ac8e35e80796/chef/cookbooks/keystone/libraries/helpers.rb#L96-L97

stefannica:
@cmurphy done

PR #1390, opened 2017-10-20
horizon: use Grafana 4

jgrassler:
This commit adds support for Grafana 4 to the Horizon barclamp.
This allows us to switch from our local Grafana 1.5 fork to
upstream Grafana 4 for the Monasca metrics dashboards.

It makes the following changes:

* Configure and deploy standalone grafana service
* Create a ProxyPass path on the Apache server to transparently
  proxy requests to standalone Grafana
* Add a new dashboard_fqdn setting for the Horizon barclamp that
  allows users to communicate the public Horizon endpoint's FQDN to
  Grafana (needed for proxying to work). If this is not set,
  Crowbar will fall back to the public cluster IP address in an HA
  cloud and to the dashboard node's public IP address.

Note: while basic functionality is there already, this still needs some more work, notably a mechanism for configuring the `monasca-grafana-datasource` plugin and updated Monasca and OpenStack dashboards. Hence I'm setting the wip flag for now.

### DIFF
--- a/chef/cookbooks/horizon/libraries/helper.rb
+++ b/chef/cookbooks/horizon/libraries/helper.rb
@@ -19,6 +19,10 @@ module MonascaUiHelper
     CrowbarHelper.get_host_for_public_url(node, ssl_enabled, ha_enabled)
   end
 
+  def self.monasca_admin_host(node)
+    CrowbarHelper.get_host_for_admin_url(node, node[:monasca][:ha][:enabled])
+  end
+
   def self.api_public_url(node)
     host = monasca_public_host(node)
     # SSL is not supported at this moment
@@ -26,5 +30,45 @@ module MonascaUiHelper
     protocol = "http"
     port = node[:monasca][:api][:bind_port]
     "#{protocol}://#{host}:#{port}/v2.0"
+  end
+
+  def self.dashboard_ip(node)
+    ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "public").address
+
+    if node[:horizon][:ha][:enabled] && !@cluster_admin_ip
+      ip = CrowbarPacemakerHelper.cluster_vip(node, "public")
+    end
+
+    ip
+  end
+
+  def self.dashboard_local_url(node)
+    public_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "public").address
+    admin_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+
+    ha_enabled = node[:horizon][:ha][:enabled]
+    ssl_enabled = node[:horizon][:apache][:ssl]
+
+    protocol = "http"
+    protocol = "https" if ssl_enabled
+
+    if ha_enabled
+      port = node[:horizon][:ha][:ports][:plain]
+      port = node[:horizon][:ha][:ports][:ssl] if ssl_enabled
+      return "#{protocol}://#{admin_ip}:#{port}"
+    end
+
+    "#{protocol}://#{public_ip}"
+  end
+
+  def self.dashboard_public_url(node)
+    protocol = "http"
+    protocol = "https" if node[:horizon][:apache][:ssl]
+
+    "#{protocol}://#{dashboard_ip(node)}"
+  end
+
+  def self.grafana_service_url(node)
+    "http://#{monasca_public_host(node)}:3000"
   end
 end
--- a/chef/cookbooks/horizon/providers/grafana_datasource.rb
+++ b/chef/cookbooks/horizon/providers/grafana_datasource.rb
@@ -1,0 +1,142 @@
+# Copyright 2017 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "json"
+require "net/http"
+require "openssl"
+require "uri"
+
+action :create do
+  connection = _build_connection(new_resource)
+  datasources = list_data_sources(connection)
+
+  datasource = find_data_source_by_name(new_resource.name, datasources)
+
+  payload = _datasource_payload(new_resource)
+
+  send_datasource(connection, payload) if datasource.nil?
+
+  if !datasource.nil? && datasource_diff?(datasource, payload)
+    payload["id"] = datasource["id"]
+    send_datasource(connection, payload)
+  end
+end
+
+def list_data_sources(conn)
+  path = ::File.join(conn["base_path"], "/api/datasources")
+  resp = conn["http"].request_get(path, conn["headers"])
+
+  return JSON.parse(resp.read_body) if resp.is_a?(Net::HTTPOK)
+
+  log_message = "Could not retrieve list of Grafana data sources"
+  _raise_error(resp, log_message, "list_data_sources()")
+end
+
+def find_data_source_by_name(name, data_sources)
+  data_sources.each do |data_source|
+    return data_source if data_source["name"] == name
+  end
+
+  nil
+end
+
+def send_datasource(conn, payload)
+  method = "POST"
+  path = ::File.join(conn["base_path"], "/api/datasources")
+
+  if payload["id"]
+    path = ::File.join(path, payload["id"].to_s)
+    method = "PUT"
+  end
+
+  headers = { "Content-Type" => "application/json" }.merge(conn["headers"])
+
+  resp = conn["http"].send_request(method, path, JSON.generate(payload), headers)
+
+  return if resp.is_a?(Net::HTTPOK)
+
+  log_message = "Could not update data source #{payload["name"]}"
+  _raise_error(resp, log_message, "send_datasource()")
+end
+
+def datasource_diff?(api, resource)
+  res = false
+  resource.each_key do |k|
+    if resource[k].is_a?(Hash)
+      res = datasource_diff?(api[k], resource[k])
+      break if res == true
+    end
+
+    unless api.key?(k) && (api[k] == resource[k])
+      res = true
+      break
+    end
+  end
+
+  res
+end
+
+def _build_connection(new_resource)
+  uri = URI(new_resource.grafana_url)
+
+  # Need to require net/https so that Net::HTTP gets monkey-patched
+  # to actually support SSL:
+  require "net/https" if uri.scheme.start_with?("https")
+
+  # Construct authentication headers
+  auth_headers = Net::HTTP::Get.new(new_resource.grafana_url)
+  auth_headers.basic_auth(new_resource.user_name, new_resource.password)
+  headers = { "Authorization" => auth_headers["authorization"] }
+
+  # Construct the http object
+  http = Net::HTTP.new(uri.host, uri.port)
+
+  http.use_ssl = true if uri.scheme.start_with?("https")
+  http.verify_mode = OpenSSL::SSL::VERIFY_NONE if new_resource.insecure
+
+  {
+    "base_path" => new_resource.grafana_url,
+    "http" => http,
+    "headers" => headers
+  }
+end
+
+def _datasource_payload(new_resource)
+  {
+    "url" => new_resource.proxy_url,
+    "access" => "direct",
+    "isDefault" => new_resource.is_default,
+    "withCredentials" => false,
+    "jsonData" => {
+      "useHorizonProxy" => true,
+      "token" => "",
+      "keystoneAuth" => false,
+      "authMode" => "Horizon"
+    },
+    "name" => new_resource.name,
+    "type" => "monasca-datasource"
+  }
+end
+
+def _log_error(resp, msg)
+  Chef::Log.error(msg)
+  Chef::Log.error("Response Code: #{resp.code}") if resp
+  Chef::Log.error("Response Message: #{resp.message}") if resp
+end
+
+def _raise_error(resp, msg, calling_function)
+  _log_error(resp, msg)
+  new_resource.updated_by_last_action(false)
+  raise "#{msg} in provider function #{calling_function}"
+end
--- a/chef/cookbooks/horizon/recipes/monasca_ui.rb
+++ b/chef/cookbooks/horizon/recipes/monasca_ui.rb
@@ -12,9 +12,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 monasca_server = node_search_with_cache("roles:monasca-server").first
+monasca_master = node_search_with_cache("roles:monasca-master").first
+monasca_host = MonascaUiHelper.monasca_admin_host(monasca_server)
+grafana_password = monasca_master[:monasca][:master][:database_grafana_password]
+
+# Used for creating data source
+grafana_base_url = ::File.join(MonascaUiHelper.dashboard_local_url(node), "/grafana")
+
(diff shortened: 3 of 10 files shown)


jgrassler:
Ok, this can create the Grafana data source now (and update it)

chef/cookbooks/horizon/providers/grafana_datasource.rb
```diff
@@ -0,0 +1,128 @@
...
+  end
+
+  headers = { "Content-Type" => "application/json" }.merge(conn["headers"])
+
+  resp = conn["http"].send_request(method, path, JSON.generate(payload), headers)
+
+  return if resp.is_a?(Net::HTTPOK)
+
+  log_message = "Could not update data source #{payload["name"]}"
+  _raise_error(resp, log_message, "post_datasource()")
```
matrixik: replace `post_datasource` with `send_datasource`

chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -13,52 +13,71 @@
 # limitations under the License.
 
 monasca_server = node_search_with_cache("roles:monasca-server").first
+monasca_master = node_search_with_cache("roles:monasca-master").first
+monasca_host = MonascaUiHelper.monasca_admin_host(monasca_server)
+grafana_password = monasca_master[:monasca][:master][:database_grafana_password]
+grafana_base_url = "#{MonascaUiHelper.dashboard_url(node)}/grafana"
```
matrixik: `::File.join(` also in this line

chef/cookbooks/horizon/templates/default/grafana.ini.erb
```diff
@@ -0,0 +1,402 @@
...
+# as seperate properties or as on string using the url propertie.
+
+# Either "mysql", "postgres" or "sqlite3", it's your choice
+type = mysql
+host = <%= @database_host %>:3306
+name = grafana
+user = grafana
+password = <%= @grafana_password %>
+# If the password contains # or ; you have to wrap it with trippel quotes. Ex """#password;"""
+;password =
```
matrixik: remove line, it's used in 72

chef/cookbooks/horizon/providers/grafana_datasource.rb
```diff
@@ -0,0 +1,128 @@
...
+end
+
+def datasource_diff?(api, resource)
```
matrixik: Sorry, but I don't fully comprehend this function. Do you check if `api` and `resource` is different here and then return `true` or you return `true` when they are the same?

jgrassler: It returns `true` if they differ. Trust me, it works...took some painful testing/bugfixing to make sure of that :-)

chef/cookbooks/horizon/templates/default/grafana.ini.erb
jgrassler: Done.

chef/cookbooks/horizon/recipes/monasca_ui.rb
jgrassler: Done.

chef/cookbooks/horizon/providers/grafana_datasource.rb
jgrassler: Done.

chef/cookbooks/horizon/providers/grafana_datasource.rb
jgrassler: Done.

chef/cookbooks/horizon/providers/grafana_datasource.rb
```diff
@@ -0,0 +1,128 @@
...
+  connection = _build_connection(new_resource)
+  datasources = list_data_sources(connection)
+
+  datasource = find_data_source_by_name(new_resource.name, datasources)
+
+  payload = _datasource_payload(new_resource)
+
+  send_datasource(connection, payload) if datasource.nil?
+
+  if !datasource.nil? && datasource_diff?(datasource, _datasource_payload(new_resource))
```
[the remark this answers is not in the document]
jgrassler: Done.

chef/cookbooks/horizon/providers/grafana_datasource.rb
```diff
@@ -0,0 +1,128 @@
+require "json"
```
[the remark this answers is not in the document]
jgrassler: Done.

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -527,13 +517,34 @@
...
+
+monasca_ui_pkgname =
+  case node[:platform_family]
+  when "suse"
+    "openstack-horizon-plugin-monasca-ui"
+  when "rhel"
+    "openstack-monasca-ui"
+  end
+
+grafana_url = ""
```
dirkmueller: whats the point of this code?

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -17,6 +17,15 @@
 include_recipe "apache2::mod_wsgi"
 include_recipe "apache2::mod_rewrite"
 
+monasca_server = node_search_with_cache("roles:monasca-server").first
+unless monasca_server.nil?
+  include_recipe "apache2::mod_proxy"
+  include_recipe "apache2::mod_proxy_http"
+
+  grafana_available = true
```
dirkmueller: whats the point of the extra ```grafana_available```  ? wouldn't it be enough to use grafana_url being non_empty for that?

chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -13,52 +13,82 @@
...
+  notifies :run, resources(execute: "grafana listening?"), :immediately
 end
 
-cookbook_file "/srv/www/grafana/app/dashboards/openstack.json" do
-  source "grafana-openstack.json"
+horizon_grafana_datasource "Monasca (Crowbar)" do
+  is_default true
+  user_name "admin"
+  password grafana_password
+  grafana_url grafana_base_url
```
dirkmueller: so in the ha case this doesn't seem to work, because it hits the vip and gets forwarded occassionally to an apache that isn't the local one, and that one doesn't have this code executed yet. 

also, if the databases are connected, we really want it to be run only on the founder node against the ip of the founder, right?

chef/cookbooks/horizon/templates/suse/openstack-dashboard.conf.erb
```diff
@@ -47,7 +47,8 @@
     Alias /media <%= @horizon_dir %>/media
     Alias /static <%= @horizon_dir %>/static
     <% if @grafana_available %>
-    Alias /grafana /srv/www/grafana
+    ProxyPass "/grafana" "<%= @grafana_url %>"
+    ProxyPassReverse "/grafana" "<%= @grafana_url %>"
```
dirkmueller: for me this doesn't work, all accesses are 403. something like 

```
    <Location /grafana>
        Require all granted
    </Location>
```

is missing. we might want to do a more correct Require statement though.

chef/cookbooks/horizon/libraries/helper.rb
```diff
@@ -27,4 +31,31 @@ def self.api_public_url(node)
...
   end
+
+  def self.dashboard_fqdn_ip(node)
+    # Use explicit dashboard FQDN if configured
+    if node[:horizon].key?("dashboard_fqdn") && !node[:horizon][:dashboard_fqdn].empty?
+      return node[:horizon][:dashboard_fqdn]
+    end
+
+    # Fall back to node's IP address without explicit dashboard FQDN
+    ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "public").address
```
dirkmueller: why is this one in non-ha case the public IP but in the HA case the admin one? inconsistent?

chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -13,52 +13,82 @@
...
+  proxy_url monasca_proxy_url
+end
+
+cookbook_file "/var/lib/grafana/dashboards/grafana-monasca.json" do
```
dirkmueller: any reason why you added the grafana- prefix into the name ? the _80_monasca_ui_settings.py still expects the old name. maybe change it back for improved compatibility?

chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -13,52 +13,82 @@
...
   owner "root"
   group "root"
   mode "0644"
-  notifies :reload, resources(service: "apache2")
+  notifies :restart, resources(service: "grafana-server")
 end
 
-cookbook_file "/srv/www/grafana/app/dashboards/monasca.json" do
-  source "grafana-monasca.json"
+cookbook_file "/var/lib/grafana/dashboards/grafana-openstack.json" do
```
dirkmueller: any reason why you added the grafana- prefix into the name ? the _80_monasca_ui_settings.py still expects the old name. maybe change it back for improved compatibility?

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -17,6 +17,15 @@
...
+  grafana_url = "#{MonascaUiHelper.dashboard_url(node)}:3000"
```
dirkmueller: this looks incorrect, ProxyPass expects a http(s):// url protocol prefix, see e.g. https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypass

the easiest solution is to use ```MonascaUiHelper.grafana_service_url(node)``` here, which avoids hardcoding
the port 3000 yet another time.

chef/cookbooks/horizon/recipes/server.rb
jgrassler: Holdover from before I used `grafana_enabled`. I'll ditch that one, move this to the proper location and used it to determine whether to configure the reverse proxy.

chef/cookbooks/horizon/recipes/server.rb
jgrassler: Superfluous, removing.

chef/cookbooks/horizon/recipes/monasca_ui.rb
jgrassler: Correct. I haven't tested this in an HA scenario, yet (the `mckloud` host I was using was slow enough without HA...). Adjusted that now.

chef/cookbooks/horizon/templates/suse/openstack-dashboard.conf.erb
jgrassler: Interesting. I'll try to reproduce this...

chef/cookbooks/horizon/libraries/helper.rb
jgrassler: Fixed.

chef/cookbooks/horizon/recipes/monasca_ui.rb
jgrassler: No reason. I can use the old name for all that will help (Horizon appends parameters to the URL, not sure how much they'll break...). I'll mess with this a bit once I've addressed all your comments and have verified the changes do not break anything.

chef/cookbooks/horizon/recipes/monasca_ui.rb
jgrassler: See above.

chef/cookbooks/horizon/templates/suse/openstack-dashboard.conf.erb
jgrassler: Uhm. That one _is_ already in there (I accidently removed it when I got rid of `grafana_enabled` which is when I started to get that same 401, but that's fixed now).

chef/cookbooks/horizon/recipes/server.rb
jgrassler: Done.

jgrassler:
Jolly good news: we no longer need this whole `dashboard_fqdn` nonsense. Confirmed to work without it now.

jgrassler:
Ok, tested and works on an HA cloud now.

chef/cookbooks/horizon/files/default/_80_monasca_ui_settings.py
```diff
@@ -59,9 +61,9 @@
 #        {'title': 'Add New Dashboard', 'fileName': 'empty.json'}]}
 #]
 
-ENABLE_KIBANA_BUTTON = getattr(settings, 'ENABLE_KIBANA_BUTTON', <%= @kibana_enabled ? "True" : "False" %>)
+ENABLE_KIBANA_BUTTON = getattr(settings, 'ENABLE_KIBANA_BUTTON', True)
 # TODO: this needs to be fixed for clustered deployment
-KIBANA_HOST = getattr(settings, 'KIBANA_HOST', 'http://<%= @kibana_host %>:5601/')
+KIBANA_HOST = getattr(settings, 'KIBANA_HOST', 'http://192.168.232.3:5601/')
```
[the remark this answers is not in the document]
jgrassler: I most assuredly do not want this to be static, thanks for spotting this. There was a from the looks of it - orphaned - template in the barclamp (not being used by anything). Guess it wasn't orphaned after all. I'll double-check where it was used (and whether I accidentally dropped it).

jgrassler:
Ok, tested on an HA cloud (with the fix for the problem @matrixik commented on in place) and works fine. I'll rebase now. I consider this ready now. Please do not merge it, yet, though. For this to work we need the packages

* `grafana`
* `monasca-grafana-datasource`
* and `grafana-natel-discrete-panel`

in our cloud repositories first (I'll remove the "do not merge yet" flag once that's the case).

jgrassler:
Ok, rebased and very much ready for review now.

chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -31,34 +50,68 @@
...
+end
+
+horizon_grafana_datasource "Monasca (Crowbar)" do
+  action :nothing
+  is_default true
+  user_name "admin"
+  password grafana_password
+  grafana_url grafana_base_url
+  proxy_url "../monitoring/proxy/"
+  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
dirkmueller: so this is a little bit problematic. this is interfacing with the monasca node, which does not start until the cluster has started, and the cluster does not start until it can talk to the monasca node. 

Basically, this is a catch 22 after a power failure, and especially problematic as the monasca node is taking the controller cluster down. can we do this differently?

jgrassler: I'm not exactly looking forward to having to touch this bit again (getting the mess of dependencies in here right took a few days), but I'll try to find a solution. Technically this doesn't need the Monasca node to be up at all since it does not take any actions whatsoever until a user actually logs in to Horizon. So maybe we could remove the dependencies on the Monasca node here.

jgrassler: Hmm, this is not interfacing with the Monasca node at all. It is interfacing with Grafana on the cluster founder. So we should be good there. So much for the good news.

The bad news: I discovered another problem while investigating this. Namely the `grafana-server` resource a little further up. It has an `action :nothing` directive and thus only ever gets triggered by changes to `grafana.ini` (due to that thicket of dependencies). That means it never gets restarted if it crashes. I'll try to find a way to plug _that_ gap...

jgrassler:
This is almost there but there's a bit of a race condition left (sometimes grafana-server gets restarted too early, interrupting database migration. That can brick grafana if the timing works out badly).

jgrassler:
Ok, this fixes the race condition by causing systemd to block upon service startup until migrations are complete: https://github.com/jgrassler/grafana/tree/systemd-sdnotify

I'll run the upstream tests on it and submit a pull request tomorrow (and add a patch to our packages probably since we're using an older Grafana version and I'd rather not switch to `master/HEAD` at this stage...)

jgrassler:
Ok, pull request submitted against upstream Grafana: https://github.com/grafana/grafana/pull/10025

jgrassler:
Backport: https://github.com/crowbar/crowbar-openstack/pull/1487

PR #1392, opened 2017-10-23
[4.0] neutron: use service account for neutron-l3-ha service

stefannica:
(backports #1389)

Update the neutron barclamp to use the neutron service user credentials for the neutron-l3-ha/neutron-ha-tool service.

The neutron-l3-ha service was previously configured to use the admin user, which:
 1. is not necessary - the neutron service user is more appropriate
 2. is problematic during a keystone admin password update, because it takes some time for the new admin password to reach the neutron barclamps in the chef-run, which means the neutron-l3-ha service may be triggered and run with the old admin password value and ultimately fail.


stefannica:
Renamed `service_project` back to `service_tenant`.

PR #1374, opened 2017-10-17
Add resource limits controls (bsc#1020922) 

Itxaka:
Some services require modifications to the system resource limits, such
as the maximum number of open files, that we cannot necessarily predict
a reasonable default for. Without this patch, there is no way to control
those limits from crowbar and they must be changed manually. This patch
adds a resource_limits parameter to a few barclamps to allow
administrators to set these limits for specific services from the
barclamp that controls those services. Currently only the LimitNOFILE
(systemd's name for the maximum number of open files) limit is supported.

This pull request is split into four commits and it will be easiest to review them each one at a time.

 - Use utils_systemd_override_limits LWRP for the cinder barclamp.
 - Use the same LWRP in the keystone and horizon barclamps to demonstrate
    how it should be used for barclamps that share a service (take the max).
 - Add the parameter to the rabbitmq barclamp, which will either use the systemd_override resource or /etc/security/limits.conf depending on whether it is deployed with pacemaker or not.
 - Add the parameter to the database barclamp, which will either use the systemd_override resource or /etc/security/limits.conf depending on whether it is deployed with pacemaker or not.

The required LWRP is https://github.com/crowbar/crowbar-core/pull/1365

PR #1393, opened 2017-10-23
[4.0]  Add resource limits controls (bsc#1020922)

Itxaka:

    Use utils_systemd_override_limits LWRP for the cinder barclamp.
    Use the same LWRP in the keystone and horizon barclamps to demonstrate
    Add the parameter to the rabbitmq barclamp, which will either use the systemd_override resource or /etc/security/limits.conf depending on whether it is deployed with pacemaker or not.
    Add the parameter to the database barclamp, which will either use the systemd_override resource or /etc/security/limits.conf depending on whether it is deployed with pacemaker or not.

The required LWRP is https://github.com/crowbar/crowbar-core/pull/1378
Backport-of: https://github.com/crowbar/crowbar-openstack/pull/1374


Itxaka:
[increase max open file limits](https://trello.com/c/E0gqtDEF/151-increase-max-open-file-limits)

PR #1395, opened 2017-10-24
[4.0] mariadb: Make HA op timeouts configurable

rsalevsky:
In big deployments it might be required to adapt the timeouts according
to the deployments. This change allows the operator to specify the
timeouts which fit best.

(cherry picked from commit 2c5d20d15f8e96ad206eddb04d726433b0f2615e)

Backport of #1384

### DIFF
--- a/chef/data_bags/crowbar/migrate/database/105_add_pacemaker_timeouts.rb
+++ b/chef/data_bags/crowbar/migrate/database/105_add_pacemaker_timeouts.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["mysql"]["ha"] = ta["mysql"]["ha"]
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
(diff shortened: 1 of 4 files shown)


chef/data_bags/crowbar/migrate/database/105_add_pacemaker_timeouts.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["mysql"]["ha"] = ta["mysql"]["ha"]
```
vuntz: As a quick note: if this is the migration script for master, then it's a bug, because the migration in master will overwrite any change that users have made when using stable/4.0. (It's missing the `unless a["mysql"].key?("ha")` bit)

rsalevsky: Will create a followup PR.

PR #1396, opened 2017-10-24
rabbitmq: Make sure rabbitmq is running on cluster HA

Itxaka:
As the resource agent for rabbitmq with cluster HA restart the rabbitmq
service several times, the current check can fail to validate rabbitmq
status, as it could do the check just on one of those times that rabbit
is up while creating/joining the cluster. Then if the check passed and
continued the chef execution, the next steps could fail as they are
dependant on having a running rabbitmq, while the rabbitmq server may
still be restarting.

Instead expand the checks to first look for a rabbit master for the
resource and expand the check for a local runing rabbit to make sure
we are checking for the local copy.

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
+++ b/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
@@ -95,12 +95,26 @@ ruby_block "wait for #{ms_name} to be started" do
   block do
     require "timeout"
     begin
-      Timeout.timeout(240) do
-        # Check that the service is running
+      Timeout.timeout(300) do
+        # Check that the service has a master
         cmd = "crm resource show #{ms_name} 2> /dev/null "
-        cmd << "| grep -q \"is running on\""
+        cmd << "| grep \"is running on\" | grep -q \"Master\""
         until ::Kernel.system(cmd)
-          Chef::Log.debug("#{ms_name} still not started")
+          Chef::Log.info("#{ms_name} still without master")
+          sleep(2)
+        end
+        # Check that the service is running on this node
+        cmd = "crm resource show #{ms_name} 2> /dev/null "
+        cmd << "| grep -q \"is running on: #{node.hostname}\""
+        until ::Kernel.system(cmd)
+          Chef::Log.info("#{ms_name} still not running locally")
+          sleep(2)
+        end
+        # Check that we dont have any pending resource operations
+        cmd = "crm resource operations #{ms_name} 2> /dev/null "
+        cmd << "| grep -q \"pending\""
+        while ::Kernel.system(cmd)
+          Chef::Log.info("resource #{ms_name} still has pending operations")
           sleep(2)
         end
 
@@ -119,12 +133,13 @@ ruby_block "wait for #{ms_name} to be started" do
         cmd = "rabbitmqctl -q status 2> /dev/null "
         cmd << "| sed -n '/{running_applications/,/\]}/p' | grep -q '{rabbit,'"
         until ::Kernel.system(cmd)
-          Chef::Log.debug("#{ms_name} still not answering")
+          Chef::Log.info("#{ms_name} still not answering")
           sleep(2)
         end
       end
     rescue Timeout::Error
-      message = "The #{ms_name} pacemaker resource is not started. Please manually check for an error."
+      message = "The #{ms_name} pacemaker resource is not started or doesn't have a master yet."
+      message << " Please manually check for an error."
       Chef::Log.fatal(message)
       raise message
     end
--- a/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
+++ b/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
@@ -17,6 +17,9 @@ pid_file = "/var/run/rabbitmq/pid"
 
 agent_name = "ocf:rabbitmq:rabbitmq-server-ha"
 
+service_name = "rabbitmq"
+ms_name = "ms-#{service_name}"
+
 # create file that will be sourced by OCF resource agent on promote
 template "/etc/rabbitmq/ocf-promote" do
   source "ocf-promote.erb"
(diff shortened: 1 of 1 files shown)


chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -100,8 +100,18 @@
         # running
         cmd = "rabbitmqctl -q status 2> /dev/null "
         cmd << "| sed -n '/{running_applications/,/\]}/p' | grep -q '{rabbit,'"
-        until ::Kernel.system(cmd)
-          Chef::Log.debug("#{ms_name} still not answering")
+        # check that at least we have confirmed that the app is up for 5 times in a row as when
+        # clustering we restart rabbit several times
+        success = 0
+        while success < 5
```
nicolasbock: I am struggling to understand why 5 is a good value here. I don't mean to say that it's necessarily a bad value, just trying to see whether there is more meaning behind it :smile:.

Itxaka: Not really but 5 * 2 seconds per check makes a total of 10 seconds where the app has been running, and that should be time enough to make sure that indeed rabbitmq is already up and not on the start-stop-reset-start phase when creating a cluster.

totally random but I think its just a good value between making the check too long and making it too short that it could lead to failures again :)

hope that makes sense?

nicolasbock: Ok, that makes sense. And don't give me this "5 was totally random" story! I don't buy it :wink:

I am concerned about what happens in CI though, especially when it is oversubscribed and timings just go to out the window. A very quick Google search turned up [this post](http://blog.clusterlabs.org/blog/2015/reliable-notifications) and I am wondering now whether it would be possible to use that to notify the `ruby_block` in the recipe so that we don't have to guess how long to wait until pacemaker is finally done restarting things. What do you think?

Itxaka: well I think that in the case of rabbit the issue cannot be solved by that as on my tests, pacemaker would return a simple "$service is running on $host" even if the service is restarting. and the service restarting is part of the agent script, which I dont think pacemaker has any visibility of, so it may not be possible :(

Itxaka: Although reading the page you linked, it seems like a nice way of pontificating that a resource is stable, although it would need more changes than this, we should definitely explore that.

Im thinking follow up PR after some investigation into that, sounds good. Thoughts @nicolasbock ?

Itxaka:
@vuntz @nicolasbock updated with a more simple check for master, then local rabbit. The combination of both should make sure that we only continue when things are calm and after restarts.

vuntz:
@Itxaka the CI is failing and that looks related to the patch

Itxaka:
This doesnt really work either, as soon as pacemaker has chosen a master the output of `crm resource show rabbitmq` returns a "valid" master but rabbit its still restarting and promoting :(

Itxaka:
time chef continued the run:
```
[2017-12-15T14:58:08+00:00] INFO: ruby_block[wait for ms-rabbitmq to be started] called
```

at that same time, rabbitmq was still restarting:
```
lrmd:   2017/12/15_14:58:25 INFO: rabbitmq[27984]: promote: Updating cluster master attribute
```

Itxaka:
parsing the output of `rabbitmqctl cluster_status` is also absurd, because for a brief moment you can get the proper output before the services start restarting

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -88,11 +88,25 @@
...
+        # Check that the service is running on this node
+        cmd = "crm resource show #{ms_name} 2> /dev/null "
+        cmd << "| grep -q \"is running on: #{node.hostname}\""
+        until ::Kernel.system(cmd)
+          Chef::Log.info("#{ms_name} still not running locally")
+          sleep(2)
+        end
+        # Check that we dont have any pending resource operations
+        cmd = "crm resource operations #{ms_name} 2> /dev/null "
+        cmd << "|grep -q \"pending\""
```
vuntz: s/|grep/| grep/

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -101,12 +115,13 @@
...
         cmd << "| sed -n '/{running_applications/,/\]}/p' | grep -q '{rabbit,'"
         until ::Kernel.system(cmd)
-          Chef::Log.debug("#{ms_name} still not answering")
+          Chef::Log.info("#{ms_name} still not answering")
           sleep(2)
         end
       end
     rescue Timeout::Error
-      message = "The #{ms_name} pacemaker resource is not started. Please manually check for an error."
+      message = "The #{ms_name} pacemaker resource is not started or doesnt have a master yet."
```
vuntz: s/doesnt/does not/ (or doesn't)

Itxaka:
IIRC this also didnt catch all scenarios as there was a brief period of time where allt he checks can pass and we are still restarting ¬_¬

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -96,11 +96,25 @@
     require "timeout"
     begin
       Timeout.timeout(240) do
-        # Check that the service is running
+        # Check that the service has a master
         cmd = "crm resource show #{ms_name} 2> /dev/null "
-        cmd << "| grep -q \"is running on\""
+        cmd << "| grep -q \"is running on\" | grep -q \"Master\""
```
ilausuch: I think that if the first grep uses -q won't pass anything to next grep and it will always fail. I think the correct line is:
cmd << "| grep \"is running on\" | grep -q \"Master\""

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -95,26 +95,46 @@
   block do
     require "timeout"
     begin
-      Timeout.timeout(240) do
-        # Check that the service is running
+      Timeout.timeout(300) do
+        # Check that the service has a master
         cmd = "crm resource show #{ms_name} 2> /dev/null "
-        cmd << "| grep -q \"is running on\""
+        cmd << "| grep \"is running on\" | grep -q \"Master\""
```
AbelNavarro: I would swap the check for Master and the check for locally running. You require rabbitmq in all nodes locally before you can get a Master.

Itxaka:
@jsuchome I will rebase this afterwards I guess, no problems there

jsuchome:
You can rebase now

Itxaka:
rebased! @jsuchome the check for upgrade was moved to the top, as the first check now checks if the service is running locally, which in the upgrade case we dont want to do so

rsalevsky:
Needs a rebase again.

jsuchome:
Was this tested during the upgrade?

Itxaka:
> Was this tested during the upgrade?

only the backport may affect the upgrade wont it?

In any case, testing it.

jsuchome:
> only the backport may affect the upgrade wont it?

Nope. This part is executed on the first upgraded node during the chef-client run.

> In any case, testing it.

Mee too. Actually not only it does not seem to break anything, it's possible it's fixing one of our other rabbitmq-related problems...

Itxaka:
> Mee too. Actually not only it does not seem to break anything, it's possible it's fixing one of our other rabbitmq-related problems...

Solving stuff and not even knowing it :dancing_men:

jsuchome:
I was too optimistic, it does not seem to solve the problem I was watching. But it also does not seem to break anything...

PR #1401, opened 2017-10-26
horizon: Fix apache cipher suite

cmurphy:
It seems that the DEFAULT_SUSE macro is not defined in the version of
openssl present for SLES SP1, and results in apache being unable to
start. Verbosely define the cipher suite list instead.

### DIFF
--- a/chef/cookbooks/horizon/templates/suse/openstack-dashboard.conf.erb
+++ b/chef/cookbooks/horizon/templates/suse/openstack-dashboard.conf.erb
@@ -26,7 +26,7 @@
 
 <VirtualHost <%= @bind_host %>:<%= @bind_port_ssl %>>
     SSLEngine On
-    SSLCipherSuite DEFAULT_SUSE
+    SSLCipherSuite EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH
     SSLProtocol all -SSLv2 -SSLv3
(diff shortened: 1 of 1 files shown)


aplanas:
@nicolasbock and myself worked on that, the fix is to update the apache2 package. I think that this was done and approved, but not sure if the package was also accepted for update

cmurphy:
@aplanas do you have links for those requests? Which repositories should it be submitted to?

aplanas:
https://trello.com/c/6KyzXOFq/347-bug-1043484-apache2-2416-sles12-sp1-modssl-does-not-expand-defaultsuse-cipher-suite

cmurphy:
The package is clearly not updated in SP1 Updates and it sounds like SP1 is not supported, so I think it makes sense to take care of this ourselves. But I found https://github.com/crowbar/crowbar-openstack/pull/1038 so we should merge that instead.

nicolasbock:
According to https://bugzilla.suse.com/show_bug.cgi?id=1043484#c16 the update made it into 12-SP1-LTSS. Instead of fixing this ourselves we should find out why the updated package is not available through the normal SP1 update channel.

cmurphy:
Aha - I found it in http://clouddata.cloud.suse.de/repos/x86_64/SUSE-OpenStack-Cloud-6-Updates/x86_64/ so maybe something is not syncing right in mkcloud

PR #1405, opened 2017-11-01
manila: Switch manila-api to run under WSGI

s-t-e-v-e-n-k:
The current release of manila no longer supports running the API service
directly while using SSL, and instead requires the use of WSGI.

---

I'm not certain this correct for HA, it deploys, but only one of the clone sets runs successfully because it binds to the VIP.

### DIFF
--- a/chef/cookbooks/manila/recipes/api.rb
+++ b/chef/cookbooks/manila/recipes/api.rb
@@ -145,8 +145,17 @@ end
 
 ssl_enabled = node[:manila][:api][:protocol] == "https"
 
+if node[:manila][:ha][:enabled]
+  admin_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+  manila_bind_host = admin_address
+  manila_port = node[:manila][:ha][:ports][:api]
+else
+  manila_bind_host = "0.0.0.0"
+  manila_port = manila_port
+end
+
 crowbar_openstack_wsgi "WSGI entry for manila-api" do
-  bind_host "0.0.0.0"
+  bind_host manila_bind_host
   bind_port manila_port
   daemon_process "manila-api"
   script_alias "/usr/bin/manila-wsgi"
--- a/chef/cookbooks/manila/recipes/api.rb
+++ b/chef/cookbooks/manila/recipes/api.rb
@@ -17,8 +17,12 @@
 # Recipe:: api
 #
 
+include_recipe "apache2"
+include_recipe "apache2::mod_wsgi"
 include_recipe "#{@cookbook_name}::common"
 include_recipe "#{@cookbook_name}::sql"
+
+package "openstack-manila-api"
 
 keystone_settings = KeystoneHelper.keystone_settings(node, :manila)
 
@@ -131,8 +135,31 @@ end
 
 crowbar_pacemaker_sync_mark "create-manila_register"
 
-use_crowbar_pacemaker_service = ha_enabled && node[:pacemaker][:clone_stateless_services]
+service "manila-api" do
+  service_name "openstack-manila-#{params[:name]}" if %w(rhel suse).include? node[:platform_family]
+  supports status: true, restart: true
+  action [:disable, :stop]
+  # allow to fail here because there may not be a service
+  ignore_failure true
+end
 
-manila_service "api" do
-  use_pacemaker_provider use_crowbar_pacemaker_service
+ssl_enabled = node[:manila][:api][:protocol] == "https"
+
+crowbar_openstack_wsgi "WSGI entry for manila-api" do
+  bind_host "0.0.0.0"
+  bind_port manila_port
+  daemon_process "manila-api"
+  script_alias "/usr/bin/manila-wsgi"
+  user node[:manila][:user]
+  group node[:manila][:group]
+  ssl_enable ssl_enabled
+  ssl_certfile node[:manila][:ssl][:certfile]
+  ssl_keyfile node[:manila][:ssl][:keyfile]
+  if node[:manila][:ssl][:cert_required]
+    ssl_cacert node[:manila][:ssl][:ca_certs]
+  end
+end
+
+apache_site "manila-api.conf" do
+  enable true
 end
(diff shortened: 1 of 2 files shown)


s-t-e-v-e-n-k:
I have confirmed that manila-api prior to this change is deployed as a clone set, so it shouldn't be making use of the VIP ...

chef/cookbooks/manila/recipes/api.rb
```diff
@@ -131,8 +133,50 @@
 
 crowbar_pacemaker_sync_mark "create-manila_register"
 
-use_crowbar_pacemaker_service = ha_enabled && node[:pacemaker][:clone_stateless_services]
+service "manila-api" do
+  service_name "manila-api"
```
vuntz: This should be:

```ruby
"openstack-manila-#{params[:name]}" if %w(rhel suse).include? node[:platform_family]
```
(according to manila/definitions/manila_service.rb)

chef/cookbooks/manila/recipes/api.rb
```diff
@@ -131,8 +133,50 @@
...
+  end
+end
+
+apache_site "manila-api-admin.conf" do
+  enable true
+end
+
+crowbar_openstack_wsgi "WSGI entry for manila-api-public" do
+  bind_host my_public_host
+  bind_port manila_port
```
vuntz: Why do we have two WSGI services? Why not just one listening on all IP addresses?

chef/cookbooks/manila/recipes/common.rb
```diff
@@ -14,6 +14,7 @@
 #
 
 package "openstack-manila"
+package "openstack-manila-api"
```
vuntz: I doubt this change is correct; manila-share nodes shouldn't have this.

chef/cookbooks/manila/recipes/controller_ha.rb
```diff
@@ -46,9 +48,25 @@
...
+  transaction_objects.push(objects)
+
+  order_name = "o-cl-apache2-manila-api"
+  pacemaker_order order_name do
+    ordering "cl-apache2 cl-manila-api"
+    score "Mandatory"
+    action :update
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+  transaction_objects << "pacemaker_order[#{order_name}]"
```
vuntz: I don't think you need the dummy resource, unless you have anything else that has a dependency on it (like was the case for keystone). I'd drop that part completely.

chef/cookbooks/manila/recipes/api.rb
s-t-e-v-e-n-k: Sorry, I was following the Keystone example, conveniently forgetting that public and admin for manila are on the same port.

chef/cookbooks/manila/recipes/api.rb
```diff
@@ -131,8 +133,50 @@
...
+  action [:disable, :stop]
+  # allow to fail here because there may not be a service
+  ignore_failure true
+end
+
+ssl_enabled = node[:manila][:api][:protocol] == "https"
+
+crowbar_openstack_wsgi "WSGI entry for manila-api-admin" do
+  bind_host my_admin_host
+  bind_port manila_port
```
rhafer: I think the bind_host/bind_port settings need to be different depending on ha/no-ha setup (to allow haproxy to listen on the main manila port). See https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/nova/recipes/placement_api.rb#L93 for an example how we do it in other places.

rhafer:
@s-t-e-v-e-n-k I've pushed some additional changes, which should address most of the review comments. With that I could get it to successfully apply in an HA deployment. I didn't really test it yet, neither did I try a SSL deployment.

jsuchome:
@s-t-e-v-e-n-k please keep the git messages from ralf's commits next time, thanks

rhafer:
> please keep the git messages from ralf's commits next time, thanks

I think he just didn't notice my commit before force-pushing. As some of my changes are missing. I'll rebase my changes and push again.

rhafer:
@s-t-e-v-e-n-k  I've added a few fixes. Feel free to squash them into your commit.

chef/cookbooks/manila/recipes/api.rb
```diff
@@ -131,8 +135,40 @@
...
+  ssl_certfile node[:manila][:ssl][:certfile]
+  ssl_keyfile node[:manila][:ssl][:keyfile]
+  if node[:manila][:ssl][:cert_required]
+    ssl_cacert node[:manila][:ssl][:ca_certs]
+  end
+end
 
-manila_service "api" do
-  use_pacemaker_provider use_crowbar_pacemaker_service
+apache_site "manila-api.conf" do
```
dirkmueller: maybe use openstack-manila-api.conf here?

chef/cookbooks/manila/recipes/api.rb
```diff
@@ -131,8 +135,40 @@
...
+  supports status: true, restart: true
+  action [:disable, :stop]
+  # allow to fail here because there may not be a service
+  ignore_failure true
+end
+
+ssl_enabled = node[:manila][:api][:protocol] == "https"
+
+if node[:manila][:ha][:enabled]
+  admin_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
```
vuntz: For the record: no need for the `Chef::Recipe::` prefix here.

chef/cookbooks/manila/recipes/api.rb
```diff
@@ -131,8 +135,40 @@
...
+  manila_bind_host = admin_address
+  manila_port = node[:manila][:ha][:ports][:api]
+else
+  manila_bind_host = "0.0.0.0"
+  manila_port = manila_port
```
vuntz: Interesting line ;-)

rhafer: Heh, I guess that was a copy'n past oversight ...

chef/cookbooks/manila/recipes/api.rb
rhafer: It's seems most other service settle for leaving the `openstack` away.

PR #1408, opened 2017-11-04
monasca: select a matching node for admin (bsc#1121778)

dirkmueller:
The admin node might not be called crowbar or have the admin role,
we need to use the select_nodes_for_role function which is using
the proper is_admin? check.

UNTESTED


kbaikov:
Retested in M9. I get the same error:
 ```I, [2019-01-23T16:12:58.235411 #18789:0x00000008c17c50]  INFO -- Completed 500 Internal Server Error in 18ms (ActiveRecord: 0.2ms)
F, [2019-01-23T16:12:58.236058 #18789:0x00000008c17c50] FATAL -- 
RuntimeError (Unable to find any node named crowbar):```

And monasca master is still not fqdn for crowbar: "elements"=>{"monasca-master"=>["crowbar"],

PR #1409, opened 2017-11-06
ceilometer: configurable default_api_return_limit (bsc#1046787)

stefannica:
The ceilometer admin dashboard in horizon doesn't specify a limit for the number of entries returned by ceilometer API calls, in which case the `[api]/default_api_return_limit` configuration option is used to limit the number of returned entries.
The default value of this option is 100, which may be too small in cases where the ceilometer database has a large number of entries (meters, samples, etc).

Making this option configurable as a raw attribute gives the crowbar admin another parameter that can be used to adjust ceilometer to larger environments.

Fixes [bsc#1046787](https://bugzilla.suse.com/show_bug.cgi?id=1046787)

NOTE: this option doesn't actually impact the horizon UI in SOC8 because the ceilometer dashboard has been removed in Pike.

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/common.rb
+++ b/chef/cookbooks/ceilometer/recipes/common.rb
@@ -95,7 +95,8 @@ template node[:ceilometer][:config_file] do
       metering_time_to_live: metering_time_to_live,
       event_time_to_live: event_time_to_live,
       instance_discovery_method: instance_discovery_method,
-      is_compute_agent: is_compute_agent
+      is_compute_agent: is_compute_agent,
+      default_api_return_limit: node[:ceilometer][:api][:default_return_limit]
     )
     if is_compute_agent
       notifies :restart, "service[nova-compute]"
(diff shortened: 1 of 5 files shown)


stefannica:
[P1: Bug 1061197 - L3: Customer updated OpenStack 7 via zypper and now recieves SSL errors in the openstack services, nova, etc](https://trello.com/c/fHO5Bm64/500-p1-bug-1061197-l3-customer-updated-openstack-7-via-zypper-and-now-recieves-ssl-errors-in-the-openstack-services-nova-etc)

stefannica:
[P2: Bug 1046787 - #L3 Resource usage showing just glance meters](https://trello.com/c/d0xJghT5/337-p2-bug-1046787-l3-resource-usage-showing-just-glance-meters)

stefannica:
Thanks, @sjamgade. I thought of using `api_return_limit` but it's already under `ceilometer.api`, so it felt kind of redundant. None of these choices is perfect, so I'll just stick with `default_return_limit` if you're ok with it.

PR #1410, opened 2017-11-06
mysql: config file tuning based on mysqltuner suggestions

jsuchome:
See http://pastebin.nue.suse.com/18851/src for mysqltuner output

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -55,7 +55,8 @@ unless node[:database][:galera_bootstrapped]
         sstuser: "root",
         sstuser_password: "",
         expire_logs_days: node[:database][:mysql][:expire_logs_days],
-        node_address: node_address
+        node_address: node_address,
+        wsrep_slave_threads: node[:database][:mysql][:wsrep_slave_threads]
       )
     end
 
@@ -135,7 +136,8 @@ template "/etc/my.cnf.d/galera.cnf" do
     sstuser: "sstuser",
     sstuser_password: node[:database][:mysql][:sstuser_password],
     expire_logs_days: node[:database][:mysql][:expire_logs_days],
-    node_address: node_address
+    node_address: node_address,
+    wsrep_slave_threads: node[:database][:mysql][:wsrep_slave_threads]
   )
 end
 
--- a/chef/cookbooks/mysql/templates/default/galera.cnf.erb
+++ b/chef/cookbooks/mysql/templates/default/galera.cnf.erb
@@ -8,7 +8,9 @@ wsrep_on=ON
 binlog_format=ROW
 expire_logs_days = <%= @expire_logs_days %>
 
-wsrep_provider_options = "gmcast.listen_addr=tcp://<%= @node_address %>:4567"
+wsrep_slave_threads = <%= @wsrep_slave_threads %>
+# values recommended by mysqltuner.pl
+wsrep_provider_options = "gmcast.listen_addr=tcp://<%= @node_address %>:4567;gcs.fc_limit = <%= @wsrep_slave_threads * 5 %>;gcs.fc_factor = 0.8"
 
 # SST method
 wsrep_sst_method = xtrabackup-v2
--- a/chef/cookbooks/mysql/templates/default/tuning.cnf.erb
+++ b/chef/cookbooks/mysql/templates/default/tuning.cnf.erb
@@ -1,11 +1,18 @@
 [mysqld]
 innodb_buffer_pool_size = <%= @innodb_buffer_pool_size %>M
 innodb_log_file_size = <%= [8, @innodb_buffer_pool_size / 4].max %>M
+innodb_buffer_pool_instances = <%= @innodb_buffer_pool_instances %>
+
 innodb_flush_method = O_DIRECT
+
+# 0 should bring better performance but has slight risk of losing some logs
+innodb_flush_log_at_trx_commit = <%= @innodb_flush_log_at_trx_commit %>
 
 max_connections = <%= @max_connections %>
 tmp_table_size = <%= @tmp_table_size %>M
 max_heap_table_size = <%= @max_heap_table_size %>M
+
+skip_name_resolve = 1
 
 # FIXME: Remove after MariaDB 10.2.X switch (new default is auto)
 thread_cache_size = <%= node['mysql']['tunable']['thread_cache_size'] %>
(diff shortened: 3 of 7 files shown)


chef/cookbooks/mysql/templates/default/galera.cnf.erb
```diff
@@ -8,6 +8,11 @@ wsrep_on=ON
 binlog_format=ROW
 expire_logs_days = <%= @expire_logs_days %>
 
+# numbers recommended by mysqltuner.pl
+<% wsrep_slave_threads = [node["cpu"]["total"] * 2, 512].min %>
```
rsalevsky: I would do the calculation part in the recipe and pass the values to the template.

jsuchome: I could do it, yes. But note that in other places we're also doing it in the template

jsuchome: Done

rhafer:
I haven't looked through all of the recommendations yet but this one looks relevant I think:

`[!!] name resolution is active : a reverse name resolution is made for each new connection and can reduce performance` 

It might also be worth to run the tuner script after a full openstack deployment (including a tempest run).

jsuchome:
So that is `skip_name_resolve` in `[mysqld]`

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -52,7 +52,8 @@
         cluster_addresses: "gcomm://",
         sstuser: "root",
         sstuser_password: "",
-        expire_logs_days: node[:database][:mysql][:expire_logs_days]
+        expire_logs_days: node[:database][:mysql][:expire_logs_days],
+        wsrep_slave_threads: [node["cpu"]["total"] * 2, 512].min
```
rhafer: Where does the `512` come from? That sound like a bit too much. Looking at https://mariadb.com/kb/en/library/tips-on-converting-to-galera/ it seem this setting is not really relevant that much for our single-writer setup anyway. (It might be better to turn that into a proposal setting an leave it at 1 by default)

chef/cookbooks/mysql/templates/default/tuning.cnf.erb
```diff
@@ -2,6 +2,7 @@
 innodb_buffer_pool_size = <%= @innodb_buffer_pool_size %>M
 innodb_log_file_size = <%= [8, @innodb_buffer_pool_size / 4].max %>M
 innodb_flush_method = O_DIRECT
+innodb_flush_log_at_trx_commit = 0
```
rhafer: Hm, are you sure about this? `0` seems to be somewhat unsafe. https://mariadb.com/kb/en/library/xtradbinnodb-server-system-variables/#innodb_flush_log_at_trx_commit

I guess we shouldn't do that by default. It might be an option for setups running into performance issues though (i.e. our qa2/qa3 hardware). So this might also be worth a proposal var.

chef/cookbooks/mysql/recipes/ha_galera.rb
jsuchome: 512 is the maximum value for this variable

chef/cookbooks/mysql/templates/default/tuning.cnf.erb
jsuchome: That was just recommendation from the script. You might be right about extra proposal variable.

rhafer:
I am still wondering what to do with the `innodb_log_file_size` and `innodb_buffer_pool_instances` setttings to the proposal. According to:  https://mariadb.com/kb/en/library/xtradbinnodb-server-system-variables/#innodb_buffer_pool_size it's recommended to increase innodb_buffer_pool_instances once innodb_buffer_pool_size goes beyond a certain limit (2GB) and mysqtuner recommends `innodb_log_file_size` to be 25% of innodb_buffer_pool_size. However I don't have any idea if these recommendation are really always valid.

jsuchome:
>  mysqtuner recommends innodb_log_file_size to be 25% of innodb_buffer_pool_size

That's what we actually have

rhafer:
> > mysqtuner recommends innodb_log_file_size to be 25% of innodb_buffer_pool_size
> 
> That's what we actually have

But we don't allow setting `innodb_buffer_pool_instances` anywhere, right?

jsuchome:
> But we don't allow setting innodb_buffer_pool_instances anywhere, right?

Right. I can add it to this PR.

chef/data_bags/crowbar/migrate/database/207_add_more_tuning.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  a["mysql"]["innodb_flush_log_at_trx_commit"] = ta["mysql"]["innodb_flush_log_at_trx_commit"]
```
rhafer: We should make these backport safe by adding `unless a["mysql"].key? "innodb_flush_log_at_trx_commit"` and so on around it.

jsuchome: done

rhafer:
> So, not sure if default value for innodb_buffer_pool_instances is correct, mysqltuner.pl suggests 1.

The mariadb documentation suggests 1 as well as long as the buffer pool is less than 2G. When more than that it recommends to have at least 1G buffer pool per instance. So I'd go with one. (https://mariadb.com/kb/en/library/xtradbinnodb-server-system-variables#innodb_buffer_pool_instances)

> Maybe join_buffer_size (> 256.0K ...) ... but that's very close to default value anyway.

Well, it does recommend how much more than 256k ;-)
But as we don't have any concrete suggestion here, let's leave it as it is for now.

chef/data_bags/crowbar/migrate/database/207_add_more_tuning.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(ta, td, a, d)
+  a["mysql"]["innodb_flush_log_at_trx_commit"] = ta["mysql"]["innodb_flush_log_at_trx_commit"] unless a["mysql"]["innodb_flush_log_at_trx_commit"]
```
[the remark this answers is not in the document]
rsalevsky: Hound is right here the line is way to long... I can't even use the side-by-side diff in GitHub to review this PR properly.

jsuchome:
@rsalevsky adapted

jsuchome:
> Additionally is it necessary to expose the tuning as config options rather than just setting them for all 
> deployments?

What do you mean? Setting some reasonable values by ourselves? That was the original approach, but then we'd thought although some kind of setting might be good for performance, it might have impact elsewhere, so let's rather let users decide.

jhesketh:
Yep, that's what I meant (setting it to something reasonable ourselves). But very happy for it to be configurable if we think it'll be useful :-)

rsalevsky:
@jsuchome Does it makes sense to cherry-pick the changes from #1445? Otherwise one or the other PR has to be rebased and they have both the same topic.

jsuchome:
Theoretically yes, but I do not have time to test https://github.com/crowbar/crowbar-openstack/pull/1445 now and I do not know which of those options from @dirkmueller are supposed to stay and which are meant just for testing...

dirkmueller:
@jsuchome lets not worry about #1445 for now, I'll rebase once your patch is in. can you rebase yours? I would like to merge it

jsuchome:
@dirkmueller rebased

PR #1411, opened 2017-11-06
neutron: fix additional external nets on compute nodes

stefannica:
On compute nodes, the `additional_external_networks` barclamp attribute used to configure custom neutron external networks was improperly taken from the local chef node instead of the remote 'neutron-network' node.

When used in combination with a DVR enabled setup, which requires the external networks to be present on compute nodes too, this mishap basically caused the custom external networks to be unreachable without a clear error indication in the neutron logs.

Fixes: [bsc#1066171](https://bugzilla.suse.com/show_bug.cgi?id=1066171)


stefannica:
[Bug 1066171 - L3-Question: Interface Binding Fails when attaching to a network](https://trello.com/c/e9YLFSw4/582-bug-1066171-l3-question-interface-binding-fails-when-attaching-to-a-network)

PR #256, opened 2016-02-04
mysql: Fix undefined attribute 'ipaddress' (bsc#921015)

ovisan:
ipddress fails, needs to be node['ipaddress']

PR #1417, opened 2017-11-07
[stable/4.0] Use httpchk in haproxy for galera

jsuchome:
By using a script provided in the galera-python-clustercheck package we
are able to ensure a node is in sync with the galera cluster rather
than the previously superficial check of just ensuring mysql is running.

Install the package and configure it to use the monitoring user.

This requires a change to crowbar-ha in PR #256 to allow extending
the haproxy pacemaker group.

(cherry picked from commit 9efe0aac04167a4d7efa77a93f3e4392df148710)

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -19,7 +19,7 @@
 
 resource_agent = "ocf:heartbeat:galera"
 
-["galera-3-wsrep-provider", "mariadb-tools", "xtrabackup", "socat"].each do |p|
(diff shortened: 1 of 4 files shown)


jsuchome:
> lgtm in terms of being a good backport, but I haven't tested it.

Required packages do not seem to be at SOC7 media, I've opened SR for that

rsalevsky:
Please don't merge until we have the packages officially out. I will create a ECO to add the new packages to Cloud 7 but it can take some time till it is approved.

vuntz:
Same backport in https://github.com/crowbar/crowbar-openstack/pull/1440 got merged

PR #1418, opened 2017-11-07
Support neutron-dhcp-agent config snippets

tpatzig:
Moved chef generated dhcp_agent.ini content to
/etc/neutron/neutron-dhcp-agent.conf.d/100-neutron-dhcp-agent.conf,
because the original /etc/neutron/dhcp_agent.ini has precedence
over config snippets in /etc/neutron/neutron-dhcp-agent.conf.d.

### DIFF
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -26,6 +26,7 @@ default[:neutron][:config_file] = "/etc/neutron/neutron.conf.d/100-neutron.conf"
 default[:neutron][:lbaas_service_file] = "/etc/neutron/neutron-server.conf.d/100-neutron_service_lbaas.conf"
 default[:neutron][:lbaas_config_file] = "/etc/neutron/neutron.conf.d/110-neutron_lbaas.conf"
 default[:neutron][:l3_agent_config_file] = "/etc/neutron/neutron-l3-agent.conf.d/100-agent.conf"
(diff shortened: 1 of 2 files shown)


tpatzig:
needs backport to stable/4.0 (SOC7)

tpatzig:
superseded by https://github.com/crowbar/crowbar-openstack/pull/1419

aplanas:
IIUC the problem exposed by @vuntz (sorry if is not the case) and issue can happen after the migration from `agent.ini` config to `agent.ini.d/100-agent.ini` if this is meet:

* the old agent.ini gets unmanaged by chef
* there is a configuration value set in agent.ini that is not over-written by `agent.ini.d/100-agent.ini`

I think that to fix this we need indeed validate the defaults in the configuration file that is in the package. In the case of neutron they are taken directly from the etc/samples directory, that is using the defaults generated by oslo.config[1], so I think that we are safe from the package side.

The spec file is not overwriting the configuration file, so that means that after an update of neutron, the old agent.ini file will contain the very old agent.ini that chef was managing. To be safe I would add a migration in #1419, that will indeed empty the now unmanaged files.

[1] https://github.com/openstack/neutron/blob/master/etc/oslo-config-generator/metadata_agent.ini

vuntz:
> there is a configuration value set in agent.ini that is not over-written by agent.ini.d/100-agent.ini

Actually the config snippet cannot overwrite the value set in agent.ini since agent.ini is explicitly passed as a `--config-file` argument, and therefore wins. That's also a reason why I tend to think that the cookbook should empty these files.

vuntz:
I added a commit to the other pull request to address that.

PR #1419, opened 2017-11-08
neutron: Fix config snippet support for neutron agents

tpatzig:
Because the dhcp/lbaasv2/metadata agents are started with an explicit
"--config-file /etc/neutron/$foo_agent.ini" flag, options in this file
will override everything that is being set in config snippets.

So do not generate /etc/neutron/$foo_agent.ini files, but instead config
snippets.

Co-Authored-By: Tom Patzig <tom.patzig@sap.com>

### DIFF
--- a/chef/cookbooks/neutron/definitions/neutron_metadata.rb
+++ b/chef/cookbooks/neutron/definitions/neutron_metadata.rb
@@ -57,6 +57,18 @@ define :neutron_metadata,
 
   keystone_settings = KeystoneHelper.keystone_settings(neutron, @cookbook_name)
 
+  # Empty the config file that is explicitly passed to the metadata agent, as
+  # its content will prevail compared to config snippets (because the file is
+  # explicitly passed as a config file, not loaded as part of the default files
+  # to load).
+  file "/etc/neutron/metadata_agent.ini" do
+    owner "root"
+    group node[:neutron][:platform][:group]
+    mode "0640"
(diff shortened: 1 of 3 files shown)


chef/cookbooks/neutron/definitions/neutron_metadata.rb
```diff
@@ -57,6 +57,18 @@
...
+  # explicitly passed as a config file, not loaded as part of the default files
+  # to load).
+  file "/etc/neutron/metadata_agent.ini" do
+    owner "root"
+    group node[:neutron][:platform][:group]
+    mode "0640"
+    content "# Please use config file snippets in /etc/neutron/neutron-metadata-agent.conf.d/.\n" \
+            "# See /etc/neutron/README.config for more details.\n"
+  end
+
```
nicolasbock: Why don't we backup the file's contents in this case so that the user doesn't lose their local modifications and settings?

vuntz: @nicolasbock chef will do the backup in /var/chef/backup :-)

vuntz: (and, well, there's no local modifications -- the file was owned by chef before)

nicolasbock: That's neat :smile:

PR #1421, opened 2017-11-08
[4.0] Backport flavor creation fixes / improvements

vuntz:
**nova: improve flavors creation performance**

- reusing the output of flavor-list
- subscribes and not_if are independent
  as a result the block will always try to create the flavors even if
they exist

This saves about 15-20 secs on non-shared hardware and about 1min on a
loaded CI

**nova: create flavors via the internal endpoint**

This is consistent with every other openstack related usage
in the chef recipies. we want to avoid traffic going over
public unnecessary in case public is unreachable or broken.

Backport of https://github.com/crowbar/crowbar-openstack/pull/1348 and https://github.com/crowbar/crowbar-openstack/pull/1400


abel-navarro:
The error in Travis looks unrelated. Rebuilding.

PR #1423, opened 2017-11-08
[4.0] nova: skip unchanged nodes chef-run

Itxaka:
By using skip_batch_for_node? we can avoid running chef on nodes that do
not change role and if there is no attribute change for the barclamp.

Does not skip remote HA nodes, nodes that have changed roles, all nodes if
attributes have changed and when deploying the barclamp for the first
time.

(cherry picked from commit ce4e336528e62cb7cf8373e6986069bd4939a008)

Backport-of: https://github.com/crowbar/crowbar-openstack/pull/1385

### DIFF
--- a/crowbar_framework/app/models/nova_service.rb
+++ b/crowbar_framework/app/models/nova_service.rb
@@ -504,7 +504,82 @@ class NovaService < OpenstackServiceObject
     super
   end
 
+  # try to know if we can skip a node from running chef-client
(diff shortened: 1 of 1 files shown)


Itxaka:
[Skip batches when applying nova barclamp](https://trello.com/c/rYkjyvHO/161-skip-batches-when-applying-nova-barclamp)

PR #1424, opened 2017-11-09
mysql: Use [:database][:mysql] attributes rather than just [:mysql]

jsuchome:
Use correct attribute hierarchy that is located under [:database] (bsc#1058876).

(This is rebasesed https://github.com/crowbar/crowbar-openstack/pull/1314)


chef/cookbooks/mysql/attributes/server.rb
```diff
@@ -17,18 +17,17 @@
...
+  default[:database][:mysql][:ebs_vol_size]            = 50
 end
 
-default["mysql"]["tunable"]["max_allowed_packet"]       = "16M"
-default["mysql"]["tunable"]["thread_cache_size"]        = 8
+default[:database][:mysql][:tunable][:max_allowed_packet]       = "16M"
+default[:database][:mysql][:tunable][:thread_cache_size]        = 8
 
 # Ports to bind to when haproxy is used
 default[:mysql][:ha][:ports][:admin_port] = 3306
```
nicolasbock: Are we moving the `[:mysql][:ha]` keys into `[:database]` as well?

jsuchome: No, I think these are really just `[:mysql]`, I think it is correct

nicolasbock: Ok, it just struck me that it looks so 'asymmetric' this way :smile:

PR #1429, opened 2017-11-13
Set socket_timeout for memcached connections to 1s

vuntz:
The default is 3 seconds, which is very high. When one memcached server
goes down, then every connection pool for memcached (in each process, on
each node) need to identify that this server is down, and that has some
dramatic impact on performance.

Lowering this to 1 second, while still being a reasonable value, helps
quite a bit. I would even consider lowering this to 0.25, but oslo.cache
doesn't support float values there (see
https://bugs.launchpad.net/oslo.cache/+bug/1731921).

Helps with https://bugzilla.suse.com/show_bug.cgi?id=1038223

### DIFF
--- a/chef/cookbooks/aodh/templates/default/aodh.conf.erb
+++ b/chef/cookbooks/aodh/templates/default/aodh.conf.erb
@@ -27,6 +27,7 @@ auth_type = password
 memcached_servers = <%= @memcached_servers.join(',') %>
 memcache_security_strategy = ENCRYPT
 memcache_secret_key = <%= node[:aodh][:memcache_secret_key] %>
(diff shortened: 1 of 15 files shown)


vuntz:
Btw, I noticed that ec2-api doesn't seem to have the memcached settings in `keystone_authtoken`. Is that intended?

vuntz:
> I'm not 100% sure, but I don't think this is needed anymore, because only one memcached server is used - the local one - even when HA is enabled. Please take a look at #1341 and let me know if I'm wrong.

Check again, we didn't merge that PR because it's actually incorrect to only use the local memcached server :-)

PR #1434, opened 2017-11-14
Set heartbeat_timeout_threshold to improve recovery when one rabbitmq node goes down

vuntz:
This will be used to configure the heartbeat timeout for clients
connecting to rabbitmq. See https://www.rabbitmq.com/heartbeats.html for
details about heartbeats.

The default used by oslo.messaging is 60 seconds, which is extremely
high. There are various recommendations out there, the rabbitmq website
has two different ranges: 5-20s [1] and 8-20s [2]. Another
recommendation is 6-12s. 10s is within these ranges, so it's a good pick
for now. Some setups may want to be more aggressive here, hence the
attribute to allow this.

[1] https://www.rabbitmq.com/heartbeats.html
[2] https://www.rabbitmq.com/networking.html
[3] http://openstack-qe-guy.blogspot.fr/2015/11/rabbitmq-best-practices.html

### DIFF
--- a/chef/data_bags/crowbar/migrate/rabbitmq/201_heartbeat_timeout.rb
+++ b/chef/data_bags/crowbar/migrate/rabbitmq/201_heartbeat_timeout.rb
@@ -1,0 +1,16 @@
+def upgrade(ta, td, a, d)
+  a["client"] ||= {}
+  unless a["client"]["heartbeat_timeout"]
+    a["client"]["heartbeat_timeout"] = ta["client"]["heartbeat_timeout"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  if ta.key?("client")
+    a["client"].delete("heartbeat_timeout") unless ta["client"].key?("heartbeat_timeout")
+  else
+    a.delete("client")
+  end
+  return a, d
+end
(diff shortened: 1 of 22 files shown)


chef/data_bags/crowbar/migrate/rabbitmq/201_heartbeat_timeout.rb
```diff
@@ -0,0 +1,16 @@
...
+  a["client"] ||= {}
+  unless a["client"]["heartbeat_timeout"]
+    a["client"]["heartbeat_timeout"] = ta["client"]["heartbeat_timeout"]
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  if ta.key?("client")
+    a["client"].delete("heartbeat_timeout") unless ta["client"].key?("heartbeat_timeout")
```
skazi0: is it possible that `"client"` doesn't exist in `a` at this point?

vuntz: No, not in `a`: we're starting from the state where the `client/heatbeat_timeout` attribute exists, and we try to remove it.

PR #1440, opened 2017-11-16
[4.0] Use httpchk in haproxy for galera (requires manual enablement from user)

vuntz:
By using a script provided in the galera-python-clustercheck package we
are able to ensure a node is in sync with the galera cluster rather
than the previously superficial check of just ensuring mysql is running.

Install the package and configure it to use the monitoring user.

This depends on https://github.com/crowbar/crowbar-ha/pull/277.

This is not usable in this branch until the galera-python-clustercheck
package is made available. However, some people may want to manually
install that package from OBS, and they can do so by simply setting an
attribute (not in the barclamp, but as an override attribute in the
cookbook).

Backport of https://github.com/crowbar/crowbar-openstack/pull/1277 with an additional commit to block the feature unless explicitly enabled.

### DIFF
--- a/chef/cookbooks/mysql/attributes/server.rb
+++ b/chef/cookbooks/mysql/attributes/server.rb
@@ -37,3 +37,6 @@ default[:mysql][:ha][:ports][:admin_port] = 3306
 # in pacemamker
 default[:mysql][:ha][:op][:monitor][:interval] = "20s"
 default[:mysql][:ha][:op][:monitor][:role]     = "Master"
+
+# Let users override this if galera-python-clustercheck is available to them
+default[:mysql][:ha][:clustercheck] = false
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -19,7 +19,7 @@
 
 resource_agent = "ocf:heartbeat:galera"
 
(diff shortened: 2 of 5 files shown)


rhafer:
Note that this would obsolete https://github.com/crowbar/crowbar-openstack/pull/1417

vuntz:
> I think we should set a default for the clustercheck attribute in cookbooks/mysql/attributes/server.rb otherwise this look good.

Done. It did work without value, but I guess it's more explicit this way, indeed.

rsalevsky:
Should we remove this explizit enablement after the package is available to everyone?

vuntz:
> Should we remove this explizit enablement after the package is available to everyone?

Yes, we'll revert the patch once the ECO is completed.

jhesketh:
> This is not usable in this branch until the galera-python-clustercheck
package is made available. However, some people may want to manually
install that package from OBS, and they can do so by simply setting an
attribute (not in the barclamp, but as an override attribute in the
cookbook).

Do we know if people actually want to use this?

If we are going to revert the second commit eventually, unless we explicitly need this now, why not wait on the ECO and then merge #1417?

vuntz:
> > This is not usable in this branch until the galera-python-clustercheck
> > package is made available. However, some people may want to manually
> > install that package from OBS, and they can do so by simply setting an
> > attribute (not in the barclamp, but as an override attribute in the
> > cookbook).
>
> Do we know if people actually want to use this?

Yes, some people want to use this, that's exactly the reason why I did this ;-)

PR #1441, opened 2017-11-16
Rework "new backend" UI for cinder and manila

vuntz:
Use the same way of presenting the new backend as the way used for
existing backends, instead of displaying this as a new section.

Old UI:
old add backend

New UI:
new add backend

### DIFF
--- a/crowbar_framework/app/views/barclamp/cinder/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/cinder/_edit_attributes.html.haml
@@ -221,15 +221,15 @@
       %div#cinder_backends
         = t(".volumes.loading_text")
 
-      %fieldset
-        %legend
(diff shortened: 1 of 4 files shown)


vuntz:
Note that CI won't work because this conflicts with a patch in our package :/

dirkmueller:
@vuntz so the previous "add backend" button placement indicated (to me at least) that the backend name and type should be set prior clicking Add. (top to bottom reading flow of a user). Now the button is "sideways" which is confusing (some users might not scroll down far enough to actually see the type and name input field, and will then be confused on how to select the type on the newly added backend). 

can we perhaps preserve that?

vuntz:
@dirkmueller this is possible, but I guess I'd prefer to make the button insensitive until the user has explicitly picked a backend type in the select menu. Let's see if I have time to play with it eventually...

PR #1450, opened 2017-11-23
monasca: use exit status from monasca-setup

jgrassler:
Currently monasca-reconfigure just forges on if one of its monasca-setup
invocations fails. This commit changes both monasca-setup scripts to run
with -e so they abort with a non-zero exit status if monasca-setup exits
non-zero.

### DIFF
--- a/chef/cookbooks/monasca/templates/default/monasca-reconfigure-server.erb
+++ b/chef/cookbooks/monasca/templates/default/monasca-reconfigure-server.erb
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -e
+
 # This script is a monasca-setup wrapper that configures monasca-agent on the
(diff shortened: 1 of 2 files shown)


jgrassler:
Do not merge this, yet: we need to make sure to delete `kibana.yaml` if it's in the broken state from https://storyboard.openstack.org/#!/story/2001311 (otherwise the fix for that bug won't work).

jgrassler:
Ok, I added some extra code for detecting and removing the broken `kibana.yaml` if required now. As far as I am concerned this is ready to merge.

dirkmueller:
personally I think an explicit "set -e" is better as the shebang only gets respected iirc when you call it directly, when you do bash xxx it doesn't reflect it then

jgrassler:
People probably won't call it that way but better to be safe than sorry, yeah :-)

Fixed.

PR #1452, opened 2017-11-23
barbican: Removes unreferenced monitor interval

scottwulf:
Removes unreferenced pacemaker monitor interval from the barbican
cookbook
Follow up to the code removal in 722dfb098aed3fa503b7f41f07627b90a5914d4a


dirkmueller:
probably the backport to SOC7 is a little bit excessive, this is purely cosmetic?

scottwulf:
Ok. I will agree it is excessive.

PR #1453, opened 2017-11-23
neutron: Add option to force metadata-proxy

skazi0:
In some cases dhcp agent disables metadata-proxy service for networks which don't have proper routed access to metadata service.
Exposed option enables users to force enabling metadata-proxy for all networks.


skazi0:
[neutron-metadata-proxy not restarted for some networks](https://trello.com/c/MGgCiHrt/212-neutron-metadata-proxy-not-restarted-for-some-networks)

vuntz:
I understood that doing this was more like a workaround, than the right fix. Do you think it is actually the right fix?

skazi0:
@vuntz IMO this is just a workaround. dhcp-agent correctly detected those networks as "non-isolated" so there must be something wrong with the routing or there is some other neutron bug which causes the computes to not have connectivity to metadata service.

vuntz:
> EDIT: as this workaround can be done by hand using a config snippet, I'm not sure if it makes sense to expose this parameter in the barclamp.

Hmm, I guess it doesn't. Let's close for now and reopen if we feel it really makes sense.

PR #1459, opened 2017-11-24
nova: Fix issues related to the use of skip_unchanged_nodes

vuntz:
**nova: Fix apply with skip_unchanged_nodes**
When there's absolutely no node impacted by the "apply" thanks to
skip_unchanged_nodes, then we hit some old micro-optimization in
apply_role_pre_chef_call that was not intended to be used that way: we
skip the whole method.

Unfortunately, we do need to run the method even with
skip_unchanged_nodes because we need to ensure that attributes, like the
HA related ones, are set.

**nova: Do not reset sync marks uselessly with skip_unchanged_nodes**
If skip_unchanged_nodes results in some control nodes to be skipped,
then we don't want to reset the sync marks as we will not run chef
through the apply. If we do the reset, then we will rely on periodic
chef runs to set the sync marks, which will take time to converge as
there's no guarantee that the periodic chef runs on the control nodes
are happening at the same time.

### DIFF
--- a/crowbar_framework/app/models/nova_service.rb
+++ b/crowbar_framework/app/models/nova_service.rb
@@ -269,7 +269,6 @@ class NovaService < OpenstackServiceObject
 
   def apply_role_pre_chef_call(old_role, role, all_nodes)
     @logger.debug("Nova apply_role_pre_chef_call: entering #{all_nodes.inspect}")
-    return if all_nodes.empty?
 
     unless hyperv_available?
       role.override_attributes["nova"]["elements"]["nova-compute-hyperv"] = []
--- a/crowbar_framework/app/models/nova_service.rb
+++ b/crowbar_framework/app/models/nova_service.rb
@@ -275,7 +275,12 @@ class NovaService < OpenstackServiceObject
     end
 
     controller_elements, controller_nodes, ha_enabled = role_expand_elements(role, "nova-controller")
-    reset_sync_marks_on_clusters_founders(controller_elements)
+    # Only reset sync marks if we are really applying on all controller nodes;
+    # if we are not, then we clearly do not intend to have some sync between
+    # them during the chef run
+    if Set.new(controller_nodes & all_nodes) == Set.new(controller_nodes)
+      reset_sync_marks_on_clusters_founders(controller_elements)
+    end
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/nova_service.rb
```diff
@@ -275,7 +275,12 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
     end
 
     controller_elements, controller_nodes, ha_enabled = role_expand_elements(role, "nova-controller")
-    reset_sync_marks_on_clusters_founders(controller_elements)
+    # Only reset sync marks if we are really applying on all controller nodes;
+    # if we are not, then we clearly do not intend to have some sync between
+    # them during the chef run
+    if Set.new(controller_nodes & all_nodes) == Set.new(controller_nodes)
```
abel-navarro: what happens if one of the controller nodes is down? we won't reset the sync marks in any controller then.

vuntz: A control node being down actually leads to the proposal to be queued in all cases (this would be handled by skip_unready_nodes, not skip_unchanged_nodes; and there we explicitly do not skip nodes for roles like nova-controller).

crowbar_framework/app/models/nova_service.rb
```diff
@@ -269,14 +269,18 @@ def set_ha_compute(node, enabled)
 
   def apply_role_pre_chef_call(old_role, role, all_nodes)
     @logger.debug("Nova apply_role_pre_chef_call: entering #{all_nodes.inspect}")
-    return if all_nodes.empty?
```
skazi0: so there's no "early return" possible anymore? neither here nor somewhere further down the method?

crowbar_framework/app/models/nova_service.rb
ilausuch: Right!

crowbar_framework/app/models/nova_service.rb
vuntz: Not really; there's also the check for ha compute later on. It could be moved and we could re-introduce this, but I fear this would be fragile and someone could introduce something that'd break it again.

PR #1463, opened 2017-11-24
rabbitmq: Improvements for clustered HA

vuntz:
First commit ensures that we monitor both the master and the slaves.

Second commit adjusts the resource configuration to match what is documented upstream.


Itxaka:
Fails with: 
```
undefined method `key?' for [{\"interval\"=\u003e\"30s\"}, {\"interval\"=\u003e\"27s\", \"role\"=\u003e\"Master\"}]:Array
```

Should `default[:rabbitmq][:ha][:clustered_op][:monitor]` be an array or a hash? Looks like a hash from the error

vuntz:
@Itxaka this fails because https://github.com/crowbar/crowbar-ha/pull/282 is not merged yet; I re-triggered with the crowbar-ha patch.

PR #1464, opened 2017-11-26
postgresql/rabbitmq: Fix drbd setup to monitor master and slaves

vuntz:
As explained in [1][2], we need to set a different monitor interval for the
slaves and the master, otherwise one of them won't be monitored.

[1] https://clusterlabs.org/doc/en-US/Pacemaker/1.1/html/Pacemaker_Explained/_monitoring_multi_state_resources.html
[2] https://www.linbit.com/en/drbd-monitor-intervals/

Requires https://github.com/crowbar/crowbar-ha/pull/282


jsuchome:
Didn't we want to get rid of DRBD ?

vuntz:
> Didn't we want to get rid of DRBD ?

Yes. But the plan is to backport this to stable/4.0 where it is needed :-)

PR #1466, opened 2017-11-28
neutron: Move ml2_conf.ini to config snippet + remove plugin.ini

skazi0:
With ml2_conf.ini explicitly passed to neutron-server (via plugin.ini
symlink) it was impossible to change some ml2 settings using config snippets.
Same file is used in other commands e.g. neutron-ovs-cleanup.
With this change, the original file is empty and the contents is moved
to /etc/neutron/neutron.conf.d/ for easier overriding.

Second commit includes similar change for nsx.ini and cleanup of plugin.ini logic.

### DIFF
--- a/chef/cookbooks/neutron/recipes/server.rb
+++ b/chef/cookbooks/neutron/recipes/server.rb
@@ -50,17 +50,32 @@ end
 
 include_recipe "neutron::common_config"
 
-# set core plugin for neutron-server
+# remove unused plugin config snippets
+all_plugin_snippets = [node[:neutron][:ml2_config_file], node[:neutron][:nsx_config_file]]
+used_plugin_snippets = []
 if node[:neutron][:networking_plugin] == "vmware"
-  core_link = "/etc/neutron/plugins/vmware/nsx.ini"
+  used_plugin_snippets << node[:neutron][:nsx_config_file]
 else
-  core_link = "/etc/neutron/plugins/ml2/ml2_conf.ini"
+  used_plugin_snippets << node[:neutron][:ml2_config_file]
 end
 
-link "/etc/neutron/plugin.ini" do
-  to core_link
-  action :create
-  notifies :restart, "service[#{node[:neutron][:platform][:service_name]}]"
+(all_plugin_snippets - used_plugin_snippets).each do |config_file|
+  file config_file do
+    action :delete
+    notifies :restart, "service[#{node[:neutron][:platform][:service_name]}]"
+  end
+end
+
+# Empty the config file that is explicitly passed to neutron-server.
+# This allows overriding of plugin settings using config snippets.
+# NOTE: if plugin.ini is a symlink it will not replace it with regular file
+#       but this is OK since all possible target files will be empty too.
+file "/etc/neutron/plugin.ini" do
+  owner "root"
+  group node[:neutron][:platform][:group]
+  mode "0640"
+  content "# Please use config file snippets in /etc/neutron/neutron.conf.d/.\n" \
+          "# See /etc/neutron/README.config for more details.\n"
 end
 
 # enable/disable ml2_conf_cisco for neutron-server
@@ -88,16 +103,6 @@ link "/etc/neutron/neutron-server.conf.d/100-ml2_conf_cisco_apic.ini.conf" do
   to "/etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini"
   action cisco_apic_link_action
   notifies :restart, "service[#{node[:neutron][:platform][:service_name]}]"
-end
-
-template "/etc/default/neutron-server" do
-  source "neutron-server.erb"
-  owner "root"
-  group node[:neutron][:platform][:group]
-  variables(
-    neutron_plugin_config: node[:neutron][:ml2_config_file]
-  )
-  only_if { node[:platform_family] == "debian" }
 end
 
 directory "/var/cache/neutron" do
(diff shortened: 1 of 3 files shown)


skazi0:
[No way to override ml2.ini settings with snippets](https://trello.com/c/ZXbwRBx0/215-no-way-to-override-ml2ini-settings-with-snippets)

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -347,7 +357,7 @@
       user node[:neutron][:user]
       group node[:neutron][:group]
       command "apic-ml2-db-manage --config-dir /etc/neutron/neutron.conf.d \
-                                  --config-file /etc/neutron/plugins/ml2/ml2_conf.ini \
+                                  --config-file #{node[:neutron][:ml2_config_file]} \
```
vuntz: Can't we skip that explicit `--config-file` now?

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -95,8 +95,8 @@
   owner "root"
   group node[:neutron][:platform][:group]
   variables(
-      neutron_plugin_config: "/etc/neutron/plugins/ml2/ml2_conf.ini"
-    )
+    neutron_plugin_config: node[:neutron][:ml2_config_file]
+  )
```
vuntz: Shouldn't we change this to not mention any file, since it's in the config dir now?

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -369,7 +379,7 @@
       user node[:neutron][:user]
       group node[:neutron][:group]
       command "gbp-db-manage --config-dir /etc/neutron/neutron.conf.d \
-                             --config-file /etc/neutron/plugins/ml2/ml2_conf.ini \
+                             --config-file #{node[:neutron][:ml2_config_file]} \
```
vuntz: Can't we skip that explicit `--config-file` now?

chef/cookbooks/neutron/recipes/server.rb
skazi0: yeah, the debian config file will obsolete if we move nsx config to snippets.

chef/cookbooks/neutron/recipes/server.rb
skazi0: done.

chef/cookbooks/neutron/recipes/server.rb
skazi0: done.

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -347,7 +366,6 @@
       user node[:neutron][:user]
       group node[:neutron][:group]
       command "apic-ml2-db-manage --config-dir /etc/neutron/neutron.conf.d \
-                                  --config-file /etc/neutron/plugins/ml2/ml2_conf.ini \
                                   --config-file /etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini \
```
rhafer: Wouldn't this be a good chance to get rid of `ml2_conf_cisco_apic.ini` as well? AFAICS we're creating it as `etc/neutron/neutron-server.conf.d/100-ml2_conf_cisco_apic.ini.conf` already anyway. See: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/neutron/recipes/cisco_apic_support.rb#L24

@mmnelemane: Is  /etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini  really still used somewhere?

skazi0: this file is still passed to `apic-ml2-db-manage` and `gbp-db-manage` so we would need to change it there or move it to `neutron.conf.d`.

rhafer: Oh I overlooked that cisco_apic_support.rb creates the file in `/etc/neutron/neutron-server.conf.d/` instead of `neutron.conf.d`. I somehow assume thinks would still work if we'd move it to  `neutron.conf.d` but I don't really new. @mmnelemane do you?

skazi0: The "README.config" file says:

$program reads the configuration files in the following order:
- /etc/$project/$project.conf
- /etc/$project/$project.conf.d/*.conf (lexically sorted)
- /etc/$project/$program.conf.d/*.conf (lexically sorted)

so it should be safe to move anything from `neutron-server.conf.d` to `neutron.conf.d` as long as there are no overrides further down the search path. One (desired?) side effect is that it will widen the scope of those settings (i.e. programs other than `neutron-server` will see them too).

chef/cookbooks/neutron/templates/default/neutron-server.erb
```diff
@@ -1 +0,0 @@
-NEUTRON_PLUGIN_CONFIG=<%= @neutron_plugin_config %>
```
vuntz: Actually, we may need to keep creating this file, but with an empty variable?

vuntz: And just to clarify: the problem is that the file is used on Debian, and if we don't make sure the content is correct, things could go wrong.

skazi0: I checked Debian packages and they don't take this variable from `/etc/default/neutron-server` at all. Ubuntu will take it if set but will not complain if it's missing. Both are fine with the file not being present at all.

Ubuntu snippet:

```
  [ -r /etc/default/neutron-server ] && . /etc/default/neutron-server
  [ -r "$NEUTRON_PLUGIN_CONFIG" ] && DAEMON_ARGS="$DAEMON_ARGS --config-file=$NEUTRON_PLUGIN_CONFIG"
```

We can leave it in but I'm pretty sure removing it will not break anything.

vuntz: Well, the point is still that if the file exists (written by the user and forgotten), then things can break. That's the same reason why we create the dummy files for other config files :-)

But I won't argue forever on this, so +1'ing

vuntz:
I think we still need to clean the cisco bits, but let's do a followup PR for this.

PR #1468, opened 2017-11-29
rabbitmq: allow to add extra users to rabbit

Itxaka:
with the introduction of the definitions file it looks like
rabbit will disregard any existing users when restarting,
using the defintions file as the blank slate to start from.

This gets in the middle of having any extra users for other
tasks like monitoring and there is no mechanism to add users.

This PR adds an extra field to crowbar to add a list of users
that will get written into the definitions file, allowing
us to specify extra users to be created

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/rabbit.rb
+++ b/chef/cookbooks/rabbitmq/recipes/rabbit.rb
@@ -134,6 +134,34 @@ execute "rabbitmqctl set_user_tags #{node[:rabbitmq][:user]} management" do
   only_if only_if_command if ha_enabled
 end
 
+node[:rabbitmq][:users].each do |user, pass|
+  # create extra users
+  rabbitmq_user "adding user #{user}" do
+    user user
+    password pass
+    address node[:rabbitmq][:management_address]
+    port node[:rabbitmq][:management_port]
+    action :add
+    only_if only_if_command if ha_enabled
+  end
+
+  # add permisions to those extra users into the default vhost
+  rabbitmq_user "setting permissions for #{user}" do
+    user user
+    vhost node[:rabbitmq][:vhost]
+    permissions "\".*\" \".*\" \".*\""
+    action :set_permissions
+    only_if only_if_command if ha_enabled
+  end
+
+  # tag those users as management
+  execute "rabbitmqctl set_user_tags #{user} management" do
+    not_if "rabbitmqctl list_users | grep #{user} | grep -q management"
+    action :run
+    only_if only_if_command if ha_enabled
+  end
+end
+
 if cluster_enabled
   set_policy_command = "rabbitmqctl set_policy -p #{node[:rabbitmq][:vhost]} " \
       " ha-all '^(?!amq\.).*' '{\"ha-mode\": \"all\"}'"
--- a/chef/cookbooks/rabbitmq/templates/default/definitions.json.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/definitions.json.erb
@@ -34,6 +34,13 @@
             "tags": ""
         },
 <% end -%>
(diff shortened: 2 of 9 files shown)


crowbar_framework/app/models/rabbitmq_service.rb
```diff
@@ -71,9 +71,32 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+    old_attrs = old_role.nil? ? nil : old_role.default_attributes[@bc_name]
+    role.default_attributes[@bc_name]["users"] ||= {}
+    role.default_attributes[@bc_name]["extra_users"].each do |user|
+      save_role = true
+      if !old_attrs.nil? && old_attrs.include?("users") && old_attrs["users"].include?(user)
+        # reuse the existing pass
+        role.default_attributes[@bc_name]["users"][user] = old_attrs["users"][user]
+      else
+        # new user, so create a random pass
+        role.default_attributes[@bc_name]["users"][user] = random_password
```
vuntz: @tpatzig I guess you wouldn't want a random password, would you?

chef/cookbooks/rabbitmq/templates/default/definitions.json.erb
```diff
@@ -33,6 +33,13 @@
             "password": <%= @json_trove_password %>,
             "tags": ""
         },
+<% end -%>
+<% @extra_users.each do |user, pass| -%>
+        {
+            "name": "<%= user %>",
+            "password": "<%= pass %>",
+            "tags": ""
```
vuntz: This is not consistent with the commands that are run (where you do `rabbitmqctl set_user_tags #{user} management`). That being said, I'm not sure we want to tag the extra users as management.

chef/cookbooks/rabbitmq/recipes/rabbit.rb
```diff
@@ -134,6 +134,34 @@
...
+    port node[:rabbitmq][:management_port]
+    action :add
+    only_if only_if_command if ha_enabled
+  end
+
+  # add permisions to those extra users into the default vhost
+  rabbitmq_user "setting permissions for #{user}" do
+    user user
+    vhost node[:rabbitmq][:vhost]
+    permissions "\".*\" \".*\" \".*\""
```
vuntz: I'm also unclear if we want to give all permissions to the vhost to the extra users... Maybe it's something that should be configurable in the barclamp?

crowbar_framework/app/models/rabbitmq_service.rb
tpatzig: random password is fine for us. As long as it is only generated once on the initial apply run 😄

chef/cookbooks/rabbitmq/templates/default/definitions.json.erb
tpatzig: The tag should be definable. E.g. for the zabbix user we need the tag 'monitoring'.

chef/cookbooks/rabbitmq/recipes/rabbit.rb
tpatzig: Yes, permissions should be definable as well. Although we need full perms for our zabbix user.

Itxaka:
@tpatzig @vuntz looks like this needs more work. My first impression was to do a quickpatch to create extra users, and then I saw that indeed we will need to be able to set at least permissions + tags for each other but was not sure if that was needed. Seeing your comments I see that they are needed so I will rework this :)

Itxaka:
[RabbitMQ add extra users](https://trello.com/c/ELZAtk53/234-rabbitmq-add-extra-users)

Itxaka:
work has been moved to https://github.com/crowbar/crowbar-openstack/pull/1506

PR #1469, opened 2017-11-29
Keystone LDAP speedup

dirkmueller:


dirkmueller:
@aplanas no other service is using that provider iirc

chef/cookbooks/crowbar-openstack/templates/default/vhost-wsgi.conf.erb
```diff
@@ -4,7 +4,6 @@ Listen <%= @bind_host %>:<%= @bind_port %>
     WSGIDaemonProcess <%= @daemon_process %> user=<%= @user %> group=<%= @group %> processes=<%= @processes %> threads=<%= @threads %> display-name=%{GROUP}
     WSGIProcessGroup <%= @process_group %>
     WSGIScriptAlias / <%= @script_alias %>
-    WSGIApplicationGroup %{GLOBAL}
```
cmurphy: How was this affecting performance?

cmurphy:
> @aplanas no other service is using that provider iirc

It is widely used https://github.com/crowbar/crowbar-openstack/search?utf8=%E2%9C%93&q=crowbar_openstack_wsgi&type=

aplanas:
the template is used here: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/crowbar-openstack/providers/wsgi.rb#L47

and is used in several services:
https://github.com/crowbar/crowbar-openstack/search?utf8=%E2%9C%93&q=WSGI+entry+for&type=

chef/cookbooks/crowbar-openstack/templates/default/vhost-wsgi.conf.erb
dirkmueller: it seems this change was incorrect. further documentation reading seems to confirm that running inside the global applciation group and only assigning one wsgi  application per processgroup is better (which is what we were doing before, so I'm dropping this commit)

dirkmueller:
@cmurphy @aplanas ok, further testing figured out that this is an issue with python-cryptography that caused keystone to hang and reload. we need to update to a newer version, and the change in wsgiprocessgroup is not needed.

PR #1470, opened 2017-11-29
galera: raise haproxy limit depending on max_connections (bsc#1068962)

dirkmueller:
Without this, haproxy is always limiting us to 1000 connections, not
exposing the full capabilities if db connections is set to 1000. Also
if db connections is set  < 1000, this allows the openstack services
to exceed the number of db connections, causing db healthcheck to fail
and pacemaker to go nuts. always reserve some spare connections
to avoid this from happening.

Requires https://github.com/crowbar/crowbar-ha/pull/279 to be merged first.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -334,6 +334,8 @@ haproxy_loadbalancer "galera" do
   address CrowbarPacemakerHelper.cluster_vip(node, "admin")
   port 3306
   mode "tcp"
+  # leave some room for pacemaker health checks
+  max_connections node[:database][:mysql][:max_connections] - 10
   options ["httpchk"]
   default_server "port 5555"
   stick ({ "on" => "dst" })


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -334,6 +334,8 @@
   address CrowbarPacemakerHelper.cluster_vip(node, "admin")
   port 3306
   mode "tcp"
+  # leave some room for pacemaker health checks
+  max_connections node[:database][:mysql][:max_connections] - 10
```
s-t-e-v-e-n-k: Why - 10 ? Surely it should be + 10 if you want to leave some headroom?

rhafer: Actually -10 is correct here. The mariadb instance should have some headroom for connection not coming via haproxy (e.g. the monitoring stuff from pacemaker).  And this line is setting the limit for the haproxy config.

vuntz: Do we need to somehow use a `min` to also limit on the haproxy max_connections?

rhafer: @vuntz You mean haproxy's global limit? Hm, according to https://github.com/crowbar/crowbar-ha/pull/279#issuecomment-350127073 haproy would take of that by itself.

Though some kind of a warning/error (in the Crowbar UI?) might be friendly, in case `node[:database][:mysql][:max_connections]` exceeds haproxy's global limit. Personally I am fine to go without that.

vuntz: My point is that the `- 10` won't be effective in that case and we would just eat all connections, while we still want the `- 10`

rhafer: @vuntz I guess I am still missing your point. If `node[:database][:mysql][:max_connections]-10` exceeds haproxy's max_connections. haproxy will not be able to use up all mysql connections.

vuntz: Heh, sorry. It's the opposite: if mysql_max_connections is higher than the general haproxy_maxconnections (so not the setting we use here, but the global one), then this frontend in haproxy can eat all connections and leave no connection for other frontend in haproxy.

But I realize it can be argued it's not a new issue... Okay, let's ignore that for now.

rhafer: At least that case would not cause pacemaker to trigger a restart festival on mariadb.

PR #1471, opened 2017-12-01
rabbitmq: remove slash from default vhost

Itxaka:
For some reason we set the default vhost for rabbit to
contain a slash. Its not that its worng but it makes no sense
to include a slash in the name and it makes it seem like the
slash is required for a vhost, but its not. Plus we end up with
rabbit urls with 2 slashes at the end wich further confuses
the proper way of writing them (i.e. guest:guest@IP:5672//nova)

So this sets the normal name minus slash to the nova+trove vhosts

### DIFF
--- a/chef/data_bags/crowbar/template-rabbitmq.json
+++ b/chef/data_bags/crowbar/template-rabbitmq.json
@@ -7,7 +7,7 @@
       "port": 5672,
       "password": "",
       "user": "nova",
-      "vhost": "/nova",
+      "vhost": "nova",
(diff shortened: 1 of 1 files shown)


vuntz:
I would argue that if you change this, then you should also rename the vhost (and the user) to `openstack`.

Itxaka:
@vuntz Im just fixing something in this PR that is wrong (see @skazi0 comment). Changing the names to different ones (which I approve!) should go into a different PR I think.

PR #1472, opened 2017-12-01
mariadb: Fixes for mariadb 10.2

rhafer:
Please don't not merge yet. As this will break 10.1 deployments.


rhafer:
I supposed we could add another attribute "mariadb-version" (set to 10.1 for now) and set the package list depending on that, to make this mergable before the final decision about 10.2 is done. (Similar what is done in the attributes file of postgresql cookbook)

rhafer:
This should be mergeable now. If we decide to go with 10.2 we just need to update the version number in the attributes.

PR #1474, opened 2017-12-05
keystone: Sync fernet keys when joining cluster

skazi0:
Newly (re)installed node will not have fernet keys causing failure
when starting keystone services. Cluster founder rotates the keys
and pushes new ones to cluster member on a regular basis.

This change creates pacemaker alert to notify founder about new node.
Keys are then pushed from founder to the node.

Preconditions:
~~https://github.com/crowbar/crowbar-core/pull/1451~~
~~https://github.com/crowbar/crowbar-ha/pull/285~~
~~https://github.com/crowbar/crowbar-ha/pull/284~~
~~https://github.com/crowbar/crowbar-ha/pull/287~~

### DIFF
--- a/chef/cookbooks/keystone/recipes/ha.rb
+++ b/chef/cookbooks/keystone/recipes/ha.rb
@@ -71,3 +71,26 @@ if node[:keystone][:frontend] == "apache" && node[:pacemaker][:clone_stateless_s
 
   crowbar_pacemaker_sync_mark "create-keystone_ha_resources"
 end
+
+template "/usr/bin/keystone-fernet-keys-sync.sh" do
+  source "keystone-fernet-keys-sync.sh"
+  owner "root"
+  group "root"
+  mode "0755"
+end
+
+# handler scripts are run by hacluster user so sudo configuration is needed
+# if the handler needs to rsync to other nodes using root's keys
+template "/etc/sudoers.d/keystone-fernet-keys-sync" do
+  source "hacluster_sudoers.erb"
+  owner "root"
+  group "root"
+  mode "0440"
+end
+
+# on founder: create/delete pacemaker alert
+pacemaker_alert "keystone-fernet-keys-sync" do
+  handler "/usr/bin/keystone-fernet-keys-sync.sh"
+  action node[:keystone][:token_format] == "fernet" ? :create : :delete
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
--- a/chef/cookbooks/keystone/templates/default/hacluster_sudoers.erb
+++ b/chef/cookbooks/keystone/templates/default/hacluster_sudoers.erb
@@ -1,0 +1,3 @@
+Defaults:hacluster !requiretty
+
+hacluster ALL = (root) NOPASSWD: /usr/bin/keystone-fernet-keys-push.sh
--- a/chef/cookbooks/keystone/templates/default/keystone-fernet-keys-push.sh
+++ b/chef/cookbooks/keystone/templates/default/keystone-fernet-keys-push.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+TARGETNODE=$1
+shift
+RSYNC_ARGS=$@
+test -z "$TARGETNODE" && { echo "usage: $0 <node-address> [extra rsync args]"; exit 1; }
+rsync -a --timeout=300 --delete-after $RSYNC_ARGS /etc/keystone/fernet-keys $TARGETNODE:/etc/keystone/
--- a/chef/cookbooks/keystone/templates/default/keystone-fernet-keys-sync.sh
+++ b/chef/cookbooks/keystone/templates/default/keystone-fernet-keys-sync.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+if [ -z "$CRM_alert_version" ]; then
+    echo "$0 must be run by Pacemaker version 1.1.15 or later"
+    exit 0
+fi
+
+# skip if alert is not triggered by node joining event
+[ "$CRM_alert_kind" = "node" -a "$CRM_alert_desc" = "member" ] || exit 0
+
+myname=$(uname -n)
+
+# skip if triggered on the re-joining node itself
+[ "$CRM_alert_node" = "$myname" ] && exit 0
+
+# sleep some random time (0-10s) to avoid all nodes hitting the new one at the same time
+sleep $(( $RANDOM % 11 ))
+
+# push keys to the joining node
+sudo /usr/bin/keystone-fernet-keys-push.sh $CRM_alert_node --ignore-existing
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -384,15 +384,21 @@ if node[:keystone][:token_format] == "fernet"
   # To be sure that rsync package is installed
   package "rsync"
   crowbar_pacemaker_sync_mark "sync-keystone_install_rsync" if ha_enabled
+
+  template "/usr/bin/keystone-fernet-keys-push.sh" do
+    source "keystone-fernet-keys-push.sh"
+    owner "root"
+    group "root"
+    mode "0755"
+  end
+
   rsync_command = ""
   if ha_enabled
     cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node)
     cluster_nodes.map do |n|
       next if node.name == n.name
       node_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address
-      rsync_command += \
-        "rsync -a --timeout=300 --delete-after /etc/keystone/fernet-keys " \
-        "#{node_address}:/etc/keystone/; "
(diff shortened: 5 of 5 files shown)


skazi0:

chef/cookbooks/keystone/recipes/ha.rb
```diff
@@ -71,3 +71,27 @@
...
+  source "hacluster_sudoers.erb"
+  owner "root"
+  group "root"
+  mode "0440"
+end
+
+# on founder: create pacemaker alert
+pacemaker_alert "keystone-fernet-keys-sync" do
+  handler "/var/lib/pacemaker/keystone-fernet-keys-sync.sh"
+  receiver "REMOVE WHEN PR#287 FROM CROWBAR-HA IS MERGED"
```
vuntz: Bah :-)

chef/cookbooks/keystone/recipes/ha.rb
```diff
@@ -71,3 +71,27 @@
 
   crowbar_pacemaker_sync_mark "create-keystone_ha_resources"
 end
+
+template "/var/lib/pacemaker/keystone-fernet-keys-sync.sh" do
```
vuntz: Note sure if `/var/lib/pacemaker` is the best location; wouldn't `/usr/lib/pacemaker` be better? Or what about `/var/lib/keystone` (may be risky, in case it becomes not world-readable)?

chef/cookbooks/keystone/templates/default/hacluster_sudoers.erb
```diff
@@ -0,0 +1,3 @@
+Defaults:hacluster !requiretty
+
+hacluster ALL = (root) NOPASSWD: /usr/bin/rsync
```
vuntz: I somehow am not 100% comfortable with allowing rsync with no further restrictions (like "which directory it can write to"), but I don't know if we can restrict further easily...

chef/cookbooks/keystone/recipes/ha.rb
skazi0: I took it from some examples but I agree that `/var` is not a good place for these.

chef/cookbooks/keystone/templates/default/hacluster_sudoers.erb
skazi0: We can create a script (e.g. `push-fernet-keys`) and give sudo access only to that. We can also re-use it in the key rotation script to have it all in one place.

skazi0: @vuntz please take a look at the second commit. It should fix the problem of too-wide-sudo.

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-push.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+TARGETNODE=$1
+test -z $TARGETNODE && exit 1
+rsync -a --timeout=300 --delete-after --ignore-existing /etc/keystone/fernet-keys $TARGETNODE:/etc/keystone/
```
vuntz: I don't think we want `--ignore-existing` in the usual sync; I would only use it for the alert mechanism.

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-push.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+TARGETNODE=$1
+test -z $TARGETNODE && exit 1
```
vuntz: Can you put quotes around `$TARGETNODE`?

skazi0: done

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-push.sh
skazi0: done

skazi0:
@vuntz in current form it will not work because on reinstallation founder still tries to do "propagate fernet keys to all nodes in the cluster"...

vuntz:
> @vuntz in current form it will not work because on reinstallation founder still tries to do "propagate fernet keys to all nodes in the cluster"...

Would it make sense to change `unless File.exist?("/etc/keystone/fernet-keys/0")` to something like:

```ruby
unless File.exist?("/etc/keystone/fernet-keys/0") &&
    (!ha_enabled ||
     (CrowbarPacemakerHelper.is_cluster_founder?(node) &&
      pseudo-code(keystone-fernet-rotate resource doesn't exist in pacemaker)))
```

skazi0:
squashed and rebased

chef/cookbooks/keystone/recipes/ha.rb
```diff
@@ -71,3 +71,26 @@
 
   crowbar_pacemaker_sync_mark "create-keystone_ha_resources"
 end
+
```
vuntz: Was wondering if we need some sync mark for this... What do you think?

Also, we probably should only do this if we use fernet; so there's a missing check here.

skazi0: I don't think we need syncmarks. The setup will be done only on founder when keystone is installed (cluster should already be working at this point). The first time when this will be used is when some of the members is reinstalled. I don't expect the time between founder setup and first reinstallation to be "close to zero" which would require synchronizing this.

The "fernet" check sounds like a good idea. In non-fernet case `keystone-fernet-keys-push.sh` would not be even installed.
Do we support changing the `token_format`? If so, then maybe we should delete the alert if current format is not "fernet"?

skazi0: Would this: `action node[:keystone][:token_format] == "fernet" ? :create : :delete` be enough?

skazi0: The scripts and sudoers config will still be installed but will not be used.

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-sync.sh
```diff
@@ -0,0 +1,19 @@
...
+# skip if alert is not triggered by node joining event
+[ "$CRM_alert_kind" = "node" -a "$CRM_alert_desc" = "member" ] || exit 0
+
+myname=$(uname -n)
+
+# skip if triggered on the re-joining node itself
+[ "$CRM_alert_node" = "$myname" ] && exit 0
+
+# sleep some random time (0-10s) to avoid all nodes hitting the new one at the same time
+sleep $(( $RANDOM % 11 ))
```
dirkmueller: so I don't understand this part.. why isn't the one owning the keystone-fernet-rotate service pushing it to the new node? why do we need all?

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-sync.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# push keys to the joining node
+sudo /usr/bin/keystone-fernet-keys-push.sh $CRM_alert_node --ignore-existing
```
dirkmueller: why the sudo?

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-sync.sh
skazi0: This is to cover cases where the node which normally runs keystone-fernet-rotate is the one being reinstalled. There was a version of this PR which did exactly as you described: only the founder pushed the keys to newly installed nodes. Then we tried to cover the founder reinstallation case and got to a point where we would need to "elect" some node which will push the keys to reinstalled founder. It turned out to be easier to just push from everywhere and forget about founder being special in this context.

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-sync.sh
skazi0: Alarm handlers are run as hacluster user. This is described when the relevant sudoers configuration is deployed.

dirkmueller: thanks. new question: why the ignore-existing? wouldn't that cause fernet-tokens/0 to not be pushed if it is already there?

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-sync.sh
dirkmueller: I see the problem in the original implementation. I need to think about this some more.

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-sync.sh
skazi0: This is supposed to further avoid problems with multiple nodes hitting the new one at the same time. As for 0-key being there already, this should not be the case for reinstalled node unless the key got there from some of the peer nodes. In such case, 0-key should already be the same as in the rest of cluster nodes. Note that key rotation mechanism doesn't have the `--ignore-existing` argument.

skazi0:
For the record: PR #1512 includes changes suggested by @vuntz

PR #1475, opened 2017-12-08
neutron: Put ML2 Cisco config files as snippets in neutron.conf.d

vuntz:
This enables people to override values in these files with snippets.

Followup to https://github.com/crowbar/crowbar-openstack/pull/1466

### DIFF
--- a/chef/cookbooks/neutron/recipes/cisco_apic_support.rb
+++ b/chef/cookbooks/neutron/recipes/cisco_apic_support.rb
@@ -20,8 +20,29 @@ if node[:neutron][:ml2_mechanism_drivers].include?("apic_gbp")
   node[:neutron][:platform][:cisco_apic_gbp_pkgs].each { |p| package p }
 end
 
+# Empty the config file to avoid confusion (it may be a symlink, due to some
+# old code we had)
+old_config = "/etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini"
+link old_config do
+  action :delete
+  only_if "test -L #{old_config}"
+end
+file old_config do
+  owner "root"
+  group node[:neutron][:platform][:group]
+  mode "0640"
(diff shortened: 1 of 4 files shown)


chef/cookbooks/neutron/recipes/cisco_apic_support.rb
```diff
@@ -20,8 +20,22 @@
   node[:neutron][:platform][:cisco_apic_gbp_pkgs].each { |p| package p }
 end
 
+# Empty the config file to avoid confusion
+file "/etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini" do
```
skazi0: same as with ml2_conf_cisco.ini

chef/cookbooks/neutron/recipes/cisco_support.rb
```diff
@@ -13,7 +13,20 @@
   switches = neutron[:neutron][:cisco_switches].to_hash
 end
 
-template "/etc/neutron/plugins/ml2/ml2_conf_cisco.ini" do
+# Empty the config file to avoid confusion
+file "/etc/neutron/plugins/ml2/ml2_conf_cisco.ini" do
```
skazi0: This is/can be a symlink to a file in neutron-server.conf.d which you delete later. The content from here will be lost and there will be a dangling symlink.

vuntz: I actually wonder how `file` deals with symlinks... Let me play with that.

vuntz: @skazi0 indeed, not as I expected. Fixed.

skazi0: @vuntz yeah, I had the same problem in my PR around this. Newer chef versions have `force_unlink` option for exactly this.

PR #1477, opened 2017-12-08
rabbitmq: Use pause_minority setting to deal with partitions better

vuntz:
We do care about data integrity, and the default setting is "ignore"
which can result in inconsistencies when the partition is over.

For two-node clusters, it could be argued that this wouldn't work nicely
as this more or less means that as soon as there's a partition, rabbitmq
stop working. But pacemaker should help here (with the fencing) and it's
really better to fail than to have inconsistencies later on.

See https://www.rabbitmq.com/partitions.html for more details

### DIFF
--- a/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
@@ -20,6 +20,9 @@
                  {certfile, "<%= node[:rabbitmq][:ssl][:certfile] %>"},
                  {keyfile, "<%= node[:rabbitmq][:ssl][:keyfile] %>"}]},
 <% end -%>
+<% if @cluster_enabled -%>
+   {cluster_partition_handling, <%= @cluster_partition_handling %>},
+<% end -%>
    {disk_free_limit, 50000000}
  ]},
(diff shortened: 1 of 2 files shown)


chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
```diff
@@ -19,6 +19,9 @@
    <% end -%>
                  {certfile, "<%= node[:rabbitmq][:ssl][:certfile] %>"},
                  {keyfile, "<%= node[:rabbitmq][:ssl][:keyfile] %>"}]},
+<% end -%>
+<% if @cluster_enabled -%>
```
Itxaka: Should we spend an extra search here to make sure that we have >2 rabbit nodes before enabling this?

vuntz: As discussed, you're right: we don't want `pause_minority` in that context as this prevents one node of the two-node cluster from going down (for maintenance, for instance).

Itxaka:
retriggered

Itxaka:
retriggered with `want_aodh_proposal=0` as aodh packages seem to be broken right now

PR #1478, opened 2017-12-11
nova: Add resource limits for api and compute (bsc#1057086)

s-t-e-v-e-n-k:
Based on exising work for allowing modification of system resource
limits, extend it to allow changing the open file limits for the
nova-api and nova-compute services.

### DIFF
--- a/chef/cookbooks/nova/recipes/api.rb
+++ b/chef/cookbooks/nova/recipes/api.rb
@@ -133,3 +133,14 @@ keystone_register "register nova_legacy endpoint" do
 end
 
 crowbar_pacemaker_sync_mark "create-nova_register" if api_ha_enabled
+
+service = "openstack-nova-api"
+if node[:nova][:resource_limits] && node[:nova][:resource_limits][service]
+  limits = node[:nova][:resource_limits][service]
+  action = limits.values.any? ? :create : :delete
+  utils_systemd_override_limits "Resource limits for #{service}" do
+    service_name service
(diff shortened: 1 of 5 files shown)


chef/cookbooks/nova/recipes/api.rb
```diff
@@ -24,10 +24,6 @@
 use_crowbar_pacemaker_service = node[:nova][:ha][:enabled] &&
   node[:pacemaker][:clone_stateless_services]
 
-nova_package "api" do
-  use_pacemaker_provider use_crowbar_pacemaker_service
-end
```
vuntz: Why do you move this? It feels wrong to define the endpoint before installing and starting the service :-)

s-t-e-v-e-n-k: It was to match the cinder api recipe, which does the same thing when setting limits.

cmurphy: I don't think that's relevant to defining the limits. In cinder the cinder_service resource happened to be declared at the end and the new utils_systemd_override_limits resource just got appended to the end of the recipe. I would move this back to avoid making irrelevant changes and keep the change small.

s-t-e-v-e-n-k:
I picked -api as well as -compute since I thought it was the next likely target -- what with connections to rabbit, the scheduler, and conductor, and servicing calls from the endpoints themselves, but I'm happy to roll this back to -compute only.

PR #1479, opened 2017-12-11
nova: Call "nova-manage api_db sync" after cell0/cell1 are managed

jsuchome:
If "nova-manage api_db sync" is called first during the upgrade from Newton, it complains about cell0/cell1 not being prepared.


jsuchome:
So it seems to work with disruptive upgrade (non-disruptive / HA is broken for different reasons)

PR #1480, opened 2017-12-11
rabbit: add mnesia dump log config

Itxaka:
As shown on big deployments, we sometimes get the warning
that mnesia is overloaded. This usually occurs when we are
dumping the log too fast and mnesia is receiving too many
messages so it cant catch up.

To fix this, we should increase the number of writes to the
transaction log before a dump is done (default is 100).
Together with this we also want to configure the maximum dump
log interval, in case the write treshold is set too high, this
will trigger a dump every X milliseconds, so we can be sure
that we are not losing anything if there is a period of low
writes or the treshold is set to high (default is 3 minutes)

Increasing the write treshold and reducing a bit the dump time
should help with mnesia being overloaded and prevent any
unintended consecuences from the overload.

### DIFF
--- a/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
@@ -1,4 +1,10 @@
 [
+ {mnesia,
+  [
+   {dump_log_write_threshold, <%= node[:rabbitmq][:mnesia][:dump_log_write_threshold] %>},
+   {dump_log_time_threshold, <%= node[:rabbitmq][:mnesia][:dump_log_time_threshold] %>}
+  ]
+ },
  {rabbit,
   [
    {tcp_listeners, [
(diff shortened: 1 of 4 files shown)


Itxaka:
[RabbitMQ restarted by pacemaker from time to time](https://trello.com/c/unGNGzw0/209-rabbitmq-restarted-by-pacemaker-from-time-to-time)

skazi0:
@Itxaka this only exposes the settings, right? the defaults are left intact?

Itxaka:
@skazi0 yep! as this is really dependent in how many messages are in the MQ per second (so Im guessing it increases along with the more nodes/services you have and the number of actions taken that require mq messages) I guess it should be configured per environment, leaving the default intact for any normal deployment (in which the default dont seem to affect anything, they work quite well)

vuntz:
CI is failing, it seems.

Itxaka:
@vuntz @ilausuch @skazi0 fixed, I was missing the correct chef way to access the variables (was using `<% -%>` instead of `<%= %>`)

PR #1481, opened 2017-12-13
nova: Fix problem with log permissions

skazi0:
`nova-manage db version` in chef's condition is run as root and
touches /var/log/nova/nova-manage.log. With --log-file argument it
will not do it and save us some permission problems.


skazi0:
[nova-manage.log with wrong permissions](https://trello.com/c/5FTD6bCx/186-nova-managelog-with-wrong-permissions)

PR #1482, opened 2017-12-13
Add *.log file for apache logrotate(bsc#1071317)

sayalilunkad:
Currently only the apache access and error logs are rotated by logrotate but /var/log/apache
also contains access and error logs from keystone, ceilometer, barbican and other services which
are not rotated. This adds another file whichh rotates all the log files in the apache directory.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -181,6 +181,13 @@ elsif node[:keystone][:frontend] == "apache"
   apache_site "keystone-admin.conf" do
     enable true
   end
+
+  template "/etc/logrotate.d/apache2-all" do
+    source "apache.logrotate.erb"
+    mode 0o644
+    owner "root"
+    group "root"
+  end
 end
 
 db_settings = fetch_database_settings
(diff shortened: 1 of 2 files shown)


sayalilunkad:
[P2 Bug 1071317 - L3: log rotation broken for ceilometer-api, keystone-admin and keystone-public logs](https://trello.com/c/xO7kK6Dr/627-bug-1071317-l3-log-rotation-broken-for-ceilometer-api-keystone-admin-and-keystone-public-logs)

rsalevsky:
Is there a reason why we add this to keystone when it effects all services using apache?

sayalilunkad:
@rsalevsky the apache access and error logs are rotated by the default configuration that is generated through the apache package. On the controller node some of the openstack services which are run via apache are logging in /var/log/apache and these logs are not rotated. Since apache is started in the keystone recipe it made sense to generate the logrotate config here since it will ensure it is created only on the controller nodes and only when using apache.

vuntz:
Can't we change the apache cookbook (in crowbar-core)?

sayalilunkad:
@vuntz I started with that approach actually https://github.com/crowbar/crowbar-core/pull/1454  but I did not find a good way to identify the controller node. Do you think that's a better approach?

vuntz:
> @vuntz I started with that approach actually crowbar/crowbar-core#1454 but I did not find a good way to identify the controller node. Do you think that's a better approach?

I don't think it matters that we identify the controller node or not: if the apache cookbook is run on a node, we know we want to setup the logrotate bits there, I think.

vuntz:
Closing in favor of https://github.com/crowbar/crowbar-core/pull/1454

PR #1484, opened 2017-12-14
[4.0] nova: Fix problem with log permissions

skazi0:
`nova-manage db version` in chef's condition is run as root and
touches /var/log/nova/nova-manage.log. With --log-file argument it
will not do it and save us some permission problems.

(cherry picked from commit b373cefdc0072ffc2756f58172a81e4f0ea66b79)

backport of #1481


skazi0:
[nova-manage.log with wrong permissions](https://trello.com/c/5FTD6bCx/186-nova-managelog-with-wrong-permissions)

PR #1485, opened 2017-12-14
[4.0] rabbit: add mnesia dump log config

Itxaka:
As shown on big deployments, we sometimes get the warning
that mnesia is overloaded. This usually occurs when we are
dumping the log too fast and mnesia is receiving too many
messages so it cant catch up.

To fix this, we should increase the number of writes to the
transaction log before a dump is done (default is 100).
Together with this we also want to configure the maximum dump
log interval, in case the write treshold is set too high, this
will trigger a dump every X milliseconds, so we can be sure
that we are not losing anything if there is a period of low
writes or the treshold is set to high (default is 3 minutes)

Increasing the write treshold and reducing a bit the dump time
should help with mnesia being overloaded and prevent any
unintended consecuences from the overload.

(cherry picked from commit fb627e670737b6d103a3d487f907e0b47b585618)
Backport-of: https://github.com/crowbar/crowbar-openstack/pull/1480

### DIFF
--- a/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
@@ -1,4 +1,10 @@
 [
+ {mnesia,
+  [
+   {dump_log_write_threshold, <%= node[:rabbitmq][:mnesia][:dump_log_write_threshold] %>},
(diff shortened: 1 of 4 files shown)


Itxaka:
[RabbitMQ restarted by pacemaker from time to time](https://trello.com/c/unGNGzw0/209-rabbitmq-restarted-by-pacemaker-from-time-to-time)

PR #1486, opened 2017-12-14
[4.0] keystone: Sync fernet keys when joining cluster

skazi0:
Newly (re)installed node will not have fernet keys causing failure
when starting keystone services. Cluster founder rotates the keys
and pushes new ones to cluster member on a regular basis.

This change creates pacemaker alert to notify all nodes about about
new member. Keys are then pushed to the new node from all other nodes
to the new node (with some random time offset).

not cherry-picked directly because of different attribute schema around `token_format`.

backport of #1474 

Preconditions:
https://github.com/crowbar/crowbar-core/pull/1457
https://github.com/crowbar/crowbar-ha/pull/288
https://github.com/crowbar/crowbar-ha/pull/289

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -480,15 +480,21 @@ if node[:keystone][:signing][:token_format] == "fernet"
   # To be sure that rsync package is installed
   package "rsync"
   crowbar_pacemaker_sync_mark "sync-keystone_install_rsync" if ha_enabled
+
+  template "/usr/bin/keystone-fernet-keys-push.sh" do
+    source "keystone-fernet-keys-push.sh"
+    owner "root"
+    group "root"
+    mode "0755"
+  end
+
   rsync_command = ""
   if ha_enabled
     cluster_nodes = CrowbarPacemakerHelper.cluster_nodes(node)
     cluster_nodes.map do |n|
       next if node.name == n.name
       node_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address
-      rsync_command += \
-        "rsync -a --timeout=300 --delete-after /etc/keystone/fernet-keys " \
-        "#{node_address}:/etc/keystone/; "
+      rsync_command += "/usr/bin/keystone-fernet-keys-push.sh #{node_address}; "
     end
     raise "No other cluster members found" if rsync_command.empty?
   end
--- a/chef/cookbooks/keystone/templates/default/keystone-fernet-keys-push.sh
+++ b/chef/cookbooks/keystone/templates/default/keystone-fernet-keys-push.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+TARGETNODE=$1
+shift
+RSYNC_ARGS=$@
+test -z "$TARGETNODE" && { echo "usage: $0 <node-address> [extra rsync args]"; exit 1; }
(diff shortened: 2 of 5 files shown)


skazi0:

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-push.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+TARGETNODE=$1
+shift
+RSYNC_ARGS=$@
+test -z "$TARGETNODE" && { echo "usage: $0 <node-address> [extra rsync args]"; exit 1; }
+rsync -a --timeout=300 --delete-after $RSYNC_ARGS /etc/keystone/fernet-keys $TARGETNODE:/etc/keystone/
```
jsuchome: I kind of think $TARGETNODE should be quoted here, but I assume you'd answer that shell has covered it by taking it as first argument....

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -480,15 +480,21 @@
...
   package "rsync"
   crowbar_pacemaker_sync_mark "sync-keystone_install_rsync" if ha_enabled
+
+  template "/usr/bin/keystone-fernet-keys-push.sh" do
+    source "keystone-fernet-keys-push.sh"
+    owner "root"
+    group "root"
+    mode "0755"
+  end
+
```
jsuchome: Why isn't this template part of `recipes/ha.rb` like the other ones? It's used in HA setup only AFAICS...

chef/cookbooks/keystone/templates/default/keystone-fernet-keys-push.sh
skazi0: Spaces are not allowed in `TARGETNODE` as this is an IP/FQDN. We can add the quotes but it will not work with spaces in `TARGETNODE` anyway.

The `TARGETNODE=$1` doesn't really help here. If you run `keystone-fernet-keys-push.sh "a b"` then `TARGETNODE="a b"`.

chef/cookbooks/keystone/recipes/server.rb
skazi0: Because it's used just below in the `rsync_command` instead of direct rsync. Maybe bigger chunk surrounding code could be moved to `ha.rb` but I'm not sure if this is not done here e.g. for ordering reasons. My idea was to not change the original behavior where not needed.

jsuchome: I've seen it's executed from this file, but that's only after `ha.rb` is included

skazi0: OK, the real "problem" is that it is deployed even if ha is not enabled. An easy fix would be to move it below the `if ha_enabled`. I can also move it to the `ha.rb` if you think it's a better phase. One way or another I would prefer to include it as a part of https://github.com/crowbar/crowbar-openstack/pull/1512 if you don't mind.

skazi0: @jsuchome please confirm that the last commit here: https://github.com/crowbar/crowbar-openstack/pull/1512 is what you meant.

rsalevsky:
Can we merge this?

vuntz:
> Can we merge this?

We really want the HA CI to be green for that one.

PR #1490, opened 2017-12-15
neutron[cisco_aci]: Authentification fixes and additional options for ACI GBP

mmnelemane:
This PR enables two features for Cisco ACI:
1. Allows usage of "apic_gbp" as ml2_mechanism_drivers and fixes the authentication for the latest version of ACI firmware + group-based-policy library combination.
2. Allows Crowbar configuration for enabling "optimized_metadata" and "optimized_dhcp" features for ACI. Until now these features were enabled by default. But since certain customer deployments wanted to test without enabling these features, it made sense to provide it as a configurable option. It was painful for the customer to change in the config file manually and avoid the chef-client from overriding the config. 

Note: The intended target of this PR is Cloud 7 and is updated here due to the standard process being followed for all PRs (master-update followed by cloud 7 backport). The tests were only done for Cloud 7 based deployments.

PR #1491, opened 2017-12-15
[4.0]neutron[Cisco_ACI]: Backport from #1490

mmnelemane:
This PR enables two features for Cisco ACI:
1. Allows usage of "apic_gbp" as ml2_mechanism_drivers and fixes the authentication for the latest version of ACI firmware + group-based-policy library combination.
2. Allows Crowbar configuration for enabling "optimized_metadata" and "optimized_dhcp" features for ACI. Until now these features were enabled by default. But since certain customer deployments wanted to test without enabling these features, it made sense to provide it as a configurable option. It was painful for the customer to change in the config file manually and avoid the chef-client from overriding the config. 

Note: The intended target of this PR is Cloud 7 and is updated here due to the standard process being followed for all PRs (master-update followed by cloud 7 backport). The tests were only done for Cloud 7 based deployments.

### DIFF
--- a/chef/cookbooks/neutron/templates/default/ml2_conf_cisco_apic.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/ml2_conf_cisco_apic.ini.erb
@@ -11,7 +11,8 @@ apic_name_mapping = use_name
 apic_clear_node_profiles = True
 enable_aci_routing = True
 apic_arp_flooding = True
-enable_optimized_metadata = True
+enable_optimized_metadata = <%= node[:neutron][:apic][:optimized_metadata] %>
+enable_optimized_dhcp = <%= node[:neutron][:apic][:optimized_dhcp] %>
 apic_provision_infra = True
 apic_provision_hostlinks = True
 <% @apic_switches.keys.each do |ip| -%>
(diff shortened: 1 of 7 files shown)


chef/cookbooks/neutron/templates/default/ml2_conf_cisco_apic.ini.erb
```diff
@@ -11,7 +11,8 @@ apic_name_mapping = use_name
 apic_clear_node_profiles = True
 enable_aci_routing = True
 apic_arp_flooding = True
-enable_optimized_metadata = True
+enable_optimized_metadata = <%= node[:neutron][:apic][:optimized_metadata] %>
+enable_optimized_dhcp = <%= node[:neutron][:apic][:optimized_dhcp] %>
```
ilausuch: If "use_apic_gbp" finally is calculated as false. set true the configuration "enable_optimized_dhcp" could be cause any incompatibility?

jcaamano:
Hello Nicolas, there is some [CISCO reference documentation](https://www.cisco.com/c/en/us/support/cloud-systems-management/application-policy-infrastructure-controller-apic/tsd-products-support-series-home.html#OpenStack_Administration). I don't know if @mmnelemane has something else.

nicolasbock:
Thanks @jcaamano !

mmnelemane:
@nicolasbock: Particularly, this [section](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/aci/apic/sw/1-x/openstack/b_ACI_with_OpenStack_OpFlex_Deployment_Guide_for_Red_Hat/b_ACI_with_OpenStack_OpFlex_Deployment_Guide_for_Red_Hat_chapter_010.html) is used as reference for GBP/ML2 configurations.

PR #1492, opened 2017-12-18
keystone: add extra rotation buffer for fernet in HA cluster

dirkmueller:
if we're unlucky and resources are migrated then we have
more than one rotation per hour:

  -rw------- 1 keystone keystone   44 18. Dez 03:15 1432
  -rw------- 1 keystone keystone   44 18. Dez 03:30 1433
  -rw------- 1 keystone keystone   44 18. Dez 04:30 1434

This is when the node that was previously owning the token
rotation is taken out/rebooted/fenced, so allocate some slack.

[UNTESTED]

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -56,9 +56,10 @@ node.normal[:apache][:listen_ports_crowbar][:keystone] = { admin: [bind_admin_po
 # service bind address.
 my_admin_host = CrowbarHelper.get_host_for_admin_url(node, ha_enabled)
 my_public_host = CrowbarHelper.get_host_for_public_url(node, node[:keystone][:api][:protocol] == "https", ha_enabled)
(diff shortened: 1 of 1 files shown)


chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -234,6 +235,11 @@
 # node[:keystone][:token_expiration] is in seconds and has to be encoded to hours
 max_active_keys = (node[:keystone][:token_expiration].to_f / 3600).ceil + 2
 
+# cron.hourly runs at a different time offset on every node in the cluster,
```
cmurphy: Why is this? I would have thought something in cron.hourly is the same as a crontab entry of "0 * * * *", so every hour on the hour?

dirkmueller: it tries to run once per hour, but not at minute 0. what it does is look every 15 minutes whether it did already run in this $interval (e.g. hour) and if not, run it. so it can be run at 0, 15, 30 or 45 depending on when the machine was booted last.

PR #1495, opened 2017-12-20
neutron: Fix soft-linking of config-files.

mmnelemane:
This change explicitly links the config files neutron.conf
and opflex-agent-conf to the override location at
/etc/neutron/neutron.conf.d/*-neutron.conf and
/etc/opflex-agent-ovs/conf.d/*-opflex-agent-ovs.conf.
The missing soft-links were discovered while testing for
in L3 services with ACI backend which failed.

Note: The test was done with Cloud 7 with ACI backend. Need this PR to ensure other cases are not affected by the change.

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_config.rb
+++ b/chef/cookbooks/neutron/recipes/common_config.rb
@@ -148,6 +148,12 @@ template neutron[:neutron][:config_file] do
     )
 end
 
+if node[:platform_family] == "suse"
+  link "/etc/neutron/neutron.conf" do
+    to node[:neutron][:config_file]
+  end
+end
+
 if neutron[:neutron][:use_lbaas]
   interface_driver = "neutron.agent.linux.interface.OVSInterfaceDriver"
   if neutron[:neutron][:networking_plugin] == "ml2" &&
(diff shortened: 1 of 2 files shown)


chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -148,6 +148,12 @@
     )
 end
 
+if node[:platform_family] == "suse"
+  link "/etc/neutron/neutron.conf" do
+    to node[:neutron][:config_file]
+  end
```
nicolasbock: If I remember correctly we install the upstream default in `/etc/neutron/neutron.conf`. Why do we need to link that to the override? Shouldn't neutron automatically take care of that?

vuntz: This one feels quite wrong, to be honest. We should not touch `/etc/neutron/neutron.conf`.

mmnelemane:
vuntz: I think thats right. The opflex agent doesn't seem to read the config dir. I will find out if its fixed and patch it in the package or fix upstream. So closing this PR for now.

vuntz:
> vuntz: I think thats right. The opflex agent doesn't seem to read the config dir. I will find out if its fixed and patch it in the package or fix upstream. So closing this PR for now.

If you look into that, you may want to check https://github.com/crowbar/crowbar-openstack/pull/1475 too ;-) This may require similar changes in the packages -- we were assuming it's all fine, but that may not be true.

PR #1496, opened 2017-12-20
neutron: Backport from #1495

mmnelemane:
This change explicitly links the config files neutron.conf
and opflex-agent-conf to the override location at
/etc/neutron/neutron.conf.d/*-neutron.conf and
/etc/opflex-agent-ovs/conf.d/*-opflex-agent-ovs.conf.
The missing soft-links were discovered while testing for
in L3 services with ACI backend which failed.

(cherry picked from commit 1b3fb1885a188409753636a3d3219289c726528d)


mmnelemane:
Closing because the related PR !1495 was closed.

PR #1499, opened 2017-12-21
neutron: Fix conditional statement in template.

mmnelemane:
The checks in openvswitch_agent.ini.erb were wrongly using
  ml2_type_drivers instead of ml2_mechanism_drivers. This was corrected
  in this commit and also removed redundant assignment lines in the
  recipe

### DIFF
--- a/chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
+++ b/chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
@@ -81,6 +81,7 @@ if node.roles.include?("nova-compute-kvm")
     mode "0640"
     variables(
       ml2_type_drivers: ml2_type_drivers,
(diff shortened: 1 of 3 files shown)


chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -23,9 +23,6 @@
   neutron = node
 end
 
-ml2_mech_drivers = neutron[:neutron][:ml2_mechanism_drivers]
-ml2_type_drivers = neutron[:neutron][:ml2_type_drivers]
```
vuntz: Don't remove that, it'll fail in the next line.

mmnelemane: oops! Thanks @vuntz. Just corrected that.

PR #1502, opened 2018-01-03
neutron[cisco_aci]: add support for apic multi-pod

jcaamano:
This PR adds multi-pod support to cisco_aci. More information on [bug 1066920](https://bugzilla.novell.com/show_bug.cgi?id=1066920).

Note: The intended target of this PR is Cloud 7 and is updated here due to the standard process being followed for all PRs (master-update followed by cloud 7 backport). The tests were only done for Cloud 7 based deployments.

### DIFF
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -71,15 +71,24 @@ default[:neutron][:apic][:password] = ""
 default[:neutron][:apic][:optimized_metadata] = true
 default[:neutron][:apic][:optimized_dhcp] = true
 
-default[:neutron][:apic][:opflex][:peer_ip] = ""
-default[:neutron][:apic][:opflex][:peer_port] = 8009
-default[:neutron][:apic][:opflex][:encap] = "vxlan"
-default[:neutron][:apic][:opflex][:vxlan][:uplink_iface] = "vlan.4093"
-default[:neutron][:apic][:opflex][:vxlan][:uplink_vlan] = 4093
-default[:neutron][:apic][:opflex][:vxlan][:encap_iface] = "br-int_vxlan0"
-default[:neutron][:apic][:opflex][:vxlan][:remote_ip] = ""
-default[:neutron][:apic][:opflex][:vxlan][:remote_port] = 8472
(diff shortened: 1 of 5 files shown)


jcaamano:
> Please see my comments for #1504

Done.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -93,6 +93,8 @@
 
   # Update config file from template
   opflex_agent_conf = "/etc/opflex-agent-ovs/conf.d/10-opflex-agent-ovs.conf"
+  opflex = neutron[:neutron][:apic][:opflex].find { |i| i[:nodes].include? node[:hostname] }
```
nicolasbock: Sorry, I should have looked up how `find()` works. So it returns the first list item that has the `hostname`. Would you never expect several overlapping pods though?

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -69,15 +69,22 @@
...
+  peer_port: "",
+  encap: "vxlan",
+  vxlan: {
+    uplink_iface: "vlan.4093",
+    uplink_vlan: 4093,
+    encap_iface: "br-int_vxlan0",
+    remote_ip: "",
+    remote_port: 8472
+  },
+  vlan: ""
```
vuntz: You're missing `encap_iface` here.

jcaamano: Good catch. Fixed.

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
jcaamano: Well is not expected but indeed can happen if the admin configures the proposal incorrectly. So I have added an additional check.

jcaamano:
Hey there. Is there anything else required on my side to push the patches through the workflow? Thanks.

rsalevsky:
@mmnelemane Can you do the first review here? I think you can best see/test if this is good to go.

mmnelemane:
@jcaamano : Please rebase to the latest version and it should be good to go.

mmnelemane:
@rsalevsky : This was again rebased after your review. Is this still +1 from your side ?

PR #1505, opened 2018-01-04
Standalone Swift proxies contain swift-account config and services

matthewoliver:
Standalone swift proxies contain a `swift-account.conf` and
swift-account init scripts. Leading to customers turning
on swift-account services that are suppose to only run on
storage nodes on the proxies.
As a side issue in bug 1073387.

The proxy recipe was including the `swift::auth` recipe which
did nothing but install the `openstack-swift-account` package.
The swift-account services has nothing to do with authentication

This patch removes the auth recipe (auth.rb) and the only 2
references to it. One commented out reference in `swift::storage`
which already goes ahead an installs `openstack-swift-account`. And
`swift::proxy` which hadn't commented it out which is the underlying
cause of the problem.


toabctl:
@matthewoliver commit msg should start with ```swift:```

matthewoliver:
@toabctl done

vuntz:
@matthewoliver Will you create the backport for stable/4.0?

PR #1506, opened 2018-01-05
RabbitMQ allow to add extra users to rabbit

ilausuch:
With the introduction of the definitions file it looks like
rabbit will disregard any existing users when restarting,
using the definitions file as the blank slate to start from.

This gets in the middle of having any extra users for other
tasks like monitoring and there is no mechanism to add users.

This PR adds an method to crowbar to add a list of users
that will get written into the definitions file, allowing
us to specify extra users to be created

### DIFF
--- a/chef/data_bags/crowbar/template-rabbitmq.schema
+++ b/chef/data_bags/crowbar/template-rabbitmq.schema
@@ -16,6 +16,20 @@
             "port": { "type": "int", "required": true },
             "password": { "type": "str", "required": true },
             "user": { "type": "str", "required": true },
+            "extra_users": {
+              "type": "map",
+              "required": false,
+              "mapping": {
+                = : {
+                  "type": "map",
+                  "required": false,
+                  "mapping": {
+                    "permissions": { "type": "seq", "required": true, "sequence": [ { "type": "str" } ] },
+                    "tags": { "type": "seq", "required": true, "sequence": [ { "type": "str" } ] }
+                  }
+                }
+              }
+            },
             "vhost": { "type": "str", "required": true },
             "ssl": {
               "type": "map", "required": true, "mapping": {
--- a/crowbar_framework/app/views/barclamp/rabbitmq/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/rabbitmq/_edit_attributes.html.haml
@@ -9,6 +9,36 @@
 
     %fieldset
       %legend
+        = t(".extra_users.title")
+
+      %table.table.table-middle{ "data-dynamic" => "#extrauser-entries", "data-namespace" => "extra_users", "data-optional" => "tags",
+      "data-invalid" => t(".extra_users.error_invalid"), "data-duplicate" => t(".extra_users.error_duplicate") }
+        %thead
+          %tr
+            %th.col-sm-2
+              = t(".extra_users.username")
+            %th.col-sm-3
+              = t(".extra_users.permissions")
+            %th.col-sm-3
+              = t(".extra_users.tags")
+            %th.col-sm-1
+        %tbody
+        %tfoot
+          %tr
+            %td
+              = text_field_tag "extrauser[name]", "", :placeholder => t(".extra_users.username"),
+              :class => "form-control", "data-name" => "name", "data-type" => "string"
+            %td
+              = text_field_tag "extrauser[permissions]", "", :placeholder => t(".extra_users.permissions"),
+              :class => "form-control", "data-name" => "permissions", "data-type" => "array-string"
+            %td
+              = text_field_tag "extrauser[tags]", "", :placeholder => t(".extra_users.tags"),
+              :class => "form-control", "data-name" => "tags", "data-type" => "array-string"
+            %td
+              = link_to t(".extra_users.add"), "#", :class => "btn btn-default btn-block", "data-add" => true
+
+    %fieldset
+      %legend
         = t(".ssl_header")
 
       = boolean_field %w(ssl enabled),
@@ -48,3 +78,26 @@
           = string_field %w(ha storage shared device)
           = string_field %w(ha storage shared fstype)
           = string_field %w(ha storage shared options)
+
+
+%script#extrauser-entries{ :type => "text/x-handlebars-template" }
+  {{#each entries}}
+  %tr.edit
+    %td
+      = text_field_tag "extrauser[name]", "{{name}}", :placeholder => t(".extra_users.username"),
+      :class => "form-control", :disabled => "disabled"
+    %td
+      = text_field_tag "extrauser[permissions]", "{{permissions}}", :placeholder => t(".extra_users.permissions"),
+      :class => "form-control", "data-update" => "extra_users/{{name}}/permissions", "data-name" => "permissions",
+      "data-type" => "array-string"
+    %td
+      = text_field_tag "extrauser[tags]", "{{tags}}", :placeholder => t(".extra_users.tags"), :class => "form-control",
+      "data-update" => "extra_users/{{name}}/tags", "data-name" => "tags", "data-type" => "array-string"
+    %td
+      = link_to t(".extra_users.record_remove"), "#", :class => "btn btn-default btn-block", "data-remove" => "{{name}}"
+  {{else}}
+  %tr
+    %td{ :colspan => 4 }
+      .empty.alert.alert-info.text-center
+        = t(".extra_users.no_records")
+  {{/each}}
--- a/crowbar_framework/config/locales/rabbitmq/en.yml
+++ b/crowbar_framework/config/locales/rabbitmq/en.yml
@@ -21,6 +21,16 @@ en:
       edit_attributes:
         vhost: 'Virtual host'
         user: 'User'
+        extra_users:
+          username: 'Username'
+          permissions: 'Permissions (3 comma separated items for configure, write, read; e.g. ".*,.*,.*")'
+          tags: 'Tags (comma separated)'
+          error_invalid: 'Username and Permissions cannot be empty'
+          error_duplicate: 'There is a user with this username'
+          title: 'Extra users'
+          add: 'Add'
+          record_remove: 'Delete'
+          no_records: 'No records'
         port: 'Port'
         ssl_header: 'SSL Support'
         ssl:
@@ -54,3 +64,4 @@ en:
         no_filesystem: 'No filesystem type specified for shared storage.'
         drbd: 'DRBD is not enabled for cluster %{cluster_name}.'
         invalid_size: 'Invalid size for DRBD device.'
+        wrong_permissions: 'Wrong permissions for user %{user}. Permissions need to be 3 comma separated items (configure, write, read)'
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -132,7 +132,8 @@ template "/etc/rabbitmq/definitions.json" do
     json_trove_password: node[:rabbitmq][:trove][:password].to_json,
     json_trove_vhost: node[:rabbitmq][:trove][:vhost].to_json,
     ha_all_policy: cluster_enabled,
-    quorum: quorum
+    quorum: quorum,
+    extra_users: node[:rabbitmq][:users]
   )
   # no notification to restart rabbitmq, as we still do changes with
   # rabbitmqctl in the rabbit.rb recipe (this is less disruptive)
--- a/chef/cookbooks/rabbitmq/recipes/rabbit.rb
+++ b/chef/cookbooks/rabbitmq/recipes/rabbit.rb
@@ -134,6 +134,35 @@ execute "rabbitmqctl set_user_tags #{node[:rabbitmq][:user]} management" do
   only_if only_if_command if ha_enabled
 end
 
+node[:rabbitmq][:users].each do |user|
+  # create extra users
+  rabbitmq_user "adding user #{user[:username]}" do
+    user user[:username]
+    password user[:password]
+    address node[:rabbitmq][:management_address]
+    port node[:rabbitmq][:management_port]
+    action :add
+    only_if only_if_command if ha_enabled
+  end
+
+  # add permisions to those extra users into the default vhost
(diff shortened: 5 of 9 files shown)


Itxaka:
@ilausuch are you sure you based this off master? the change in the template from 101 to 202 makes it seem like it was based off an outdated base, as on master the schema is already in 202 -> https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/template-rabbitmq.json#L60

Itxaka:
[RabbitMQ add extra users](https://trello.com/c/ELZAtk53/234-rabbitmq-add-extra-users)

Itxaka:
Tested and works but a couple of missing things.

 - [ ] tags and permissions should be `array_string_field` on the view so it automatically rejects them if they are not separated by comma and returns a string instead of us just casting them to be a string manually
 - [ ] tags resource needs to check for each tag to check if the user has that tag. Rigth now it does a search for all tags, which probably fails
 - [ ] password reuse is not working rigth now, not sure why as it worked before. I need to check this further.

skazi0:
@ilausuch @Itxaka does it make the whole user "database" managed by crowbar? i.e. if someone creates users by hand they will be lost on the restart, right?

ilausuch:
I think is true. We don't have any code to recover the users from rabbit every time we access to the proposal. @Itxaka do you agree? Will this be the behavior?

Itxaka:
> @ilausuch @Itxaka does it make the whole user "database" managed by crowbar? i.e. if someone creates users by hand they will be lost on the restart, right?

Indeed, which is also the case right now in case of a reset (which happens sometimes on clustered rabbits)

> I think is true. We don't have any code to recover the users from rabbit every time we access to the proposal. @Itxaka do you agree? Will this be the behavior?

Nope, I dont agree with that. The only source of truth that we can guarantee the user for its settings is the barclamp. If they modify manually they may lost everything and there is nothing we can do for them :)

chef/data_bags/crowbar/template-rabbitmq.schema
```diff
@@ -16,6 +16,19 @@
...
             "user": { "type": "str", "required": true },
+            "extra_users": {
+              "type": "seq",
+              "required": true,
+              "sequence": [ {
+                "type": "map",
+                "required": true,
+                "mapping": {
+                  "username": { "type": "str", "required": true },
+                  "permissions": { "type": "str", "required": true },
```
skazi0: should be a sequence e.g.: `"permissions": { "type": "seq", "required": true, "sequence": [ { "type": "str" } ] }`

chef/data_bags/crowbar/template-rabbitmq.schema
```diff
@@ -16,6 +16,19 @@
...
+                  "tags": { "type": "str", "required": true }
```
skazi0: ditto e.g.: `"tags": { "type": "seq", "required": true, "sequence": [ { "type": "str" } ] }`

crowbar_framework/app/assets/javascripts/barclamps/rabbitmq/application.js
```diff
@@ -14,3 +14,104 @@
...
+   redisplay_extra_users();
+  }
+
+  // when add new user button is clicked
+  $('#add-extra_user-button').click(function() {
+
+    // create the new user
+    var new_extra_user = {
+      'username': $('#extra_users_index_username').val(),
+      'permissions': $('#extra_users_index_permissions').val(),
```
skazi0: `split(',')` is needed here to convert the string to array.
NOTE: it will return `[""]` for empty input string but this should not hurt in our case.

crowbar_framework/app/assets/javascripts/barclamps/rabbitmq/application.js
```diff
@@ -14,3 +14,104 @@
...
+      'tags': $('#extra_users_index_tags').val()
```
skazi0: ditto

crowbar_framework/app/views/barclamp/rabbitmq/_edit_attributes.html.haml
```diff
@@ -7,6 +7,41 @@
...
+        %li.list-group-item.active
+          %h3.list-group-item-heading
+            = t('.extra_users.new_user_title')
+            .form-group.pull-right
+              %input{ :id => "add-extra_user-button", :class => "form-control btn-default",
+                      :type => "button", :value => t('.extra_users.add_button_label') }
+
+        %li.list-group-item
+          = string_field %w(extra_users index username)
+          = string_field %w(extra_users index permissions)
```
skazi0: this should be `array_string_field` as above

crowbar_framework/app/views/barclamp/rabbitmq/_edit_attributes.html.haml
```diff
@@ -7,6 +7,41 @@
...
+          = string_field %w(extra_users index tags)
```
skazi0: ditto

crowbar_framework/config/locales/rabbitmq/en.yml
```diff
@@ -21,6 +21,16 @@ en:
       edit_attributes:
         vhost: 'Virtual host'
         user: 'User'
+        extra_users:
+          index:
+            username: 'Username'
+            permissions: 'Permissions (need to be 3 comma separated items (configure, write, read))'
+            tags: 'tags'
```
skazi0: please include hint about expected format (i.e. comma separated)

chef/data_bags/crowbar/template-rabbitmq.schema
Itxaka: +1

chef/data_bags/crowbar/template-rabbitmq.schema
Itxaka: +1

crowbar_framework/app/assets/javascripts/barclamps/rabbitmq/application.js
```diff
@@ -14,3 +14,104 @@
...
+
+  var extra_users_template;
+
+  function cb_extra_users_delete()
+  {
+    // get the index of deleted in users array
+    extra_user_index = $(this).data("extra_user_id");
+
+    // delete the extra_user from the JSON attributes
+    var extra_users = $('#proposal_attributes').readJsonAttribute('extra_users', {});
```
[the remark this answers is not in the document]
Itxaka: +1

crowbar_framework/app/assets/javascripts/barclamps/rabbitmq/application.js
```diff
@@ -14,3 +14,104 @@
...
+       $('#extra_users_entries').html()
+      );
+    }
+
+    //Hide extra user list and show loading
+    $('#rabbitmq_extra_users').hide();
+    $('#rabbitmq_extra_users_loading').show();
+
+    // Load extra users
+    var extra_users = $('#proposal_attributes').readJsonAttribute('extra_users', {});
```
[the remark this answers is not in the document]
Itxaka: +1

crowbar_framework/config/locales/rabbitmq/en.yml
```diff
@@ -21,6 +21,16 @@ en:
...
+          title: 'Extra users'
+          entry_label: 'User'
+          loading_text: 'Loading extra users...'
+          add_button_label: 'Add extra user'
```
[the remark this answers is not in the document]
Itxaka: Yes, this is very confusing indeed. The button is just supposed to be for adding an extra box to add more users, maybe a `Add another user` would be better?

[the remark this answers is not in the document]
skazi0: "Add another user" will not work either as you need to hit this button to even add one user. If you just fill the form and save the proposal, the user will not be created.

Itxaka:
@skazi0 fixed all comments and squashed all commits.

crowbar_framework/config/locales/rabbitmq/en.yml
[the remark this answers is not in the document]
Itxaka: umm, that is true, as it only gets pushed to the proposal during that `writeJsonAttribute` stuff.

changing it to `Add this user` which is more clear

Itxaka:
mmmh, not working as expected, some javascript weirdness

@skazi0 @ilausuch can you guys have a look? You are the js masters around :D

This is what you get on the barclamp load:
image


Note that the message for users loading is visible, the 2 fields have some trash text on it.

Also when clicking "Add this user" nothing happens and there is a message in the console indicating `Empty string passed to getElementById().`

crowbar_framework/app/views/barclamp/rabbitmq/_edit_attributes.html.haml
```diff
@@ -7,6 +7,41 @@
...
+        %li.list-group-item.active
+          %h3.list-group-item-heading
+            = t('.extra_users.new_user_title')
+            .form-group.pull-right
+              %input{ :id => "add-extra_user-button", :class => "form-control btn-default",
+                      :type => "button", :value => t('.extra_users.add_button_label') }
+
+        %li.list-group-item
+          = string_field %w(extra_users index username)
+          = array_string_field %w(extra_users {{@index}} permissions), :only_comma => true
```
skazi0: `{{@index}}` should be just `index` here and in the following line.

skazi0:
@ilausuch @Itxaka @vuntz I'm wondering... maybe we drop the "new user" form completely. Currently you can click "Add this user" without filling the form and it will add new "empty" user which needs to be filled before saving to pass validation. What if we replace this form with basic "Add user" button which will add empty entry above/below and user will need to fill those entries? wdyt?

skazi0:
...alternative would be to add validation to the "new user" form to not allow adding of empty users (like e.g. in nfs-client barclamp). If we decide to keep it I would align the style with cinder barclamp ("Add new Cinder Backend" form).

crowbar_framework/app/views/barclamp/rabbitmq/_edit_attributes.html.haml
Itxaka: ahh, I think I changed that in my last push for no reason. good catch @skazi0 👍

Itxaka:
> ...alternative would be to add validation to the "new user" form to not allow adding of empty users (like e.g. in nfs-client barclamp). If we decide to keep it I would align the style with cinder barclamp ("Add new Cinder Backend" form).

I think this is the way to go, aligning how everything else works in the other barclamps is a must for a decent UX

ilausuch:
WIP

ilausuch:
I have tested this version. 
- It creates the users in the rabbitmq configuration and appears in rabbitmqctl list_users
- The UI checks that there are something in permissions and in tags
- The barclamp checks that the permissions has three elements
- The users can be modified (only permissions and tags) or deleted

However I found this case: From the proposal a user is created and it's applied the user is created as expected, and then if we delete this user from the proposal it is deleted in the rabbit configuration (definitions.json) but with the command 'rabbitmqctl list_users' still appears

crowbar_framework/config/locales/rabbitmq/en.yml
```diff
@@ -21,6 +21,16 @@ en:
       edit_attributes:
         vhost: 'Virtual host'
         user: 'User'
+        extra_users:
+          username: 'Username'
+          permissions: 'Permissions (need to be 3 comma separated items for configure, write, read; i.e. ".*, .*, .*")'
+          tags: 'Tags (Comma separated)'
+          error_invalid: 'Any field can be empty'
```
skazi0: This should say "No field can be empty (except tags)" or "Username and Permissions cannot be empty".

crowbar_framework/config/locales/rabbitmq/en.yml
```diff
@@ -21,6 +21,16 @@ en:
       edit_attributes:
         vhost: 'Virtual host'
         user: 'User'
+        extra_users:
+          username: 'Username'
+          permissions: 'Permissions (need to be 3 comma separated items for configure, write, read; i.e. ".*, .*, .*")'
+          tags: 'Tags (Comma separated)'
```
skazi0: I would make it "comma" instead of "Comma".

crowbar_framework/config/locales/rabbitmq/en.yml
```diff
@@ -21,6 +21,16 @@ en:
       edit_attributes:
         vhost: 'Virtual host'
         user: 'User'
+        extra_users:
+          username: 'Username'
+          permissions: 'Permissions (need to be 3 comma separated items for configure, write, read; i.e. ".*, .*, .*")'
```
skazi0: Please shorten it to something like `Permissions (3 comma separated items for configure, write, read; e.g. ".*,.*,.*")`.

crowbar_framework/config/locales/rabbitmq/en.yml
```diff
@@ -54,3 +64,4 @@ en:
         no_filesystem: 'No filesystem type specified for shared storage.'
         drbd: 'DRBD is not enabled for cluster %{cluster_name}.'
         invalid_size: 'Invalid size for DRBD device.'
+        wrong_permissions: 'Wrong permissions for user %{user}. Permissions need to be at least 3 comma separated items (configure, write, read)'
```
skazi0: "at least 3" or "exactly 3"?

crowbar_framework/app/views/barclamp/rabbitmq/_edit_attributes.html.haml
```diff
@@ -7,6 +7,36 @@
     = integer_field :port
     = string_field :user
 
+    %fieldset
+      %legend
+        = t(".extra_users.title")
+
+      %table.table.table-middle{ "data-dynamic" => "#extrauser-entries", "data-namespace" => "extra_users", "data-optional" => "",
```
skazi0: tags should be optional (i.e. `"data-optional" => "tags"`)

ilausuch:
Thanks @skazi0

ilausuch:
@skazi0 I forgot to comment that all suggested changes are done

Itxaka:
@ilausuch what is missing from here? Seems ok to me now, should we remove the "do not merge" tag and dismiss @skazi0 review if he is not available to move forward?

PR #1512, opened 2018-01-10
Fernet keys sync polish

skazi0:
Some followup changes for fernet-keys-sync feature. See commit messages for details.

### DIFF
--- a/chef/cookbooks/keystone/recipes/ha.rb
+++ b/chef/cookbooks/keystone/recipes/ha.rb
@@ -78,12 +78,22 @@ end
 # node is reinstalled. We assume that time between keystone installation and
 # possible node reinstallation is high enough to run this safely without
 # syncmarks.
+fernet_resources_action = node[:keystone][:token_format] == "fernet" ? :create : :delete
+
+template "/usr/bin/keystone-fernet-keys-push.sh" do
+  source "keystone-fernet-keys-push.sh"
+  owner "root"
+  group "root"
+  mode "0755"
+  action fernet_resources_action
+end
+
 template "/usr/bin/keystone-fernet-keys-sync.sh" do
   source "keystone-fernet-keys-sync.sh"
   owner "root"
   group "root"
   mode "0755"
-  action node[:keystone][:token_format] == "fernet" ? :create : :delete
+  action fernet_resources_action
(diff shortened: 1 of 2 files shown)


skazi0:

vuntz:
@skazi0 you can rebase now that the other patch is merged.

skazi0:
@vuntz thx! rebased.

chef/cookbooks/keystone/recipes/ha.rb
```diff
@@ -72,11 +72,28 @@
...
+  mode "0755"
+  action fernet_resources_action
+end
+
 template "/usr/bin/keystone-fernet-keys-sync.sh" do
   source "keystone-fernet-keys-sync.sh"
   owner "root"
   group "root"
   mode "0755"
+  action fernet_resources_action
```
ilausuch: We always will take an action create or delete. Sometimes will create a file over existing one, or other ways delete an non existing one? Could it generates a problem? The same for the flowing actions.

skazi0: Chef's meaning for those actions is more like "create" = "make sure it exists and matches the template" and "delete" = "make sure it doesn't exist". I don't expect any problems with these.

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -391,13 +391,6 @@
   package "rsync"
   crowbar_pacemaker_sync_mark "sync-keystone_install_rsync" if ha_enabled
 
-  template "/usr/bin/keystone-fernet-keys-push.sh" do
-    source "keystone-fernet-keys-push.sh"
-    owner "root"
-    group "root"
-    mode "0755"
-  end
-
```
vuntz: Given that this script is ued a few line below, I don't know of moving it is really a good idea (unless you move the all thing).

skazi0: This script is used in two places. Directly in `rsync_command` below and inside `keystone-fernet-keys-sync.sh` alert handler. We could move the whole HA related code from `server.rb` to `ha.rb` but it is mixed with some non-HA code so some duplication would be needed.

PR #1513, opened 2018-01-15
[4.0] swift: Standalone proxies contain swift-account

matthewoliver:
Standalone swift proxies contain a `swift-account.conf` and
swift-account init scripts. Leading to customers turning
on swift-account services that are suppose to only run on
storage nodes on the proxies.
As a side issue in bug 1073387.

The proxy recipe was including the `swift::auth` recipe which
did nothing but install the `openstack-swift-account` package.
The swift-account services has nothing to do with authentication

This patch removes the auth recipe (auth.rb) and the only 2
references to it. One commented out reference in `swift::storage`
which already goes ahead an installs `openstack-swift-account`. And
`swift::proxy` which hadn't commented it out which is the underlying
cause of the problem.

(cherry picked from commit 4dcc7d7a3c049c9d04a207d9fad3baf037a8ea8e)

Backport of https://github.com/crowbar/crowbar-openstack/pull/1505


matthewoliver:
@vuntz  Here is the backport.

PR #1517, opened 2018-01-18
[3.0] nova: Add resource limits for api and compute (bsc#1057086)

s-t-e-v-e-n-k:
Based on exising work for allowing modification of system resource
limits, extend it to allow changing the open file limits for the
nova-api and nova-compute services.

(cherry picked from commit 650e08f80a2c51646d22128a4a85fc1e3dd70397)


s-t-e-v-e-n-k:
Force pushing due to Jenkins misbehaving about rebuilding the HA job that keeps hanging, but I finally worked out why the job times out and is then killed -- 'modprobe drbd' hangs in finit_module(), and I can not work out why.

PR #1518, opened 2018-01-18
[WIP] Compute HA refresh

aspiers:
This coupled with many updates to openstack-resource-agents and fence-agents should in theory bring our compute HA stack up to date with upstream, but it doesn't quite work yet.


aspiers:
Just realised the main part of this change was totally unnecessary - dropped.

PR #1519, opened 2018-01-18
rabbitmq: Various changes to enable the use of rabbitmq 3.6.x

vuntz:
We need to adapt to how rabbitmq packages now use the systemwide epmd service, and then we can enjoy some new features from rabbitmq 3.6.x.

### DIFF
--- a/chef/cookbooks/rabbitmq/attributes/default.rb
+++ b/chef/cookbooks/rabbitmq/attributes/default.rb
@@ -57,10 +57,12 @@ default[:rabbitmq][:ha][:clustered_op][:monitor] = [
   { interval: "30s" }, { interval: "27s", role: "Master" }
 ]
 
+default[:rabbitmq][:hipe_compile] = false
 default[:rabbitmq][:ha][:clustered_rmq_features] = false
 case node[:platform_family]
 when "suse"
   if node[:platform] != "suse" || node[:platform_version].to_f > 12.2
+    default[:rabbitmq][:hipe_compile] = true
     default[:rabbitmq][:ha][:clustered_rmq_features] = true
   end
 end
--- a/chef/cookbooks/rabbitmq/attributes/default.rb
+++ b/chef/cookbooks/rabbitmq/attributes/default.rb
@@ -56,3 +56,11 @@ default[:rabbitmq][:ha][:clustered_op][:notify][:timeout] = "180s"
 default[:rabbitmq][:ha][:clustered_op][:monitor] = [
   { interval: "30s" }, { interval: "27s", role: "Master" }
 ]
+
+default[:rabbitmq][:ha][:clustered_rmq_features] = false
+case node[:platform_family]
+when "suse"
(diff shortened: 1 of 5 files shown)


chef/cookbooks/rabbitmq/attributes/default.rb
```diff
@@ -56,3 +56,17 @@
 default[:rabbitmq][:ha][:clustered_op][:monitor] = [
   { interval: "30s" }, { interval: "27s", role: "Master" }
 ]
+
+case node[:platform_family]
+when "suse"
+  if node[:platform] == "suse" && node[:platform_version].to_f <= 12.2
```
Itxaka: why not set those values to false by default and only change in the case of suse instead of this more complicated approach?

```
default[:rabbitmq][:hipe_compile] = false
default[:rabbitmq][:ha][:clustered_rmq_features] = false

if node[:platform] == "suse" && node[:platform_version].to_f > 12.2
  default[:rabbitmq][:hipe_compile] = true
  default[:rabbitmq][:ha][:clustered_rmq_features] = true
end
```

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -37,6 +37,45 @@
 package "rabbitmq-server"
 package "rabbitmq-server-plugins" if node[:platform_family] == "suse"
 
+if node[:platform_family] == "suse"
```
Itxaka: dont we need to check for `node[:platform_version].to_f > 12.2` as well?

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -37,6 +37,45 @@
...
+  # With new erlang packages, we move to a system-wide epmd service, with a
+  # epmd.socket unit. This is enabled by default but only listens on 127.0.0.1,
+  # while we need it to listen on the admin network too.
+
+  directory "/etc/systemd/system/epmd.socket.d" do
+    owner "root"
+    group "root"
+    mode 0o755
+    action :create
+    only_if "grep -q Requires=epmd.service /usr/lib/systemd/system/rabbitmq-server.service"
```
Itxaka: I dont really get this `only_if`

isnt it easier to only do this on platforms we know its needed like `node[:platform_version].to_f > 12.2` instead? This seems very flimsy, especially because cookbooks and the package are not really synced so when modifying the package there is no visibility into this.

chef/cookbooks/rabbitmq/recipes/default.rb
vuntz: No, that's why I added the `only_if` parts in the resources here. The rationale is that it'll make it easier to backport to stable/4.0 for people who need newer rabbitmq there ;-)

chef/cookbooks/rabbitmq/attributes/default.rb
vuntz: better now?

PR #1520, opened 2018-01-18
rabbitmq: Do not mirror queues to all nodes

vuntz:
Mirroring queues is known to have an impact on performance [1] [2]. We
were looking at mirroring queues to all nodes, but what we really need
is to mirror to a quorum of nodes. This should improve things a bit as a
first step. We may choose to not mirror some queues as a later step.

[1] "To How Many Nodes to Mirror?" in https://www.rabbitmq.com/ha.html
[2] http://fuel-ccp.readthedocs.io/en/latest/design/ref_arch_1000_nodes.html#replication

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -21,6 +21,7 @@
 ha_enabled = node[:rabbitmq][:ha][:enabled]
 # we only do cluster if we do HA
 cluster_enabled = node[:rabbitmq][:cluster] && ha_enabled
+quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
 crm_resource_stop_cmd = cluster_enabled ? "force-demote" : "force-stop"
 crm_resource_start_cmd = cluster_enabled ? "force-promote" : "force-start"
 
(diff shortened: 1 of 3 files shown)


Itxaka:
Seems to work as expected, when the policy is applied one of the mirrors "dies" so we get 2 mirrors on a 3 node cluster:

```
Mirrored queue 'neutron-vo-SecurityGroupRule-1.0.d52-54-77-77-01-04' in vhost 'nova': Master <rabbit@d52-54-77-77-01-01.1.6716.0> saw deaths of mirrors <rabbit@d52-54-77-77-01-02.3.6890.0>
```

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -21,6 +21,7 @@
 ha_enabled = node[:rabbitmq][:ha][:enabled]
 # we only do cluster if we do HA
 cluster_enabled = node[:rabbitmq][:cluster] && ha_enabled
+quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
```
dirkmueller: this doesn't need to be protected by a cluster_enabled check?

vuntz: If there's no cluster, then it will be the same as `0 / 2 + 1`

PR #1523, opened 2018-01-19
Enable tempest cinder snapshot tests for NFS backend

flaviodsr:
With the added support for cinder snapshot when using a NFS
backend, this change enables cinder snapshot tempest tests
when 'nfs_snapshot' is set to True.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -409,7 +409,7 @@ cinders[0][:cinder][:volumes].each do |volume|
     break
   elsif volume[:backend_driver] == "nfs"
     storage_protocol = "nfs"
-    cinder_snapshot = false
+    cinder_snapshot = volume[:nfs][:nfs_snapshot]
     break
   elsif volume[:backend_driver] == "vmware"
     vendor_name = "VMware"


chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -409,7 +409,7 @@
     break
   elsif volume[:backend_driver] == "nfs"
     storage_protocol = "nfs"
-    cinder_snapshot = false
+    cinder_snapshot = false if volume[:nfs][:nfs_snapshot] == false
```
nicolasbock: We don't want to turn it on if `nfs_snapshot` is `true`?

flaviodsr: Prior to these checkings, cinder_snapshot is assigned to true, see https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/tempest/recipes/config.rb#L388

nicolasbock: I see. Thanks! Wouldn't it be easier to just `cinder_snapshot = volume[:nfs][:nfs_snapshot]`?

nicolasbock: Why the conditional?

flaviodsr: I agree, however I wanted to keep the same pattern used before, for example in: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/tempest/recipes/config.rb#L408

If you think its better to remove the conditional, it is also fine for me.

nicolasbock: Personally I think it would be more readable :smile: This construction seems a bit overly complicated for getting a simple Boolean into a variable :wink:

aplanas: I had the same concern, but actually is following the same patter that the rest

nicolasbock: I agree, that being consistent is good. @flaviodsr could you have a quick look how often we use this over engineered approach? Maybe it's not too much work to change all of them.

flaviodsr: Ok, taking a better look on it. In fact, I am wrong saying that it was used before.

In https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/tempest/recipes/config.rb#L408 is a different case. As it is, for any value, besides `nfs`, assigned to `volume[:netapp][:storage_protocol]` `storage_protocol` will be "iSCSI", so it makes sense to have the conditional.

On the case of this commit, as you guys were suggesting, it is safe to remove the conditional as `cinder_snapshot` will always have the same value of `volume[:nfs][:nfs_snapshot]` and it will be more readable.

Hope it is clear enough! 🤔

nicolasbock: Thanks!

PR #1525, opened 2018-01-19
rabbitmq: Fixes to changes for rabbitmq 3.6.x support

vuntz:
This is a followup to https://github.com/crowbar/crowbar-openstack/pull/1519 to fix the CI (we hadn't tested the non-HA mode, and for some reason HA is significantly slower in the CI than with our own mkcloud).

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -67,11 +67,16 @@ if node[:platform_family] == "suse"
     subscribes :run, "template[/etc/systemd/system/epmd.socket.d/port.conf]", :immediate
   end
 
-  # Enable epmd.socket so that even when we don't use the rabbitmq systemd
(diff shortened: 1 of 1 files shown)


chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -96,6 +96,16 @@
   notifies :restart, "service[rabbitmq-server]"
 end
 
+virtualized = [
+  "KVM", "QEMU", "Bochs",
+  "VMWare Virtual Platform", "VMware Virtual Platform",
+  "VirtualBox"
+]
+hipe_compile = node[:rabbitmq][:hipe_compile] &&
```
dirkmueller: personally I'd prefer either a check on ```systemd-detect-virt``` return value or disable this by default alltogether

vuntz: done in https://github.com/crowbar/crowbar-openstack/pull/1527

PR #1526, opened 2018-01-20
[4.0] rabbitmq: Various changes to enable the use of rabbitmq 3.6.x

vuntz:
This is still compatible with current behavior (as in "no change") for old rabbitmq, but enables people to play with rabbitmq 3.6.x on that branch.

Backport of https://github.com/crowbar/crowbar-openstack/pull/1519, https://github.com/crowbar/crowbar-openstack/pull/1525, https://github.com/crowbar/crowbar-openstack/pull/1527.

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -37,6 +37,45 @@ end
 package "rabbitmq-server"
 package "rabbitmq-server-plugins" if node[:platform_family] == "suse"
 
+if node[:platform_family] == "suse"
+  # With new erlang packages, we move to a system-wide epmd service, with a
+  # epmd.socket unit. This is enabled by default but only listens on 127.0.0.1,
+  # while we need it to listen on the admin network too.
+
+  directory "/etc/systemd/system/epmd.socket.d" do
+    owner "root"
+    group "root"
+    mode 0o755
+    action :create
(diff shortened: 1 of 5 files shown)


chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -52,14 +96,25 @@
   notifies service_action, "service[rabbitmq-server]"
 end
 
+virtualized = [
```
dirkmueller: here I would prefer a check on the return value of systemd-detect-virt instead of hardcoding this here

vuntz: @dirkmueller done

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -52,14 +96,19 @@
   notifies service_action, "service[rabbitmq-server]"
 end
 
+`systemd-detect-virt -v -q`
+virtualized = $?.exitstatus.zero?
+hipe_compile = node[:rabbitmq][:hipe_compile] && !virtualized
```
Itxaka: this needs a forwardport to master

vuntz: That was https://github.com/crowbar/crowbar-openstack/pull/1527

PR #1528, opened 2018-01-22
rabbitmq: add some performance configs

Itxaka:
Adds some performance configs to tweak rabbitmq:

- backlog: number of connections that can be initialized at the
same time. Useful for a high number of connections or for a temporal
high numbers of connections, like in a server restart. Default is 128,
recommended is 4096.

- sndbuf: increase buffer for TCP connections of consumers, on linux
this value is set automatically between 80KB and 120 KB.
The higher this value is, the higher the throughput and the RAM consumed.
Recommended values for this config depend on the number of connections
and the amount of RAM available.

- recbuf: same as sndbuf but for publishers and protocol operations.

- collect_statistics_interval: the interval in milliseconds to collect
statistics, mainly by the management plugin. Increasing it will reduce
the load on the node that has the stats collector DB. Default is 5000,
recommended is 30000 or even 60000, depending on how recent you need
the stats.

### DIFF
--- a/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
@@ -7,6 +7,13 @@
  },
  {rabbit,
   [
+   <% if node[:rabbitmq][:collect_statistics_interval] -%>
+   {collect_statistics_interval, <%= node[:rabbitmq][:collect_statistics_interval] %>},
+   <% end -%>
+   {tcp_listen_options, [
+                         <%= node[:rabbitmq][:tcp_listen_options].reject { |k,v| v.nil? }.map { |k,v| "{#{k}, #{v}}" }.join(",") %>
+                        ]
+   },
    {tcp_listeners, [
                     <%= node[:rabbitmq][:addresses].map { |address| "{\"#{address}\", #{node[:rabbitmq][:port]}}" }.join(", ") %>
                    ]},
--- a/chef/data_bags/crowbar/template-rabbitmq.json
+++ b/chef/data_bags/crowbar/template-rabbitmq.json
@@ -50,14 +50,20 @@
       "mnesia": {
         "dump_log_write_threshold": 100,
         "dump_log_time_threshold": 180000
-      }
+      },
+      "tcp_listen_options": {
+        "backlog": 128,
+        "sndbuf": null,
+        "recbuf": null
+      },
+      "collect_statistics_interval": 5000
     }
   },
   "deployment": {
     "rabbitmq": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 202,
+      "schema-revision": 203,
       "element_states": {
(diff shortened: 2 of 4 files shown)


Itxaka:

chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -50,14 +50,18 @@
       "mnesia": {
         "dump_log_write_threshold": 100,
         "dump_log_time_threshold": 180000
-      }
+      },
+      "backlog": 128,
+      "sndbuf": null,
+      "recbuf": null,
```
vuntz: Should we put these three options together in a "tcp_options" hash?

Itxaka: we should and I will. that will probably fix also the trailing comma that cannot be there

chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
```diff
@@ -7,6 +7,13 @@
  },
  {rabbit,
   [
+   <% if node[:rabbitmq][:collect_statistics_interval] %>
```
vuntz: s/`%>`/`-%>`/

chef/data_bags/crowbar/template-rabbitmq.schema
```diff
@@ -95,7 +95,17 @@
...
+            "tcp_listen_options": {
+              "type": "map",
+              "required": false,
+              "mapping": {
+                "backlog": { "type": "int", "required": false },
+                "sndbuf": { "type": "int", "required": false },
+                "recbuf": { "type": "int", "required": false }
+              }
+            },
+            "collect_statistics_interval": { "type": "int", "required": false}
```
vuntz: Any reason to put them as not required, but still defining them in the template and adding a migration for them?

chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -50,14 +50,20 @@
       "mnesia": {
         "dump_log_write_threshold": 100,
         "dump_log_time_threshold": 180000
-      }
+      },
+      "tcp_listen_options": {
+        "backlog": 128,
+        "sndbuf": null,
+        "recbuf": null
```
vuntz: Why not putting the default value for these two? (I mean the upstream default)

chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
Itxaka: 👍

chef/data_bags/crowbar/template-rabbitmq.json
Itxaka: I dunno, let me change that ¯\\\_(ツ)_/¯

chef/data_bags/crowbar/template-rabbitmq.schema
Itxaka: I dunno, let me change that ¯\\\_(ツ)_/¯

PR #1530, opened 2018-01-22
crowbar: change postgresql connection to avoid deadlocks

ilausuch:
Change connection string scheme to postgresql+psycopg2 in case of
postgresql connections. This prevent the deadlocks in the DB particulary
using cinder and neutron.

Not mentioning a specific driver in the sql connection string configuration makes openstack use the wrapper driver instead of the native one. This causes deadlocks due to the wrapper using serialization vs the native driver using parallelism.

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -137,6 +137,8 @@ class CrowbarOpenStackHelper
       end
     end
 
+    db_conn_scheme = "postgresql+psycopg2" if db_conn_scheme == "postgresql"
+
     "#{db_conn_scheme}://" \
     "#{db_auth['user']}:#{db_auth['password']}@#{db_settings[:address]}/" \
     "#{db_auth['database']}" \


ilausuch:
[Change postgresql backend in sqlalchemy to avoid DB deadlocks](https://trello.com/c/oLP5h9SA/85-change-postgresql-backend-in-sqlalchemy-to-avoid-db-deadlocks)

Itxaka:
@ilausuch would be possible to add more info to the commit message? For example, if this commit is checked out in a couple of years, we wont know where the issues comes really from unless github and trello are still up and being used :)

As far as I understand the problem this is due to the following

```
Not mentioning a specific driver in the sql connection string configuration makes openstack use the wrapper driver instead of the native one. This causes deadlocks due to the wrapper using serialization vs the native driver using parallelism. 
```

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -137,6 +137,8 @@ def self.database_connection_string(db_settings, db_auth_attr)
       end
     end
 
+    db_conn_scheme = "postgresql+psycopg2" if db_conn_scheme == "postgresql"
```
rsalevsky: Can you please do it in a similar way as for mysql? the code looks a bit too hacky this way.

ilausuch:
I close this pull request because openstack is already using the driver psycopg2.

After checking in the documentation and the oslo.db code. My conclusion is that by default openstack in SOC8 and SOC6CD are using by default psycopg2 as the driver for postgresql. To be sure that I don't miss something, I would like to share with you the reasons for this conclusion:

In SOC8 and SOC6CD the SQLalchemy library is using by default psycopg2
in SOC6CD we are usign the version of SQLalchemy 1.0.8
http://docs.sqlalchemy.org/en/rel_1_0/core/engines.html

In SOC8 we are using 1.1.12
http://docs.sqlalchemy.org/en/rel_1_1/core/engines.html

In jenkins jobs during the python libraries are not references to the other drivers only psycopg2
3.There aren't references in the code of oslo.db to others drivers : pg8000, psycopg2cffi, pypostgresql, pygresql, zxjdbc only psycopg2

4.It seems that in the code the connection string is not parsed before creating the connection. I reviewed enginefacade.py and engines.py

PR #1532, opened 2018-01-24
keystone: Update endpoint before chef run

sayalilunkad:
Without this change we run into scenarios where if the chef run fails for any reason
when we modify the endpoints in keystone the data is not updated/saved as apply_role_post_chef_call
is only called after the chef run. To avoid this case we should save the endpoint data in
apply_role_pre_chef_call as well.

### DIFF
--- a/crowbar_framework/app/models/keystone_service.rb
+++ b/crowbar_framework/app/models/keystone_service.rb
@@ -149,6 +149,20 @@ class KeystoneService < OpenstackServiceObject
       allocate_virtual_ips_for_any_cluster_in_networks_and_sync_dns(server_elements, vip_networks)
     end
 
+    # Save current keystone endpoints to all keystone-server nodes.
+    all_nodes.each do |n|
+      node = NodeObject.find_by_name(n)
+      unless node[:keystone].nil?
+        node[:keystone][:endpoint] = {
+          insecure: node[:keystone][:ssl][:insecure],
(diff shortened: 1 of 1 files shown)


nicolasbock:
Failing in CI due to the grub issue covered in https://github.com/crowbar/crowbar-core/pull/1485. We should wait until that PR is merged before rerunning CI on this PR.

vuntz:
> Without this change we run into scenarios where if the chef run fails for any reason
> when we modify the endpoints in keystone the data is not updated/saved as apply_role_post_chef_call
> is only called after the chef run

Then maybe we should do the update in pre_chef_call instead of post_chef_call?

sayalilunkad:
@vuntz We need it in the post_chef_call as well for the one scenario when the barclamp is applied the first time because the node attributes do not exist at that point and so can't be updated in the pre_chef_call.

sayalilunkad:
[Issues when keystone endpoint changed](https://trello.com/c/FhYLJT1t/676-issues-when-keystone-endpoint-changed)

jsuchome:
Is there a bug report for that described behavior? If so, please include the bsc in the commit message...

sayalilunkad:
@jsuchome there isn't any bug directly but I ran into this issue while working on bsc#1075394.

sayalilunkad:
@vuntz are you also ok with this change?

PR #1533, opened 2018-01-24
apache: don't collect Listen ports from wsgi vhosts (bsc#1077234)

stefannica:
The listen_ports_crowbar attribute controls which ports are added
as Listen directives to the /etc/apache2/listen.conf file.

However, all wsgi vhosts aside from horizon already have an explicit
Listen directive in the vhost configuration file, so there's no need
to add these ports to the /etc/apache2/listen.conf file.

To align horizon to the other WSGI vhosts, Listen directives are
added to the horizon WSGI vhost configuration file. Since these don't
also have to be added to the /etc/apache2/listen.conf file anymore, the
usage of listen_ports_crowbar can be removed altogether.

Fixes [bsc#1077234](https://bugzilla.suse.com/show_bug.cgi?id=1077234)

Additional notes:
 - `listen_ports_crowbar` is still being used by the HA apache recipe [1] to determine which port to use for the status URL. If no `listen_ports_crowbar` entries are configured (which is indeed the case unless the horizon barclamp is deployed) then the default 80 port is used and the apache HA status URL is `http://127.0.0.1:80/server-status` [2]. This could be a problem (to be tested).
UPDATE: the server_status module isn't even configured in apache, so this part can be safely ignored.

[1] https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb#L33
[2] https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb#L46

### DIFF
--- a/chef/cookbooks/aodh/recipes/aodh.rb
+++ b/chef/cookbooks/aodh/recipes/aodh.rb
@@ -145,9 +145,6 @@ else
   bind_port = node[:aodh][:api][:port]
 end
 
-node.normal[:apache][:listen_ports_crowbar] ||= {}
-node.normal[:apache][:listen_ports_crowbar][:aodh] = { plain: [bind_port] }
-
 memcached_servers = MemcachedHelper.get_memcached_servers(
   ha_enabled ? CrowbarPacemakerHelper.cluster_nodes(node, "aodh-server") : [node]
 )
--- a/chef/cookbooks/ceilometer/recipes/server.rb
+++ b/chef/cookbooks/ceilometer/recipes/server.rb
@@ -195,9 +195,6 @@ else
   bind_port = node[:ceilometer][:api][:port]
 end
 
-node.normal[:apache][:listen_ports_crowbar] ||= {}
-node.normal[:apache][:listen_ports_crowbar][:ceilometer] = { plain: [bind_port] }
-
 if ceilometer_protocol == "https"
   ssl_setup "setting up ssl for ceilometer" do
     generate_certs node[:ceilometer][:ssl][:generate_certs]
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -484,20 +484,7 @@ else
   bind_port_ssl = 443
 end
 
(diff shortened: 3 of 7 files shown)


jsuchome:
_However, all wsgi vhosts aside from horizon already have an explicit
Listen directive in the vhost configuration file_

Could you point to the example?

stefannica:
>> However, all wsgi vhosts aside from horizon already have an explicit
>> Listen directive in the vhost configuration file
>
> Could you point to the example?

@jsuchome , all wsgi vhost configuration files (aside from horizon's) are generated via the `crowbar_openstack_wsgi` helper (e.g. [1]) and the template always includes a `Listen` directive [2].
[1] https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/keystone/recipes/server.rb#L140
[2] https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/crowbar-openstack/templates/default/vhost-wsgi.conf.erb#L1

jsuchome:
Oh, thanks! I was trying to look for some templates :-)

vuntz:
@stefannica does it help if I tell you that the OCF RA is not used in SLE12 for apache? So the whole thing about `listen_ports_crowbar` may actually be completely dropped...

stefannica:
> @stefannica does it help if I tell you that the OCF RA is not used in SLE12 for apache? So the whole thing about listen_ports_crowbar may actually be completely dropped...

That's a very useful hint, @vuntz . If you're saying that this crowbar-ha code is never reached, then I will remove the usage of `listen_ports_crowbar` even in the horizon barclamp: https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/recipes/apache.rb#L46

vuntz:
> That's a very useful hint, @vuntz . If you're saying that this crowbar-ha code is never reached

Don't blindly trust me, you probably want to double-check ;-)

stefannica:
rebased

stefannica:
Gate failures should be fixed by #1491

stefannica:
@jsuchome , @vuntz , please review this when you get a chance ! Thanks !

PR #1535, opened 2018-01-26
neutron-ha-tool: Add insecure flag (bsc#1075394)

sayalilunkad:
The insecure flag was missing from the status command which was causing the neutron-l3-ha-service to
fail.


sayalilunkad:
[P2: Bug 1075394 - #L3 neutron barclamp can't be deployed setting "Generate (self-signed) certificates (implies insecure)"](https://trello.com/c/uw0Qkh4w/659-p2-bug-1075394-l3-neutron-barclamp-cant-be-deployed-setting-generate-self-signed-certificates-implies-insecure)

PR #1536, opened 2018-01-26
rabbitmq: fix queue_master_locator value

Itxaka:
This specific options needs to be an erlang binary string in
the config as mentioned on the docs[0]

[0] http://previous.rabbitmq.com/v3_6_x/configure.html


Itxaka:

PR #1537, opened 2018-01-26
[4.0] rabbitmq: fix queue_master_locator value

Itxaka:
It needs to be an erlang binary

(cherry picked from commit 5f266b46e9f1e36cd605a6b4046accd3a0b3c20e)

Backport-of: https://github.com/crowbar/crowbar-openstack/pull/1536

### DIFF
--- a/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
@@ -28,7 +28,7 @@
 <% end -%>
 <% if @cluster_enabled -%>
    {cluster_partition_handling, <%= @cluster_partition_handling %>},
-   {queue_master_locator, "min-masters"},
+   {queue_master_locator, <<"min-masters">>},
 <% end -%>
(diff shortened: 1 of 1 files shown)


Itxaka:
> If tested

tested on sap deployment ;D

Itxaka:
> I don't quite understand your commit message to be honest 😉 How is that change referring to a binary?

yeah its terrible, like erlang.

So on erlang there is no string manipulation (well there kind of is but its terrible) so to use strings as values, they sometimes use binary strings, which are represented by `<<"string goes here">>` and erlang will transform them into binary, send them over the wire and they will be transformed back into a string representation.

For some unknown reason, in the rabbit config, this option needs to be set as a binary string, instead of an atom (erlang kind of constants) like the other configs.

I now can see that I did only mentioned `binary` but not `binary string`. Let me fix that :)

nicolasbock:
Ok, now that is crazy! Thanks for the explanation! And sorry you had to dive into this madness ;)

Itxaka:

PR #1543, opened 2018-01-31
[4.0] rabbitmq: add some performance configs

Itxaka:
Adds some performance configs to tweak rabbitmq:

- backlog: number of connections that can be initialized at the
same time. Useful for a high number of connections or for a temporal
high numbers of connections, like in a server restart. Default is 128,
recommended is 4096.

- sndbuf: increase buffer for TCP connections of consumers, on linux
this value is set automatically between 80KB and 120 KB.
The higher this value is, the higher the throughput and the RAM consumed.
Recommended values for this config depend on the number of connections
and the amount of RAM available.

- recbuf: same as sndbuf but for publishers and protocol operations.

- collect_statistics_interval: the interval in milliseconds to collect
statistics, mainly by the management plugin. Increasing it will reduce
the load on the node that has the stats collector DB. Default is 5000,
recommended is 30000 or even 60000, depending on how recent you need
the stats.

(cherry picked from commit 25f8f347f385594f4d8742bc898f23473304001c)

Backport-of: https://github.com/crowbar/crowbar-openstack/pull/1528

### DIFF
--- a/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
@@ -7,6 +7,13 @@
  },
  {rabbit,
   [
+   <% if node[:rabbitmq][:collect_statistics_interval] -%>
+   {collect_statistics_interval, <%= node[:rabbitmq][:collect_statistics_interval] %>},
+   <% end -%>
+   {tcp_listen_options, [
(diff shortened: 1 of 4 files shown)


Itxaka:
[More rabbitmq tweaks for performance](https://trello.com/c/HP5w4bjs/254-more-rabbitmq-tweaks-for-performance)

Itxaka:

Itxaka:
This wont work unless we update rabbit to 3.6 like in master, leaving it here in case we do in the future

vuntz:
Let's close it, we can always reopen if needed.

PR #1544, opened 2018-01-31
ironic: Expose pxe_append_params as attribute + migration fix

skazi0:
pxe_append_params option is used to add extra kernel parameters to IPA runs.
Example use:
`sshkey="<public key>"` param gives ssh access to running IPA image.
`coreos.autologin` enables console auto login (CoreOS based image)
`ipa-debug=1` enables debug logging in IPA
Great for troubleshooting.

EDIT: additional commit fixes small bug causing migrations to fail.

### DIFF
--- a/chef/data_bags/crowbar/migrate/ironic/202_add_pxe_append_params.rb
+++ b/chef/data_bags/crowbar/migrate/ironic/202_add_pxe_append_params.rb
@@ -1,0 +1,12 @@
+def upgrade(ta, td, a, d)
+  unless a.key? "pxe_append_params"
+    a["pxe_append_params"] = ta["pxe_append_params"]
+  end
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("pxe_append_params")
+  return a, d
+end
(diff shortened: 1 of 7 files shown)


skazi0:
@nicolasbock the line you linked does provide the logger parameter.
The problem I had during migrations had this trace:

```
wrong number of arguments (0 for 1)
/opt/dell/crowbar_framework/app/models/ironic_service.rb:19:in `initialize'
/opt/dell/crowbar_framework/lib/schema_migration.rb:63:in `new'
/opt/dell/crowbar_framework/lib/schema_migration.rb:63:in `block in run_for_bc'
/opt/dell/crowbar_framework/lib/schema_migration.rb:49:in `run_for_bc'
/opt/dell/crowbar_framework/lib/tasks/crowbar.rake:30:in `block (3 levels) in <top (required)>'
/opt/dell/crowbar_framework/lib/tasks/crowbar.rake:29:in `each'
/opt/dell/crowbar_framework/lib/tasks/crowbar.rake:29:in `block (2 levels) in <top (required)>'
```

nicolasbock:
Yes, you are right. Thanks for the backtrace.

chef/data_bags/crowbar/migrate/ironic/202_add_pxe_append_params.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["pxe_append_params"] = ta["pxe_append_params"]
```
toabctl: If we want to backport this, a check if the key is already there is needed.

skazi0: I don't think we need to backport this but I added the check just in case.

PR #1545, opened 2018-02-01
Disallow postgresql and unclustered HA for rabbitmq for new deployments

vuntz:
We don't want new deployments to use these old ways of deploying the database and rabbitmq.

Old deployments will still work fine, as the check is only done when the barclamp is not applied yet.

### DIFF
--- a/crowbar_framework/app/models/database_service.rb
+++ b/crowbar_framework/app/models/database_service.rb
@@ -120,6 +120,20 @@ class DatabaseService < PacemakerServiceObject
       db_engine: db_engine
     ) unless %w(mysql postgresql).include?(db_engine)
 
+    # for new deployments, only allow mysql
+    case attributes["sql_engine"]
+    when "postgresql"
+      proposal_id = proposal["id"].gsub("#{@bc_name}-", "")
+      proposal_object = Proposal.where(barclamp: @bc_name, name: proposal_id).first
+      if proposal_object.nil? || !proposal_object.active_status?
+        validation_error I18n.t(
+          "barclamp.#{@bc_name}.validation.no_new_postgresql"
+        )
+      end
+    when "mysql"
+      # we're good
+    end
+
     # HA validation
     servers = proposal["deployment"][@bc_name]["elements"]["database-server"]
     unless servers.nil? || servers.first.nil? || !is_cluster?(servers.first)
--- a/crowbar_framework/config/locales/database/en.yml
+++ b/crowbar_framework/config/locales/database/en.yml
@@ -56,6 +56,7 @@ en:
               options: 'Mount Options'
       validation:
         invalid_db_engine: 'Invalid database engine: %{db_engine}.'
+        no_new_postgresql: 'PostgreSQL is not allowed for new deployments.'
         unknown_mode_ha: 'Unknown mode for HA storage: %{storage_mode}.'
         no_device: 'No device specified for shared storage.'
         no_filesystem: 'No filesystem type specified for shared storage.'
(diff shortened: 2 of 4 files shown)


crowbar_framework/app/models/database_service.rb
```diff
@@ -120,6 +120,19 @@ def validate_proposal_after_save(proposal)
...
+    # for new deployments, only allow mysql
+    case attributes["sql_engine"]
+    when "postgresql"
+      proposal_id = proposal["id"].gsub("#{@bc_name}-", "")
+      if RoleObject.find_role_by_name("#{@bc_name}-config-#{proposal_id}").nil?
+        validation_error I18n.t(
+          "barclamp.#{@bc_name}.validation.no_new_postgresql"
+        )
+      end
+    when "mysql"
```
Itxaka: should this just be removed? actually not even a case is needed here just a `if attributes["sql_engine"] == "postgresql"`

aspiers: Well not a big deal but I think it's kind of helpful to list the other possible value, so personally I'd keep it.

crowbar_framework/config/locales/database/en.yml
```diff
@@ -56,6 +56,7 @@ en:
               options: 'Mount Options'
       validation:
         invalid_db_engine: 'Invalid database engine: %{db_engine}.'
+        no_new_postgresql: 'PostgreSQL is not allowed for new deployments.'
```
aspiers: A message as terse as this is inevitably going to cause shock / confusion and a flurry of questions from people who didn't read the release notes.  So let's pre-empt that and avoid wasting a load of our own time by following the [Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and making the message tell them to read the release notes :-)

crowbar_framework/config/locales/rabbitmq/en.yml
```diff
@@ -49,6 +49,7 @@ en:
               fstype: 'Filesystem Type'
               options: 'Mount Options'
       validation:
+        no_new_unclustered: 'HA without RabbitMQ clustering is not allowed for new deployments.'
```
aspiers: Same comment here - make it easy / obvious for operators to find out *why* it's not allowed.

crowbar_framework/app/models/database_service.rb
vuntz: Hardly believable, I know, but I did it exactly for the reason @aspiers wrote down here :-)

crowbar_framework/config/locales/database/en.yml
vuntz: @aspiers as mentioned in https://github.com/crowbar/crowbar-ha/pull/301, that would be a downstream patch.

(also, there would not be too much confusion as only people editing the raw json would hit this)

aspiers: > @aspiers as mentioned in crowbar/crowbar-ha#301, that would be a downstream patch.

s/would be/will be/ and I'll be happy - let's make this a reality rather than a theoretical possibility ;-)

> (also, there would not be too much confusion as only people editing the raw json would hit this)

Nope, they could also achieve it by reusing old batch yaml files, and I suspect this is a very likely scenario (admittedly more likely with our own SEs etc. than with customers, but that would still generate support queries we have to deal with).

aspiers:
As with https://github.com/crowbar/crowbar-ha/pull/301, I'll approve if it's agreed that the downstream patch will be made :-)

vuntz:
> As with crowbar/crowbar-ha#301, I'll approve if it's agreed that the downstream patch will be made :-)

I'm happy to commit to that :-)

vuntz:
Rebased

rsalevsky:
@aspiers Can you please re-review this PR?

PR #1550, opened 2018-02-02
neutron: enable configuring L3 HA

AbelNavarro:
Enable configuring L3 HA on Neutron. This option is still not supported in our product but some customers my want to enable it. The options we allow to configure are the minimum desirable to configure this functionality.

Using L3 HA is not compatible with running neutron-ha-tool. When L3 HA is enabled neutron-ha-tool service will be disabled.

There's no conversion fron Non-HA to HA routers. That step has to be done manually by the user.


chef/data_bags/crowbar/template-neutron.schema
```diff
@@ -24,6 +24,11 @@
                       "agent_boot_time": { "type" : "int", "required" : true }
                     }},
                     "use_dvr": { "type": "bool", "required": true },
+                    "use_l3_ha": { "type": "bool", "required": true },
+                    "l3_ha_password": { "type": "str", "required": true },
+                    "max_l3_agents_per_router": { "type": "int", "required": true },
+                    "ha_vrrp_advert_int": { "type": "int", "required": true },
+                    "ha_vrrp_health_check_interval": { "type": "int", "required": true },
```
ilausuch: I am thinking that if all these parameters are related and doesn't work isolated, could be inside a l3_ha structure. For instance:

l3_ha:{"type": "map", "required": true, "mapping": {
                     "enable": { "type": "bool", "required": true },
                     "password": { "type": "str", "required": true },
                     "max_l3_agents_per_router": { "type": "int", "required": true },
                     "ha_vrrp_advert_int": { "type": "int", "required": true },
                     "ha_vrrp_health_check_interval": { "type": "int", "required": true }
}}

AbelNavarro:
@nicolasbock amended commit messages to make it more descriptive.

@ilausuch although your proposal looks reasonable, we are not using such grouping in this or other schema. For now I'd prefer to keep it as it.

AbelNavarro:
[L3 HA](https://trello.com/c/OJyN9gjj/155-l3-ha)

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -19,6 +19,11 @@
         "agent_boot_time": 180
       },
       "use_dvr": false,
+      "use_l3_ha": false,
+      "l3_ha_password": "",
+      "max_l3_agents_per_router": 3,
+      "ha_vrrp_advert_int": 2,
+      "ha_vrrp_health_check_interval": 0,
```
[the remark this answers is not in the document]
AbelNavarro: right, these parameters are from upstream. Change `ha_vrrp_health_check_interval` as recommended in the config file.

chef/data_bags/crowbar/template-neutron.schema
AbelNavarro: mkay, I put all these parameters inside the `l3_ha` group.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -258,28 +258,39 @@
...
+    pacemaker_primitive ha_service_primitive_name do
+      agent "systemd:neutron-l3-ha-service"
+      op node[:neutron][:ha][:neutron_l3_ha_resource][:op]
+      action [:stop, :delete]
+      only_if "crm configure show #{ha_service_primitive_name}"
+      only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+    end
   end
 
   rabbit_settings = fetch_rabbitmq_settings
```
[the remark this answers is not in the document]
AbelNavarro: Right, will move them inside the `if`.

AbelNavarro:
@vuntz @nicolasbock @ilausuch comments have been addressed. Build is green. Mind another review?

AbelNavarro:
A bit more detailed explanation about why we don't need to configure `keepalived`: it is configured by the L3 agent dynamically for every network. If we look at a `ps` output from a network node we can see this:

```
keepalived -P -f /var/lib/neutron/ha_confs/d9705638-5f14-45d4-9487-e2dc04f4b892/keepalived.conf -p /var/lib/neutron/ha_confs/d9705638-5f14-45d4-9487-e2dc04f4b892.pid -r /var/lib/neutron/ha_confs/d9705638-5f14-45d4-9487-e2dc04f4b892.pid-vrrp
```

The config file doesn't live at `/etc`  but is stored separately per network

AbelNavarro:
rebased.

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -19,6 +19,13 @@
         "agent_boot_time": 180
       },
       "use_dvr": false,
+      "l3_ha": {
+        "enable": false,
```
vuntz: s/enable/enabled/

This is our usual convention, and you actually use `enabled` in chef/cookbooks/neutron/recipes/common_agent.rb ;-)

AbelNavarro: deal

vuntz:
Quick comment (without having reviewed the whole patch again): it'd be nice to add the setting in the webui. Could be a followup PR if preferred, though.

chef/data_bags/crowbar/migrate/neutron/207_add_use_l3_ha.rb
```diff
@@ -0,0 +1,41 @@
...
+    a["l3_ha"]["ha_vrrp_advert_int"] = ta["l3_ha"]["ha_vrrp_advert_int"]
+  end
+
+  unless a["l3_ha"].key?("ha_vrrp_health_check_interval")
+    a["l3_ha"]["ha_vrrp_health_check_interval"] = ta["l3_ha"]["ha_vrrp_health_check_interval"]
+  end
+
+  unless a["l3_ha"].key?("max_l3_agents_per_router")
+    a["l3_ha"]["max_l3_agents_per_router"] = ta["l3_ha"]["max_l3_agents_per_router"]
+  end
```
vuntz: No big deal, but you could also write it like this:

```ruby
["enabled", "password", "ha_vrrp_advert_int",
 "ha_vrrp_health_check_interval", "max_l3_agents_per_router"].each do |attribute|
  a["l3_ha"][attribute] = ta["l3_ha"][attribute] unless a["l3_ha"].key?(attribute)
end
```

chef/data_bags/crowbar/migrate/neutron/207_add_use_l3_ha.rb
```diff
@@ -0,0 +1,41 @@
...
+        a["l3_ha"].delete("ha_vrrp_health_check_interval")
+      end
+
+      unless ta["l3_ha"].key?("max_l3_agents_per_router")
+        a["l3_ha"].delete("max_l3_agents_per_router")
+      end
+
+      a.delete("l3_ha")
+    end
+  end
```
vuntz: Same comment here about the loop that could be used.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -258,45 +258,58 @@
...
+      only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+    end
+
+  else
+    pacemaker_primitive ha_service_primitive_name do
+      agent "systemd:neutron-l3-ha-service"
+      op node[:neutron][:ha][:neutron_l3_ha_resource][:op]
+      action [:stop, :delete]
+      only_if "crm configure show #{ha_service_primitive_name}"
+      only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
vuntz: I'd change the order of these two lines: that will ensure that will only run the `crm configure show` on the founder node.

vuntz:
Oh, and don't we need to install the keepalived package on the nodes where neutron-l3-agent runs?

vuntz:
And another question: do the compute nodes (in DVR mode) also need the keepalived package (and therefore the vrrp config bits)? If no, we should make sure to not set these config bits on compute nodes (especially the password)

chef/data_bags/crowbar/migrate/neutron/207_add_use_l3_ha.rb
AbelNavarro: agree, you way looks much cleaner.

chef/data_bags/crowbar/migrate/neutron/207_add_use_l3_ha.rb
AbelNavarro: again, will fix it too.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
AbelNavarro: agree with that.

chef/data_bags/crowbar/migrate/neutron/210_add_use_l3_ha.rb
```diff
@@ -0,0 +1,24 @@
+def upgrade(ta, td, a, d)
+  a["l3_ha"] ||= {}
+  ["enabled", "password", "ha_vrrp_advert_int",
```
vuntz: Sorry, I failed to notice this earlier: for the password attribute, you need to actually set it to some real value.

This can be done like this:

```ruby
  # We use a class variable to set the same password in the proposal and in the
  # role, though
  unless defined?(@@neutron_l3_ha_password)
    service = ServiceObject.new "fake-logger"
    @@neutron_l3_ha_password = service.random_password
  end
  a["l3_ha"]["password"] = @@neutron_l3_ha_password unless a["l3_ha"].key?("password")
end
```

AbelNavarro: thanks for noticing this, I'll implement it like that.

AbelNavarro:
@vuntz yes, the `keepalived` install is still missing, but not forgotten.

AbelNavarro:
@vuntz about the question you asked back then about DVR+HA if `keepalived` needed to be installed in the computes as well the answer will be no. Only network nodes run the SNAT namespaces which are the ones monitored and balanced. Compute nodes don't run SNAT namespaces:

https://docs.openstack.org/ocata/networking-guide/config-dvr-ha-snat.html

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -338,6 +338,10 @@
         periodic_interval: 40,
         periodic_fuzzy_delay: 5,
         dvr_enabled: neutron[:neutron][:use_dvr],
+        l3_ha: neutron[:neutron][:l3_ha][:enabled],
```
vuntz: @AbelNavarro based on your comment, I think that this variable should actually be set to `false` if the code is not running on a network node. We have some test here and there for this (using `node.roles.include?("neutron-network")`)

chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -28,6 +28,10 @@
   end
 end
 
+if node[:neutron][:ha][:network][:enabled]
```
vuntz: Didn't you mean `node[:neutron][:l3_ha][:enabled]`?

AbelNavarro: right, I did confuse it with the network agents HA variable. Fixing that.

chef/cookbooks/neutron/recipes/common_agent.rb
vuntz: For the record, it does harm because it adds a password on a node where it's not needed :-)

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -338,6 +338,10 @@
         periodic_interval: 40,
         periodic_fuzzy_delay: 5,
         dvr_enabled: neutron[:neutron][:use_dvr],
+        l3_ha: node.roles.include?("neutron-network") ? neutron[:neutron][:l3_ha][:enabled] : false,
```
vuntz: Based on our discussion, the test should be:

`node.roles.include?("neutron-network") && node[:neutron][:ha][:network][:enabled] && node[:neutron][:l3_ha][:enabled]`

chef/data_bags/crowbar/migrate/neutron/210_add_use_l3_ha.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  unless ta.key?("l3_ha")
+    if a.key("l3_ha")
+      ["enabled", "password", "ha_vrrp_advert_int",
+       "ha_vrrp_health_check_interval", "max_l3_agents_per_router"].each do |attribute|
+         a["l3_ha"].delete(attribute) unless ta["l3_ha"].key?(attribute)
```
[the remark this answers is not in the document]
vuntz: Can you fix this hound comment? I think the other ones (that I see) are fine to ignore.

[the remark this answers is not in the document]
AbelNavarro: sure thing, I hope this is the one blocking the PR.

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -142,6 +142,8 @@
       service_plugins: service_plugins,
       allow_overlapping_ips: neutron[:neutron][:allow_overlapping_ips],
       dvr_enabled: neutron[:neutron][:use_dvr],
+      l3_ha_enabled: neutron[:neutron][:l3_ha][:enabled],
```
vuntz: Don't we need the same check on HA being actually used for network nodes?

chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -28,6 +28,10 @@
   end
 end
 
+if node[:neutron][:l3_ha][:enabled]
```
vuntz: Not that it hurts, but `node[:neutron][:ha][:network][:enabled]` could be added here. (Feel free to ignore)

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -333,6 +334,12 @@ def validate_dvr(proposal)
     end
   end
 
+  def validate_l3ha(proposal)
+    if proposal["attributes"]["neutron"]["l3_ha"]["enabled"]
+      validate_at_least_n_for_role proposal, "neutron-network", 2
```
vuntz: I don't think that will work, unfortunately: when you have a cluster, it's probably perceived as "one" node for this check. I guess what you need to check is that you have a cluster assigned to the role.

AbelNavarro: I didn't trust you and had to check myself. You were completely right, cluster is considered a single item. This was confusing since the function comments mentioned it counted nodes, not clusters.

Anyway, I'm not finding a nice way to filter this, so probably will just look for the `cluster` string in the `elements` section. Sample:

```
elements: {"neutron-server"=>["cluster:data"], "neutron-network"=>["cluster:data"]}
```

AbelNavarro: I think the best way to deal with it is to add a new validator for this situation. I created a PR for that: https://github.com/crowbar/crowbar-core/pull/1546

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -333,6 +334,13 @@ def validate_dvr(proposal)
     end
   end
 
+  def validate_l3ha(proposal)
+    if proposal["attributes"]["neutron"]["l3_ha"]["enabled"]
+      network_elements, network_nodes, network_ha_enabled = role_expand_elements(role, "neutron-network")
```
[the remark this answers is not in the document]
AbelNavarro: with this new validator we don't need anymore the role expansion, so removing it completely.

chef/cookbooks/neutron/recipes/common_config.rb
vuntz: As mentioned, I think you need to look at `neutron[:neutron][:ha][:network][:enabled]` too here.

AbelNavarro: i have to agree here.

chef/cookbooks/neutron/recipes/network_agents.rb
AbelNavarro: will make you happy here, adding that extra check.

chef/data_bags/crowbar/migrate/neutron/210_add_use_l3_ha.rb
```diff
@@ -0,0 +1,30 @@
+def upgrade(ta, td, a, d)
```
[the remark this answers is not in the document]
AbelNavarro: skipping this.

chef/data_bags/crowbar/migrate/neutron/210_add_use_l3_ha.rb
```diff
@@ -0,0 +1,30 @@
...
+  a["l3_ha"] ||= {}
+  ["enabled", "ha_vrrp_advert_int", "ha_vrrp_health_check_interval",
+   "max_l3_agents_per_router"].each do |attribute|
+     a["l3_ha"][attribute] = ta["l3_ha"][attribute] unless a["l3_ha"].key?(attribute)
+  end
+
+  unless defined?(@@neutron_l3_ha_password)
+    service = ServiceObject.new "fake-logger"
+    @@neutron_l3_ha_password = service.random_password
```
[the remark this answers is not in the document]
AbelNavarro: this `@@` is being used the same way for `neutron_db_password`, so I guess for simplicity it is fine here.

chef/data_bags/crowbar/migrate/neutron/210_add_use_l3_ha.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+  a["l3_ha"]["password"] = @@neutron_l3_ha_password unless a["l3_ha"].key?("password")
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
```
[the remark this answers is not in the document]
AbelNavarro: skipping this.

chef/data_bags/crowbar/migrate/neutron/210_add_use_l3_ha.rb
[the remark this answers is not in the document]
AbelNavarro: I think this was already fixed, let's see what hound says again.

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -333,6 +334,12 @@ def validate_dvr(proposal)
     end
   end
 
+  def validate_l3ha(proposal)
+    if proposal["attributes"]["neutron"]["l3_ha"]["enabled"]
```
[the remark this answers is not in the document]
AbelNavarro: the way it is written is easier to read and matches other clauses style in the same file.

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
```diff
@@ -6,6 +6,12 @@ agent_mode = <%= @dvr_mode %>
 <% if @force_metadata -%>
 enable_metadata_proxy = False
 <% end -%>
+<% if @l3_ha -%>
+external_network_bridge =
+ha_vrrp_auth_password = <%= @l3_ha_password %>
+ha_vrrp_advert_int = <%= @ha_vrrp_advert_int %>
+ha_vrrp_health_check_interval = <%= @ha_vrrp_health_check_interval %>
```
dirkmueller: do we really want to enable the vrrp_health checking? I think this causes some issues. I would remove this part

AbelNavarro: the time I created this PR the proposed change was mimicking our customer's config. I don't know if that changed recently, but if so I can remove that part.

AbelNavarro: Afaik, neutron needs Keepalived for the L3 HA implementation, that means VRRP should be running on those routers.

I removed though the VRRP parameters because our customer's current implementation has those as default.

aspiers: Yeah that's my understanding that keepalived+VRRP is a fundamental part of the architecture.  I can't imagine how it would work without this.

AbelNavarro:
After the rebase we were still using `l3_agent_clone` which was removed at https://github.com/crowbar/crowbar-openstack/pull/1926 . I guess rebase didn't take this right.

chef/data_bags/crowbar/migrate/neutron/206_add_use_l3_ha.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  a["use_l3_ha"] = ta["use_l3_ha"] unless a.key? "use_l3_ha"
+  a["l3_ha_password"] = ta["l3_ha_password"] unless a.key? "l3_ha_password"
```
dirkmueller: we need to initialize it with a random password not empty

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
dirkmueller: @AbelNavarro our customer config is not using *health* check. VRRP is integral, but the health check is optional, and it seems to be net benefit negative. I wouldn't set it. 

To make it explitcit, I am talking about *not* setting ha_vrrp_health_check_interval*. you're correct that VRRP itself needs to be configured as it is integral part of the solution.

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -20,6 +20,11 @@
         "agent_boot_time": 180
       },
       "use_dvr": false,
+      "l3_ha": {
+        "enabled": false,
```
dirkmueller: this needs to be a tristate: enabled, disabled, and "pacemake" or "neutron-ha-tool". so that we have the ability to do scale up without any ha failover.

chef/data_bags/crowbar/migrate/neutron/206_add_use_l3_ha.rb
AbelNavarro: The current file already creates a random password if none was defined (`304_add_use_l3_ha.rb`)

chef/data_bags/crowbar/template-neutron.json
AbelNavarro: if `l3_ha` is not enabled, then defaults to `neutron-ha-tool` (https://github.com/crowbar/crowbar-openstack/pull/1550/files#diff-3327575d1764c639cecb592c09c5694cR138). And `neutron-ha-tool` is always there in our case. I don't think we allow it to be disabled at all. That would change lots of things for the user and may have further implications, even in documentation.

About the `pacemaker` option I don't fully understand this? is it about neutron-ha-tool? `l3_ha` uses keepalived and doesn't need pacemaker at all.

aspiers: [**@dirkmueller**](https://github.com/dirkmueller) [commented](https://github.com/crowbar/crowbar-openstack/pull/1550/files#r248964117 "2019-01-18T08:48:49Z - Replied by Github Reply Comments"):
> this needs to be a tristate: enabled, disabled, and "pacemake" or "neutron-ha-tool".

That's quad-state ;-)

[**@AbelNavarro**](https://github.com/AbelNavarro) [commented](https://github.com/crowbar/crowbar-openstack/pull/1550/files#r249109475 "2019-01-18T16:41:45Z - Replied by Github Reply Comments"):
> if `l3_ha` is not enabled, then defaults to `neutron-ha-tool` (https://github.com/crowbar/crowbar-openstack/pull/1550/files#diff-3327575d1764c639cecb592c09c5694cR138).

That makes sense.

> And `neutron-ha-tool` is always there in our case.

Yes `neutron-ha-tool` is always there (in case it's useful for manual maintenance) but the `neutron-ha-tool` *service* which handles *automatic* failover is a different thing.  We need to be careful to avoid ambiguous terminology in this discussion.

> About the `pacemaker` option I don't fully understand this? is it about neutron-ha-tool? `l3_ha` uses keepalived and doesn't need pacemaker at all.

Yeah I think we only need 3 options:

1. no HA (which should happen if the Chef role isn't assigned to a cluster)
2. Pacemaker + neutron-ha-tool service
3. upstream HA implementation (keepalived+VRRP, Pacemaker only for fencing as per column E of [my spreadsheet](https://ethercalc.openstack.org/Pike-Neutron-L3-HA))

chef/data_bags/crowbar/migrate/neutron/304_add_use_l3_ha.rb
```diff
@@ -0,0 +1,30 @@
...
+  unless defined?(@@neutron_l3_ha_password)
+    service = ServiceObject.new "fake-logger"
+    @@neutron_l3_ha_password = service.random_password
+  end
+  a["l3_ha"]["password"] = @@neutron_l3_ha_password unless a["l3_ha"].key?("password")
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
```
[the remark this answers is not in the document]
Itxaka: nooo, fix them please!

So much nicer to read `template_attributes, template_deployment, attributes, deployment` than `ta, td, a, d` ¬_¬

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -135,28 +135,56 @@
 
 if use_l3_agent
   # Add pacemaker resource for neutron-l3-ha-service
-  ha_service_transaction_objects = []
+  # only if l3_ha is not enabled
   ha_service_primitive_name = "neutron-l3-ha-service"
+  if !node[:neutron][:use_l3_ha]
```
dirkmueller: how#s that better than ```unless``` ? maybe just change the order of the alternatives?

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -20,6 +20,11 @@
...
+        "password": "",
+        "max_l3_agents_per_router": 3
```
dirkmueller: why does that need to be hardcoded? can we use automatic determination based on the cluster size?

chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -29,6 +29,10 @@
   end
 end
 
+if node[:neutron][:l3_ha][:enabled] && node[:neutron][:l3_ha][:enabled]
```
dirkmueller: eh, what?

chef/cookbooks/neutron/recipes/common_agent.rb
```diff
@@ -335,6 +335,10 @@
         periodic_interval: 40,
         periodic_fuzzy_delay: 5,
         dvr_enabled: neutron[:neutron][:use_dvr],
+        l3_ha: node.roles.include?("neutron-network") &&
+               node[:neutron][:ha][:network][:enabled] &&
+               node[:neutron][:l3_ha][:enabled],
+        l3_ha_password: neutron[:neutron][:l3_ha][:password],
```
dirkmueller: where is this setting being used?

chef/cookbooks/neutron/recipes/network_agents_ha.rb
AbelNavarro: old habits never die. Updating.

chef/cookbooks/neutron/recipes/network_agents.rb
AbelNavarro: at some point in the past one of those conditions was `node[:neutron][:ha][:network][:enabled]`. Later that was considered unnecessary and I probably replaced it with those two identical conditions. Will leave just one.

chef/cookbooks/neutron/recipes/common_agent.rb
AbelNavarro: It was originally used in `l3_agent.ini.erb`:

```ha_vrrp_auth_password = <%= @l3_ha_password %>```

I removed VRRP configuration bits from that file (`ha_vrrp_advert_int` and `ha_vrrp_health_check_interval`) because customer was using defaults. Probably I went to far and removed the password field as well. That one may be still necessary. I have to take a deeper look.

chef/data_bags/crowbar/template-neutron.json
AbelNavarro: The `max_l3_agents_per_router` determine the number of backup DVR/SNAT routers which will be instantiated [1]. The default value is 3 (one master, two standby). I haven't seen any documentation or proposals with other recommendations for this number. Even if we had 20 network nodes, three l3_agents look a reasonable number. That number is configurable in the barclamp as the user may find convenient. I don't think making an automatic calculation for this number would add much benefits so I advocate to keep this default.

[1] https://docs.openstack.org/mitaka/networking-guide/config-dvr-ha-snat.html

AbelNavarro:
Closing in favor of https://github.com/crowbar/crowbar-openstack/pull/2005

PR #1552, opened 2018-02-07
[POC] rabbit: use clones and manual clustering 

Itxaka:
make the pacemaker resources simple clones and manage ourselves the clustering


vuntz:
I think you should not use any of the OCF resource agents in that case, but simply the systemd service. Or is there some special thing the OCF RA can help with still?

Itxaka:
> I think you should not use any of the OCF resource agents in that case, but simply the systemd service. Or is there some special thing the OCF RA can help with still?

mainly for the health checks I guess, in case of a stuck/non-working/hanging node

I will test this today manually and resend the patch

Itxaka:
not working

PR #1554, opened 2018-02-09

ilausuch:
This script blocks the connection to the rabbitmq cluster in case the number of nodes decay bellow the half of the total. In this case the remain rabbit nodes won't accept new connections. The problem to solve is that when a rabbit node has too much clients with low timeout, the final scenario could be a DOS when the clients try to constantly reconnect to rabbitmq and forcing more timeouts.

It takes advantage of the pacemaker notifications that notify when a rabbitmq has failed or has restored. All nodes rabbitmq ports will be blocked if the total number of alive nodes are below to the half of nodes of the cluster, or unblock if its over this value. This prevents the overload of alive nodes and the total block of service.

Only the node that have problems will receive the alert, so it is necessary to notify other nodes to block their ports if necessary. The only case when this script won't work is when the node fails completely and is not able to notify the others.

The script is divided in two parts: On one hand, the alert handler that manages the alerts discarding non interesting alerts, launches to the blocker script in this node and other cluster nodes (via SSH). And in the other hand, the blocker script than checks the condition of running nodes and blocks or unblocks the rabbitmq client port.

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
+++ b/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
@@ -120,3 +120,71 @@ ruby_block "wait for #{ms_name} to be started" do
     end
   end # block
 end # ruby_block
+
+if CrowbarPacemakerHelper.cluster_nodes(node).size > 2
+  # create the directory to lock rabbitmq-port-blocker
+  cookbook_file "/etc/tmpfiles.d/rabbitmq.conf" do
+    owner "root"
+    group "root"
+    mode "0644"
+    action :create
+    source "rabbitmq.tmpfiles"
+  end
+
+  bash "create tmpfiles.d files for rabbitmq" do
+    code "systemd-tmpfiles --create /etc/tmpfiles.d/rabbitmq.conf"
+    action :nothing
+    subscribes :run, resources("cookbook_file[/etc/tmpfiles.d/rabbitmq.conf]"), :immediately
+  end
+
+  # create the scripts to block the client port on startup
+  template "/usr/bin/rabbitmq-alert-handler.sh" do
+    source "rabbitmq-alert-handler.erb"
+    owner "root"
+    group "root"
+    mode "0755"
+    variables(node: node, nodes: CrowbarPacemakerHelper.cluster_nodes(node))
+  end
+
+  template "/usr/bin/rabbitmq-port-blocker.sh" do
+    source "rabbitmq-port-blocker.erb"
+    owner "root"
+    group "root"
+    mode "0755"
+    variables(total_nodes: CrowbarPacemakerHelper.cluster_nodes(node).size)
+  end
+
+  template "/etc/sudoers.d/rabbitmq-port-blocker" do
+    source "hacluster_sudoers.erb"
+    owner "root"
+    group "root"
+    mode "0440"
+  end
+
+  # create the alert
+  pacemaker_alert "rabbitmq-alert-handler" do
+    handler "/usr/bin/rabbitmq-alert-handler.sh"
+    action :create
+  end
+else
+  pacemaker_alert "rabbitmq-alert-handler" do
+    handler "/usr/bin/rabbitmq-alert-handler.sh"
+    action :delete
+  end
+
+  cookbook_file "/etc/tmpfiles.d/rabbitmq.conf" do
+    action :delete
+  end
+
+  file "/usr/bin/rabbitmq-alert-handler.sh" do
+    action :delete
+  end
+
+  file "/usr/bin/rabbitmq-port-blocker.sh" do
+    action :delete
+  end
+
+  file "/etc/sudoers.d/rabbitmq-port-blocker" do
+    action :delete
+  end
+end
--- a/chef/cookbooks/rabbitmq/templates/default/rabbitmq-alert-handler.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/rabbitmq-alert-handler.erb
@@ -1,0 +1,19 @@
+#!/bin/sh
+if [ -z "$CRM_alert_version" ]; then
+  echo "$0 must be run by Pacemaker version 1.1.15 or later"
+  exit 0
+fi
+
+# exit if isn't a rabbitmq alert or is not a monitor task
+[ "$CRM_alert_kind" = "resource" -a "$CRM_alert_rsc" = "rabbitmq" -a "$CRM_alert_task" = "monitor" ] || exit 0
+
+# launch the blocker in exclusive mode
+nohup sudo flock /var/lock/rabbit /usr/bin/rabbitmq-port-blocker.sh &
+
+# for each node in the cluster
+<% @nodes.each do |cluster_node| -%>
+  # unless this is the current node launch remotely the blocker in exclusive mode
+  <% unless cluster_node.name==@node.name -%>
+    nohup /usr/bin/timeout 15 sudo /usr/bin/ssh -o TCPKeepAlive=no -o ServerAliveInterval=15 root@<%= cluster_node.name %> nohup flock /var/lock/rabbit /usr/bin/rabbitmq-port-blocker.sh &
+  <% end -%>
+<% end -%>
--- a/chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
@@ -1,0 +1,31 @@
+#!/bin/sh
+
+# calcules the blocking level applying the formula
+total_nodes=<%= @total_nodes %>
+blocking_level=$(expr $total_nodes / 2)
+comment_text="rabbitmq port blocker (no quorum)"
+
+# get the number of running nodes of rabbitmq in the current cluster
+function running_nodes()
+{
+  rabbitmqctl cluster_status 2>/dev/null | tr -d "\n" | sed -e 's/running_nodes,/\nrunning_nodes/g'| grep running_nodes | cut -d "[" -f2 | cut -d "]" -f1 | tr "," "\n" | wc -l
+}
+
+# check if exists the blocking rule for rabbitmq clients
+function check_rule()
+{
+  iptables -L -n | grep -F "tcp dpt:5672 /* $comment_text */" | grep DROP | wc -l
+}
+
+# if the running nodes is les that the blocking level, then...
+if [ $(running_nodes) -le $blocking_level ]; then
+  # if rule not exists the rule will be added to block the clients port
+  if [ $(check_rule) -eq 0 ]; then
+    iptables -A INPUT -p tcp --destination-port 5672 -m comment --comment "$comment_text" -j DROP
+  fi
+else
+  # finally if the rule exists it will be deleted. If there are more than one, will remove all
+  if [[ $(check_rule) -gt 0 ]]; then
+    iptables -D INPUT -p tcp --destination-port 5672 -m comment --comment "$comment_text" -j DROP
+  fi
+fi
--- a/chef/cookbooks/rabbitmq/files/default/rabbitmq.tmpfiles
+++ b/chef/cookbooks/rabbitmq/files/default/rabbitmq.tmpfiles
@@ -1,0 +1,1 @@
+d /var/lock/rabbit  0755  root  root  -
--- a/chef/cookbooks/rabbitmq/templates/default/hacluster_sudoers.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/hacluster_sudoers.erb
@@ -1,0 +1,4 @@
+Defaults:hacluster !requiretty
+
+hacluster ALL = (root) NOPASSWD: /usr/bin/ssh
+hacluster ALL = (root) NOPASSWD: /usr/bin/flock


chef/data_bags/crowbar/template-rabbitmq.schema
```diff
@@ -63,7 +63,8 @@
                         "options": { "type": "str", "required": true }
                       }
                     }
-                  }
+                  },
+                  "block_port_on_startup": { "type" : "bool", "required" : false }
```
bmwiedemann: I think you need to add that one line below. Test failed with
```
crowbar/template-rabbitmq.schema: 67:19 [/mapping/attributes/mapping/rabbitmq/mapping/ha/mapping/storage/block_port_on_startup] key 'block_port_on_startup:' is undefined.
```

ilausuch: Yes! thanks, a copy paste mistake

ilausuch:
Now I am checking if there is a way to solve the problem without using ssh

ilausuch:
All test I did was passed. Ready for review

bmwiedemann:
commit message needs some fixes:

> To prevent the problem of overhead the running nodes and prevent an execive

excessive ?
what is 'the problem of overhead the running nodes'?

> movement of queues when the number of running rabbitmq nodes in a cluster
> is inferior to a critical level

inferior to => below?

> When pacemaker alerts to a node that the service of rabbitmq has failed
> it blocks the rabbit client port if the number of running nodes in cluster

in **the** cluster

> is less the a critical level and notify to others to do the same.

less the**n**

> When the number of running nodes is superior to the critical level all nodes
> will unblock the client port.

superior to => above ?

ilausuch:
Thanks @bmwiedemann. I refactored the description

bmwiedemann:
you only updated the text in the PR, but not the commit itself, which will be easier to find later because it is directly in the git history. can you update the commit message, too?

bmwiedemann:
It does not apply cleanly on the package - needs rebase?
```
patching file chef/data_bags/crowbar/template-rabbitmq.json
Hunk #2 FAILED at 58.
```

chef/data_bags/crowbar/migrate/rabbitmq/204_ha_block_client_port_on_startup.rb
```diff
@@ -0,0 +1,10 @@
+def upgrade(ta, td, a, d)
+  a["ha"]["block_port_on_startup"] = ta["ha"]["block_port_on_startup"] \
+    unless a["ha"]["block_port_on_startup"]
```
bmwiedemann: is this the same as
```ruby
a["ha"]["block_port_on_startup"] ||= ta["ha"]["block_port_on_startup"]
```

IMHO that would be easier to read and have less duplication

ilausuch: Oh! thanks! I'll change it

ilausuch:

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -120,3 +120,49 @@
...
+  mode "0755"
+  variables(total_nodes: CrowbarPacemakerHelper.cluster_nodes(node).size)
+end
+
+template "/etc/sudoers.d/rabbitmq-port-blocker" do
+  source "hacluster_sudoers.erb"
+  owner "root"
+  group "root"
+  mode "0440"
+end
```
vuntz: Assuming we keep this optional, I would only do all of the above if the option is enabled.

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -120,3 +120,49 @@
     end
   end # block
 end # ruby_block
+
+# removes all previous rules
+execute "remove_all_previous_drop_rules" do
+  command <<-RUBY
+    rule="-A INPUT -p tcp -m tcp --dport 5672 -j DROP"
+    iptables-save | grep -v -e "$(rule)" | iptables-restore
```
vuntz: I'm slightly confused about this: why is it needed? Does that mean that the periodic chef-client (the one that runs every 15 minuets) will break the feature by dropping the rule that the script adds?

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -120,3 +120,49 @@
...
+  RUBY
+end
+
+# create the directory to lock rabbitmq-port-blocker
+directory "/var/lock/rabbitmq" do
```
vuntz: I think you need to create a systemd tmpfile (in /etc/tmpfiles.d) for this (and actually use `/run/lock`), as this will get dropped automatically on reboot otherwise.

chef/cookbooks/rabbitmq/templates/default/rabbitmq-alert-handler.erb
```diff
@@ -0,0 +1,19 @@
...
+if [ -z "$CRM_alert_version" ]; then
+  echo "$0 must be run by Pacemaker version 1.1.15 or later"
+  exit 0
+fi
+
+# exit if isn't a rabbitmq alert or is not a monitor task
+[ "$CRM_alert_kind" = "resource" -a "$CRM_alert_rsc" = "rabbitmq" -a "$CRM_alert_task" = "monitor" ] || exit 0
+
+# launch the blocker in exclusive mode
+sudo flock /var/lock/rabbitmq/rabbitmq-port-blocker.lock /usr/bin/rabbitmq-port-blocker.sh &
```
vuntz: Any reason to require sudo? Why not make the directory owned by the hacluster user? (I guess we'd need to rename the directory to be slightly less generic, because I would expect that the lock directory could be used by the rabbitmq user)

vuntz: Ah, the ssh calls below connects as root. This could be a good reason.

chef/cookbooks/rabbitmq/templates/default/rabbitmq-alert-handler.erb
```diff
@@ -0,0 +1,19 @@
...
+
+# for each node in the cluster
+<% @nodes.each do |cluster_node| -%>
+  # unless this is the current node launch remotelly the blocker in exclusive mode
+  <% unless cluster_node.name==@node.name -%>
+    sudo /usr/bin/ssh root@<%= cluster_node.name %> flock /var/lock/rabbitmq/rabbitmq-port-blocker.lock /usr/bin/rabbitmq-port-blocker.sh &
```
vuntz: I'm slightly concerned about what's happening when a node is down here. Does it make sense to specify a timeout? For instance, crowbar usually has `-o TCPKeepAlive=no -o ServerAliveInterval=15` for this reason.

Also do we possibly need to use `nohup`?

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
+# get the number of running nodes of rabbitmq in the current cluster
+function running_nodes()
+{
+  local running_nodes=$(rabbitmqctl cluster_status 2>/dev/null | grep running_nodes | cut -d "," -f2- | cut -d "}" -f1 | tr -d "[" | tr -d "]" | tr -d "'" | tr "," "\n" | wc -l)
+  echo $running_nodes
```
vuntz: You don't need a variable here, just run the command with all the pipes :-)

vuntz: Btw, I'm assuming the command will work fine even when a cluster has 5 nodes.

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
```diff
@@ -0,0 +1,36 @@
...
+  echo $exists_rule
+}
+
+# calcules the blocking level applying the formula
+total_nodes=<%= @total_nodes %>
+blocking_level=$(expr $total_nodes / 2)
+
+# if the running nodes is les that the blocking level, then...
+if [ $(running_nodes) -le $blocking_level ]; then
+
```
vuntz: I don't think you need to put an empty line here (and after the `else`)

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
```diff
@@ -0,0 +1,36 @@
...
+  # if rule not exists the rule will be added to block the clients port
+  if [ $(check_rule) -eq 0 ]; then
+    iptables -A INPUT -p tcp --destination-port 5672 -j DROP
+    exists_rule=1
```
vuntz: This variable doesn't exist in this scope, and it's not used later anyway, so you can drop that line.

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
```diff
@@ -0,0 +1,36 @@
...
+  fi
+else
+
+  # finally if the rule exists it will be deleted. If there are more than one, will remove all
+  if [[ $(check_rule) -gt 0 ]]; then
+    iptables-save | grep -v -e "-A INPUT -p tcp -m tcp --dport 5672 -j DROP" | iptables-restore
```
vuntz: Why not do `iptables -D INPUT -p tcp --destination-port 5672 -j DROP`?

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
Itxaka: fixed!

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
Itxaka: Im guessing this is a protection against leftover rules, in case something fails and the block rule is left this will remove it, but it could also mean that it removes the rule while its active :/

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
Itxaka: done

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
Itxaka: done and will check for 5 nodes

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
Itxaka: dropped

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
Itxaka: done

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
Itxaka: can you have more than one of the same rule? I guess its to protect against having multiple identical rules but not sure if its possible, will check

Itxaka: yup, its possible to have the same rule several times and deleting with that command will only delete one. Ivan's approach here is the safe one to remove all rules in case they were inserted more than once

ilausuch: Yes, sometimes I found more than one rule (the same rule) at the same time

chef/cookbooks/rabbitmq/templates/default/rabbitmq-alert-handler.erb
ilausuch: Done

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -120,3 +120,50 @@
...
+    owner "root"
+    group "root"
+    mode "0440"
+  end
+
+  # create the alert
+  pacemaker_alert "rabbitmq-alert-handler" do
+    handler "/usr/bin/rabbitmq-alert-handler.sh"
+    action :create
+  end
```
vuntz: Can you add a `else` statement where you delete what is created in the `if` statement, especially the pacemaker alert? (just in case a cluster somehow moves from 3 to 2 nodes).

chef/cookbooks/rabbitmq/templates/default/rabbitmq-alert-handler.erb
```diff
@@ -0,0 +1,19 @@
...
+[ "$CRM_alert_kind" = "resource" -a "$CRM_alert_rsc" = "rabbitmq" -a "$CRM_alert_task" = "monitor" ] || exit 0
+
+# launch the blocker in exclusive mode
+sudo flock /var/lock/rabbit /usr/bin/rabbitmq-port-blocker.sh &
+
+# for each node in the cluster
+<% @nodes.each do |cluster_node| -%>
+  # unless this is the current node launch remotely the blocker in exclusive mode
+  <% unless cluster_node.name==@node.name -%>
+    /usr/bin/timeout 15 sudo /usr/bin/ssh -o TCPKeepAlive=no -o ServerAliveInterval=15 root@<%= cluster_node.name %> nohup flock /var/lock/rabbit /usr/bin/rabbitmq-port-blocker.sh &
```
vuntz: I think the nohup should be at the beginning of the command, not in the middle.

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+
+# get the number of running nodes of rabbitmq in the current cluster
+function running_nodes()
+{
+  echo $(rabbitmqctl cluster_status 2>/dev/null | grep running_nodes | cut -d "," -f2- | cut -d "}" -f1 | tr -d "[" | tr -d "]" | tr -d "'" | tr "," "\n" | wc -l)
```
vuntz: You don't need to do `echo $(foo)`, just do `foo` :-)

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
```diff
@@ -0,0 +1,30 @@
...
+}
+
+# check if exists the bloquing rule for rabbitmq clients
+function check_rule()
+{
+  echo $(iptables -L -n | grep "tcp dpt:5672" | grep DROP | wc -l)
```
vuntz: Same thing here, no need for `echo $(foo)`

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
```diff
@@ -0,0 +1,30 @@
...
+# if the running nodes is les that the blocking level, then...
+if [ $(running_nodes) -le $blocking_level ]; then
+  # if rule not exists the rule will be added to block the clients port
+  if [ $(check_rule) -eq 0 ]; then
+    iptables -A INPUT -p tcp --destination-port 5672 -j DROP
+  fi
+else
+  # finally if the rule exists it will be deleted. If there are more than one, will remove all
+  if [[ $(check_rule) -gt 0 ]]; then
+    iptables-save | grep -v -e "-A INPUT -p tcp -m tcp --dport 5672 -j DROP" | iptables-restore
```
vuntz: Continuing discussion from https://github.com/crowbar/crowbar-openstack/pull/1554#discussion_r173436330 here:

>>> Why not do iptables -D INPUT -p tcp --destination-port 5672 -j DROP?
>>
>> can you have more than one of the same rule? I guess its to protect against having multiple identical rules but not sure if its possible, will check
>
> Yes, sometimes I found more than one rule (the same rule) at the same time

First, let's do a simple trick that would be good: use `-m comment --comment 'rabbitmq port blocker (no quorum)'`. See for instance https://github.com/rabbitmq/rabbitmq-server-release/blob/master/scripts/rabbitmq-server-ha.ocf#L772

This way, we would explicitly identify the rule that we add, and we can remove that rule only.

Then, second question: if the PR somehow makes it possible to have the rule more than once, then there's a bug to be fixed :-)

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
ilausuch: yes, sure

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
ilausuch: done

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
ilausuch: done

chef/cookbooks/rabbitmq/templates/default/rabbitmq-alert-handler.erb
ilausuch: Yes! and maybe we can maintain the second one, because, something I didn't check is what will happen if the node than calls the second one fails. Will the script in the second node continue the execution, or it will be closed?

ilausuch: Done, I've added in the first (local) execution of script the nohup because it is executed in background too

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
ilausuch: Ok, it seems that the lock was the solution, I did some tests and I couldn't reproduce it. @Itxaka, did you know how this problem could be reproduced?

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -120,3 +120,71 @@
...
+
+  file "/etc/sudoers.d/rabbitmq-port-blocker" do
+    action :delete
+  end
+
+  pacemaker_alert "rabbitmq-alert-handler" do
+    handler "/usr/bin/rabbitmq-alert-handler.sh"
+    action :delete
+  end
```
vuntz: You should revert the order of deletion: delete the alert first, so that you don't create a race where the alert is still here (and used) but the scripts from the alerts are gone.

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
```diff
@@ -0,0 +1,32 @@
...
+# get the number of running nodes of rabbitmq in the current cluster
+function running_nodes()
+{
+  rabbitmqctl cluster_status 2>/dev/null | grep running_nodes | cut -d "," -f2- | cut -d "}" -f1 | tr -d "[" | tr -d "]" | tr -d "'" | tr "," "\n" | wc -l
+}
+
+# check if exists the bloquing rule for rabbitmq clients
+function check_rule()
+{
+  iptables -L -n | grep "tcp dpt:5672" | grep DROP | wc -l
```
vuntz: This needs to be updated to check for the comment too.

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
Itxaka: done

chef/cookbooks/rabbitmq/templates/default/rabbitmq-port-blocker.erb
Itxaka: done

Itxaka:
@vuntz comments addressed in a different commit, in case we dont want them like that or a different approach is needed, can be amended on merge.

ilausuch:
All commits together

chef/cookbooks/rabbitmq/templates/default/rabbitmq-alert-handler.erb
```diff
@@ -0,0 +1,19 @@
...
+fi
+
+# exit if isn't a rabbitmq alert or is not a monitor task
+[ "$CRM_alert_kind" = "resource" -a "$CRM_alert_rsc" = "rabbitmq" -a "$CRM_alert_task" = "monitor" ] || exit 0
+
+# launch the blocker in exclusive mode
+nohup sudo flock /var/lock/rabbit /usr/bin/rabbitmq-port-blocker.sh &
+
+# for each node in the cluster
+<% @nodes.each do |cluster_node| -%>
```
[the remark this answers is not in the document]
ilausuch: During first tests, I checked that only the node that had generated the problem received a quickly notification (monitor type) about the problem, and the others had to wait longer. So from my experience the best way was that the node that receive the alert, had to notify the others to homogenize the global status of the IPTABLES block.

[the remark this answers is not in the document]
ilausuch: I've just checked it again and when a node falls (tested with a reboot and halt commands) other nodes execute the script. However If the problematic host doesn't send a message it will be a problem. I can fix it easily running the script of block against all rabbit alert messages from pacemaker. The frequency of these message depends, but during a reboot I checked two or tree per minute. I cannot avoid the remote execution, because sometimes other nodes doesn't receive any message if one has failed. A solution will be only to execute remote scripts if the type of the alert is monitor, and only locally if the type is notify. I do not know if it's worth it. What do you think?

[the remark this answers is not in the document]
AbelNavarro: I don't think it would be necessary to capture all rabbitmq alerts. If the problematic node fails, then you are as good as without these scripts, but if the message is sent then you improve the situation.

I can also see now that is mainly one node sending all block messages. I guess that is in a 3-node cluster. What about in a 5-node cluster? it may happen that two nodes are triggering the blocks at once. I guess that `flock` should work here. What's your take?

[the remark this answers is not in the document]
ilausuch: Yes, empirically I checked that it was necessary protecting the block script against concurrent executions in the same node to maintain a final stable situation with the IPTABLES rules. Other way, It could happen that more than one rule was added to IPTABLES. Or remove too soon the rules. In fact It can happen because the rabbitmqctr cluster_status shows all the nodes running sooner than the pacemaker alert been a false positive (pacemaker can reset the service again).

ilausuch:
@AbelNavarro commit message updated. Thanks

PR #1555, opened 2018-02-11
mysql: skip checking constrainst for mysql 10.2.x upgrade

dirkmueller:


chef/cookbooks/mysql/templates/default/my.cnf.erb
```diff
@@ -10,6 +10,9 @@ user                   = mysql
 datadir                = <%= node[:database][:mysql][:datadir] %>
 tmpdir                 = <%= node[:database][:mysql][:tmpdir] %>
 
+# workaround for incorrect constraints handling with mysql 10.2
+check_constraint_checks = 0
```
jsuchome: I believe a check for mariadb version is needed...

PR #1558, opened 2018-02-13
ironic: Store configdrive in Swift if possible

skazi0:
Allow storage of configdrive in Swift if it is available.
If Swift is available but user doesn't want to store the configdrive there it can be disabled using added attribute. It's value is ignored if Swift is not available.

Additional commit fixes username/password fields in swift settings.

### DIFF
--- a/chef/cookbooks/ironic/libraries/helpers.rb
+++ b/chef/cookbooks/ironic/libraries/helpers.rb
@@ -62,6 +62,8 @@ module IronicHelper
         glance_container: glance[:glance][:swift][:store_container],
         glance_account: glance_account,
         protocol: swift[:swift][:ssl][:enabled] ? "https" : "http",
+        service_user: swift[:swift][:service_user],
+        service_password: swift[:swift][:service_password],
         host: swift_address,
         port: swift[:swift][:ports][:proxy]
       }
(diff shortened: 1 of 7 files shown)


chef/cookbooks/ironic/libraries/helpers.rb
```diff
@@ -62,6 +62,8 @@ def swift_settings(node, glance)
         glance_container: glance[:glance][:swift][:store_container],
         glance_account: glance_account,
         protocol: swift[:swift][:ssl][:enabled] ? "https" : "http",
+        service_user: swift[:swift][:service_user],
+        service_password: swift[:swift][:service_password],
```
vuntz: Are these really used?

skazi0: Here they are:
https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/ironic/templates/default/ironic.conf.erb#L45

skazi0: The template might drop the conditions but I'm not sure why they were added there.

vuntz: Thanks, that wasn't obvious to me.

PR #1560, opened 2018-02-13
[4.0] Switch to MariaDB 10.2

rsalevsky:
Revert the temporary disable galera-clustercheck commit to make backporting easier (the package is released so this is not needed anymore) and add the missing backports to make MariaDB 10.2 work.


rhafer:
> Revert the temporary disable galera-clustercheck commit to make backporting easier (the package is released so this is not needed anymore)

Are you sure about that? IIRC we specifically introduced that for some customer. @vuntz Is this ok to go away?

rsalevsky:
@rhafer We did this for SAP because the galera-clustercheck package wasn't available in the channel back then. But now it is available so it should be ok to remove it.

vuntz:
> But now it is available so it should be ok to remove it.

If it's available, yes, it's okay.

rhafer:
The ci failure is still cause by the missing xtrabackup 2.4 it seems

PR #1563, opened 2018-02-16
[4.0] rabbitmq: block client port on startup

ilausuch:
This script blocks the connection to the rabbitmq cluster in case the number of nodes decay bellow the half of the total. In this case the remain rabbit nodes won't accept new connections. The problem to solve is that when a rabbit node has too much clients with low timeout, the final scenario could be a DOS when the clients try to constantly reconnect to rabbitmq and forcing more timeouts.

It takes advantage of the pacemaker notifications that notify when a rabbitmq has failed or has restored. All nodes rabbitmq ports will be blocked if the total number of alive nodes are below to the half of nodes of the cluster, or unblock if its over this value. This prevents the overload of alive nodes and the total block of service.

Only the node that have problems will receive the alert, so it is necessary to notify other nodes to block their ports if necessary. The only case when this script won't work is when the node fails completely and is not able to notify the others.

The script is divided in two parts: On one hand, the alert handler that manages the alerts discarding non interesting alerts, launches to the blocker script in this node and other cluster nodes (via SSH). And in the other hand, the blocker script than checks the condition of running nodes and blocks or unblocks the rabbitmq client port.

(cherry picked from commit 22d6a98)

Backport from #1554

### DIFF
--- a/chef/cookbooks/rabbitmq/files/default/rabbitmq.tmpfiles
+++ b/chef/cookbooks/rabbitmq/files/default/rabbitmq.tmpfiles
@@ -1,0 +1,1 @@
+d /var/lock/rabbit  0755  root  root  -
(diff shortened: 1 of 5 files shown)


ilausuch:

ilausuch:

PR #1564, opened 2018-02-16
MariaDB 10.2 cleanups

rsalevsky:
After the MariaDB 10.2 switch the new defaults make these entries
redundant so they are not needed anymore.

### DIFF
--- a/chef/cookbooks/mysql/attributes/server.rb
+++ b/chef/cookbooks/mysql/attributes/server.rb
@@ -26,9 +26,6 @@ if node[:database].key? "ec2"
   default[:database][:mysql][:ebs_vol_size]            = 50
 end
 
-default[:database][:mysql][:tunable][:max_allowed_packet]       = "16M"
-default[:database][:mysql][:tunable][:thread_cache_size]        = 8
-
 # Ports to bind to when haproxy is used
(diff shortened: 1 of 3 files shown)


chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -95,23 +95,23 @@
   end
 end
 
-template "/etc/my.cnf.d/openstack.cnf" do
+template "/etc/my.cnf.d/101-openstack.cnf" do
```
rhafer: If we really want to do this, you need to remove the old files (using the `file` resource). Otherwise we endup with the old `openstack.cnf` lying around and overriding settings from 101-openstack.cnf.

rsalevsky: This would be only interesting for Cloud 7 as this is already released there but in Cloud 8 this theoretically doesn't happen. So I will add this when I backport this change.

rhafer: I'd say it would still be safer to have it in master as well. 
a) to make the backport (and maybe future backports easier)
b) just to be safe in case this is not backported or the Cloud 7 backport is released too long after Cloud 8

PR #1567, opened 2018-02-22
ceilometer: Remove mongodb suport

jsuchome:
Our mongodb version is not properly maintained and has vulnerabilities which are not going to be fixed.

Requires https://build.suse.de/request/show/155990 (already merged)


aplanas:
can we try 3.6 before dropping mongodb for soc8? or is this the best option?

jsuchome:
I guess we can, but do you have time for that?

jsuchome:
[Drop mongodb support from ceilometer barclamp](https://trello.com/c/SlnFclzY/709-drop-mongodb-support-from-ceilometer-barclamp)

jsuchome:
Testbuild is failing because of 'technology preview' patch that is part of crowbar-openstack package

jsuchome:
@vuntz another review, please?

crowbar_framework/app/models/ceilometer_service.rb
```diff
@@ -243,20 +193,6 @@ def mongodb_ha(new_members, role)
   # primary.
   def validate_minimum_three_nodes_in_cluster(proposal)
```
vuntz: I think you should simply drop that method as it's doing nothing anymore (so no need to call it anymore).

jsuchome: Oh, true.

PR #1570, opened 2018-02-23
mariadb: Add prefix to configs

rsalevsky:
All default MariaDB configs have now a prefix to allow overwriting of
the options. Let's follow the same to not conflict with the defaults.


rsalevsky:
@nicolasbock @Itxaka Can you please re-review this PR? The gating issues are fixed.

PR #1571, opened 2018-02-23
rabbitmq: active the haproxy balancing

ilausuch:

The balancer gets the default client ports specified in the proposal and each rabbitmq node uses crowbar/crowbar-ha#304

### DIFF
--- a/chef/cookbooks/rabbitmq/providers/vhost.rb
+++ b/chef/cookbooks/rabbitmq/providers/vhost.rb
@@ -19,6 +19,9 @@
 
 action :add do
   unless Kernel::system("rabbitmqctl list_vhosts | grep -q #{new_resource.vhost}")
+    # wait for service to have a master, and to be active
+    CrowbarPacemakerHelper.wait_until_rabbitmq_ready
+
     Chef::Log.info "Adding RabbitMQ vhost '#{new_resource.vhost}'."
     execute "rabbitmqctl add_vhost #{new_resource.vhost}"
     new_resource.updated_by_last_action(true)
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -22,8 +22,16 @@ ha_enabled = node[:rabbitmq][:ha][:enabled]
 # we only do cluster if we do HA
 cluster_enabled = node[:rabbitmq][:cluster] && ha_enabled
 quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
-crm_resource_stop_cmd = cluster_enabled ? "force-demote" : "force-stop"
-crm_resource_start_cmd = cluster_enabled ? "force-promote" : "force-start"
+crm_resource_stop_cmd = "force-stop"
+crm_resource_start_cmd = "force-start"
+
+if node[:rabbitmq][:ha][:haproxy_enabled] && cluster_enabled
+  rabbitmq_port = node[:rabbitmq][:ha][:ports][:api]
+  rabbitmq_ssl_port = node[:rabbitmq][:ha][:ports][:api_ssl]
+else
+  rabbitmq_port = node[:rabbitmq][:port]
+  rabbitmq_ssl_port = node[:rabbitmq][:ssl][:port]
+end
 
 cluster_partition_handling = if cluster_enabled
   if CrowbarPacemakerHelper.num_corosync_nodes(node) > 2
@@ -110,7 +118,9 @@ template "/etc/rabbitmq/rabbitmq.config" do
   variables(
     cluster_enabled: cluster_enabled,
     cluster_partition_handling: cluster_partition_handling,
-    hipe_compile: hipe_compile
+    hipe_compile: hipe_compile,
+    rabbitmq_port: rabbitmq_port,
+    rabbitmq_ssl_port: rabbitmq_ssl_port
   )
   notifies :restart, "service[rabbitmq-server]"
 end
--- a/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
+++ b/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
@@ -108,6 +108,13 @@ end
 
 transaction_objects = []
 
+rabbitmq_port = if node[:rabbitmq][:ha][:haproxy_enabled]
+  node[:rabbitmq][:ha][:ports][:api]
+else
+  node[:rabbitmq][:port]
+end
+
+service_name = "rabbitmq"
 pacemaker_primitive service_name do
   agent agent_name
   # nodename is empty so that we explicitly depend on the config files
@@ -117,7 +124,8 @@ pacemaker_primitive service_name do
     "policy_file" => "/etc/rabbitmq/ocf-promote",
     "rmq_feature_health_check" => node[:rabbitmq][:ha][:clustered_rmq_features],
     "rmq_feature_local_list_queues" => node[:rabbitmq][:ha][:clustered_rmq_features],
-    "default_vhost" => node[:rabbitmq][:vhost]
+    "default_vhost" => node[:rabbitmq][:vhost],
+    "node_port" => rabbitmq_port
   })
   op node[:rabbitmq][:ha][:clustered_op]
   meta ({
@@ -127,6 +135,7 @@ pacemaker_primitive service_name do
   })
   action :update
   only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  notifies :restart, "service[rabbitmq-server]", :immediately
 end
 transaction_objects.push("pacemaker_primitive[#{service_name}]")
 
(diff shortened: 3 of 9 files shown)


ilausuch:
[RabbitMQ connections rebalancing](https://trello.com/c/bMDQwg0D/246-rabbitmq-connections-rebalancing)

ilausuch:
[RabbitMQ connections rebalancing](https://trello.com/c/bMDQwg0D/246-rabbitmq-connections-rebalancing)

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -37,6 +39,27 @@
   )
 end
 
+haproxy_loadbalancer "rabbitmq-api" do
```
vuntz: Why not just `rabbitmq` (same a few lines below for the ssl variant)?

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
```diff
@@ -37,6 +39,27 @@
...
+  address "0.0.0.0"
```
vuntz: Are you sure about that? I think we want to only have it listen on the admin network.

chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -24,6 +24,7 @@
       },
       "cluster": true,
       "ha": {
+        "enabled": false,
```
vuntz: That setting should be named `haproxy`, I guess.

chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
ilausuch: I agree, maybe I should use listen_public (is a proposal value) to control that.

vuntz: Probably, yes.

chef/cookbooks/rabbitmq/recipes/ha.rb
```diff
@@ -220,6 +220,8 @@
 # Now we can get the rabbitmq process to start since we know the directory is
 # writable, so we can create the primitive for rabbitmq.
 
+include_recipe "rabbitmq::haproxy" if node[:rabbitmq][:ha][:haproxy_enabled]
```
vuntz: No, not here :-) That's the old HA and in that mode, we don't use haproxy. Which actually means that... (see other comment)

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -25,6 +25,14 @@
 crm_resource_stop_cmd = cluster_enabled ? "force-demote" : "force-stop"
 crm_resource_start_cmd = cluster_enabled ? "force-promote" : "force-start"
 
+if node[:rabbitmq][:ha][:haproxy_enabled]
```
vuntz: (that's the other comment ;-))

We should only do this if we're also using clustering. If we're using the active/passive HA, we should not use haproxy.

chef/cookbooks/rabbitmq/recipes/haproxy.rb
```diff
@@ -0,0 +1,25 @@
...
+  haproxy_loadbalancer "rabbitmq-ssl" do
+    address node[:rabbitmq][:listen_public] ? "0.0.0.0" : cluster_admin_ip
+    port node[:rabbitmq][:ssl][:port]
+    use_ssl true
+    mode "tcp"
+    servers CrowbarPacemakerHelper.haproxy_servers_for_service(node, "rabbitmq", "rabbitmq-server",
+      "api_ssl")
+    action :nothing
+  end.run_action(:create)
+end
```
vuntz: Because this is only needed for HA clustering, I would probably not add a new file, but that's up to you.

chef/cookbooks/rabbitmq/recipes/ha.rb
Itxaka: shouldnt we? This was my idea as even if we got non-cluster HA we could take advantage of that, shouldnt we?

FAKE EDIT: Oh wait, so in drbd ha we have only 1 active rabbit server with the drbd as active and reachable via the VIP, while the others are just dormant and waiting??? Agh, yes in that case this should not be set here, sorry @ilausuch my mistake.

ilausuch: @Itxaka no worries.

ilausuch:
@vuntz thaks for your comments. WIP!

chef/cookbooks/rabbitmq/recipes/default.rb
ilausuch: Maybe is better a double condition. Check to do optinal in SOC7 haproxy_enabled and cluster flag. Do you agree?

ilausuch:
Recently the CI was failing. The reason was that with a rabbitm clusterand with haproxy enabled the services stil was using the list of rabbitmq nodes individually. Due to the change of port all services was failing to connect to rabbitmq. To solve that, if we are in cluster and haproxy enabled scenario, all barclamps will be configurated using the haproxy addess instead of the list.

chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
```diff
@@ -15,11 +15,11 @@
                         ]
    },
    {tcp_listeners, [
-                    <%= node[:rabbitmq][:addresses].map { |address| "{\"#{address}\", #{node[:rabbitmq][:port]}}" }.join(", ") %>
+                    <%= node[:rabbitmq][:addresses].map { |address| "{\"#{address}\", #{@rabbitmq_port}}" }.join(", ") %>
```
Itxaka: Just thinking out loud here, if we enable ssl should we disable the normal port as well? All clients will be connected to the ssl port, so I dont know if it makes sense to have the defautl port still up. Maybe for the pacemaker agents?

ilausuch: I am thinking we could have some scenarios:

A. public/private network <—–ssl—–> HAProxy <— ssl—-> rabbitmq
B. public/private network <—–no-ssl—–> HAProxy <—-no-ssl—-> rabbitmq
C. public/private network <—–ssl—–> HAProxy <—-no-ssl—-> rabbitmq ** it's not implemented yet

- As you said, I am not sure if having A and B at the same time is a good idea. I will check if disabling the non-ssl port could be a problem.
- Implementing C we could have more speed without sacrifice the security. Do you think it is a good way to solve that?

crowbar_framework/app/models/rabbitmq_service.rb
```diff
@@ -134,6 +134,17 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
+    vip_networks = ["admin"]
+    if role.default_attributes["rabbitmq"]["listen_public"]
+      vip_networks << "public"
+    end
+
+    allocate_virtual_ips_for_any_cluster_in_networks_and_sync_dns(rabbitmq_elements, vip_networks)
+
+    dirty = prepare_role_for_ha_with_haproxy(role, ["rabbitmq", "ha", "haproxy"],
+      rabbitmq_ha_enabled, rabbitmq_elements, vip_networks)
+    role.save if dirty
```
vuntz: All of this should only be done if the setting is set to true.

ilausuch: You're right
Done

vuntz:
@ilausuch I'd like to wait on that until we get the active/active haproxy parts, as this has a big impact on how things are configured. Would that be fine with you?

ilausuch:
@vuntz of course, in fact, am researching about active-active already

ilausuch:
Working in haproxy active-active

Itxaka:
@dirkmueller probably the follow up PR to make haproxy active/active as to avoid delays when a node goes down makes more sense to get the whole picture of this PR? https://trello.com/c/7oHIy01J/271-make-haproxy-active-active

vuntz:
> is there a rationale on "why" somewhere? If I understand https://insidethecpu.com/2014/11/17/load-balancing-a-rabbitmq-cluster/ properly then this doesn't sound like a good approach - basically queues already have a designated master and if you get redirected to the mirror it will send you back to the master of that queue.

Short summary is that it isn't about the queues, it's about balancing the number of client connections, to avoid a DDoS with every client reconnecting on one single server.

chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
AbelNavarro: I do wonder if the default round-robin algorithm to balance the connections will work with rabbitmq. Do we need some sort of stickiness? (i.e. by src IP)?

Itxaka: should be ok, rabbit will do internal routing itself. As long as we dont have all the connections in the same node the rest should be ok

crowbar_framework/app/models/rabbitmq_service.rb
```diff
@@ -100,6 +100,15 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
 
     role.save if save_role
 
+    haproxy_enabled = Rails.application.config.experimental
```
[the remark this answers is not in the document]
ilausuch: @Itxaka Done. Thanks

ilausuch:
Do no review yet, some tests has failed

ilausuch:
New approach, doing some additional tests

ilausuch:
Ready to review.
Tests:
- On a new installation. The haproxy active-active rabbitmq is configured
- On a old installation the haproxy active-active for rabbitmq is omitted
- On a new installation when the first attempt to apply the barclamp fails, finally it is configured too

crowbar_framework/app/models/rabbitmq_service.rb
```diff
@@ -100,6 +100,17 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
 
     role.save if save_role
 
+    # if no previous configuration of haproxy_enable, active if the proposal is not already applied
+    if old_role.nil?
```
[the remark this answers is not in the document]
ilausuch: vuntz idea!

chef/cookbooks/rabbitmq/providers/vhost.rb
```diff
@@ -19,6 +19,9 @@
 
 action :add do
   unless Kernel::system("rabbitmqctl list_vhosts | grep -q #{new_resource.vhost}")
+    # wait for service to have a master, and to be active
+    CrowbarPacemakerHelper.wait_until_rabbitmq_ready
```
ilausuch: To have the rabbitmq running and stable was important to try to add the vhost using rabbitmqctl. It was failing so I have used @Itxaka #1396 solution here

ilausuch:
@Itxaka Solved and tested. Now the check to ensure the ha for rabbit is only using #1396

ilausuch:
This is ready for review

aojea:
@aspiers That's exactly what we have to do, we should follow common architectures based on best practices and test results, i.e. [disabling queue mirroring](https://www.mirantis.com/blog/best-practices-rabbitmq-openstack/), [removing haproxy to balance connections rabbitmq and leveraging the clients to do it ](https://bugs.launchpad.net/fuel/+bug/1465263)

```Davanum Srinivas (DIMS) (dims-v) wrote on 2015-07-08: | #3
upstream oslo.messaging team and ops folks do *NOT* recommend using HAProxy. rabbit_hosts is what everyone is using and testing with. (Example - http://markmail.org/message/dzyokzmtu2khdish)```

aspiers:
@aojea Thanks, that helps a lot!  But I'll repeat:

> please let's document that research in a discoverable location

I don't count a comment in a PR as being an easily discoverable location ;-)  Which leads me back to the same request that [I made yesterday](https://github.com/SUSE-Cloud/automation/pull/2831#issuecomment-427016659): for [the commit message to document the *why*, not just the *what* and *how*](https://github.com/SUSE/cloud-specs/pull/177/files#diff-4c8f5d18098a22653ea34e4b9c4b5e7fR150).

dirkmueller:
@ilausuch can we close this?

ilausuch:
Yes, let's close it

PR #1574, opened 2018-02-28
horizon: fix apache logrotate

bmwiedemann:
/etc/init.d/apache2 does not exist anymore since SLE12

### DIFF
--- a/chef/cookbooks/horizon/templates/default/openstack-dashboard.logrotate.erb
+++ b/chef/cookbooks/horizon/templates/default/openstack-dashboard.logrotate.erb
@@ -8,7 +8,7 @@
     missingok
     create 644 root root
     postrotate
-     /etc/init.d/apache2 reload
+     systemctl reload apache2.service
     endscript
 }
 
@@ -22,6 +22,6 @@
     missingok
     create 644 root root
     postrotate
-     /etc/init.d/apache2 reload
+     systemctl reload apache2.service
     endscript
 }


chef/cookbooks/horizon/templates/default/openstack-dashboard.logrotate.erb
```diff
@@ -8,7 +8,7 @@
     missingok
     create 644 root root
     postrotate
-     /etc/init.d/apache2 reload
+     service apache2 reload
```
jsuchome: logrotate config for apache2 uses `systemctl reload apache2.service`, so let's keep it same here

bmwiedemann:
@jsuchome updated to use systemctl

bmwiedemann:
@rsalevsky please re-review

vuntz:
@bmwiedemann see also https://github.com/crowbar/crowbar-openstack/pull/1581

rsalevsky:
@bbobrov I guess we should do both.

@bmwiedemann The HA gating failed.

bmwiedemann:
https://github.com/crowbar/crowbar-core/commit/b74fcf9e9ff978e6d8f483d3aea4622a4c1ce46b should cover these dashboard logfiles too, so we could even drop this specialized rotation.

bmwiedemann:
except that the other logrotate file covers `*.log` and not `_log`

PR #1575, opened 2018-03-01
neutron: enable automatic lbaas reschedule

Itxaka:
Enable the ability to automatically reschedule load balancers from LBaaS
agents the server detects to have died.
Previously, load balancers could be scheduled and realized across
multiple LBaaS agents, however if a hypervisor died, the load
balancers scheduled to that node would cease operation.
Now, these load balancers will be automatically rescheduled to a
different agent.


Itxaka:
[Re-schedule LBs from failed lbaasv2-agents](https://trello.com/c/mHgY3bnW/253-re-schedule-lbs-from-failed-lbaasv2-agents)

PR #1577, opened 2018-03-02
[4.0] cinder: Set os_privileged_* values

jsuchome:
When the privileged user is not set, cinder does not provide
correct credentials to nova client and migration of attached
volumes fails.

See https://bugs.launchpad.net/cinder/+bug/1614344 for upstream bug.
The relevant code was rewritten for Pike so providing those options
won't be necessary in newer releases. But as the work was not
backported, we have to provide os_privileged values in Newton codebase
if we want to make the migration work (bsc#1079763).

### DIFF
--- a/chef/cookbooks/cinder/templates/default/cinder.conf.erb
+++ b/chef/cookbooks/cinder/templates/default/cinder.conf.erb
@@ -12,6 +12,12 @@ wsgi_keep_alive = false
 state_path = /var/lib/cinder
 my_ip = <%= node[:cinder][:my_ip] %>
 
+# os_privileged_* values are required for migrations of attached volumes
+# See bsc#1079763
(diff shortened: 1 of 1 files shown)


jsuchome:
[Bug 1079763 #L3 cinder migrate: volume goes into state "error" instantly](https://trello.com/c/CRLrBQHb/686-bug-1079763-l3-cinder-migrate-volume-goes-into-state-error-instantly)

jsuchome:
@dirkmueller do you have anything against? 

This is simpler than the alternative solution (of having 2 cinder users, as you've suggested in the bug), and I think enough because it won't be needed in SOC8

jsuchome:
@rsalevsky could you please flag this one for next maintenance update? Thanks

PR #1579, opened 2018-03-02
rabbitmq: allow pacemaker restart

Itxaka:
Up until now we were relaying on the restart for rabbit while on
HA to silently fail due to the resource service name called
rabbitmq-server while the pacemaker reasource is called rabbitmq.

This triggered a restart but it silently failed. Instead pass the
correct service name to allow for the services to be properly
promoted/demoted/restarted even on HA.


Itxaka:

rsalevsky:
But the HA gating failed. :/

PR #1580, opened 2018-03-02
[4.0] rabbitmq: allow pacemaker service restart

Itxaka:
Up until now we were relaying on the restart for rabbit while on
HA to silently fail due to the resource service name called
rabbitmq-server while the pacemaker reasource is called rabbitmq.

This triggered a restart but it silently failed. Instead pass the
correct service name to allow for the services to be properly
promoted/demoted/restarted even on HA.

(cherry picked from commit 00c9abd8c4f0067090a8a25faf268edcaa0b71b4)

Backport-of: https://github.com/crowbar/crowbar-openstack/pull/1579

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -22,6 +22,8 @@ ha_enabled = node[:rabbitmq][:ha][:enabled]
 # we only do cluster if we do HA
 cluster_enabled = node[:rabbitmq][:cluster] && ha_enabled
 quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+crm_resource_stop_cmd = cluster_enabled ? "force-demote" : "force-stop"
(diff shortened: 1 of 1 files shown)


Itxaka:

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -21,6 +21,8 @@
 ha_enabled = node[:rabbitmq][:ha][:enabled]
 # we only do cluster if we do HA
 cluster_enabled = node[:rabbitmq][:cluster] && ha_enabled
+crm_resource_stop_cmd = cluster_enabled ? "force-demote" : "force-stop"
+crm_resource_start_cmd = cluster_enabled ? "force-promote" : "force-start"
 quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
```
vuntz: To be in sync with master: the "quorum" line should be before the two new lines.

PR #1581, opened 2018-03-06
horizon: copytruncate apache logs bsc#1083093

bbobrov:
apache2 reload causes responses 406 from keystone. This change was
verified on a customer's setup.

bsc#1083093

### DIFF
--- a/chef/cookbooks/horizon/templates/default/openstack-dashboard.logrotate.erb
+++ b/chef/cookbooks/horizon/templates/default/openstack-dashboard.logrotate.erb
@@ -1,27 +1,21 @@
 /var/log/apache2/openstack-dashboard-access_log {
     compress
+    copytruncate
     dateext
     maxage 365
     rotate 99
     size=+4096k
     notifempty
     missingok
-    create 644 root root
-    postrotate
-     systemctl reload apache2.service
-    endscript
 }
 
 /var/log/apache2/openstack-dashboard-error_log {
     compress
+    copytruncate
     dateext
     maxage 365
     rotate 99
     size=+1024k
     notifempty
     missingok
-    create 644 root root
-    postrotate
-     systemctl reload apache2.service
-    endscript
 }


vuntz:
@bbobrov see also https://github.com/crowbar/crowbar-openstack/pull/1574

Itxaka:
@bmwiedemann is this still a NOK from you? Bugzilla was updated, should we move forward with this?

Itxaka:
[Logrotate block Keystone Service](https://trello.com/c/iyHul6QR/268-logrotate-block-keystone-service)

bbobrov:
Done.

> weird that susecommitpolice does not find it

Especially weird given that i have not done it. I wonder if github could do it automagically.

vuntz:
For the record, there are reasons that `copytruncate` is usually not used (as an example, see https://serverfault.com/questions/688658/rsyslog-with-logrotate-reload-rsyslog-vs-copytruncate).

AbelNavarro:
@vuntz looks like for apache this is an accepted mechanism to avoid reload: https://unix.stackexchange.com/questions/47688/how-to-avoid-apache-reload-when-rotating-logs . Since this change is only affecting apache it should be ok.

vuntz:
@AbelNavarro I know it's widely documented, but it's also not the default in any distribution for the apache logrotate files.

Not having dived in the topic enough, I don't know what's right. Just wanted to make this explicit (as the commit is not mentioning that).

bbobrov:
> but it's also not the default in any distribution for the apache logrotate files

It is the default for keystone at least in Debian. Opensuse also uses copytruncate for gnocchi. It also seems to be the default in RH, but i cannot find a good proof for that. We use copytruncate everywhere in Ardana-based cloud.

toabctl:
> It also seems to be the default for openstack services in RH, but i cannot find a good proof for that. 

for the record - RDO: https://github.com/rdo-packages/keystone-distgit/blob/rpm-master/openstack-keystone.logrotate (not using `copytruncate`).

bbobrov:
> for the record - RDO: https://github.com/rdo-packages/keystone-distgit/blob/rpm-master/openstack-keystone.logrotate (not using copytruncate).

right. I was basing my guess on https://bugzilla.redhat.com/show_bug.cgi?id=1123298, but i guess i was wrong.

vuntz:
> > but it's also not the default in any distribution for the apache logrotate files
> 
> It is the default for keystone in Debian. Opensuse also uses copytruncate for gnocchi at least.

I meant the logrotate files for the apache service itself (in the apache package), not for openstack services behind apache.

Since the logrotate file from the apache service will not use `copytruncate`, the same issue may still happen (although it's going to be less likely since it's for less logs).

PR #1583, opened 2018-03-07
Fixes needed for adding a node into existing cluster

jsuchome:
These are various bits found during testing the case of new node being added to existing pacemaker cluster (bsc#1083427)

- At least in the special case of adding new member to existing cluster,  Openstack::HA.set_controller_role is called and node[:pacemaker] accessed.
- cinder: Do not access role attributes with symbol keys 
- neutron: Check network presence before accessing it

### DIFF
--- a/crowbar_framework/app/models/cinder_service.rb
+++ b/crowbar_framework/app/models/cinder_service.rb
@@ -243,14 +243,13 @@ class CinderService < OpenstackServiceObject
     # Generate secrets uuid for libvirt rbd backend
     dirty = false
     proposal = Proposal.find_by(barclamp: "cinder", name: role.inst)
-    role.default_attributes[:cinder][:volumes].each_with_index do |volume, volid|
-      next unless volume[:backend_driver] == "rbd"
-      if volume[:rbd][:secret_uuid].empty?
-        secret_uuid = `uuidgen`.strip
-        volume[:rbd][:secret_uuid] = secret_uuid
-        proposal[:attributes][:cinder][:volumes][volid][:rbd][:secret_uuid] = secret_uuid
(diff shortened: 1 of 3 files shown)


jsuchome:
Hi guys, any chance for another review?

crowbar_framework/app/models/cinder_service.rb
```diff
@@ -243,14 +243,13 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
...
     proposal = Proposal.find_by(barclamp: "cinder", name: role.inst)
-    role.default_attributes[:cinder][:volumes].each_with_index do |volume, volid|
-      next unless volume[:backend_driver] == "rbd"
-      if volume[:rbd][:secret_uuid].empty?
-        secret_uuid = `uuidgen`.strip
-        volume[:rbd][:secret_uuid] = secret_uuid
-        proposal[:attributes][:cinder][:volumes][volid][:rbd][:secret_uuid] = secret_uuid
-        dirty = true
-      end
+    role.default_attributes["cinder"]["volumes"].each_with_index do |volume, volid|
```
Itxaka: any reason of changing the sym to string here? AFAIK it should be ok to get it with either as this was patched on our chef server

jsuchome: @Itxaka I do not exactly know why is it required, only that it fails with symbols. We're doing strings in other places as well (models accessing role attributes).

Now, when you're mentioning it, it's possible it works with symbols in master, I've tested this on 4.0 actually

Itxaka: :+1:

PR #1586, opened 2018-03-13
Remove nagios leftovers

rsalevsky:
Nagios got fully replaced by the new cloud monitoring solution so we
don't need this anymore.


Itxaka:
So can we now remove the fake stuff for nagios that we do on scripts/install-chef-suse.sh?  https://github.com/crowbar/crowbar/blob/master/scripts/install-chef-suse.sh#L778 and https://github.com/crowbar/crowbar/blob/master/scripts/install-chef-suse.sh#L908

rsalevsky:
@Itxaka I think there is also some stuff in crowbar-core but yes we can remove all of this now.

PR #1587, opened 2018-03-14
[4.0] mariadb: Add prefix to configs

rsalevsky:
All default MariaDB configs have now a prefix to allow overwriting of
the options. Let's follow the same to not conflict with the defaults.

(cherry picked from commit 4cc520ab12bc5d9b0a488652742a2afb074bf879)

Backport of https://github.com/crowbar/crowbar-openstack/pull/1570


rsalevsky:
@Itxaka We will deliver this together with the MariaDB 10.1 to MariaDB 10.2 upgrade. Therefore a customer will have to restart anyway.

PR #1588, opened 2018-03-14
[4.0] mariadb: Remove redundant config values

rsalevsky:
After the MariaDB 10.2 switch the new default makes these entrys
redundant so there are not needed anymore.

(cherry picked from commit 1aaa8c20d1e34022354ca2f0e39e4d1532910c8a)

Backport of https://github.com/crowbar/crowbar-openstack/pull/1564


rsalevsky:
This is ready to merge as the MariaDB 10.2 update is in the process to get released.

@Itxaka: It's a tech preview feature so we don't give an guaranty and we claim that we still do changes to the functionality.

PR #1592, opened 2018-03-19
rabbitmq: Expire unused non-durable queues after 1h by default

dirkmueller:
When a queue has no listeners anymore and is not marked
as durable, it can be deleted after an hour of no usage.


dirkmueller:
this might actually cause harm, I've given up on it.

PR #1593, opened 2018-03-19
barbican: Add missing roles used in policy.json (bsc#1081573)

JanZerebecki:
Add the key-manager:service-admin, observer and audit roles. These roles
are used in the policy.json file for the barbican API.


JanZerebecki:
Done.

PR #1594, opened 2018-03-19
travis: separate the different checks

rsalevsky:
Until now, we are just running the default rake tasks which would do a
syntaxcheck and also run any tests that rspec found. Instead of that,
lets separate those into different travis jobs to have more visibility
in case of failure.

Also it enhances the rspec tests by setting the output format to a nicer
way for easy checking and tags the travis builds by using the
environment variables to give the jobs a easy visual distiction ont he
travis page.

(cherry picked from commit 269c32deaef8744fda35beea8fba7c338e5b78b1)

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,1 @@
+--format documentation --color
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,12 @@ cache: bundler
 dist: trusty
 
 rvm: 2.1.9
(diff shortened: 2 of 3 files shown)


Itxaka:
Probably needs https://github.com/crowbar/crowbar-openstack/commit/7a18075d9c58221d732754db9f3f422642c4d86b forwardported before this

PR #1596, opened 2018-03-20
[4.0] Backport Nova Placement API

dirkmueller:
[UNTESTED]

The Placement API is a prerequisite for switching to Ocata/Pike ,so we need to release it as a maintenance update for SOC7 in order to enable upgrading to Cloud 8.

### DIFF
--- a/chef/cookbooks/nova/attributes/default.rb
+++ b/chef/cookbooks/nova/attributes/default.rb
@@ -21,6 +21,7 @@
 default[:nova][:debug] = false
 default[:nova][:max_header_line] = 16384
 default[:nova][:config_file] = "/etc/nova/nova.conf.d/100-nova.conf"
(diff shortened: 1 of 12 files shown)


stefannica:
@dirkmueller , you should kickoff an SSL job for this too, because I recall there being some complications when we did this for Ardana (see https://gerrit.suse.provo.cloud/#/c/2863/)

stefannica:
> @dirkmueller , you should kickoff an SSL job for this too, because I recall there being some complications when we did this for Ardana (see https://gerrit.suse.provo.cloud/#/c/2863/)

It might not be the case here, as we're using python-cryptography 1.3.1 in SOC7, but better safe than sorry.

dirkmueller:
@stefannica added a with_ssl run

PR #1597, opened 2018-03-20
magnum: Remove SLES image upload (bsc#1082278)

sayalilunkad:
This removes the upload of the SLES based magnum image to glance since we no longer
support this image in Cloud 8. Only the images that are supported upstream can
be used with magnum going forward. This also removes the specific flavor we
created for the mangum image since it does not make sense to leave that in
there.


stefannica:
@sayalilunkad just a thought, but shouldn't we replace the SLES based magnum image with one of the upstream images instead of removing it altogether ? Can we do any automated testing for magnum otherwise ?

sayalilunkad:
@stefannica Currently the two drivers maintained actively upstream are kubernetes and docker swarm which are based on a fedora atomic image. I am not sure we want to ship a fedora image to the customers since we won't be the one's maintaining it. Maybe after we have a opensuse based image we could ship that. As for testing magnum has functional tests for API and each of it's COE. If they work as they are supposed to (something I am looking into now) that should be good enough from the testing perspective I think.

rsalevsky:
We should have automated testing for magnum but this has to be done in the automation repo. In crowbar we will not offer/support/upload any pre-build image going forward.

rsalevsky:
What will happen if a customer upgrades from SOC7? Should we handle this case?

sayalilunkad:
@rsalevsky I would not, because they could have clusters running based on the sles image and removing the image will cause problems in that case. The customers only need to be made aware that we will not be supporting this image in SOC8.

stefannica:
> What will happen if a customer upgrades from SOC7? Should we handle this case?

That's a good point. Can the magnum clusters already running on the SOC7 SUSE image still be used after an upgrade ? And if not, does that mean that all the magnum clusters need to be removed ?

stefannica:
> > What will happen if a customer upgrades from SOC7? Should we handle this case?

> That's a good point. Can the magnum clusters already running on the SOC7 SUSE image still be used after an upgrade ? And if not, does that mean that all the magnum clusters need to be removed ?

Thanks @sayalilunkad , I think you just answered that in full.

sayalilunkad:
@stefannica Well as discussed with the containers team it was decided not to support this image for long since we did not have any major customers using it but not sure if that has changed and if we need to maintain this image. As for the upgrade part, it isn't specifically tested but since it is only nova instances it should be fine.

rsalevsky:
I will mention the removal in the release notes so that customers can plan accordingly.

PR #1599, opened 2018-03-21
neutron: [cisco-aci]: Add support for vpc_pairs.

mmnelemane:
Cisco ACI supports connecting to the fabric either through direct port mappings that in which case, we need to specify the exact ports under [apic-switches] configuration. The pattern usually looks like "1/34" for interface Eth1 and Port 34. Associating these ports creates new policy profiles and groups for the integration. This is not feasible if the ACI is already running another production cloud and we need to integrate openstack as a supplement (which is the case with most ACI customers).
However, ACI also allows for associating pre-existing profiles using VPC_Pairs. This can be configured by specifying the apic_vpc_pairs in ml2_conf_cisco_apic.ini.
This commit provides an option to test with either modes (VPC Pairs or direct connection) as the use-case may be.

Backport: https://github.com/crowbar/crowbar-openstack/pull/1600

### DIFF
--- a/chef/cookbooks/neutron/recipes/cisco_apic_support.rb
+++ b/chef/cookbooks/neutron/recipes/cisco_apic_support.rb
@@ -49,6 +49,7 @@ template node[:neutron][:ml2_cisco_apic_config_file] do
   owner "root"
   group node[:neutron][:platform][:group]
   variables(
+    vpc_pairs: node[:neutron][:apic][:vpc_pairs],
     apic_switches: aciswitches,
     ml2_mechanism_drivers: node[:neutron][:ml2_mechanism_drivers],
     policy_drivers: "implicit_policy,apic",
--- a/chef/cookbooks/neutron/templates/default/ml2_conf_cisco_apic.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/ml2_conf_cisco_apic.ini.erb
@@ -15,6 +15,10 @@ enable_optimized_metadata = <%= node[:neutron][:apic][:optimized_metadata] %>
 enable_optimized_dhcp = <%= node[:neutron][:apic][:optimized_dhcp] %>
 apic_provision_infra = True
 apic_provision_hostlinks = True
+<%  unless @vpc_pairs.nil? -%>
+apic_vpc_pairs = <%= @vpc_pairs %>
+<% end -%>
+
 <% @apic_switches.keys.each do |ip| -%>
 [apic_switch:<%=ip%>]
 <%    if @apic_switches[ip].key?(:switch_ports) -%>
--- a/chef/data_bags/crowbar/template-neutron.schema
+++ b/chef/data_bags/crowbar/template-neutron.schema
@@ -52,6 +52,7 @@
                       "password": { "type" : "str", "required": true },
                       "optimized_metadata": { "type" : "bool", "required": true },
                       "optimized_dhcp": { "type" : "bool", "required": true }, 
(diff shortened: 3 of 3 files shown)


chef/cookbooks/neutron/templates/default/ml2_conf_cisco_apic.ini.erb
```diff
@@ -15,6 +15,10 @@ enable_optimized_metadata = <%= node[:neutron][:apic][:optimized_metadata] %>
 enable_optimized_dhcp = <%= node[:neutron][:apic][:optimized_dhcp] %>
 apic_provision_infra = True
 apic_provision_hostlinks = True
+<%  unless @vpc_pairs.nil? -%>
+apic_vpc_pairs = <%= @vpc_pairs %>
+<% end -%>
+
```
rhafer: Is it really ok to have vpc_pairs and the apic_switches[ip][:switch_ports] setting at the same time? Or are the mutually exclusive.

mmnelemane: We still need apic_switches for each node to be specified. The difference is that when apic_vpc_pairs is enabled, we are able to specify the VPC data instead of port numbers under the switch_ports section. So both fields need to co-exist. If there is no vpc_pair specified, then switch_ports will contain port numbers (like 1/25).

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -47,6 +47,7 @@
         "password": "",
         "optimized_metadata": true,
         "optimized_dhcp": true,
+        "vpc_pairs": "101:102",
```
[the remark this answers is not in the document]
mmnelemane: @rhafer: Thats right. Its better we have this nil by default and allow the user to set it if needed.

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -42,11 +42,12 @@
       },
       "apic": {
         "hosts": "",
-        "system_id": "soc",
+        "system_id": "",
         "username": "admin",
         "password": "",
         "optimized_metadata": true,
         "optimized_dhcp": true,
+        "vpc_pairs": "",
```
rhafer: As this attribute is not required I guess you could also just leave it unset here. In that case it would just be `nil` in the recipe I think.

chef/cookbooks/neutron/recipes/cisco_apic_support.rb
```diff
@@ -41,6 +41,7 @@
 end
 
 aciswitches = node[:neutron][:apic][:apic_switches].to_hash
+apic_vpc_pairs = node[:neutron][:apic][:vpc_pairs].nil? ? nil : node[:neutron][:apic][:vpc_pairs]
```
rhafer: Hm, somewhat my earlier commit got lost on this. I guess I clicked the wrong button somewhere.  So I try to repeat it here:

1. Isn't this basically the same as: `apic_vpc_pairs = node[:neutron][:apic][:vpc_pairs]`? What is the `nil?` check good for if it just sets `nil`

2. Currently you set `""` as the default in the template. But `"".nil?` is false. So an empty string would endup in the config file.

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -42,11 +42,12 @@
       },
       "apic": {
         "hosts": "",
-        "system_id": "soc",
+        "system_id": "",
```
skazi0: why change this?

PR #1601, opened 2018-03-27
Revert "rabbitmq: allow pacemaker restart"

Itxaka:
Reverts crowbar/crowbar-openstack#1579

Restarting rabbit seems to bring instability to the rest of the services so we better not restart.


Itxaka:

PR #1602, opened 2018-03-27
Revert "[4.0] rabbitmq: allow pacemaker service restart"

Itxaka:
Reverts crowbar/crowbar-openstack#1580

Restarting rabbit seems to bring instability to the rest of the services so we better not restart.


rsalevsky:
We are currently in the process of releasing a maintenance update which includes this feature. Should we stop the update?

Itxaka:
@rsalevsky I would say yes, unless it includes this revert. Usually rabbitmq is not changed once is running but if the customer changes it afterwards it could lead to instability due to all the pacemaker shenanigans.

Itxaka:

PR #1603, opened 2018-03-27
[cisco_aci]neutron: Add support for L3Out External Connectivity

mmnelemane:
This commit provides a way to configure external connectivity
to the opnenstack instances by allowing the user to provide
ml2 config file with the L3 Out network and EPG details.

The commit adds a new section [apic_external_network] in
ml2_conf_cisco_apic.ini. The name of the section and the EPG provided
in the input reflects the Bridge domain, EPG and the External
Routed Network elements created on the ACI fabric.

When the external network of the ACI fabric is mapped to ML2,
the creation of neutron subnets get mapped to the subnets on the ACI
fabric and therefore will be subject to the Contracts and Policies
defined on the tenant. 
.

### DIFF
--- a/chef/cookbooks/neutron/recipes/cisco_apic_support.rb
+++ b/chef/cookbooks/neutron/recipes/cisco_apic_support.rb
@@ -53,7 +53,7 @@ template node[:neutron][:ml2_cisco_apic_config_file] do
     apic_switches: aciswitches,
     ml2_mechanism_drivers: node[:neutron][:ml2_mechanism_drivers],
     policy_drivers: "implicit_policy,apic",
-    default_ip_pool: "192.168.0.0/16",
+    default_ip_pool: "192.168.0.0/16"
   )
   notifies :restart, "service[#{node[:neutron][:platform][:service_name]}]"
 end
--- a/chef/cookbooks/neutron/templates/default/ml2_conf_cisco_apic.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/ml2_conf_cisco_apic.ini.erb
@@ -21,11 +21,11 @@ apic_vpc_pairs = <%= @vpc_pairs %>
 
 <% @apic_switches.keys.each do |ip| -%>
 [apic_switch:<%=ip%>]
-<%    if @apic_switches[ip].key?(:switch_ports) -%>
-<%      @apic_switches[ip][:switch_ports].each do |name, values| -%>
-<%=       name %> = <%= values[:switch_port] %>
-<%      end -%>
-<%    end -%>
+  <% if @apic_switches[ip].key?(:switch_ports) -%>
+    <% @apic_switches[ip][:switch_ports].each do |name, values| -%>
+<%= name %> = <%= values[:switch_port] %>
+    <% end -%>
+  <% end -%>
 <% end -%>
 <% if @ml2_mechanism_drivers.include?("apic_gbp") -%>
 [group_policy]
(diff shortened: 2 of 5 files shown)


chef/data_bags/crowbar/template-neutron.schema
```diff
@@ -53,6 +53,13 @@
                       "optimized_metadata": { "type" : "bool", "required": true },
                       "optimized_dhcp": { "type" : "bool", "required": true }, 
                       "vpc_pairs": { "type": "str", "required": false },
+                      "ext_net": { "type" : "map", "required" : false, "mapping" : {
+                          "name": { "type" : "str", "required" : true },
+                          "preexisting": { "type" : "bool", "required" : true },
```
rsalevsky: You need to add a migration if you add something with `"required": true`.

mmnelemane: @rsalevsky: As I understood, the mapping itself is set to "required": false, even though the internal elements of the mapping are required (makes the fields mandatory if the mapping is defined). For this reason, I intentionally skipped migration. I guess, infoblox had the same approach where the mapping itself was not mandatory while the internal elements were mandatory if the mapping was used. I could be wrong. Please confirm.

mmnelemane:
@rsalevsky: on second thought, I made the [apic_external_network] section as mandatory as there is no other option to provide L3 external connectivity to the instances and the user should use the options with ACI to isolate traffic based on his policies.

chef/cookbooks/neutron/templates/default/ml2_conf_cisco_apic.ini.erb
```diff
@@ -33,3 +33,13 @@ policy_drivers = <%= @policy_drivers %>
 [group_policy_implicit_policy]
 default_ip_pool = <%= @default_ip_pool %>
 <% end -%>
+
+<% if @use_apic_ext -%>
+[apic_external_network:<%=node[:neutron][:apic][:ext_net][:name]%>]
+preexisting = <%= node[:neutron][:apic][:ext_net][:preexisting] %>
+<%   unless node[:neutron][:apic][:ext_net][:nat_enabled].nil? -%>
```
rsalevsky: please use only one whitespace here

mmnelemane: @rsalevsky: Updated with single space. Also corrected the earlier section on apic_switches for consistency in the file

chef/cookbooks/neutron/recipes/cisco_apic_support.rb
```diff
@@ -41,6 +41,7 @@
 end
 
 aciswitches = node[:neutron][:apic][:apic_switches].to_hash
+use_ext_net = !node[:neutron][:apic][:ext_net].nil?
```
rhafer: As [neutron][:apic][:ext_net] is a required attribute this condition will always be true. If you need a way to disable use_ext_net  you need to find something else I think.

mmnelemane:
@rhafer: The message is indeed confusing. I have corrected the same along with the redundant ext_net check (making it mandatory) in the code.

mmnelemane:
@rhafer, @rsalevsky: This code was verified on the customer test environment as well as our own lab and found to provide external connectivity to the instances as expected. Only thing missing now is a suitable documentation explaining the detailed steps and reasons which I am working on.

PR #1605, opened 2018-03-27
magnum: Set insecure option in keystone_authtoken

sayalilunkad:
This is needed when ssl is enabled. Just having it in the keystone_auth section
is not enough.


toabctl:
Typo in commit msg: kesytone

PR #1606, opened 2018-03-28
trove: Hide trove barclamp

toabctl:
SOC8 does not support Trove so hide the barclamp.

### DIFF
--- a/trove.yml
+++ b/trove.yml
@@ -19,7 +19,7 @@ barclamp:
   display: 'Trove'
   description: 'OpenStack Database: Scalable and reliable Database-as-a-Service provisioning'
   version: 0
-  user_managed: true
+  user_managed: false
   requires:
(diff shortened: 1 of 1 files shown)


dirkmueller:
its currently marked as tech preview though, not really unsupported..

toabctl:
> its currently marked as tech preview though, not really unsupported..

I know. But I thought we will remove Trove and no longer add it as tech preview.

vuntz:
I'm fine with this, but @rsalevsky will remember better what we chose to do with it.

rsalevsky:
Trove is still planned to be tech-preview in Crowbar

toabctl:
Trove will stay tech preview according to https://github.com/crowbar/crowbar-openstack/pull/1606#issuecomment-377182425

PR #1609, opened 2018-03-29
horizon: load monasca from databag

dirkmueller:
The monasca databag config is accessing the monasca node attributes
unconditionally, which causes crashes when the monasca barclamp
is deactivated. Converting this to use the databag reduces
unnecessary nodesearches as well as avoiding those issues


dirkmueller:
this one makes the work on the engcloud rather clumsy hence I would like to backport it

PR #1611, opened 2018-04-05
tempest: install optional tempest plugins

dirkmueller:
In order to increase test coverate we want to install
the available tempest plugins for services that have splitted
them into a separate package.

### DIFF
--- a/chef/cookbooks/tempest/recipes/install.rb
+++ b/chef/cookbooks/tempest/recipes/install.rb
@@ -31,7 +31,7 @@ package "euca2ools"
 
 package "openstack-tempest-test"
 
-["cinder", "designate", "ironic", "keystone", "magnum",
+["barbican", "cinder", "designate", "heat", "ironic", "keystone", "magnum",
  "manila", "neutron"].each do |service|
   package "python-#{service}-tempest-plugin"
 end


chef/cookbooks/tempest/recipes/install.rb
```diff
@@ -36,9 +36,18 @@
   package "python-#{component}client"
 end
 
+if !node_search_with_cache("roles:barbican-controller").empty?
```
guangyee: We probably don't need to do node search here. We can just add "barbican" to the [service_available] section here.
https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/tempest/templates/default/tempest.conf.erb#L152-#L164

chef/cookbooks/tempest/recipes/install.rb
```diff
@@ -36,9 +36,18 @@
...
+if !node_search_with_cache("roles:heat-server").empty?
+  package "python-heat-tempest-plugin"
+end
+
 if node[:kernel][:machine] == "x86_64" &&
     !node_search_with_cache("roles:monasca-server").empty?
   ["api", "log-api"].each do |component|
     package "python-monasca-#{component}"
   end
+  package "python-monasca-tempest-plugin"
```
guangyee: Same as above. Notice that monasca-tempest-plugin has three service available configurations. See
https://github.com/openstack/monasca-tempest-plugin/blob/master/monasca_tempest_tests/config.py#L20-#L32
We'll need to figure out how to determine if elastic search and log API are available.

cmurphy:
@jgrassler can you revisit your -1?

dirkmueller:
seems to increase test coverate to 169 smoke tests and passing. +1

toabctl:
There are lots of config options for at least for the heat plugin (see https://git.openstack.org/cgit/openstack/heat-tempest-plugin/tree/heat_tempest_plugin/config.py ). Interesting that this just works...

PR #1612, opened 2018-04-05
framework: save attributes to databag during pre chef run

Itxaka:
Instead of waiting until the post chef method to be called to save
the attributes, just do it before as some of those attributes are
used during the chef run and if they are not available they will
be defaulted to false which means that we are configuring services
with the wrong parameters, at least on the first chef run

### DIFF
--- a/crowbar_framework/app/models/ceilometer_service.rb
+++ b/crowbar_framework/app/models/ceilometer_service.rb
@@ -180,6 +180,9 @@ class CeilometerService < OpenstackServiceObject
                                      ["ceilometer", "ha", "central", "enabled"],\
                                      central_ha_enabled)
 
+    # save attributes into the databag
+    save_config_to_databag(old_role, role)
+
     @logger.debug("Ceilometer apply_role_pre_chef_call: leaving")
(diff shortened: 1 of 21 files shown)


crowbar_framework/app/models/ceilometer_service.rb
```diff
@@ -180,6 +180,8 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
                                      ["ceilometer", "ha", "central", "enabled"],\
                                      central_ha_enabled)
 
+    # save attributes into the databag
+    save_config_to_databag(old_role, role)
     @logger.debug("Ceilometer apply_role_pre_chef_call: leaving")
```
nicolasbock: Just to make this beautiful PR even more beautiful :wink: Could you add an empty line to make the code formatting consistent? Thanks!

Itxaka: I tried to keep it consistent, I promise :P

Itxaka:
doesnt seem to be needed feel free to restore if so

PR #1614, opened 2018-04-06
[tempest]: Make public network name configurable.

mmnelemane:
Tempest barclamp expects the public network name to be
"floating" and is hard coded in the recipe. Some special
use-cases (for e.g: ACI) require that the public network
be named as per the configuration setup in the SDN solution.

This commit is allowing the user to define the public network
name as a variable from the Crowbar UI so that its possible
to test other SDN backends requiring such flexibility.


rsalevsky:
Seems like tempest is failing with this changes.

PR #1615, opened 2018-04-06
[tempest]: Backport from #1614. Add public_network_name.

mmnelemane:
Tempest barclamp expects the public network name to be
"floating" and is hard coded in the recipe. Some special
use-cases (for e.g: ACI) require that the public network
be named as per the configuration setup in the SDN solution.

This commit is allowing the user to define the public network
name as a variable from the Crowbar UI so that its possible
to test other SDN backends requiring such flexibility.

(cherry picked from commit c5ccc84310ba8398c3d51fc11f6f56de773dd7ca)


JanZerebecki:
Backport of #1614 which is not merged yet.
SOC7 should not get any changes any more besides important fixes, which this seems not to be. Please reopen if you think otherwise.

PR #1619, opened 2018-04-09
nova: disable progress timeout for live migration (bsc#1091190)

dirkmueller:
The progress timeout is unreliable and was already disabled
by default in pike, however it is still enabled in Newton, hence
this is going to be backported. Also be a little bit more generous
to slow compute nodes and allow migration to as slow as 1MByte/s
before aborting it. this tremendeously helps live migration to
succeed on very busy compute nodes (where migrating away busy
workload is the best thing to do)

Also remove an outdated setting that is defaulting to 30 already.

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -200,6 +200,9 @@ live_migration_flag = VIR_MIGRATE_UNDEFINE_SOURCE, VIR_MIGRATE_PEER2PEER, VIR_MI
     <% else %>
 block_migration_flag = VIR_MIGRATE_UNDEFINE_SOURCE, VIR_MIGRATE_PEER2PEER, VIR_MIGRATE_NON_SHARED_INC, VIR_MIGRATE_LIVE
     <% end %>
+# Timeout migration if less than 1MB/s RAM can be copied
+live_migration_progress_timeout=0
+live_migration_completion_timeout=1000
   <% end %>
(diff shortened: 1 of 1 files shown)


jsuchome:
bsc?

dirkmueller:
@jsuchome bug reference added

chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -200,6 +200,9 @@ live_migration_flag = VIR_MIGRATE_UNDEFINE_SOURCE, VIR_MIGRATE_PEER2PEER, VIR_MI
     <% else %>
 block_migration_flag = VIR_MIGRATE_UNDEFINE_SOURCE, VIR_MIGRATE_PEER2PEER, VIR_MIGRATE_NON_SHARED_INC, VIR_MIGRATE_LIVE
     <% end %>
+# Timeout migration if less than 1MB/s RAM can be copied
```
jsuchome: Thist comment belongs to the second option

dirkmueller: thats debateable. without *also* disabling the progress timeout it would be only half the truth (the progress timeout kicks in faster, so the "if less than 1MB/s can be copied" would not be true anymore)

I had it both ways and settled for the current one. I can change it if you want but strictly both settings are needed for the comment to be true.

jsuchome: I see

PR #1621, opened 2018-04-09
mysql: Set monitor to 10s for galera-python-clustercheck

vuntz:
This wasn't set and the default is 0, which means it wasn't monitored at
all.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -289,8 +289,13 @@ end
 transaction_objects = []
 service_name = "galera-python-clustercheck"
 
+clustercheck_op = {}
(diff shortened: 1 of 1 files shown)


vuntz:
> Also, isn't 10 seconds the default anyway?

Nope, see the commit message :-)

jhesketh:
>>Also, isn't 10 seconds the default anyway?

>Nope, see the commit message :-)

Right, sorry. I must be looking in the wrong place then (I assumed it came from here: https://github.com/crowbar/crowbar-ha/blob/master/chef/cookbooks/crowbar-pacemaker/attributes/default.rb#L34)

toabctl:
stable/4.0 backport is https://github.com/crowbar/crowbar-openstack/pull/1622

PR #1624, opened 2018-04-10
nova: avoid scheduling conflicts on HA CP

dirkmueller:
When all schedulers are picking the "best" host, in a 3 node
HA case they will all pick the same host, which is suboptimal.
Try to scale out a bit and increase retries in case of conflicts.

Even worse, a node that had still capacity when the scheduling
started might be out of memory at the time the VM is launched,
causing failed starts of VMs.

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -323,6 +323,9 @@ keep_alive = false
 
 [scheduler]
 discover_hosts_in_cells_interval = <%= node[:nova][:scheduler][:discover_hosts_in_cells_interval] %>
+max_attempts = 9
+# Avoid scheduler conflicts when using HA
+host_subset_size = 4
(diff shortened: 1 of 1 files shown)


chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -318,6 +318,9 @@ keep_alive = false
 
 [scheduler]
 discover_hosts_in_cells_interval = <%= node[:nova][:scheduler][:discover_hosts_in_cells_interval] %>
+max_attempts = 9
+# Avoid scheduler conflicts when using HA
+host_subset_size = 4
```
vuntz: Wondering if the value should be more dynamic (something that depends on the amount of nodes in the cluster, for instance).

rsalevsky: @dirkmueller What do you think?

PR #1630, opened 2018-04-12
neutron: enable trunk service plugin

rhafer:
This enables the creation of trunk ports in neutron e.g. to be able to
use tagged vlans inside VM instances.


rhafer:
FYI I just tested the trunking feature (vlan aware VMs) in engcloud where Dirk applied this patch. Seems to work just fine!

PR #1631, opened 2018-04-13
ironic: Use proper Swift address in HA mode

skazi0:
If Swift was deployed with HA client should always access the cluster
IP not any of the individual nodes.

### DIFF
--- a/chef/cookbooks/ironic/libraries/helpers.rb
+++ b/chef/cookbooks/ironic/libraries/helpers.rb
@@ -54,9 +54,6 @@ module IronicHelper
       get_tempurl_key = "#{swift_command} stat | grep -m1 'Meta Temp-Url-Key:' | awk '{print $3}'"
       tempurl_key = Mixlib::ShellOut.new(get_tempurl_key, environment: env).run_command.stdout.chomp
 
-      # use IP as this will be used by agent which can have no DNS configured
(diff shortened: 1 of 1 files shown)


chef/cookbooks/ironic/libraries/helpers.rb
```diff
@@ -54,17 +54,14 @@ def swift_settings(node, glance)
...
-
       {
         tempurl_key: tempurl_key,
         glance_container: glance[:glance][:swift][:store_container],
         glance_account: glance_account,
         protocol: swift[:swift][:ssl][:enabled] ? "https" : "http",
         service_user: swift[:swift][:service_user],
         service_password: swift[:swift][:service_password],
-        host: swift_address,
+        host: CrowbarHelper.get_host_for_admin_url(swift, swift[:swift][:ha][:enabled]),
```
toabctl: Switching from `public` to `admin` is correct?

skazi0: I think so, these are be used only for provisioning so admin connectivity should be available. It can also result in DNS names instead of IPs but these seem to work just fine (at least in the tests we did so far with @gosipyan ).

dirkmueller:
@skazi0 why is this marked as "do not merge yet" Looks obvious enough to merge it for now..

PR #1632, opened 2018-04-13
rabbitmq: restart epmd on config change

Itxaka:
Epmd service could very well be active by the time we add the new
port.conf config so it would ignore the setting.
This just adds a subscribe event to that config file which will fire
a restart of the service if the file is changed.

This may or may not be affecting the nodes upgrade, in which after upgrading the OS and restarting, makes epmd be up and running, so the new ports config will not be applied.


toabctl:
Same - don't merge stuff at this point in time...

PR #1635, opened 2018-04-16
Moving S390X to SOC7

aplanas:


rsalevsky:
Do we still need this?

aplanas:
indeed

rhafer (commented):
I am pretty sure my review isn't needed here anymore :-)

PR #1638, opened 2018-04-23
monasca: fix monasca-installer for InfluxDB (bsc#1090413)

jgrassler:
monasca-installer chokes on there not being a cassandra
Ansible group defined if InfluxDB is the time series database
being used. This commit removes the conditionals in the
template that prevent this group from being defined in the
InfluxDB scenario.


jgrassler:
Sorry, forgot the "do not merge, yet" label earlier. Please do not merge this right away. I want to first test it in both the InfluxDB and Cassandra scenario before we merge.

jgrassler:
Ok, I just learned that we're on a rather tight schedule with this, so I'm withdrawing my "do not merge,  yet": if CI passes (the CI job has Monasca enabled and uses the default time series database setting of  InfluxDB) we can merge. In the worst case Cassandra will be broken, but I don't think it will be.

toabctl:
monasca installation passed CI. merging.

PR #1639, opened 2018-04-23
keystone: Add retry loop to _get_token (bsc#1087466)

skazi0:
Sometimes _get_token fails with 502 HTTP code because Apache is being
restarted while chef-client tries to access it. Small retry loop will
give it a bit more chance to complete the operation.

### DIFF
--- a/chef/cookbooks/keystone/providers/register.rb
+++ b/chef/cookbooks/keystone/providers/register.rb
@@ -513,16 +513,27 @@ def _get_token(http, user_name, password, project = "")
   path = "/v3/auth/tokens"
   headers = _build_headers
   body = _build_auth(user_name, password, project)
-  resp = http.send_request("POST", path, JSON.generate(body), headers)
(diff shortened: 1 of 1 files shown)


jsuchome:
Why minute, isn't it too long?

skazi0:
Yeah, wakeup has a retry loop (50x 1s) but it can fail before it gets to this loop, inside `_build_connection()`. `_get_token()` looks like a first common point where we could check the HTTP status codes to ensure API is accessible. Of course this doesn't solve all problems (i.e. apache restart might be triggered after we got the token but before we use it but it should solve at least some of the synchronization problems we have now (e.g. in Ironic barclamp).

skazi0:
This bug appears in CI tests pretty often so final confirmation should be done by the CI jobs after merging. I haven't seen the bug in my mkcloud test runs since I started adding this patch (3 runs iirc). Another run ongoing.

toabctl:
@skazi0 I guess we want to backport this to SOC7?

PR #1640, opened 2018-04-23
trove: Hide trove barclamp

rsalevsky:


toabctl:
I added the patch to the package for now.

PR #1641, opened 2018-04-24
galera: change listening address/port with haproxy

AbelNavarro:
During our work to make haproxy active in all controller nodes, we found
out Galera is not listening to all interfaces (0.0.0.0) but the VIP
address. This scenario works when there's only one haproxy checking for
Galera but fails on multiple haproxy scenario.

By changing the address and port where Galera is listening we aim to
avoid the conflict for service listening.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -347,7 +347,7 @@ ha_servers = ha_servers.each do |n|
 end
 
 haproxy_loadbalancer "galera" do
-  address CrowbarPacemakerHelper.cluster_vip(node, "admin")
+  address "0.0.0.0"
   port 3306
   mode "tcp"
   # leave some room for pacemaker health checks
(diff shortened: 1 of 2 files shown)


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -347,7 +347,7 @@
 end
 
 haproxy_loadbalancer "galera" do
-  address CrowbarPacemakerHelper.cluster_vip(node, "admin")
+  address "0.0.0.0"
```
rhafer: I am pretty sure we don't want HAproxy to be listening on all interfaces for database traffic. That should really be limited to the admin network. We don't want the database to be accessible e.g. on the public network.

vuntz:
@AbelNavarro that's why we have this https://github.com/crowbar/crowbar-ha/pull/304/files#diff-2b5b3bc6cce6c520ff94fb6439d8ade9R47 actually.

Can you give more details on the issue you're seeing?

chef/cookbooks/mysql/recipes/ha_galera.rb
rhafer: Yeah, `net.ipv4.ip_nonlocal_bind = 1` should avoid that as @vuntz pointed out.

AbelNavarro:
Abandoning this change in favor of https://github.com/crowbar/crowbar-openstack/pull/1647 . We may need to revisit this when/if we have multiple VIPs for the services.

PR #1643, opened 2018-04-24
nova: metadata to listen to all interfaces

AbelNavarro:
In our work to make haproxy active on all controllers we found some
services are listening on the VIP address rather than on all interfaces.
When it listens on the VIP adress and haproxy is configured in several
nodes then there's a problem binding the socket to that VIP address.

This change makes nova-metadata to use a listen-to-all interfaces
(0.0.0.0) as it is configured in all other nova services.


AbelNavarro:
[Make haproxy active/active](https://trello.com/c/7oHIy01J/271-make-haproxy-active-active)

AbelNavarro:
Abandoning this change in favor of https://github.com/crowbar/crowbar-openstack/pull/1647 . We may need to revisit this when/if we have multiple VIPs for the services.

PR #1644, opened 2018-04-24
sahara: Fix trust delete problem (bsc#1033735) 

charleswang007:
According to https://bugzilla.suse.com/show_bug.cgi?id=1033735, a Hadoop-Cluster can not be deleted and is stuck in "deleting" state because of **'failed to delete trust'** error as described in https://bugs.launchpad.net/mos/+bug/1521184. This patch adopts the workaround proposed upstream by setting **use_identity_api_v3** to **_false_**.


charleswang007:
Looks like the original bug https://bugs.launchpad.net/keystone/+bug/1524849 has fix merged in keystone: https://review.openstack.org/#/c/257478/. I will ask the bug reporter to reverify if the bug is still valid.

PR #1647, opened 2018-04-27
mysql: swap haproxy and cluster-check creation

AbelNavarro:
While working in the haproxy active/active feature, we found out that
galera clustercheck script was failing on non founder nodes. We believe
the cause is clustercheck trying to get the status of the service while
haproxy has not still been activated.

By swapping clustercheck and haproxy creation we aim to give time for
clustercheck to find the service active and thus avoid errors.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -290,6 +290,43 @@ crowbar_pacemaker_sync_mark "sync-database_root_password" do
   revision node[:database]["crowbar-revision"]
 end
 
+include_recipe "crowbar-pacemaker::haproxy"
+
+ha_servers = CrowbarPacemakerHelper.haproxy_servers_for_service(
+  node, "mysql", "database-server", "admin_port"
+)
+
+# Let all nodes but one act as backup (standby) servers.
+# Backup server is only used when non-backup one is down. Thus we prevent possible deadlocks
+# from OpenStack services writing to the database on different nodes at once.
+ha_servers = ha_servers.each do |n|
+  # Let the current node be non-backup one, so haproxy running on this node does
+  # not direct traffic elsewhere by default
+  n["backup"] = n["name"] != node["hostname"]
+  # lower the number of unsuccessful checks needed for declaring server DOWN
+  n["fall"] = 2
+  # lower the interval checking after first failure is found
+  n["fastinter"] = 1000
+end
+
+haproxy_loadbalancer "galera" do
+  address CrowbarPacemakerHelper.cluster_vip(node, "admin")
+  port 3306
+  mode "tcp"
+  # leave some room for pacemaker health checks
+  max_connections node[:database][:mysql][:max_connections] - 10
+  options ["httpchk", "clitcpka"]
+  default_server "port 5555"
+  stick ({ "on" => "dst" })
+  servers ha_servers
+  action :nothing
+end.run_action(:create)
+
+# wait for all nodes to have haproxy configured before clustercheck
+crowbar_pacemaker_sync_mark "sync-database_before_clustercheck" do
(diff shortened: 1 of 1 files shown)


AbelNavarro:
[Make haproxy active/active](https://trello.com/c/7oHIy01J/271-make-haproxy-active-active)

jsuchome:
Isn't it contradicting https://github.com/crowbar/crowbar-openstack/pull/1627 ?

AbelNavarro:
This seems to work together with https://github.com/crowbar/crowbar-ha/pull/304 . I had errors before but I couldn't find the exact cause for them, so I blame the environment. On new tries `galera` doesn't show anymore `pacemaker` fail errors.

vuntz:
I'm wondering if there's a need for a sync mark after the haproxy config and before the pacemaker resources, though. I would expect that we would usually have that, and we're probably lucky that it works without this.

vuntz:
> Once haproxy is up, it takes some time for haproxy to discover that the configured services are up and running. So it may happen that haproxy is up but then clustercheck to fail. Would it be fine to add a sleep after the sync mark?

I would think it's not needed, if only because with the sync mark, there will be a sleep in nearly all cases (and it's really only needed on the initial setup, not later).

rhafer:
> While working in the haproxy active/active feature, we found out that
> galera clustercheck script was failing on non founder nodes. We believe
> the cause is clustercheck trying to get the status of the service while
> haproxy has not still been activated.

Can you give some more details on this? This sounds somewhat unexpected. IIRC clustercheck
should not be depending on haproxy to be up and running it should be contacting the mariadb instances
via the nodes admin IPs. Or am I missing your point?

AbelNavarro:
@rhafer you are indeed right sir. My assumption was that `clustercheck` was listening on the haproxy port but I was wrong. Still I think it's worth trying to set the barrier and keep investigating, because this works sometimes (and some don't).

rhafer:
> you are indeed right sir. My assumption was that clustercheck was listening on the haproxy port but I was wrong.

But then clustercheck needs to be started **before** haproxy. As haproxy needs to connect to it. So somehow I think this PR is wrong.

AbelNavarro:
AFAIK haproxy does not connect to clustercheck. The service configured in haproxy connects to the database directly, so does clustercheck. In that sense the order of the two of them should be irrelevant.

Nevertheless, I never got clustercheck to work when it was configured before haproxy. It may be chance. I'm rebuilding the environment and will try to reproduce the problem. I'll ping you so you can take a look.

rhafer:
> AFAIK haproxy does not connect to clustercheck. The service configured in haproxy connects to the database directly, so does clustercheck. In that sense the order of the two of them should be irrelevant.

The "httpchk"  configured for the "galera" backend in haproxy connects to clustercheck (port 5555). So if "clustercheck" is not running on a node, haproxy will consider mariadb on that node to be down and not direct any traffic to it. (That's the whole purpose of the clustercheck service)

AbelNavarro:
The failure I'm observing at the moment is due to galera_monitor pacemaker resource failing, not clustercheck. The monitor ends up working properly but the pacemaker fail count is non-zero, so the deployment aborts. This may be a networking error during installation because rabbitmq has similar problems moments before.

About the clustercheck failure, I'd say I saw an error regarding that specific service, but cannot now. This means this patch may be unnecessary. I will try again only with https://github.com/crowbar/crowbar-ha/pull/304 which was the one showing this problem initially.

AbelNavarro:
I have not seen again that clustercheck failure. About the pacemaker failcount it was due to a galera sync error that happened often. Thanks to @rhafer we could see that the problem was actually in the underlay rather in the change (https://github.com/crowbar/crowbar-ha/pull/304) itself. I'm then closing this PR because the change is considered unnecessary.

PR #1650, opened 2018-05-02
nova: Remove the checks that were testing if we're not running Newton

jsuchome:
Current code applies only to Pike already.


jsuchome:
Ping @toabctl -  this is changing your original patch

PR #1651, opened 2018-05-02
rabbitmq: fix variable typo from users to extra_users

jsuchome:


jsuchome:
Template mentions only `extra_users`, I'm also not sure now if this is the right fix...

vuntz:
I must admit that we still have the code for the non-extra users in the cookbook, so it feels like it should work to use extra_users there indeed.

FWIW, a very simple fix could be to add an empty `users` hash to the default attributes in rabbitmq/attributes/default.rb.

JanZerebecki:
Superseded by https://github.com/crowbar/crowbar-openstack/pull/1652

PR #1654, opened 2018-05-02
monasca: increase virtual memory

zreigz:
Elasticsearch uses a mmapfs directory by default to store its indices.
The default operating system limits on mmap counts is likely to be too low,
which may result in out of memory exceptions.


witekest:
Official Elasticsearch RPM packages [configure this setting automatically](https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html). I'm wondering if we shouldn't follow the same path. @jgrassler What do you think?

jgrassler:
One slight modification: please write all the `limits.conf` settings to a systemd override file for elasticsearch (`/etc/systemd/system/elasticsearch.service.d/override.conf`) with a template that looks like this:

```
[Service]
LimitNPROC={{elasticsearch_max_procs}}
LimitNOFILE={{elasticsearch_max_files_hard_limit}}
LimitMEMLOCK={{elasticsearch_max_locked_memory}}
```

That way we do not need to mess with the global `limits.conf`.

PR #1656, opened 2018-05-03
nova scheduler: default_filters initialization

sandonovsuse:
Initializes nova scheduler default_filters variable.


jsuchome:
Please fixup the 2nd commit to the first one

toabctl:
please squash the 2 commits

dirkmueller:
repository was deleted and the bugreport says that a previous maintenance update arleady fixed the situation, so it looks like this change is no longer needed (the key patched here about default_filters has already been renamed to enabled_filters some time ago)

PR #1657, opened 2018-05-03
Remove SSH-based drivers from ironic (bsc#1087472)

sayalilunkad:
Removing all SSH based drivers from config file and added validaion from ironic
proposal to not allow these drivers.

### DIFF
--- a/chef/cookbooks/ironic/attributes/default.rb
+++ b/chef/cookbooks/ironic/attributes/default.rb
@@ -29,8 +29,6 @@ when "rhel", "suse"
       "openstack-ironic-conductor"
     ],
     driver_dependencies: {
-      pxe_ssh: ["python-paramiko"],
-      agent_ssh: ["python-paramiko"]
     }
   }
   default[:ironic][:api][:service_name] = "openstack-ironic-api"
@@ -50,8 +48,6 @@ when "debian"
       "ironic-api-conductor"
     ],
     driver_dependencies: {
-      pxe_ssh: ["python-paramiko"],
-      agent_ssh: ["python-paramiko"]
     }
   }
   default[:ironic][:api][:service_name] = "ironic-api"
--- a/chef/data_bags/crowbar/migrate/ironic/204_remove_ssh_based_drivers.rb
+++ b/chef/data_bags/crowbar/migrate/ironic/204_remove_ssh_based_drivers.rb
@@ -1,0 +1,5 @@
+def upgrade(ta, td, a, d)
+  a["ironic"]["enabled_drivers"] =
+    a["ironic"]["enabled_drivers"].reject { |driver| driver.include?("ssh") }
+  return a, d
+end
--- a/crowbar_framework/app/models/ironic_service.rb
+++ b/crowbar_framework/app/models/ironic_service.rb
@@ -94,6 +94,10 @@ class IronicService < ServiceObject
       validation_error I18n.t("barclamp.#{@bc_name}.validation.no_drivers")
     end
 
+    if proposal["attributes"][@bc_name]["enabled_drivers"].any? { |d| d.include?("ssh") }
+      validation_error I18n.t("barclamp.#{@bc_name}.validation.wrong_drivers")
+    end
+
     # additional soft dependencies for agent_* drivers
     if proposal["attributes"][@bc_name]["enabled_drivers"].any? { |d| d.start_with?("agent_") }
       swift_svc = SwiftService.new @logger
--- a/chef/data_bags/crowbar/template-ironic.json
+++ b/chef/data_bags/crowbar/template-ironic.json
@@ -31,7 +31,7 @@
   "deployment": {
     "ironic": {
       "crowbar-revision": 0,
-      "schema-revision": 203,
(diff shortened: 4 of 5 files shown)


chef/cookbooks/ironic/recipes/server.rb
```diff
@@ -54,9 +54,13 @@
 end
 
 node[:ironic][:enabled_drivers].each do |d|
-  driver_dependencies = node[:ironic][:platform][:driver_dependencies][d] || []
-  driver_dependencies.each do |p|
-    package p
+  if d.include?("ssh")
+    node[:ironic][:enabled_drivers].delete(d)
```
toabctl: Not sure, but shouldn't that be done via a crowbar migration?

toabctl: Yes. But don't we want to delete these drivers also from the proposal? My understanding is, that theses drivers are no longer valid so they should not stay in the crowbar ironic proposal.

skazi0: I was going to suggest the same. The question is: would it only fix the proposal or also node attributes (i.e. would users need to apply the proposal)? Maybe both would be needed?
@sayalilunkad if you insist on doing it in the cookbook, I would copy `node[:ironic][:enabled_drivers]` to a local variable, filter that and use it in the following places to make it clear that you don't modify the node attributes.

sayalilunkad: @toabctl @skazi0 yes that makes sense that we remove it from the proposal. I think this should update the node attribute as well but I will check.

chef/data_bags/crowbar/migrate/ironic/204_remove_ssh_based_drivers.rb
```diff
@@ -0,0 +1,12 @@
+def upgrade(ta, td, a, d)
+  a["ironic"]["enabled_drivers"] = ta["ironic"]["enabled_drivers"]
+  a["ironic"]["enabled_drivers"].each do |driver|
+    a["ironic"]["enabled_drivers"].delete(driver) if driver.include?("ssh")
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["ironic"]["enabled_drivers"] = ta["ironic"]["enabled_drivers"]
```
skazi0: I don't think we should do anything in downgrade.

chef/data_bags/crowbar/migrate/ironic/204_remove_ssh_based_drivers.rb
```diff
@@ -0,0 +1,12 @@
+def upgrade(ta, td, a, d)
+  a["ironic"]["enabled_drivers"] = ta["ironic"]["enabled_drivers"]
```
skazi0: Why overwrite user's drivers with default?

chef/data_bags/crowbar/migrate/ironic/204_remove_ssh_based_drivers.rb
```diff
@@ -0,0 +1,12 @@
...
+  a["ironic"]["enabled_drivers"].each do |driver|
```
skazi0: Maybe something like this?
`a["ironic"]["enabled_drivers"] = a["ironic"]["enabled_drivers"].reject { |d| d.include? "ssh" }`

chef/cookbooks/ironic/attributes/default.rb
```diff
@@ -28,10 +28,6 @@
       "openstack-ironic-api",
       "openstack-ironic-conductor"
     ],
-    driver_dependencies: {
```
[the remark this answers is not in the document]
sayalilunkad: Leaving a empty hastable for now incase there are other dependencies in the future.

crowbar_framework/app/models/ironic_service.rb
```diff
@@ -94,6 +94,10 @@ def validate_proposal_after_save(proposal)
       validation_error I18n.t("barclamp.#{@bc_name}.validation.no_drivers")
     end
 
+    if proposal["attributes"][@bc_name]["enabled_drivers"].any? { |d| d.include?("ssh") }
```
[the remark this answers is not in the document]
sayalilunkad: @skazi0 I agree.

chef/data_bags/crowbar/migrate/ironic/204_remove_ssh_based_drivers.rb
```diff
@@ -0,0 +1,5 @@
+def upgrade(ta, td, a, d)
+  a["ironic"]["enabled_drivers"] =
+    a["ironic"]["enabled_drivers"].reject { |driver| driver.include?("ssh") }
+  return a, d
```
toabctl: Did you test the downgrade? Does it work if the function is not defined? Maybe we need an empty function?

crowbar_framework/app/models/ironic_service.rb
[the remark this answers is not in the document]
toabctl: oki. thanks for clarification

PR #1658, opened 2018-05-03
mysql: make overrides for mysql version work again

seife:
Apparently, attributes/server.rb is evaluated fully before overrides.rb
is considered. This breaks overriding mysql.mariadb.version, because
mariadb-galera is always added mysql.galera.packages.
Move the version check out of attributes/ into recipes/ha_galera.rb to
fix this.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -19,6 +19,14 @@
 
 resource_agent = "ocf:heartbeat:galera"
 
+# this was in attributes/server.rb where it did not work, because it
+# was evaluated too early for overrides to be useful
+#
+# newer version need an additional package on SLES
+unless node[:mysql][:mariadb][:version] == "10.1"
+  node[:mysql][:galera_packages] << "mariadb-galera"
+end
+
 node[:mysql][:galera_packages].each do |p|
   package p
 end
--- a/chef/cookbooks/mysql/attributes/server.rb
+++ b/chef/cookbooks/mysql/attributes/server.rb
@@ -48,8 +48,4 @@ default[:mysql][:galera_packages] = [
   "socat",
   "galera-python-clustercheck"
 ]
-
-# newer version need an additional package on SLES
-unless node[:mysql][:mariadb][:version] == "10.1"
-  default[:mysql][:galera_packages] << "mariadb-galera"
-end
+# for versions != 10.1, mariadb-galera is added in the ha_galera.rb recipe


AbelNavarro:
in this particular case @seife tried to set mysql version to 10.1 in an override file, but 10.2 kept being applied and this `mariadb-galera` package was installed. My guess is that at [this](https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/mysql/attributes/server.rb#L50) point the override hasn't yet kicked in so the package list doesn't change. I think @seife approach is in the right direction.

AbelNavarro:
[Fix override for mysql version](https://trello.com/c/hInJKzwD/280-fix-override-for-mysql-version)

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -19,6 +19,14 @@
 
 resource_agent = "ocf:heartbeat:galera"
 
+# this was in attributes/server.rb where it did not work, because it
+# was evaluated too early for overrides to be useful
+#
+# newer version need an additional package on SLES
+unless node[:mysql][:mariadb][:version] == "10.1"
+  node[:mysql][:galera_packages] << "mariadb-galera"
```
vuntz: I don't have a good solution (right now) for your problem, but I don't think this fix is good: this basically makes this attribute a normal node attribute, instead of being a default attribute, and each run will just add the package again and again, ending up with a huge list.

seife: Ok, I did of course not think of *that* problem. Will try to fix that somehow.

rhafer:
@seife can you share some details about how/where you added the override attribute? It works just fine here to add an "override.rb" to attributes/ in the mysql cookbook with this content:

```
override[:mysql][:mariadb][:version] = "10.1"
```

re-upload the cookbook and apply the database barclamp. (That was on a completely clean installation with no barclamps deployed yet)

seife:
I did it exactly the same (override.rb), this is the excerpt from the mkphsycloud log file, the function that called this was `pre_do_installcrowbar`:
```
test -e /opt/dell/chef/cookbooks/mysql/attributes/override.rb || echo '\''override[:mysql][:ha][:clustercheck] = true
override[:mysql][:mariadb][:version] = "10.1"'\'' > /opt/dell/chef/cookbooks/mysql/attributes/override.rb'
```

The funny thing: after an hint by @vuntz, I renamed the override.rb to `aaa_i_am_first.rb` and then it worked. Actually, bothe "override.rb" and "aaa_i_am_first.rb" both sort alphabetically before "server.rb".

My guess is, that chef pull this in in `opendir(3)` / `readdir(3)` order and it is just luck if it works or not.
I'll try a few more runs with the file named override.rb to see if it consistently fails for me or if it is a random failure.

rsalevsky:
@seife AFAIK you are building packages correct? If yes, I suggest to take the latest version from https://build.opensuse.org/package/show/systemsmanagement:crowbar:4.0/crowbar-openstack with the revert-switch-to-mariadb-10.2.patch and enable line 95 in the spec. That's the way how we deliver functional updates to our customers.

seife:
Ok, fine with me. I had something similar in our OBS instance, but @vuntz reminded us, that overrides are the proposed way to go ;-) It puzzles me that they seem to work only sometimes.

rhafer:
@seife Is this still needed, or did you go a with a different solution? I did multiple test runs (using various file names for the overide) and it always worked for me. Maybe I was just lucky.

seife:
No, as @rsalevsky pointed out we adopted the "revert to 10.1" patch in our package, so this PR is obslete.

PR #1660, opened 2018-05-07
upgrade: Do not check if rabbitmq is running when it is not supposed to run

jsuchome:
pre-upgrade attribute is set to true to indicate the case that the upgrade of the node has
not been finished yet. In such case, services cannot start on the node
(there's a location constraint that prevents that).
See disable_pre_upgrade_attribute_for method in models/api/node.rb for more info

See also commit 54aad4646798bf933d4ed0d9d937342a12644e0c

This should fix the cases for "shared storage" option. Referenced commit targeted only the rabbitmq clustered setup.


jsuchome:
I believe this was replaced by some better solution

PR #1663, opened 2018-05-07
rabbitmq: do not set the blocking mechanism if upgrade ongoing

Itxaka:
When we upgrade we can end up with only one rabbit node up on the first
chef join, so blocking rabbit until the cluster is up may not be
the most smart thing to do

Instead check if we are on upgrade status and if so, do not deploy
the blocking mechanism and instead remove it

(cherry picked from commit a32f607fb575420d9b09883ca720c08aeed187bd)


Itxaka:
not yet

PR #1665, opened 2018-05-07
Revert "rabbitmq: block client port on startup"

Itxaka:
Reverts crowbar/crowbar-openstack#1554 due to unintended issues:

 - Random failures while trying to create the pacemaker alerts -> http://paste.suse.de/17495
 - Blocks upgrade as rabbit is left constantly blocked and the iptable rule is not cleaned up properly on resource deletion


jsuchome:
https://bugzilla.suse.com/show_bug.cgi?id=1092180 is the upgrade related bug

PR #1669, opened 2018-05-08
[4.0] rabbitmq: do not set the blocking mechanism if upgrade ongoing

Itxaka:
When we upgrade we can end up with only one rabbit node up on the first
chef join, so blocking rabbit until the cluster is up may not be
the most smart thing to do

Instead check if we are on upgrade status and if so, do not deploy
the blocking mechanism and instead remove it

(cherry picked from commit aa8ed14bb6c4853e1d782da524b54519c50e15cc)

Backport-of: https://github.com/crowbar/crowbar-openstack/pull/1661


Itxaka:
Reverting the whole thing instead

PR #1672, opened 2018-05-21
nova: Do not do partial online migrations, that was Newton specific

jsuchome:
We needed to do step by step upgrade (sync up to certain version + online migrations + new synce + new online migrations) when upgrading
to Newton. This code should not be needed any more in master branch.

I *think* this is correct, but it's not yet tested neither during installation nor during upgrade... 
**UPDATE** gating success confirms that it does not break normal installation...

### DIFF
--- a/chef/cookbooks/nova/recipes/database.rb
+++ b/chef/cookbooks/nova/recipes/database.rb
@@ -139,22 +139,19 @@ execute "nova-manage api_db sync" do
   end
 end
 
-execute "nova-manage db sync up to revision 329" do
+execute "nova-manage db sync" do
   user node[:nova][:user]
   group node[:nova][:group]
-  command "nova-manage db sync 329"
+  command "nova-manage db sync"
   action :run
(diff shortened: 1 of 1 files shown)


jsuchome:
Works for me during the upgrade as well. 

I did have 3 instances running before the upgrade, but I'm not sure if those potentially problematic tables had any entries on Newton side.

jsuchome:
@jgrassler @matthewoliver please see 2nd commit.

Alternatively, I can subscribe the action to different block by
`subscribes :create, "execute[nova-manage cell_v2 map_instances]", :immediately`

But somehow I do not see a reason why we're using `:immediately` in this place.

jgrassler:
The `:immediately` seems superfluous indeed, but as long as it works I'd lean towards leaving it in there.

jsuchome:
@jgrassler so it seems current state works well with installation, and worked for me during upgrade as well

jsuchome:
@jgrassler @matthewoliver another round of review please?

PR #1673, opened 2018-05-24
rabbitmq: install and configure yarb to launch automatically

Itxaka:
Installs and configures yarb to balance the rabbit queues
automagically, by using ClusterMon to catch cluster actions and
react to any rabbitmq coming back from a restart or promotion to
lunach yarb and balance the queues.


chef/cookbooks/rabbitmq/recipes/yarb.rb
```diff
@@ -0,0 +1,75 @@
...
+  agent "ocf:pacemaker:ClusterMon"
+  op clustermon_op
+  params clustermon_params
+  action :update
+  only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+pacemaker_ms "ms-autoyarb" do
+  rsc "autoyarb"
+  meta ({
```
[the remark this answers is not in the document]
Itxaka: This needs to be like this, not a choice

Itxaka:
[Investigate yarb automation after node outage](https://trello.com/c/4HgSYQ05/282-investigate-yarb-automation-after-node-outage)

chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -57,14 +57,19 @@
...
+        "wait_time": 1
+      }
     }
   },
   "deployment": {
     "rabbitmq": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 204,
+      "schema-revision": 300,
```
AbelNavarro: I think it's a non-written rule that the schema revision keeps the first digit on the same SOC version. So in this case it should be a 2xx number.

chef/cookbooks/rabbitmq/recipes/yarb.rb
```diff
@@ -0,0 +1,77 @@
...
+#
+# Sets yarb to auto run after a rabbit cluster failure in order to auto level the
+# rabbit queues
+#
+
+clustermon_op = { "monitor" => [{ "interval" => "10s" }] }
+clustermon_params = { "extra_options" => "-E /usr/bin/yarb-alert-pacemaker.sh --watch-fencing" }
+name = "autoyarb"
+
+package "python-yarb"
```
AbelNavarro: is this package already in our repositories? and for customers?

Itxaka: yup!

chef/data_bags/crowbar/template-rabbitmq.json
Itxaka: but I thought that numbers where increased on each version so cloud6 -> 000, cloud7 -> 100, cloud8 -> 200, and the future cloud9 will start at 300, which this PR should be part of, as we already released cloud8

Not like this changes anything but its a visual indication of where the migration comes from. Migrations are run anyway.

rsalevsky:
Merging?

Itxaka:
rebased!

Itxaka:
Removing fails because the zypper resource on chef does not deal properly with empty versions. Im trying to submit a patch to obs but Im having issues applying it properly so this is blocked for now (I wonder if I can pass an empty version to the resource in order to avoid this issue)

Itxaka:
patch for fixing zypper resource removing is merged, needs retesting

chef/cookbooks/rabbitmq/recipes/yarb.rb
```diff
@@ -0,0 +1,142 @@
...
+
+  pacemaker_location "l-#{name}" do
+    definition OpenStackHAHelper.controller_only_location("l-#{name}", "g-#{name}")
+    action :delete
+    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
+
+  pacemaker_group "g-#{name}" do
+    members [name]
+    action :stop
```
AbelNavarro: you can compact the `:stop` and `:delete` actions as `[:stop, :delete]`. See https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/rabbitmq/recipes/ha.rb#L353

Itxaka: done!

chef/cookbooks/rabbitmq/recipes/rabbit.rb
```diff
@@ -116,6 +116,12 @@
   only_if only_if_command if ha_enabled
 end
 
+execute "rabbitmqctl set_user_tags #{node[:rabbitmq][:user]} administrator" do
+  not_if "rabbitmqctl list_users | grep #{node[:rabbitmq][:user]} | grep -q administrator"
```
dirkmueller: I would like not_if near the only_if

chef/cookbooks/rabbitmq/recipes/yarb.rb
```diff
@@ -0,0 +1,122 @@
...
+group_running = "crm resource show g-#{name}"
+
+if node[:rabbitmq][:yarb][:enabled]
+
+  package "python-yarb"
+
+  template "#{ENV["HOME"]}/.yarb.conf" do
+    source "yarb.conf.erb"
+    owner "root"
+    group "root"
```
dirkmueller: contains credentials, so we need to set mode to 0700 imho.

chef/cookbooks/rabbitmq/templates/default/yarb-alert-pacemaker.sh.erb
```diff
@@ -0,0 +1,18 @@
...
+else
+  node_type="slave"
+fi
+
+logger -t "yarb-alert-pacemaker" "A rabbit ${node_type} node is back: ${CRM_notify_node}. Launching yarb..."
+
+# lock in case there is 2 node recoveries really close
+# we dont care on launching yarb multiple times, because everytime we get that alert it
+# means there is a new node up so the queues should be rebalanced
+flock /var/lock/yarb /usr/bin/yarb
```
dirkmueller: so this directory doesn't exist by default? and how does that lock across nodes?

chef/cookbooks/rabbitmq/recipes/yarb.rb
Itxaka: done

chef/cookbooks/rabbitmq/recipes/rabbit.rb
Itxaka: ok

chef/cookbooks/rabbitmq/recipes/yarb.rb
```diff
@@ -0,0 +1,122 @@
...
+
+  template "/usr/bin/yarb-alert-pacemaker.sh" do
+    source "yarb-alert-pacemaker.sh.erb"
+    owner "root"
+    group "root"
+    mode "0755"
+  end
+
+  pacemaker_primitive name do
+    agent "ocf:pacemaker:ClusterMon"
```
dirkmueller: whats this resource doing? is that cloning it on every node or only on one of the node in the cluster? where is the "if not use_systemd for cloning" part?

chef/cookbooks/rabbitmq/templates/default/yarb-alert-pacemaker.sh.erb
Itxaka: >  and how does that lock across nodes?
doesnt matter across nodes as the service can only run in one of them at the time

> so this directory doesn't exist by default?
/var/lock exists by default, this creates the yarb file inside it.

chef/cookbooks/rabbitmq/recipes/yarb.rb
Itxaka: > whats this resource doing?

```The ocf:pacemaker:ClusterMon resource can monitor the cluster status and triggers alerts on each cluster event. This resource runs crm_mon in the background at regular intervals (configurable) and uses crm_mon capabilities to send emails (SMTP), SNMP traps or to execute an external program via the extra_options parameter. ```

> is that cloning it on every node or only on one of the node in the cluster? 

only on one


> where is the "if not use_systemd for cloning" part?

I dont understand this part sorry. Is this regarding the way of managing the cluster resources by using systemd instead of pacemaker? If its that in this case we want to use pacemaker as we want only one of this running and we want it on HA.

PR #1675, opened 2018-05-31
Add elasticsearch tunables

jgrassler:
This commits exposes various `elasticsearch` tunables that have recently been added to `monasca-installer`. Please do not merge, yet. I'm still adapting this and `monasca-installer` to each other. Also, I'll rebase and clean up the commit message once it actually works :-)

### DIFF
--- a/chef/cookbooks/monasca/recipes/master.rb
+++ b/chef/cookbooks/monasca/recipes/master.rb
@@ -125,6 +125,7 @@ template "/opt/monasca-installer/crowbar_vars.yml" do
     curator_cron_config: [curator_cron_config].to_yaml.split("\n")[1..-1],
     curator_excluded_index: curator_excluded_index.to_yaml.split("\n")[1..-1],
     elasticsearch_repo_dir: node[:monasca][:elasticsearch][:repo_dir].to_yaml.split("\n")[1..-1],
+    elasticsearch_tunables: node[:monasca][:elasticsearch][:tunables],
     monitor_libvirt: node[:monasca][:agent][:monitor_libvirt],
     delegate_role: node[:monasca][:delegate_role]
   )
--- a/chef/cookbooks/monasca/templates/default/crowbar_vars.yml.erb
+++ b/chef/cookbooks/monasca/templates/default/crowbar_vars.yml.erb
@@ -64,6 +64,9 @@ monasca_log_api_url: "http://<%= @pub_net_ip %>:<%= @log_api_settings['bind_port
 memcached_nodes: ["<%= @monasca_net_ip %>:11211"]
 elasticsearch_nodes: ["<%= @monasca_net_ip %>"]
 elasticsearch_hosts: <%= @monasca_net_ip %>
+<%- @elasticsearch_tunables.each_key do |t| %>
+elasticsearch_<%= t %>: <%= @elasticsearch_tunables[t] %>
+<%- end %>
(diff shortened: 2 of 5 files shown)


jgrassler:
Ok, tested and works. Feel free to review but please do not merge, yet: This requires https://gitlab.com/monasca-installer/ansible-monasca-elasticsearch/merge_requests/4 in place to work.

chef/cookbooks/monasca/recipes/master.rb
```diff
@@ -108,6 +108,18 @@
...
+
+["heap_size",
+ "max_locked_memory",
+ "max_open_files_hard_limit",
+ "max_open_files_soft_limit",
+ "max_procs",
+ "memory_lock",
+ "vm_max_map_count"].each do |k|
+  elasticsearch_tunables[k] = node[:monasca][:elasticsearch][k]
+end
```
nicolasbock: It would be easier to simply loop through the `keys()` instead of hardcoding the keys here. This also makes a future change in the tunables easier to maintain.

jgrassler: It would, if `node[:monasca][:elasticsearch]` only contained the tunables. If you prefer, I can remove the non-tunable items from the `keys()` list, but then we'll have the same problem if we add any fields that are not tunables in the future...

jgrassler:
I fixed the schema revision now, thanks for spotting!

chef/cookbooks/monasca/recipes/master.rb
nicolasbock: There is only one non-tunable item in there, isn't it (`repo_dir`)? I see your point though, I had missed that. You could create another nesting level such as `node[:monasca][:elasticsearch][:tunables]` so that you can loop. I think the less duplication of the keys we have the better.

jgrassler: Yeah indeed...that shortens the code considerably. I moved them to a dedicated key now. I'll create a fresh cloud now and test it tomorrow to make sure I didn't break anything. Once I'm sure, I'll rebase. Until then I'll leave the extra commit around.

jgrassler:
Ok, tested and works fine with the tunables moved to a dedicated `:tunables` key. I rebased already.

chef/cookbooks/monasca/templates/default/crowbar_vars.yml.erb
```diff
@@ -64,9 +64,13 @@ monasca_log_api_url: "http://<%= @pub_net_ip %>:<%= @log_api_settings['bind_port
 memcached_nodes: ["<%= @monasca_net_ip %>:11211"]
 elasticsearch_nodes: ["<%= @monasca_net_ip %>"]
 elasticsearch_hosts: <%= @monasca_net_ip %>
+<%- @elasticsearch_tunables.each_key do |t| %>
+elasticsearch_<%= t %>: <%= @elasticsearch_tunables[t] %>
+<%- end %>
 monasca_api_log_level: <%= @api_settings['log_level'] %>
 log_api_log_level: <%= @log_api_settings['log_level'] %>
 
+
```
dirkmueller: NIT: extra line?

jgrassler: Sorry, missed that one. Fixed.

PR #1677, opened 2018-06-04
Add rate limiting for glance api (bsc#1005886)

JanZerebecki:
Disabled by default default. It can be set to avoid filling up image related tables.

Though the tables are only filled by POST requests this limit is for all
request types.

See https://wiki.openstack.org/wiki/OSSN/OSSN-0076 for details.

### DIFF
--- a/chef/data_bags/crowbar/template-glance.json
+++ b/chef/data_bags/crowbar/template-glance.json
@@ -65,14 +65,17 @@
       "service_user": "glance",
       "database_instance": "none",
       "rabbitmq_instance": "none",
-      "memcache_secret_key": ""
+      "memcache_secret_key": "",
+      "ha_rate_limit": {
+        "glance-api": 0
+      }
     }
   },
   "deployment": {
     "glance": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 201,
+      "schema-revision": 202,
       "element_states": {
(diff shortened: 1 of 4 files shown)


JanZerebecki:
I haven't checked yet, if 2 is a sane default.

nicolasbock:
@JanZerebecki Why not set the default to `0` as in the referenced PR so that the operator can choose a sensible value? For instance, if the rate applies across all users then it will have to scale somewhat with the number of users.

JanZerebecki:
Changed default to 0.

chef/data_bags/crowbar/template-glance.json
```diff
@@ -65,14 +65,17 @@
...
+        "glance-api": 0
+      }
     }
   },
   "deployment": {
     "glance": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 201,
+      "schema-revision": 202,
```
Itxaka: I think now we need to jump to 300 as we had 100 for cloud6 and 200 for cloud8, it makes sense to start with 300 for cloud9?

Sorry for the nitpick, feel free to ignore and merge :)

JanZerebecki: This patch will be backported to Cloud7 and there is no branch for Cloud8 yet, so lets see if can be merged before the branch is created.

PR #1678, opened 2018-06-08
rabbitmq: optimize for max connections

Itxaka:
Allows marking a setting to optimize the underlying erlang and
kernel settings to deal with a higher number of connections,
especially with burst connections.
This will require more resources and may affect throughput.

Adds extra erlangs arguments:
 - +A: Sets the number of threads in async thread pool.
Valid range is 0-1024. Defaults to 128. Set to double of that.

 - +P: Sets the maximum number of simultaneously existing
processes for this system. Valid range for is [1024-134217727].
Set to half of maximum value.

Adds extra kernel params:
 - net.ipv4.ip_local_port_range: increase the range of ports available.
Default is 27018 to 64999. Set to 15000 to 65000

 - net.ipv4.tcp_fin_timeout: reduces the amount of time closed connections
will stay in the TIME_WAIT state. Default is 60. Set to 10

 - net.core.somaxconn: size of the listen queue at the same time.
Default is 128. Set to 4096.

 - net.ipv4.tcp_max_syn_backlog: maximum number of remembered connection
requests which did not receive an acknowledgment yet from connecting client.
Default is 128. Mainly for throughput but it may help with the burst.
Set to 4096.


Itxaka:
Requires backport to cloud7 but it wont work as the previous patch this is based on was not ported due to cloud7 only having rabbbit 3.4.

So to backport this, this needs to be included: https://github.com/crowbar/crowbar-openstack/pull/1543

Itxaka:
[Investigate why rabbitmq cannot deal with all connections when one node goes down](https://trello.com/c/O47WjPyI/292-investigate-why-rabbitmq-cannot-deal-with-all-connections-when-one-node-goes-down)

vuntz:
Two questions:

 - can you think of any impact if rabbitmq is running on the same node as openstack API or on an openstack neutron network node?

 - is it possibly useful to make the values configurable? Or are we good with hard-coded?

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -115,6 +124,24 @@
...
+    source "50-rabbitmq.conf"
+    notifies :restart, "service[rabbitmq-server]"
+  end
+
+  bash "reload sysctl" do
+    code "/sbin/sysctl -e -p /etc/sysctl.d/50-rabbitmq.conf"
+    action :nothing
+    subscribes :run, "cookbook_file[/etc/sysctl.d/50-rabbitmq.conf]", :immediate
+  end
+
```
vuntz: Don't you need a `else` to remove the sysctl file?

Itxaka:
> can you think of any impact if rabbitmq is running on the same node as openstack API or on an openstack neutron network node?

Yes. Anything running in the same node as rabbit will have to share RAM and increasing max connections also increases the RAM by a good margin. So mainly resource-wise.
Rabbit should always be run by itself if you want to use this max connections setting.

> is it possibly useful to make the values configurable? Or are we good with hard-coded?

Yep, will probably do that next, this was a first round to make sure I didnt break anything :)

chef/cookbooks/rabbitmq/recipes/default.rb
Itxaka: Good point, will add!

chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -57,14 +57,21 @@
...
+        "tcp_max_syn_backlog": 4096
+      }
     }
   },
   "deployment": {
     "rabbitmq": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 204,
+      "schema-revision": 300,
```
AbelNavarro: same comment as with the other PR. Shouldn't be this number in the 2xx range?

chef/cookbooks/rabbitmq/templates/default/rabbitmq-env.conf.erb
```diff
@@ -12,3 +12,6 @@ LOG_BASE=<%= node[:rabbitmq][:logdir] %>
 <% if node[:rabbitmq][:mnesiadir] -%>
 MNESIA_BASE=<%= node[:rabbitmq][:mnesiadir] %>
 <% end -%>
+<% if node[:rabbitmq][:optimize_max_conn][:enabled] -%>
+SERVER_ADDITIONAL_ERL_ARGS="+A 256 +P 67108862"
```
AbelNavarro: This is a bit obscure and difficult to trace why it is needed. I would add a comment explaining those numbers.

chef/data_bags/crowbar/migrate/rabbitmq/205_add_optimize_max_conn.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(template_attributes, template_deployment, attributes, deployment)
```
AbelNavarro: nice to see the full names for the variables instead of the usual 'ta, td, a, d'.

chef/cookbooks/rabbitmq/templates/default/50-rabbitmq.conf.erb
```diff
@@ -0,0 +1,14 @@
+# Local IP port range, define as a pair of values. The range must provide enough entries for the peak number of concurrent connections.
```
AbelNavarro: Changing the range of the local ports can be dangerous because some services may be living within that range (i.e. libvirtd at port 16509). Unfortunately we don't have a means to show a warning to the user about this. I cannot think of better alternatives to address this.

chef/cookbooks/rabbitmq/templates/default/rabbitmq-env.conf.erb
Itxaka: Sounds reasonable :+1:

chef/cookbooks/rabbitmq/templates/default/rabbitmq-env.conf.erb
```diff
@@ -12,3 +12,11 @@ LOG_BASE=<%= node[:rabbitmq][:logdir] %>
 <% if node[:rabbitmq][:mnesiadir] -%>
 MNESIA_BASE=<%= node[:rabbitmq][:mnesiadir] %>
 <% end -%>
+<% if node[:rabbitmq][:optimize_max_conn][:enabled] -%>
+# optimization for erlang to server more connections in exchange for more resources
+# +A: Sets the number of threads in async thread pool.
+# Valid range is 0-1024. Defaults to 128. Set to double of that.
+# +P: Sets the maximum number of simultaneously existing processes for this system
+# Valid range for is [1024-134217727]. Set to half of maximum value.
+SERVER_ADDITIONAL_ERL_ARGS="+A 256 +P 67108862"
```
dirkmueller: The default seems to be autotuning actually, see 

https://github.com/rabbitmq/rabbitmq-server/issues/151

Are we doing something good by hardcoding this?

Itxaka: As far as I understood `+A` adds additional threads to the already configured ones, so you could set this to 0 and you would still get `16 * cores` threads anyway.

dirkmueller: I don't think thats what it does (based on my testing)

chef/cookbooks/rabbitmq/templates/default/50-rabbitmq.conf.erb
dirkmueller: The network barclamp is also setting this value, so overwriting it here doesn't make sense. also the local_port_range should not matter for the *server* side should it?

chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -57,14 +57,21 @@
         "sndbuf": null,
         "recbuf": null
       },
-      "collect_statistics_interval": 5000
+      "collect_statistics_interval": 5000,
+      "optimize_max_conn": {
+        "enabled": false,
+        "ip_local_port_range": [10000, 65000],
+        "tcp_fin_timeout": 10,
```
dirkmueller: this is a dangerous default from what I can see. a default is 60s, and the rabbitmq tuning guide recommends ranges for that in the range of 15-30s.

chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -57,14 +57,21 @@
...
+        "somaxconn": 4096,
+        "tcp_max_syn_backlog": 4096
```
dirkmueller: I think this is a dangerous default. the SLES default is 256 with syn cookies enabled, which means it will accept up to 256 new connections before user space accepts it. keeping that large means a hanging / busy rabbitmq process is masked by the kernel keeping the tcp connect "stalling" in the open phase, which is not expected by the amp client side libraries. the tunables you found on rabbitmq is for "IOT" scenarios, aka extremely high latency extremely high concurrency ("hundreds of thousands of parallel connections"). the openstack usecase is very low latency and not high concurrency (connection counts in the order of 10000). I do not believe this is a useful setting.

chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -57,14 +57,21 @@
         "sndbuf": null,
         "recbuf": null
       },
-      "collect_statistics_interval": 5000
+      "collect_statistics_interval": 5000,
+      "optimize_max_conn": {
+        "enabled": false,
+        "ip_local_port_range": [10000, 65000],
```
dirkmueller: this is overlapping with what the network barclamp already configures. we should not set this in rabbitmq.

chef/cookbooks/rabbitmq/templates/default/rabbitmq-env.conf.erb
Itxaka: Indeed it doesnt as confirmed on SAP side

Itxaka:
This do not seem like a good idea

PR #1680, opened 2018-06-12
upgrade: cinder: run live migrations at correct rev

matthewoliver:
The Pike release of OpenStack Cinder requires the live migrations
in Ocata to be complete. A check in the database sync at version
97 enforces this.

This patch follows the conventions established in our Nova upgrade:

 1. db sync up to version 96
 2. Run the required live-migrations
 3. Complete the database sync

We are skipping the Ocata release and instead running the migrations
from a Pike codebase. However, Pike's sqlalchemy models include an
extra column to the groups database table that is therefore required
for the live migrations. Thus, we will copy the column creation from                                                                                                                                               
db version 102 into a placeholder at 96. This will ensure the column
is created so that the Pike code is capable of running the live
migration.

### DIFF
--- a/chef/cookbooks/cinder/recipes/sql.rb
+++ b/chef/cookbooks/cinder/recipes/sql.rb
@@ -61,6 +61,37 @@ database_user "grant database access for cinder database user" do
     only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
+# Some live migrations need to happen before revision 97, so we first need to
+# sync all revisions up to 96
+execute "cinder-manage db sync up to revision 96" do
+  command "cinder-manage db sync 96"
+  user node[:cinder][:user]
+  group node[:cinder][:group]
+  ignore_failure true
+  # We only do the sync the first time, and only if we're not doing HA or if we
+  # are the founder of the HA cluster (so that it's really only done once).
+  only_if do
+    !node[:cinder][:db_synced] &&
+      (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)) &&
+      (`cinder-manage --log-file /dev/null db version`.to_i < 97)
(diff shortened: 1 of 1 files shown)


matthewoliver:
This is the first version that works, but now that I discovered the craziness of db version 97 I'm thinking it might be better to add the adding of the missing column (thanks pike) to db version 96. Db version 96 is a place holder version, so was unsure if it was smart to use it or not. Seeing as it isn't used, it seems like the best way forward, then we can:

  db sync -> 96
  live_migrations
  db_sync

Which is much easier to grok then what the code currently does.

Thoughs?

jsuchome:
I think going to 96 before running online migrations really makes sense

rhafer:
@matthewoliver 

> Seeing as it isn't used, it seems like the best way forward, then we can:
> 
> db sync -> 96
> live_migrations
> db_sync
> 
> Which is much easier to grok then what the code currently does.
> 
> Thoughs?

I agree. That approach sounds somewhat cleaner.

jhesketh:
I'm fine with this as well, but as discussed in standup it is bad downstream practice to be modifying those placeholders. That said, the risk of them being changed is minimal (it's more likely to happen during an N+1 development cycle) and also something that we can work around should it become necessary.

chef/cookbooks/cinder/recipes/sql.rb
```diff
@@ -61,6 +61,34 @@
     only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
+execute "cinder-manage db sync up to revision 96" do
```
jhesketh: Lets add a comment to explain why we're only syncing up this far.

jsuchome:
I can confirm it has worked for me with the right cinder packages installed on time

PR #1681, opened 2018-06-13
[5.0] rabbitmq: block client port on startup

Itxaka:
This script blocks the connection to the rabbitmq cluster in case the number of nodes decay below the half
of the total. In this case the remaining rabbit nodes won't accept new connections until quorum is reached.

It takes advantage of the ClusterMon agent that notifies when a rabbitmq has failed or has restored.

The reasoning behind this patch is to allow waiting for more than one node in the cluster to be ready,
which is useful for two reasons:

 - clients would connect to more than one node
 - master for queues would be spread more evenly

Currently most of the connections and master queues are created on the master node, as its the first
to come up when the cluster is restarted/new master promoted.

co-authored-by: Ivan Lausuch <ilausuch@suse.com>

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
+++ b/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
@@ -159,3 +159,141 @@ pacemaker_transaction "rabbitmq service" do
 end
 
 crowbar_pacemaker_sync_mark "create-rabbitmq_ha_resources"
+
+clustermon_op = { "monitor" => [{ "interval" => "10s" }] }
+clustermon_params = { "extra_options" => "-E /usr/bin/rabbitmq-alert-handler.sh --watch-fencing" }
+name = "rabbitmq-port-blocker"
+clone_name = "cl-#{name}"
+location_name = "l-#{name}-controller"
+node_upgrading = CrowbarPacemakerHelper.being_upgraded?(node)
+clone_running = "crm resource show #{clone_name}"
+primitive_running = "crm resource show #{name}"
+port = node[:rabbitmq][:port]
+ssl_port = node[:rabbitmq][:ssl][:port]
+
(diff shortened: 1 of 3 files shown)


jsuchome:
So this was already reverted once, am I right? What's the fix?

Itxaka:
> So this was already reverted once, am I right? What's the fix?

no fix yet @jsuchome , Im just testing it and cant get the failure to happen at all :/

Itxaka:

ilausuch:
The SSL ports has to be blocked too. WIP

ilausuch:
Tested & ready for review

Itxaka:
Needs to be tested on an upgrade case.

Itxaka:
> I think the commit message is somewhat misleading. Or maybe refering to an older variant of this code:
>The script is divided in two parts: On one hand, the alert handler that manages the alerts discarding non interesting alerts, launches to the blocker script in this node and other cluster nodes (via SSH).
>AFAICS there is nothing that does anything with SSH on other nodes in here. Or am I missing something?

Indeed, this is my fault. This was an original PR by @ilausuch but he was out so when I restored it (original was reverted) I kept the same commit text and worked on top of it without ever updating it.

> Also, would you mind to adding some information about why this is actually needed?
Will do.

Itxaka:
@rhafer is clearer now? Anything you feel is missing or does it convey enough info as to understand the PR?

Itxaka:
> I wonder a bit whether it is a good idea to have such "magic" in the pacemaker resources, but I also lack ideas how to implement it in a less magic way.

Same same, that is why this is having a really heavy testing session again and again to make sure we dont break anything.

ilausuch:
The upgrade test from SOC 7 to SOC 8 was successful.

- During the upgrade the script to block the rabbitmq wasn't launched.
- The upgrade works without any related problem
- At the end of upgrade, I did a test to check the PR and it worked too.

PR #1683, opened 2018-06-14
rabbitmq: Stop standalone epmd before systemd socket start (bsc#1092162)

skazi0:
It can happen that epmd is already running before systemd epmd.socket unit
starts. This leads to "Address already in use" error. Try to stop all epmd
processes before starting systemd socket unit.

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -68,6 +68,19 @@ if node[:platform_family] == "suse"
     subscribes :run, "template[/etc/systemd/system/epmd.socket.d/port.conf]", :immediate
   end
 
+  # Before we pass epmd socket management to systemd, we need to make sure that no
+  # "standalone" epmd is running to avoid "Address already in use" errors. Epmd can
+  # be implicitly started by any erlang command e.g. rabbitmqctl (also used by OCF
+  # monitoring agents in HA setup).
+  # Note that epmd stopped here can still be started by some command during the
+  # short window before systemd socket is started.
+  bash "stop any running epmd before systemd socket start" do
+    code "epmd -kill"
+    action :run
+    only_if "! systemctl is-active --quiet epmd.socket && " \
+      "grep -q Requires=epmd.service /usr/lib/systemd/system/rabbitmq-server.service"
+  end
+
   # Enable epmd.socket for two reasons:
   # 1. when we don't use the rabbitmq systemd service (in HA, for instance),
   #    this will enable the use of the system-wide epmd


chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -68,6 +68,18 @@
...
+  # Before we pass epmd socket management to systemd, we need to make sure that no
+  # "standalone" epmd is running to avoid "Address already in use" errors. Epmd can
+  # be implicitly started by any erlang command e.g. rabbitmqctl (also used by OCF
+  # monitoring agents in HA setup).
+  # Note that epmd stopped here can still be started by some command during the
+  # short window before systemd socket is started.
+  bash "stop any running epmd before systemd socket start" do
+    code "epmd -kill"
+    action :nothing
+    only_if "grep -q Requires=epmd.service /usr/lib/systemd/system/rabbitmq-server.service"
```
rhafer: Is this really correct? I'm not entirely sure what this test is actually good for, but don't you want to do the kill only if the `epmd.socket` is not enabled yet?

jsuchome:
Is it right to fix it in chef recipe? Isn't it actually epmd bug?

chef/cookbooks/rabbitmq/recipes/default.rb
skazi0: Right. I keep forgetting that this is not a simple script where you can just add some commands :)

skazi0: As for the test, this is to check if rabbitmq-server has systemd level dependency on epmd. I think this was introduced in some new version. It is not present in Cloud7. Not sure if epmd even has a systemd service there. The check was copied from other related resources

skazi0:
@jsuchome I think this is "by design" on epmd/erlang side. They simply start epmd when any erlang command is started and leave it running. If we start one beforehand it will be used but we can't assure that nobody starts some erlang command before we pass it to systemd.

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -68,6 +68,19 @@
...
+  # Before we pass epmd socket management to systemd, we need to make sure that no
+  # "standalone" epmd is running to avoid "Address already in use" errors. Epmd can
+  # be implicitly started by any erlang command e.g. rabbitmqctl (also used by OCF
+  # monitoring agents in HA setup).
+  # Note that epmd stopped here can still be started by some command during the
+  # short window before systemd socket is started.
+  bash "stop any running epmd before systemd socket start" do
+    code "epmd -kill"
+    action :nothing
+    only_if "! systemctl is-active --quiet epmd.socket && " \
```
Itxaka: Not sure about this only_if. If epmd was already running, and we execute the `reload systemd for epmd.socket extension` resource, it will do a `daemon-reload` and leave epmd hanging *but* systemd will think that the epmd services are stopped due to the reload.

skazi0: Hmm... was the old epmd already managed by systemd or standalone? Note that we check socket unit here so we want to kill epmd only during the runs where we will start the socket unit. Subsequent runs should have socket unit active and all future epmds should be managed by systemd.

skazi0:
After some testing it looks like it might be more problematic... we can't do `epmd -kill` if there are any rabbits running already and using the standalone epmd instance. Not sure if it's safe to do a plain `kill` on such epmd.

skazi0:
The docs say:

> -kill
> 
> Kills the currently running epmd.
> 
> Killing the running epmd is only allowed if epmd -names shows an empty database or if -relaxed_command_check was specified when the running instance of epmd was started.
> 
> Notice that -relaxed_command_check is specified when starting the daemon that is to accept killing when it has live nodes registered. When running epmd interactively, -relaxed_command_check has no effect. **A daemon** that is started without relaxed command checking **must be killed using, for example, signals or some other OS-specific method if it has active clients registered.**

PR #1685, opened 2018-06-15
WIP/POC Dual DB support for migrations

jhesketh:
This is to allow users to migrate from postgres to mariadb. It enables
multiple proposals for the database barclamp so that the deployer can
choose between migrating on the same node or over to a new node.

This assumes there are three scenarios:
 - 1 single instance of postgres is deployed (keep using in legacy)
 - 1 single instance of mariadb is deployed (migration finished)
 - 1 instance of each postgresql and mariadb is deployed. (In this case
postgres is still used by the control plane until the migration is
complete).

TODO:
 - Handle deploying to a new node
 - Handle deploying postgres+HA and mariadb+galera
 - Handle the 2 node drdb case
 - Ensure/enforce only the above 3 scenarios are possible. ie, disallow
a second postgres to be deployed.

This code is kinda hacky but only intended as a temporary measure for
the migration in cloud 7. This should not be in cloud 8.

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -60,7 +60,13 @@ end
 
 class CrowbarOpenStackHelper
   def self.database_settings(node, barclamp)
-    instance = node[barclamp][:database_instance] || "default"
+    if barclamp == "mysql" or barclamp == "postgresql"
+      # We're called from one of the database cookbooks, which doesn't reference
+      # another database instance.
+      instance = node[:database][:config][:environment].gsub(/^database-config-/, "")
+    else
+      instance = node[barclamp][:database_instance] || "default"
+    end
 
     # Cache the result for each cookbook in an instance variable hash. This
     # cache needs to be invalidated for each chef-client run from chef-client
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -24,6 +24,25 @@ class Chef
       CrowbarOpenStackHelper.database_settings(node, barclamp)
     end
 
+    # The following are special cases of fetch_database_settings.
+    # To allow users to migrate from postgresql to mariadb it is possible to
+    # have multiple databases deployed. Most barclamps will use the above
+    # `fetch_database_settings` to get the currently active connection details.
+    # However, for the database barclamps themselves, they may need access to
+    # one set of settings or the other regardless of which one is active.
+    # These helper functions return the appropriate settings for the database
+    # recipes.
+    # NOTE: These are probably redundant as the barclamp parameter is actually
+    # the cookbook name, not the barclamp, so we can likely use that. Leaving
+    # them for now while this patch is WIP
+    def fetch_database_mysql_settings(barclamp=@cookbook_name)
+      CrowbarOpenStackHelper.database_settings(node, barclamp, 'mysql')
+    end
+
+    def fetch_database_postgresql_settings(barclamp=@cookbook_name)
+      CrowbarOpenStackHelper.database_settings(node, barclamp, 'postgresql')
+    end
+
     def fetch_database_connection_string(db_auth, barclamp = @cookbook_name)
       db_settings = CrowbarOpenStackHelper.database_settings(node, barclamp)
       CrowbarOpenStackHelper.database_connection_string(db_settings, db_auth)
@@ -59,7 +78,7 @@ class Chef
 end
 
 class CrowbarOpenStackHelper
-  def self.database_settings(node, barclamp)
+  def self.database_settings(node, barclamp, engine=nil)
     if barclamp == "mysql" or barclamp == "postgresql"
       # We're called from one of the database cookbooks, which doesn't reference
       # another database instance.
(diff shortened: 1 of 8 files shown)


jsuchome:
So, I've 1. deployed simple setup (just db+keystone), 2. applied your patch 3. added new database proposal with mariadb.

It failed with
```
FATAL:  password authentication failed for user "db_maker"
```

Without looking too closely at the patch (so I could be wrong), I think you expect db_maker_password to be under `database["database"][backend_name]` , while my first (unpatched!) deployment left it under usual `database["database"]`

jsuchome:
I've left it at `crowbar.vi7.cloud.suse.de`, you can use it for debugging

jhesketh:
Hmm, I thought I fixed that. I think I know why it might have gone wrong, will push an untested fix in a minute.

jhesketh:
Should be able to try this new one. I'm rebuilding an env to test it.

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -59,8 +78,14 @@ def fetch_rabbitmq_settings(barclamp=@cookbook_name)
 end
 
 class CrowbarOpenStackHelper
-  def self.database_settings(node, barclamp)
-    instance = node[barclamp][:database_instance] || "default"
+  def self.database_settings(node, barclamp, engine=nil)
```
rhafer: I really wonder if this additional "backend" parameter is really needed. As each barclamp needing a database is referencing a specific database proposal by name it might make thing a lot easier to just load the "database-config-<proposalname>" role. And get the database related attribute from there. Instead of loading the node(s) that have the "database-server" roles assigned.

E.g. (we already get the name of the database proposal a few lines lower with:
```
instance = node[barclamp][:database_instance] || "default"
```

reading the database proposals config role should just be a matter of:

```
db_config = Chef::Search::Query.new.search(:role,"name:database-config-#{instance}")
```

With that you have access to all the config values of the selected database. (I.e. you shouldn't  need to make assumptions on the used db backend based on the number of deployed dbs)

It might also avoid the need to add those `fetch_database_<backend>_settings` methods.

chef/data_bags/crowbar/template-database.schema
```diff
@@ -13,11 +13,11 @@
           "required": true,
           "mapping": {
             "sql_engine": { "type": "str", "required": true },
-            "db_maker_password": { "type": "str" },
```
rhafer: This needs a schema migration. With that you should also be able to avoid the`|| node[:database][:db_maker_password]` fallback in other parts of the code.

rhafer:
Regarding the deployment section. After assigning two database proposals to the same set of nodes we end up with a merged (chef's own internal merging, nothing that crowbar is doing) deployment section on the node looking like this:

```
database: 
[..]
  element_order:         
      database-server
      database-server
  element_states:        
    database-server: 
        readying
        ready
        applying
        readying
        ready
        applying
  elements:              
    database-server: 
        d52-54-77-77-01-01.virtual.cloud.suse.de
        d52-54-77-77-01-01.virtual.cloud.suse.de
```

which doesn't look too nice. (Basically what was already raised in the mailing list discussion some time ago)

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
jhesketh: Right, the `fetch_database_<backend>_settings` isn't really needed for other reasons too. But yes, this section would probably benefit from being rewritten with the proposals settings. I might look at that shortly as it'll likely make the 'which node to use' logic simpler (or disappear) and just have the exceptions for the database recipes themselves.

chef/data_bags/crowbar/template-database.schema
jhesketh: Yep, was totally a hack ;-)

jhesketh:
Yeah the merging isn't great, but that's what this POC was to show.

We probably need to compare it to multiple roles on the one proposal. However, I feel like that has some big disadvantages that we can consider before doing another POC:
 - We'd likely need to carry or handle parts of the multiple proposals in SOC8
 - We'd have to reconsider how node assigning works
 - It'll probably take changes to the UI and more migrations

rhafer:
> Yeah the merging isn't great, but that's what this POC was to show.

Sure. And it's good to have that POC to discuss this kind of issues.

BTW, there is another issue related to chef's attribute merging. Once you have two database proposals created the database cookbook will still only ever run the recipes for one of the backends. (https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/database/recipes/server.rb#L24)

I am not really sure if it is really deterministic which backend it will take (probably the one which was created last). But it really means that after creating a second proposal, only one of them ever gets its cookbooks executed.
Even if you click apply on the other proposal chef might just see the backend (`sql_engine`) attribute of the other proposal.

rhafer:
> We probably need to compare it to multiple roles on the one proposal. However, I feel like that has some big disadvantages that we can consider before doing another POC:
> 
>     We'd likely need to carry or handle parts of the multiple proposals in SOC8

Why? If we go for mulitple roles. We don't need to worry about multiple proposals at all. There would always just be a single database proposal.
It might of course require some code changes on the SOC8 side (those might be limited to just schema migrations). But I am pretty sure the same we need some changes in master of this PR as well.

>     We'd have to reconsider how node assigning works

Yes. OTOH the whole node assignment and attribute merging isn't completely solved in this PR either. And especially the merging would be a non-issue if we had multiple roles.

>     It'll probably take changes to the UI and more migrations

UI changes, yes certain. More migrations, not sure. I'd say just different migrations.

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -59,8 +78,14 @@ def fetch_rabbitmq_settings(barclamp=@cookbook_name)
...
+    if barclamp == "mysql" or barclamp == "postgresql"
```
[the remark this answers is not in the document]
skazi0: `if ["mysql", "postgresql"].include? barclamp` ??

rhafer:
> I am not really sure if it is really deterministic which backend it will take (probably the one which was created last).

Actually seems to depend in which order the roles appear in the run list. And that depends on the "priority" of the barclamp and the name of the proposal. So if the name of the 2nd database proposal (e.g. "aaaa")  is lexically before the one of the 1st (usually "default") the 2nd proposal is never being applied.

jsuchome:
we can change `recipes/server.rb` to apply recipes for all present database proposals, right? What would be the problem then?

rhafer:
> We can change recipes/server.rb to apply recipes for all present database proposals, right? What would be the problem then?

Currently it does not even see that there are multiple proposals. I would need to search for the database-config-roles manually. 
And it is not just a matter of including both recipes. The recipes itself will also need to be modifed to read the attributes from the database-config-roles directly (instead of the node object). This will make the database cookbook a very special snowflake compared to the other cookbooks we have. That's why I still have some doubts about this approach. Of course it's all solvable somehow. It's just another hack away :). 
Of course I can't currently promise that the approach with split roles will be simpler/better in every regard. But at least in this area it has clear advantages since we only have to deal with a single proposal.

> It does not seem to be that the order would be an issue.

The order in which the recipes are excuted? Right, that shouldn't be a problem. The ordering (or naming) is currently just what prevents the recipe from seeing one or the other backend.

Just for completeness. This is what is inserting the stuff in to the run_list:
https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/service_object.rb#L1338
the ordering is done here:
https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/node.rb#L539

rhafer:
> The recipes itself will also need to be modifed to read the attributes from the database-config-roles directly (instead of the node object)

Just to explain in some more detail what exactly the problem is:

Consider you have the proposals "default" and "mariadb". Both will have the attribute `mysql.innodb_tunings` (it's a list of config lines that end up in some mysql.cnf file. The "default" proposal (being for postgresql) has the attribute set to the default values. If the "mariadb" set the attribute to a custom value. The (merged) node attributes will the get boths lists merged and add the values from both proposal to the config file.

jhesketh:
We're going with an alternative approach: https://github.com/crowbar/crowbar-openstack/pull/1691

PR #1689, opened 2018-06-20
aodh: Add config for alarm_history_ttl (bsc#1073703)

aplanas:
The alarm_history_ttl config option for aodh was
not previously configurable.

(cherry picked from commit 47a7d26c864e7ca1a21053e6b50b3916558f1fcc)


aplanas:
[P2 Bug 1073703 - SOC7 L3: openstack-aodh-expirer.cron question](https://trello.com/c/aGbpX57n/648-p2-bug-1073703-soc7-l3-openstack-aodh-expirercron-question)

aplanas:
There is no 100 migration here, so I used this one as the first one (aodh is empty of migrations). There are other cases like barbican that start in 100, but there are other that choose 101 here.

I do not mind to use 101 as the first one.

PR #1691, opened 2018-06-22
[4.0] WIP: split database-server role into backend specific roles

rhafer:
This is still very rough an missing a couple of things:

- [x] UI changes to allow configuring both backends in parallel: https://github.com/crowbar/crowbar-openstack/pull/1705
- [x] validation of role assignment (node assigned to the correct role for the selected backend
- [x] Likely needs some update HA
- [x] database_settings caching is disabled and needs to be fixed
- [ ] related changes on Cloud 8 (maybe just a schema migration back to the "database-server" role)
- [ ] ...

~~Note: The UI is currently broken! Only the "raw" mode is displayed for the attributes.~~

Note: The current implementation only adds the "mysql-server" role. postgresql based deployments continue to use the "database-server" role. This seems to have a couple of benefits:
* Most of the existing postgresql based ci jobs (especially the "batch" jobs) can be used without changes
* Fewer changes will be needed in "master". Basically we "just" need to rename "mysql-server" to database-server. (That would however mean the "master" will lose postgresql support)

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -79,7 +79,19 @@ class CrowbarOpenStackHelper
       Chef::Log.info("Database server found at #{@database_settings[instance][:address]} [cached]")
     else
       @database_settings ||= Hash.new
-      database = get_node(node, "database-server", "database", instance)
+      db_roles, = Chef::Search::Query.new.search(
+                          :role,
+                          "name:database-config-#{instance}"
+                        )
+      db_proposal_role = db_roles.first unless db_roles.empty?
+      sql_engine = db_proposal_role.default_attributes["database"]["sql_engine"]
+      db_role = if sql_engine == "postgresql"
+                  "database-server"
+                else
+                  "mysql-server"
+                end
+
+      database = get_node(node, db_role, "database", instance)
 
       if database.nil?
         Chef::Log.warn("No database server found!")
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -72,38 +72,45 @@ class CrowbarOpenStackHelper
                        "on behalf of #{barclamp}")
       end
       @database_settings = nil
+      @sql_engine_cache = nil
       @database_settings_cache_time = node[:ohai_time]
     end
 
-    if @database_settings && @database_settings.include?(instance) && false
-      Chef::Log.info("Database server found at #{@database_settings[instance][:address]} [cached]")
+    if barclamp == "mysql" || barclamp == "postgresql"
+      sql_engine = barclamp
+    elsif @sql_engine_cache && @sql_engine_cache.include?(instance)
+      sql_engine = @sql_engine_cache[instance]
+    else
+      db_roles, = Chef::Search::Query.new.search(
+        :role,
+        "name:database-config-#{instance}"
+      )
+      db_proposal_role = db_roles.first unless db_roles.empty?
+      # TODO(jhesketh): What if db_roles is empty here?
+      sql_engine = db_proposal_role.default_attributes["database"]["sql_engine"]
+
+      @sql_engine_cache ||= Hash.new
+      @sql_engine_cache[instance] = sql_engine
+    end
+
+    if @database_settings && @database_settings.include?(instance) && @database_settings[instance].include?(sql_engine)
+      Chef::Log.info("Database server found at #{@database_settings[instance][sql_engine][:address]} [cached]")
     else
       @database_settings ||= Hash.new
-      db_roles, = Chef::Search::Query.new.search(
-                          :role,
-                          "name:database-config-#{instance}"
-                        )
-      db_proposal_role = db_roles.first unless db_roles.empty?
-      sql_engine = db_proposal_role.default_attributes["database"]["sql_engine"]
-      if barclamp == "mysql" || barclamp == "postgresql"
-        sql_engine = barclamp
-      end
       db_role = if sql_engine == "postgresql"
                   "database-server"
                 else
                   "mysql-server"
                 end
-
       database = get_node(node, db_role, "database", instance)
 
       if database.nil?
         Chef::Log.warn("No database server found!")
       else
         address = CrowbarDatabaseHelper.get_listen_address(database)
-        backend_name = sql_engine
 
         ssl_opts = {}
-        if backend_name == "mysql"
+        if sql_engine == "mysql"
           ssl_opts = {
             enabled: database["database"]["mysql"]["ssl"]["enabled"],
             ca_certs: database["database"]["mysql"]["ssl"]["ca_certs"],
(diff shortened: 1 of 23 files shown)


chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -72,52 +72,60 @@ def self.database_settings(node, barclamp)
                        "on behalf of #{barclamp}")
       end
       @database_settings = nil
+      @sql_engine_cache = nil
       @database_settings_cache_time = node[:ohai_time]
     end
 
-    if @database_settings && @database_settings.include?(instance) && false
-      Chef::Log.info("Database server found at #{@database_settings[instance][:address]} [cached]")
+    if barclamp == "mysql" || barclamp == "postgresql"
```
jsuchome: (@skazi0  has seen this first) This does not seem to make sense; there's still only one (database) barclamp, right?

skazi0: @jsuchome I think the parameter name is confusing. See: `def fetch_database_settings(barclamp=@cookbook_name)`

chef/data_bags/crowbar/migrate/database/109_separate_db_roles.rb
```diff
@@ -0,0 +1,55 @@
+def upgrade(ta, td, a, d)
+  db_engine = a["sql_engine"]
+  a[db_engine]["ha"] = a["ha"]
+  a.delete("ha")
+  if db_engine == "postgresql"
+    a["mysql"]["ha"] = ta["mysql"]["ha"]
+  else
+    a["postgresql"]["ha"] = ta["postgresql"]["ha"]
+  end
```
jsuchome: Hm, migration failed from me when I had mysql deployment in SOC7. Don't we need here actually both
```
a["mysql"]["ha"] = ta["mysql"]["ha"]
a["postgresql"]["ha"] = ta["postgresql"]["ha"]
```

skazi0: I think the lines above is supposed to take care of the other engine, i.e. copy from old attributes to new for used engine and use template for the not-used one.
The problem might be that there are different set of HA params for each engine.

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -79,7 +79,19 @@ def self.database_settings(node, barclamp)
       Chef::Log.info("Database server found at #{@database_settings[instance][:address]} [cached]")
     else
       @database_settings ||= Hash.new
-      database = get_node(node, "database-server", "database", instance)
+      db_roles, = Chef::Search::Query.new.search(
+                          :role,
+                          "name:database-config-#{instance}"
+                        )
+      db_proposal_role = db_roles.first unless db_roles.empty?
```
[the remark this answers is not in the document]
skazi0: I see @jhesketh already marked this.

chef/data_bags/crowbar/migrate/database/109_separate_db_roles.rb
```diff
@@ -0,0 +1,55 @@
...
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+
+  if a["sql_engine"] == "mysql"
+    d["elements"]["mysql-server"] = d["elements"]["database-server"]
+    d["elements"]["atabase-server"] = []
```
skazi0: typo.

chef/cookbooks/database/recipes/role_mysql_server.rb
```diff
@@ -1,24 +1,18 @@
...
 # limitations under the License.
 #
-
-backend = node[:database][:sql_engine]
-
-Chef::Log.info("Running database::server for #{backend}")
-
-include_recipe "#{backend}::server"
+if CrowbarRoleRecipe.node_state_valid_for_role?(node, "database", "mysql-server")
+  include_recipe "mysql::server"
```
skazi0: Please add the log similar as in psql role to make things consistent.

chef/cookbooks/database/libraries/crowbar.rb
```diff
@@ -1,10 +1,10 @@
 module CrowbarDatabaseHelper
-  def self.get_ha_vhostname(node)
-    if node[:database][:ha][:enabled]
+  def self.get_ha_vhostname(node, sql_engine=node[:database][:sql_engine])
+    if node["database"][sql_engine]["ha"]["enabled"]
```
skazi0: Shouldn't these be symbols not strings?

chef/data_bags/crowbar/migrate/database/109_separate_db_roles.rb
skazi0: So maybe it should be something like:
```
if db_engine == "postgresql"
# copy old ha settings for psql
  a["postgresql"]["ha"] = a["ha"]
# there was a["mysql"]["ha"] already before, no need to touch that
else
# fill psql ha settings from template
  a["postgresql"]["ha"] = ta["postgresql"]["ha"]
# only use "enabled" flag from old (psql) settings as it was used for both engines
  a["mysql"]["ha"]["enabled"] = a["ha"]["enabled"]
end
a.delete("ha")
```

chef/cookbooks/database/libraries/crowbar.rb
skazi0: eh... ruby magic... will this work properly? with normal dicts it will not. is `node` somehow special?

chef/data_bags/crowbar/migrate/database/109_separate_db_roles.rb
skazi0: @jsuchome  sorry I missed your comment. I still think `a["mysql"]["ha"]` can be left as is.

jsuchome: True, I forgot about the `enabled` part

jsuchome: No, I get
`Failed to validate migrated proposal default for database: key 'enabled:' is undefined.`
with `a["mysql"]["ha"]["enabled"] = a["ha"]["enabled"]`
Now I think it should actually be filled correctly when reapplying the proposal...

chef/cookbooks/database/libraries/crowbar.rb
jsuchome: I think strings work everywhere. It's the magic that allows symbols in their place...

skazi0:
```
irb(main):001:0> x={a:1}
=> {:a=>1}
irb(main):002:0> x["a"]
=> nil
irb(main):003:0> x[:a]
=> 1
```

chef/data_bags/crowbar/migrate/database/109_separate_db_roles.rb
skazi0: Yeah, it will be added on applying but if it was there before it should be copied. So some `if a["ha"].key? "enabled"` would be needed.

chef/cookbooks/database/libraries/crowbar.rb
jsuchome: OK, I did not mean everywhere like really _everywhere_, but at least with `node` object.

skazi0: OK, so `node` has some magic for that.

jsuchome:
Closing. Was already merged as https://github.com/crowbar/crowbar-openstack/pull/1713

PR #1693, opened 2018-06-27
Add security settings (bsc#1094134)

nicolasbock:
Using default settings any Zookeeper instance can join the ensemble
and inject malicious requests. This change adds settings to enable
SASL DIGEST-MD5 based authentication only allowing instances to join
that know the shared secret.

### DIFF
--- a/chef/cookbooks/monasca/recipes/master.rb
+++ b/chef/cookbooks/monasca/recipes/master.rb
@@ -166,3 +166,20 @@ execute "run ansible" do
   cwd "/opt/monasca-installer"
   action :nothing unless actual_versions != previous_versions
 end
+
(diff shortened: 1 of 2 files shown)


jgrassler:
Sorry for only looking at this now, missed the notification!

Why are you submitting this against `stable/4.0` (rather than master)? We have the same problem in `master`, don't we?

Also, given how much of a headache running `monasca-installer` only in the cases where it is needed has been in the past, I'd rather avoid launching additional Ansible runs from Crowbar. Can't this be done in [ansible-zookeeper](https://gitlab.com/monasca-installer/ansible-zookeeper)? For two different Ansible playbooks modifying the same `/etc/zookeeper/zoo.cfg` sounds a bit dodgy to me...

PR #1697, opened 2018-07-03
[4.0]  rabbitmq: Make sure rabbitmq is running on cluster HA

Itxaka:
As the resource agent for rabbitmq with cluster HA restart the rabbitmq
service several times, the current check can fail to validate rabbitmq
status, as it could do the check just on one of those times that rabbit
is up while creating/joining the cluster. Then if the check passed and
continued the chef execution, the next steps could fail as they are
dependant on having a running rabbitmq, while the rabbitmq server may
still be restarting.

Instead expand the checks to first look for a rabbit master for the
resource and expand the check for a local runing rabbit to make sure
we are checking for the local copy.

Backport-of: https://github.com/crowbar/crowbar-openstack/pull/1396

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
+++ b/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
@@ -95,12 +95,26 @@ ruby_block "wait for #{ms_name} to be started" do
   block do
     require "timeout"
     begin
-      Timeout.timeout(240) do
-        # Check that the service is running
+      Timeout.timeout(300) do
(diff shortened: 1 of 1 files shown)


Itxaka:
Do not merge until we know if https://github.com/crowbar/crowbar-openstack/pull/1637 needs backport or not (as its included here as part of the backport, if its not needed we should remove it from this PR)

vuntz:
Ignoring hound and merging at the request of @ilausuch (there were two +1 already)

PR #1699, opened 2018-07-05
Add cisco-aci role for neutron.

mmnelemane:
So far, the implementation assumed that only compute nodes
would be attached to the ACI leaf ports and the controller
nodes are independently attached to the network not connected
to the ACI fabric itself. The only form of communication from
the controller to the ACI is through the agents. However,
recent customer POCs have changed our assumption since these
setups expect all or most of the nodes attached the ACI leaf
as the ACI fabric forms the central network infrastructure
of the entire deployment. In such scenarios, we need a mechanism
where we can know which of the nodes are attached to the ACI
leaf and which nodes are not. The agents should be enabled
and run on each node attached to the ACI leaf ports.
This commit tries to achieve exactly this result, so that
future deployments can flexibly allow any node attached to
the fabric.


chef/data_bags/crowbar/template-neutron.json
```diff
@@ -185,19 +185,22 @@
...
          ["neutron-server" ],
-         ["neutron-network" ]
+         ["neutron-network" ],
+         ["neutron-cisco-aci" ]
       ],
       "element_run_list_order": {
         "neutron-server": 94,
-        "neutron-network": 95
+        "neutron-network": 95,
+        "neutron-cisco-aci": 96
```
dirkmueller: same prio like neutron-network

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -185,19 +185,22 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 209,
+      "schema-revision": 210,
       "element_states": {
         "neutron-server": [ "readying", "ready", "applying" ],
-        "neutron-network": [ "readying", "ready", "applying" ]
+        "neutron-network": [ "readying", "ready", "applying" ],
+        "neutron-cisco-aci": [ "readying", "ready", "applying" ]
```
dirkmueller: So I understand this role is for any node that needs to have sdn integration enabled, so we should maybe not call it neutron-* specific?

How about calling it 

neutron-sdn-managed-cisco-aci

I know thats a really long and clumsy name but maybe we can come up with something in that area that is more descriptive.

dirkmueller: neutron-cisco-aci-agents

or neutron-sdn-cisco-aci-agents might be good names. I would like those

chef/data_bags/crowbar/migrate/neutron/211_add_cisco_aci_role.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+def upgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+  d["element_run_list_order"] = td["element_run_list_order"]
```
rhafer: In order to keep existing setups working wouldn't you at least need to assign the new role to `nova-compute-kvm` nodes here, if the aci driver is enabled in the proposal? Or is that something you intend to have the users do manually after upgrading to the new code?

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -57,7 +57,7 @@
 end
 
 # apply configurations to compute node
-if node.roles.include?("nova-compute-kvm")
+if node.roles.include?("neutron-cisco-aci")
```
rhafer: Will any node that doesn't have the "neutron-cisco-aci" role assigned ever get to execute the recipe? If yes, which node(s) would that be and what do they need it for?

Currently there is a `include_recipe "neutron::cisco_apic_agents"` call in `./chef/cookbooks/neutron/recipes/common_agent.rb` but if I understand it correctly that is no longer needed if all required nodes (including compute- and network-nodes) get the new role assigned. In that case you can remove the `if` statement here, can't you?

chef/data_bags/crowbar/migrate/neutron/211_add_cisco_aci_role.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  nodes = NodeObject.find("roles:nova-compute-kvm")
+  nodes.each do |node|
+    next unless node["neutron"]["networking_plugin"] == "ml2" &&
+        (node["neutron"]["mechanism_drivers"].include?("cisco_apic_ml2") ||
+        node["neutron"]["mechanism_drivers"].include?("apic_gbp"))
```
rhafer: This will likely not work . The compute nodes don't have the `neutron` attributes as they don't have any "neutron-*" role assigned before running the migration. You need to read those attributes from the neutron proposal. l think replacing `node["neutron"]` with  `a` should work here

chef/cookbooks/neutron/recipes/cisco_apic_agents.rb
```diff
@@ -57,7 +57,7 @@
 end
 
 # apply configurations to compute node
-if node.roles.include?("nova-compute-kvm")
+if node.roles.include?("neutron-sdn-cisco-aci-agents")
```
rhafer: Isn't this superfluous now? The whole recipe is only ever executed on nodes that have the `neutron-sdn-cisco-aci-agents` assigned.

rhafer:
@mmnelemane BTW, if this is supposed to land in Cloud8 as well you need to cherry-pick into the stable/5.0-pike branch as well.

mmnelemane:
@rhafer : Thats right. The PRs for Cloud 7 and Cloud 8 are at https://github.com/crowbar/crowbar-openstack/pull/1700 and https://github.com/crowbar/crowbar-openstack/pull/1750.

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -188,19 +188,22 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 210,
+      "schema-revision": 211,
```
dirkmueller: please change it to rev 300 for master

crowbar_framework/app/models/neutron_service.rb
```diff
@@ -61,6 +61,16 @@ def role_constraints
             "windows" => "/.*/"
           },
           "cluster" => true
+        },
+        "neutron-sdn-cisco-aci-agents" => {
+          "unique" => false,
+          "count" => -1,
+          "admin" => false,
+          "exclude_platform" => {
+            "suse" => "< 12.3",
```
dirkmueller: master is at 12.4

rhafer: Hm, all the role constraints are also still at 12.3 in master? Wouldn't updating those be something for a separate pull request?

mmnelemane:
@rhafer: Tested on Cloud7. Cloud 8 does not support the regular plugin mechanism and needs unified mode. Another PR is on its way for Cloud 8 to enable this mode. But it should be merged on top of this PR.

rhafer:
> Tested on Cloud7. Cloud 8 does not support the regular plugin mechanism and needs unified mode. Another PR is on its way for Cloud 8 to enable this mode. But it should be merged on top of this PR.

@mmnelemane The code looks ok to me from a crowbar perspective (I don't know much about cisco-aci).

But I am a bit confused about the above statement. You are saying this only work for Cloud 7 and not for Cloud8 (neither for master, which is basically Cloud 9). I assume this doesn't have todo much with this patch itself and even the unpatched code in master doesn't work on Cloud8/9, right? If that is the case I am fine merging this into master and backporting it into Cloud 7 (and 8).

mmnelemane:
@rhafer: True, the unpatched version also does not work with cloud 8 and cloud 9. The ml2 driver used for cloud 7 is not supported after OpenStack Ocata. So testing this directly will not be possible without major changes with the updated driver. However, the logic of using aci-role will still be valid for cloud 8 and cloud 9.

chef/data_bags/crowbar/migrate/neutron/301_add_cisco_aci_role.rb
```diff
@@ -0,0 +1,35 @@
...
+
+def upgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+  d["element_run_list_order"] = td["element_run_list_order"]
+
+  if a["networking_plugin"] == "ml2" &&
+      (a["ml2_mechanism_drivers"].include?("cisco_apic_ml2") ||
+      a["ml2_mechanism_drivers"].include?("apic_gbp"))
+    nodes = NodeObject.find("roles:nova-compute-kvm")
```
rhafer: Since you're going to backport this change to Cloud7 (and 8) we need to make the migration backport safe. I.e. if it already ran on a certain setup in Cloud 7 it has to basically a no-op when executed on that same setup after an upgrade to Cloud 8. That mean you only should add the neutron-sdn-cisco-aci-agents role to any run list, if you haven't done so before.

I guess the easiest way to check that is by checking if the currently deployment section (`d`) already has the new role in `element_states`.

I.e. by adding `d["element_states"].key? "neutron-sdn-cisco-aci-agents"` add the beginning of the migration. 

The downgrade migration needs a similar (but reversed check). If  `td` still contains the new role than don't touch any run_lists.

chef/data_bags/crowbar/migrate/neutron/301_add_cisco_aci_role.rb
```diff
@@ -0,0 +1,39 @@
...
+                             td["element_states"]["neutron-sdn-cisco-aci-agents"])
+        node.save
+      end
+    end
+  end
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  if d["element_states"].key?("neutron-sdn-cisco-aci-agents")
```
rhafer: This needs to be: `unless td["element_states"].key?("neutron-sdn-cisco-aci-agents")` 

That means of the proposal template of the revision that you're downgrading to already contains the added role, you're downgrading to a release that already has the change backported.

(It doesn't matter much since downgrades are not really supported, but ...)

PR #304, opened 2016-02-22
Fix error handling when comparing timeouts (bsc#967677)

rhafer:

PR #1702, opened 2018-07-17
[4.0] rabbitmq: active the haproxy balancing

ilausuch:

(cherry picked from commit a4ebcfe603bc03c1c9c1fa32857987e4b40c7666)


ilausuch:
It couldn't be implemented at this moment because the requisite #304 isn't implemented in SOC7 because it would be disruptive in the upgrade

PR #1704, opened 2018-07-18

ilausuch:
This script blocks the connection to the rabbitmq cluster in case the number of nodes decay bellow the half
of the total. In this case the remain rabbit nodes won't accept new connections if the quorum is reached.

It take advantage of the pacemaker notifications that notify when a rabbitmq has failed or has restored.

The script is divided in two parts: On one hand, the alert handler that manages the alerts discarding non
interesting alerts, launches to the blocker script in this node and other cluster nodes (via SSH).
And in the other hand, the blocker script than checks the condition of running nodes and blocks or unblocks the rabbitmq client port.

backported from #1681


ilausuch:

ilausuch:
Tested & ready for review

Itxaka:
I cant approve this as I was involved in the code wrinting but the 5.0 version was merged so this shouldbe too @rhafer @jsuchome 


Mkcloud failure is non applicable here as this only deals with HA

PR #1705, opened 2018-07-18
[4.0] database: UI for mysql and/or postgresql

skazi0:
Barclamp UI was modified to show custom view of MySQL and/or
PostgreSQL depending on assigned roles.

Extra features:
- automatic assignment of active sql engine based on roles (new proposals only)
- model validation on save
- added borders around each engine parameters section to avoid confusion

### DIFF
--- a/crowbar_framework/app/models/database_service.rb
+++ b/crowbar_framework/app/models/database_service.rb
@@ -84,6 +84,10 @@ class DatabaseService < PacemakerServiceObject
     end
   end
 
+  def already_applied?(proposal_name="default")
+    !RoleObject.find_role_by_name("#{@bc_name}-config-#{proposal_name}").nil?
+  end
+
   def validate_ha_attributes(attributes, cluster, sql_engine)
     role = available_clusters[cluster]
 
@@ -151,7 +155,7 @@ class DatabaseService < PacemakerServiceObject
 
     validation_error I18n.t(
       "barclamp.#{@bc_name}.validation.new_proposal_multi_engine"
-    ) if selected_engines.length > 1 && !proposal["deployment"]["crowbar-applied"]
+    ) if selected_engines.length > 1 && !already_applied?
 
     validation_error I18n.t(
       "barclamp.#{@bc_name}.validation.engine_roles_mismatch",
--- a/crowbar_framework/app/models/database_service.rb
+++ b/crowbar_framework/app/models/database_service.rb
@@ -162,10 +162,6 @@ class DatabaseService < PacemakerServiceObject
       db_engine: active_engine
     ) unless selected_engines.include?(active_engine)
 
-    validation_error I18n.t(
-      "barclamp.#{@bc_name}.validation.secondary_psql"
-    ) if selected_engines.length > 1 && active_engine == "mysql"
-
     selected_engines.each do |engine|
       db_role = role_for_engine engine
       validate_one_for_role proposal, db_role
@@ -177,6 +173,25 @@ class DatabaseService < PacemakerServiceObject
         validate_ha_attributes(attributes, cluster, engine)
       end
     end
+    super
+  end
+
+  def validate_proposal_elements(proposal_elements)
+    old_proposal = proposals_raw.first
+
+    return super if old_proposal.nil?
+
+    # disallow adding psql when mysql is already deployed
+    old_psql_nodes = old_proposal.elements["database-server"] || []
+    old_mysql_nodes = old_proposal.elements["mysql-server"] || []
+    new_psql_nodes = proposal_elements["database-server"] || []
+    raise I18n.t(
+      "barclamp.#{@bc_name}.validation.secondary_psql"
+    ) if already_applied? &&
+        !old_mysql_nodes.empty? &&
+        old_psql_nodes.empty? &&
+        !new_psql_nodes.empty?
+
     super
   end
 
--- a/crowbar_framework/app/models/database_service.rb
+++ b/crowbar_framework/app/models/database_service.rb
@@ -76,10 +76,18 @@ class DatabaseService < PacemakerServiceObject
     base
   end
 
-  def validate_ha_attributes(attributes, cluster)
+  def role_for_engine(engine)
+    if engine == "postgresql"
+      "database-server"
+    else
+      "mysql-server"
+    end
+  end
+
+  def validate_ha_attributes(attributes, cluster, sql_engine)
     role = available_clusters[cluster]
 
-    case attributes["sql_engine"]
+    case sql_engine
     when "postgresql"
       ha_attr = attributes["postgresql"]["ha"]
       storage_mode = ha_attr["storage"]["mode"]
@@ -129,26 +137,38 @@ class DatabaseService < PacemakerServiceObject
 
   def validate_proposal_after_save(proposal)
     attributes = proposal["attributes"][@bc_name]
-    sql_engine = attributes["sql_engine"]
-    db_role = if sql_engine == "postgresql"
-      "database-server"
-    else
-      "mysql-server"
-    end
-    validate_one_for_role proposal, db_role
+    active_engine = attributes["sql_engine"]
 
     validation_error I18n.t(
       "barclamp.#{@bc_name}.validation.invalid_db_engine",
-      db_engine: sql_engine
-    ) unless ["mysql", "postgresql"].include?(sql_engine)
+      db_engine: active_engine
+    ) unless ["mysql", "postgresql"].include?(active_engine)
 
-    # HA validation
-    servers = proposal["deployment"][@bc_name]["elements"][db_role]
-    unless servers.nil? || servers.first.nil? || !is_cluster?(servers.first)
-      cluster = servers.first
-      validate_ha_attributes(attributes, cluster)
+    selected_engines = ["postgresql", "mysql"].select do |engine|
+      nodes = proposal["deployment"][@bc_name]["elements"][role_for_engine engine]
+      !nodes.nil? && !nodes.first.nil?
     end
 
+    validation_error I18n.t(
+      "barclamp.#{@bc_name}.validation.engine_roles_mismatch",
+      db_engine: active_engine
+    ) unless selected_engines.include?(active_engine)
+
+    validation_error I18n.t(
+      "barclamp.#{@bc_name}.validation.secondary_psql"
+    ) if selected_engines.length > 1 && active_engine == "mysql"
+
+    selected_engines.each do |engine|
+      db_role = role_for_engine engine
+      validate_one_for_role proposal, db_role
+
+      # HA validation
+      servers = proposal["deployment"][@bc_name]["elements"][db_role]
+      unless servers.nil? || servers.first.nil? || !is_cluster?(servers.first)
+        cluster = servers.first
+        validate_ha_attributes(attributes, cluster, engine)
+      end
+    end
     super
   end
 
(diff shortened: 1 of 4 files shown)


skazi0:
Some description of what is going on there would not hurt probably. So it would be something like:
- postgres role only - "Before upgrading to SOC8 please assign mysql-server role to some node/cluster and apply the proposal."
- postgres and mysql role - "Note that parallel deployment of MySQL and PostgreSQL is supported only for the migration purposes. PostgreSQL will not be supported in SOC8."
- mysql role only - is it possible?

jsuchome:
I was thinking about something simpler - just general message shown for the whole barclamp that allows these 2 roles; nothing specific, based on current roles allocation.
Such message would be removed in SOC8

jsuchome:
Also, could you please move postgresql section to the top? Seems more logical to me as user already has postgresql deployed (it's on top) and is going to add mariadb (reads from top to bottom)

skazi0:
psql settings moved up.

jsuchome:
With the last commit you've broken some changes that were in the original PR, see https://github.com/crowbar/crowbar-openstack/pull/1691/commits/aa10d7d89cf54f1525bbb8ece42f006c3850841b

skazi0:
@jsuchome I split the changes into two commits but I think I will need to rebase it on top of the base PR to make it clean.

jsuchome:
@skazi0 there are still some conflicts when merging this on top of https://github.com/crowbar/crowbar-openstack/pull/1713 ...

skazi0:
@jsuchome I rebased this PR on top of #1713.

jsuchome:
Shouldn't the sql_engine widget be editable when there's no DB deployed? 
Now when I enter the barclamp, postgres is the first option. That's fine, but when I use only mysql-server role, sql_engine still shows postgresql...

jsuchome:
2. Now, I've deployed postgresql first, then reloaded barclamp and tried to add cluster role for mariadb-server. Unfortunately the MariaDB parts of UI were not shown...

jhesketh:
Something also isn't working for me with this patchset. I'm also unable to get the mariadb UI shown and the active sql_engine doesn't switch automatically.

skazi0:
@jsuchome I don't think the engine should be editable at all. It is just to show what is going on under the hood. If your really want to change it, you can use raw view.

@jhesketh for now we don't have a clear idea on how the switch of `sql_engine` will be performed during migration/upgrade. We can remove/adjust those validations at any time. Current code is supposed to be a replacement for the "disabled-psql-droppable" we had before.

jsuchome:
It should not be editable after the deployment, true. But my case is different. The value of sql_engine still shows Postgresql if I have only mysql-server role assigned.

jhesketh:
> @jhesketh for now we don't have a clear idea on how the switch of sql_engine will be performed during migration/upgrade. We can remove/adjust those validations at any time. Current code is supposed to be a replacement for the "disabled-psql-droppable" we had before.

Okay, that's a good point.

jsuchome:
Scratch my previous comments. I did not have the patch applied correctly :-(

crowbar_framework/config/locales/database/en.yml
```diff
@@ -21,7 +21,8 @@ en:
   barclamp:
     database:
       edit_attributes:
-        sql_engine: 'SQL Engine'
+        engine_upgrade: 'Deployment of multiple database engines at the same time is supported only during upgrade to support migration from PostgreSQL to MariaDB. Refer to the SUSE OpenStack Cloud documentation for more information about upgrade procedure.'
```
jhesketh: Deployment of multiple database engines at the same time is only supported before SUSE OpenStack Cloud 8 upgrade to allow the migration from PostgreSQL to MariaDB. Please refer to the SUSE OpenStack CLoud documentation for more information about the upgrade procedure.

jhesketh:
Also, when I go to create a new database barclamp (rather than edit etc), by default the database-server (ie postgres) role has a node selected for it. We should probably default to mysql?

jsuchome:
@jhesketh you mean showing different value in the UI, but not renaming the role? Sounds super ugly to me

jhesketh:
> @jhesketh you mean showing different value in the UI, but not renaming the role? Sounds super ugly to me

Yes, that is what I mean. At the moment it's unclear how 'database-server' is different to 'mysql-server' or what exactly the role is. Additionally 'database-server' will have two different meanings depending if you're in cloud 7 or 8. It could also be confusing that postgres options appear when you assign a node to 'database-server'.

This probably will persist outside of the UI so I'm not sure what the best answer is. Is it just a matter of really good documentation and user training?

jsuchome:
> At the moment it's unclear how 'database-server' is different to 'mysql-server' or what exactly the role
> is

I think this needs to be properly documented, yes.
We decided mysql-server is the temporary role for upgrade only. Normal SOC8 users should not see it. And upgrading SOC7 users should see it only when they are performing DB part of the upgrade. Again, I think we need to document it properly. 

But having database-server role and having it look like postgresql-server in the UI is IMHO only path to bigger confusion.

jhesketh:
> I think this needs to be properly documented, yes.
We decided mysql-server is the temporary role for upgrade only. Normal SOC8 users should not see it. And upgrading SOC7 users should see it only when they are performing DB part of the upgrade. Again, I think we need to document it properly.
> 
> But having database-server role and having it look like postgresql-server in the UI is IMHO only path to bigger confusion.

Okay, I think that's fair enough.

Is there a case where a user would update to the latest cloud7 codebase, without then proceeding to upgrade to cloud8? ie, will others see this patch and do we need to do something to help the confusion there too?

skazi0:
I think the upgrade code will be released as normal maintenance update so everyone on SOC7 will see these changes. The warning on top of the custom view is supposed to inform the user about weird stuff happening below and point to the docs for more details. The docs should explain the changes around roles and what they mean.

jsuchome:
Yes, that's the reason for that warning line. I don't know, maybe it could be somehow more prominent?

Anyway even users having SOC7 fully upgraded and not doing the upgrade for months ideally should not even look what's inside database barclamp, right?

jhesketh:
Okay, that makes sense.

crowbar_framework/config/locales/database/en.yml
skazi0: updated (except the typo in Cloud)

skazi0:
> Also, when I go to create a new database barclamp (rather than edit etc), by default the database-server (ie postgres) role has a node selected for it. We should probably default to mysql?
This is tricky as the default proposal is created by the model and I think changes there would affect the CI runs too. I might be wrong though. Maybe we should change the default there and update the CI scripts to override it back to psql for SOC7 runs.

crowbar_framework/config/locales/database/en.yml
```diff
@@ -21,7 +21,8 @@ en:
   barclamp:
     database:
       edit_attributes:
-        sql_engine: 'SQL Engine'
+        engine_upgrade: 'Deployment of multiple database engines at the same time is only supported before SUSE OpenStack Cloud 8 upgrade to allow the migration from PostgreSQL to MariaDB. Please refer to the SUSE OpenStack Cloud documentation for more information about the upgrade procedure.
```
jhesketh: This is missing a closing '

skazi0: fixed, sorry

jsuchome:
You can rebase now

jsuchome:
I think it would be nice to forbid deploying both databases when there's a new proposal. In other words, end result of 2 databases should be possible by adding mysql-server role into existing proposal with database-server

skazi0:
@jsuchome rebased

crowbar_framework/app/views/barclamp/database/_edit_attributes.html.haml
```diff
@@ -3,6 +3,9 @@
     = header show_raw_deployment?, true
 
   .panel-body
+    .alert.alert-warning
+      = t(".engine_upgrade")
```
rhafer: IMO this warning should only be visible if both database roles have nodes assigned.

crowbar_framework/config/locales/database/en.yml
```diff
@@ -21,6 +21,7 @@ en:
   barclamp:
     database:
       edit_attributes:
+        engine_upgrade: 'Deployment of multiple database engines at the same time is only supported before SUSE OpenStack Cloud 8 upgrade to allow the migration from PostgreSQL to MariaDB. Please refer to the SUSE OpenStack Cloud documentation for more information about the upgrade procedure.'
```
rhafer: How about: "Deploying multiple database engines at the same time is only supported in preparation for a migration from PostgreSQL to MariaDB.  Please refer to the SUSE OpenStack Cloud documentation for more information about the upgrade procedure."

But I guess we should get some tuning from our native speakers here :) @matthewoliver  @s-t-e-v-e-n-k ?

crowbar_framework/app/models/database_service.rb
```diff
@@ -124,26 +131,38 @@ def validate_ha_attributes(attributes, cluster)
...
     end
 
+    validation_error I18n.t(
+      "barclamp.#{@bc_name}.validation.engine_roles_mismatch",
+      db_engine: active_engine
+    ) unless selected_engines.include?(active_engine)
+
+    validation_error I18n.t(
+      "barclamp.#{@bc_name}.validation.secondary_psql"
+    ) if selected_engines.length > 1 && active_engine == "mysql"
```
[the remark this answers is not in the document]
rhafer: I agree. I think we want to avoid the user from deploying a **new** postgresql when mariadb is already installed. But I think we need to have it possible to have postgresql and mysql  installed for some limited time even after the switch to mariadb.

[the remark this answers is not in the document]
rhafer: I guess this could be implemented by overwriting `validate_proposal_elements` and having it load the saved proposal to compare new vs old settings.

crowbar_framework/app/views/barclamp/database/_edit_attributes.html.haml
skazi0: I don't remember exactly but the idea was that this is on top of the page and visible all the time so that user knows it before starting to play around with the roles. @jsuchome any comment on that?

jsuchome: Yes, I wanted to show some warning to users visible even before applying the roles.

rhafer: That doesn't really conflict with what I wrote (or at least with what I meant to say :-) ). Popup the warning as soon as somebody drops a node (or cluster) on the 2nd role (before hitting apply).

But in the end I don't have a very strong opinion on this. If you prefer it this way, just keep it.

skazi0:
Rebased. It would be good if someone does a quick test-run of this version before final approval.

jsuchome:
Hm, I have an error at `undefined method `empty?' for nil:NilClass
/opt/dell/crowbar_framework/app/models/database_service.rb:189`

It's the check for `old_proposal.elements["database-server"].empty?`

My setup had mysql-server from start and I - just out of curiosity - added nodes to database-server role. I actually think this should not be allowed, but no warning occured, just the error

jsuchome:
@skazi0 what about the part that this case should not be even allowed? (Maybe you have it covered, I just do not have that setup any more)

skazi0:
@jsuchome the part which failed should take care of exactly this case: if there was mysql and there were no psql and you want to add psql then it should not allow this.

PR #1710, opened 2018-07-20
Set a proper ACL for shared certificates (bsc#1101154)

aplanas:
If the user is sharing a certificate between different OpenStack
services, this needs to be readed for different system users.  For
example, if the client store the certificate in /etc/apache2/ssl.key,
this directory and all the parents needs to be read for all the
OpenStack service users, and also the certificates contained there.

To address that we use the ACL from the operating system, via the
`setfacl` command.

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb
@@ -72,6 +72,15 @@ class Chef
           end
           # We do not check for existence of keyfile, as the private key is
           # allowed to be in the certfile
+
+          # If we do no generate the certificate, we need to be sure
+          # that is readable for the user.  In some configurations we
+          # need to share the same certificate for multiple services,
+          # so needs to be readable for multiple different users and
+          # groups (for example, if we share the apache certificate
+          # for Nova and the Dashboard)
+          _fix_acl @current_resource.certfile, @current_resource.group
+          _fix_acl @current_resource.keyfile, @current_resource.group
         end # if generate_certs
 
         if @current_resource.cert_required && ! ::File.size?(@current_resource.ca_certs)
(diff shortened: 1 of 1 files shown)


aplanas:
[P2 Bug 1101154: L3: after applying patches vnc-console is not working anymore (gendorf.de)](https://trello.com/c/332gLEG7/1033-p2-bug-1101154-l3-after-applying-patches-vnc-console-is-not-working-anymore-gendorfde)

chef/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb
```diff
@@ -80,6 +89,23 @@ def action_setup
...
         end
       end
+
+      def _fix_acl(certificate, group)
+        partial = "/"
+        directory.split(File::SEPARATOR).each do |entry|
+          next if entry.empty?
+
+          partial = File.join(partial, entry)
+          next if File.world_readable?(partial)
```
JanZerebecki: This misses directories that are world readable but not world executable.

JanZerebecki:
Looks good otherwise.

chef/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb
aplanas: Uhm right, but not sure that this is a problem as for reading the certificate we do not need to list the content ...

JanZerebecki: On Linux this is not true:
```
# ls -la /home/jan/testx
total 12
drwxr-xr--  2 root root 4096 Jul 26 17:48 .
drwxr-xr-x 66 jan  jan  4096 Jul 26 17:47 ..
-rw-r--r--  1 root root    4 Jul 26 17:48 bar
$ cat ~/testx/bar 
cat: /home/jan/testx/bar: Permission denied
$ ls -la ~/testx/
ls: cannot access '/home/jan/testx/.': Permission denied
ls: cannot access '/home/jan/testx/..': Permission denied
ls: cannot access '/home/jan/testx/bar': Permission denied
total 0
d????????? ? ? ? ?            ? .
d????????? ? ? ? ?            ? ..
-????????? ? ? ? ?            ? bar
```

aplanas:
@JanZerebecki I try to address the commented case

chef/cookbooks/crowbar-openstack/libraries/provider_ssl_setup.rb
```diff
@@ -80,6 +89,29 @@ def action_setup
...
         end
       end
+
+      def _fix_acl(certificate, group)
+        partial = "/"
+        directory.split(File::SEPARATOR).each do |entry|
+          next if entry.empty?
+
+          partial = File.join(partial, entry)
+          next if File.world_readable?(partial) && File.file?(partial)
```
JanZerebecki: Putting the conditions of this line and the next two into one function should appease the hound.

PR #1711, opened 2018-07-23
/etc/sysctl.d/99-sysctl.conf is a symlink to /etc/sysctl.conf

rsalevsky:
We should really use the direct file instead of relying on a symlink.

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents.rb
@@ -39,9 +39,9 @@ ruby_block "edit /etc/sysconfig/sysctl for IP_FORWARD" do
 end
 
 # Enable ip forwarding on network node for new SUSE platforms
-ruby_block "edit /etc/sysctl.d/99-sysctl.conf for net.ipv4.ip_forward" do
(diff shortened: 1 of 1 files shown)


rsalevsky:
I'm not sure about this change. Since some days the symlink is gone and the CI fails.

jsuchome:
See e.g. https://ci.suse.de/job/openstack-mkcloud/125743/ for the failure

rsalevsky:
@jsuchome The file comes from aaa_base but for this package was no update since January. The symlink exists on the admin node but is not owned by any package; on the control nodes the symlink doesn't exist.

jsuchome:
Yeah, I've found the origin of /etc/sysctl.conf. But who creates (created) the symlink is unclear

aplanas:
I can see the fails in the CI, but for me is not clear if this is a wrong test package from SLE

jsuchome:
@aplanas so do you see where is this coming from?

(Anyway I think that Rick and Abel are right and the patch would make sense even if the link would exist)

jsuchome:
@aplanas oh, you've added the original code! Now you have to decide ;-)

PR #1712, opened 2018-07-23
database: Drop temporary mysql-server role, used for upgrade only

jsuchome:
During upgrade there was a time where both different database engines
were deployed to the same noes and we needed extra role.
Once back on SOC8, there's again need for only one database role.

See https://github.com/crowbar/crowbar-openstack/pull/1746 for spliting the roles in SOC7

### DIFF
--- a/chef/data_bags/crowbar/migrate/database/209_merge_db_roles.rb
+++ b/chef/data_bags/crowbar/migrate/database/209_merge_db_roles.rb
@@ -1,0 +1,45 @@
+def upgrade(ta, td, a, d)
+  a["ha"] = ta["ha"] unless a.key? "ha"
+
+  # Cookbooks expect the 'enabled' flag under high level 'ha' map
+  # - only during the upgrade it was moved under 'ha/mysql'
+  if a["mysql"] && a["mysql"]["ha"]
+    a["ha"]["enabled"] = a["mysql"]["ha"]["enabled"] if a["mysql"]["ha"].key?("enabled")
+    a["mysql"]["ha"].delete("enabled")
+  end
+
+  a["postgresql"].delete("ha") if a["postgresql"] && a["postgresql"].key?("ha")
+
+  if d["elements"].key? "mysql-server"
+    d["elements"]["database-server"] = d["elements"]["mysql-server"]
+    d["elements"].delete("mysql-server")
+    if d.fetch("elements_expanded", {}).key? "mysql-server"
+      d["elements_expanded"]["database-server"] = d["elements_expanded"]["mysql-server"]
+      d["elements_expanded"].delete("mysql-server")
+    end
+
+    # Make sure mysql-server role is gone from all places
+    d["element_states"] = td["element_states"]
+    d["element_order"] = td["element_order"]
+
+    chef_order = BarclampCatalog.chef_order("database")
+    nodes = NodeObject.find("run_list_map:mysql-server")
+    nodes.each do |node|
+      node.add_to_run_list("database-server", chef_order,
+                           td["element_states"]["database-server"])
+      node.delete_from_run_list("mysql-server")
+      node.save
+    end
+  end
+
+  # Delete mysql-server role if it exists
+  role = Chef::Role.load("mysql-server") rescue nil
+  role.destroy unless role.nil?
+
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  # No role splitting needed now; it already has suppport in SOC7 via migration 109
+  return a, d
+end
--- a/chef/data_bags/crowbar/template-database.json
+++ b/chef/data_bags/crowbar/template-database.json
@@ -83,7 +83,7 @@
     "database": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 208,
+      "schema-revision": 209,
       "element_states": {
         "database-server": [ "readying", "ready", "applying" ]
       },


chef/data_bags/crowbar/migrate/database/209_merge_db_roles.rb
```diff
@@ -0,0 +1,27 @@
+def upgrade(ta, td, a, d)
+  # there's still an option to deploy PostgreSQL in SOC8
+  if a["sql_engine"] == "mysql"
+    d["elements"]["database-server"] = d["elements"]["mysql-server"]
```
jhesketh: I think we'll need to check if d['elements']['mysql-server'] exists. If somebody runs this against a clean C8 deploy in a maintenance update they'll have an sql_engine of mysql, but not the odd schema

jsuchome: True

chef/data_bags/crowbar/migrate/database/210_drop_postgreql.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
+  a.delete("postgresql") if a.key? "postgresql"
+  a.delete("ha") if a.key? "ha"
```
rhafer: You can just remove that `if a.key? ....` part. Deleting an non-existing key wont throw an error in ruby.

chef/data_bags/crowbar/migrate/database/209_merge_db_roles.rb
```diff
@@ -0,0 +1,26 @@
...
+    chef_order = BarclampCatalog.chef_order("database")
+    nodes = NodeObject.find("run_list_map:mysql-server")
+    nodes.each do |node|
+      node.add_to_run_list("database-server", chef_order,
+                           td["element_states"]["database-server"])
+      node.delete_from_run_list("mysql-server")
+      node.save
+    end
+  end
+
```
rhafer: I am not entirely sure here. But don't we need to do something with the ha-enabled attribute for mysql? I think for crowbar/4.0 we moved that from `[ha][enabled]` to `[mysql][ha][enabled]`.

We probably should port that part to master and 5.0 as well :/ (or migrate it back to `[ha][enabled]`)

jsuchome: My idea was to port this part to 5.0 (check the discussion in https://jira.suse.de/browse/SCRD-3943)
I'm surprised it is not part of this PR, I probably forgot or intended to do it separately

rhafer: I am not entirely sure what you mean here. But currently master (and 5.0) is still using `[database][ha][enabled]` for galera setups (https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/mysql/recipes/server.rb#L23) while stable/4.0 will soon switch to `[database][mysql][ha][enabled] (https://github.com/crowbar/crowbar-openstack/pull/1746/files#diff-f521d41b9a8ee2ff558c030e1b5b1c73). So we need to address this as part of this migraton. But maybe that is what you have been saying already :)

jsuchome: I tried (not very clearly).

But thinking about it again, might be better to do a small fix in migration and go back to `[database][ha][enabled]` and not try to port crowbar code that is accessing this value (that would be parts of https://github.com/crowbar/crowbar-openstack/pull/1746 I believe)

rhafer: Yeah, I think both approaches are fine. And I agree that going back to `[database][ha][enabled]` seems simpler.

jhesketh:
I think my previous concern is still valid here. It's perhaps a moot issue if the migrations are ran after the C7 upgrade and before C8 is turned on, but I'll note them again for completeness:

> My concern here is that we can end up at this migration in two ways. 1) coming from a fresh C8 install (either from scratch, or from a maintenance update), 2) coming after doing a C7 upgrade. In each case the previous schema is different.
> 
> Perhaps this should not be a general migration but rather a migration step done as part of the upgrade process?

Another thought to bring the 2 versions into line, could be to place this as migration 199, but as mentioned above I feel it should perhaps be a step of the migration itself to ensure C7 is in an expected state.

rhafer:
> I think my previous concern is still valid here. It's perhaps a moot issue if the migrations are ran after the C7 upgrade and before C8 is turned on, but I'll note them again for completeness:
> 
>     My concern here is that we can end up at this migration in two ways. 1) coming from a fresh C8 install (either from scratch, or from a maintenance update), 2) coming after doing a C7 upgrade. In each case the previous schema is different.

Yes. Though migration 209 is basically a no-op when coming from a fresh (or maintenance updated) C8 deployment. So it's whole purpose is to bring systems upgrading from C7 into the correct shape for C8. So I think the current approach is already taking care of that fact.

chef/data_bags/crowbar/migrate/database/209_merge_db_roles.rb
```diff
@@ -0,0 +1,33 @@
+def upgrade(ta, td, a, d)
+  a["ha"] = ta["ha"] unless a.key? "ha"
+  a["ha"]["enabled"] = true if a["mysql"]["ha"]["enabled"]
```
rhafer: II think we should delete `a["mysql"]["ha"]["enabled"]` here to avoid more confusion. Also this line would benefit from a comment on why this reassignement to a["ha"]["enabled"] is actually needed.

chef/data_bags/crowbar/migrate/database/209_merge_db_roles.rb
```diff
@@ -0,0 +1,43 @@
+def upgrade(ta, td, a, d)
+  a["ha"] = ta["ha"] unless a.key? "ha"
+
+  # Cookbooks expect the 'enabled' flag under high level 'ha' map
+  # - only during the upgrade it was moved under 'ha/mysql'
+  if a["mysql"]["ha"]["enabled"]
+    a["ha"]["enabled"] = true
+    a["mysql"]["ha"].delete("enabled")
```
rhafer: This has to happen regardless of the value of ` a["mysql"]["ha"]["enabled"]`. If it is `false` we also should delete it. If it is not present, it doesn't hure either.

chef/data_bags/crowbar/migrate/database/209_merge_db_roles.rb
```diff
@@ -0,0 +1,45 @@
+def upgrade(ta, td, a, d)
+  a["ha"] = ta["ha"] unless a.key? "ha"
+
+  # Cookbooks expect the 'enabled' flag under high level 'ha' map
+  # - only during the upgrade it was moved under 'ha/mysql'
+  if a["mysql"] && a["mysql"]["ha"]
```
skazi0: why not do the same `fetch` trick as below?

PR #1713, opened 2018-07-23
[4.0] split database-server role into backend specific roles

jsuchome:
This is the extension of https://github.com/crowbar/crowbar-openstack/pull/1691 with followup fixes

Please read the original PR for more details and TODO.


jhesketh:
Looking good to me. It would be helpful to have some comments on what is happening. In particular the helper function and the migrations.

jsuchome:
I've added some comments into the migration code

PR #1714, opened 2018-07-23
rabbitmq: set client timout to default value

Itxaka:
A very low client timeout can lead to multiple disconnections of
the clients due to missing heartbeats, depending on the load on the
rabbit nodes. Instead of configuring the clients with a very low
value, lets just use the default openstack values for safety


bbobrov:
the jobs are failing for a unrelated reason

PR #1716, opened 2018-07-24
[4.0] database: Move PostgreSQL settings above MySQL

skazi0:
Both database forms can be shown at the same time now so order is
more important. PostgreSQL is the "old" engine so it should be on top.

Will need a rebase after #1705 is merged

### DIFF
--- a/crowbar_framework/app/views/barclamp/database/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/database/_edit_attributes.html.haml
@@ -5,29 +5,6 @@
   .panel-body
     = select_field :sql_engine, :collection => :engines_for_database
 
-    #mysql_container
-      %fieldset
-        %legend
-          = t('.mysql_attributes')
-
-        = integer_field %w(mysql max_connections)
-        = integer_field %w(mysql expire_logs_days)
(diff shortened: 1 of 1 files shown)


jsuchome (approved):

jhesketh (approved):

rsalevsky:
I would argue different and would place MySQL on top as we want everyone to use MySQL. Further I would place a depreciation warning on top of PostgreSQL-

jsuchome:
The point is that this is intended for SOC7 where users already have PostgreSQL. With our new set of patches they will be able to apply mariadb, but not remove postgres, that's a preparation for migration step.
I find it more logical to see on top the stuff that is already deployed and new options below.

rsalevsky:
If there is already a database deployed then this makes sense, can we easily switch the default behavior to something like: If a no database or MariaDB is deployed or show MariaDB first otherwise show PostgreSQL first?

jsuchome:
For such kind of logic there's https://github.com/crowbar/crowbar-openstack/pull/1705 ... you can follow the discussion out there.

rsalevsky:
Thanks, I will follow-up there.

PR #1718, opened 2018-07-26
[4.0] database: Make sure correct role is assigned when creating new proposal

jsuchome:
If sql_engine is already set before creating proposal (e.g. by mkcloud)
it should be honored when auto-assigning the database role.


jsuchome:
This is not gonna work as expected.

PR #1720, opened 2018-07-26
neutron: add force_metadata attribute

ilausuch:
This changes allows to force_metadata from the proposal. Previously, if this attribute was added directly to configuration in neutron network nodes, it was removed from the config in each chef-client execution

### DIFF
--- a/chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
@@ -4,6 +4,7 @@ resync_interval = <%= @resync_interval %>
 dhcp_driver = <%= @dhcp_driver %>
 enable_isolated_metadata = <%= @enable_isolated_metadata %>
 enable_metadata_network = <%= @enable_metadata_network %>
+force_metadata = <%= @force_metadata %>
 dns_domain = <%= @dns_domain %>
 <% if @nameservers -%>
 dnsmasq_dns_servers = <%= @nameservers %>
--- a/chef/data_bags/crowbar/migrate/neutron/210_add_force_metadata_attributes.rb
+++ b/chef/data_bags/crowbar/migrate/neutron/210_add_force_metadata_attributes.rb
@@ -1,0 +1,9 @@
+def upgrade(ta, td, a, d)
+  a["metadata"] = ta["metadata"] unless a.key? "metadata"
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a.delete("metadata") unless ta.key? "metadata"
+  return a, d
+end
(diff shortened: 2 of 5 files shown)


ilausuch:
[Disable Service on compute nodes](https://trello.com/c/pIBnFo2W/296-disable-service-on-compute-nodes)

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
```diff
@@ -4,6 +4,7 @@ resync_interval = <%= @resync_interval %>
 dhcp_driver = <%= @dhcp_driver %>
 enable_isolated_metadata = <%= @enable_isolated_metadata %>
 enable_metadata_network = <%= @enable_metadata_network %>
+force_matadata = <%= @force_metadata %>
```
Itxaka: typo :+1:

chef/data_bags/crowbar/migrate/neutron/210_add_force_metadata_attributes.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
```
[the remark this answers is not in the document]
Itxaka: I do agree with hound here, this parameters are shameful, but I guess its the way it has been done before ¯\\\_(ツ)_/¯

chef/data_bags/crowbar/template-neutron.schema
```diff
@@ -221,6 +221,11 @@
                       "type": "map", "required": true, "mapping": {
                         "neutron-server": { "type": "int", "required": true }
                       }
+                    },
+                    "metadata": {
+                      "type": "map", "required": false, "mapping": {
```
Itxaka: do we really need a map here instead of a simple `force_metadata` bool? Is it expected to add more keys to the metadata?

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
ilausuch: Ups! error moving the code to PR

chef/data_bags/crowbar/template-neutron.schema
ilausuch: There are two more flags that are fixed in the configuration:

enable_isolated_metadata: "True",
enable_metadata_network: "False",

So I assumed that maybe in the future one of this could be configured in the proposal.

chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
ilausuch: Fixed! Thanks

chef/data_bags/crowbar/template-neutron.schema
Itxaka: :+1:

ilausuch:
After @Itxaka I tested it again

PR #1721, opened 2018-07-27
mariadb: Set mariadb as the default database

dirkmueller:
In Cloud 8 we will use maridb as the default and only supported database
for OpenStack. This change sets the database to mysql (mariadb).

(cherry picked from commit b72823f3abbb8349582188d6af62824dab6021e3)

### DIFF
--- a/chef/data_bags/crowbar/template-database.json
+++ b/chef/data_bags/crowbar/template-database.json
@@ -3,7 +3,7 @@
   "description": "Installation for Database",
   "attributes": {
     "database": {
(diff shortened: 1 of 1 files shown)


dirkmueller:
Need to check documentation situation before switching default

jsuchome:
Note that this breaks current SOC7 CI

rhafer:
Seems that something similar landed as part of this: https://github.com/crowbar/crowbar-openstack/pull/1746

PR #1723, opened 2018-08-03
rabbitmq: fix extra users password regeneration

ilausuch:
Every time the rabbitmq proposal was applied the extra users
passwords was regenerated. Featured introduced in #1506

This fix mantain the paswords.


ilausuch:
[rabbitmq extra_users always get new random password](https://trello.com/c/mIXlq4Lu/298-rabbitmq-extrausers-always-get-new-random-password)

PR #1729, opened 2018-08-03
[5.0] nova: reload nova-placement-api (bsc#1103383)

aplanas:
nova-placement-api service is behing WGSI (apache2), and in order
to reload it when nova-placement.conf file change, we need to
create a proxy service.

This proxy will signal the WSGI apache modules for a reload,
without a full restart of apache service.

(cherry picked from commit db813e1ce3042eb17214532063ac8de9557fec98)

### DIFF
--- a/chef/cookbooks/nova/recipes/config.rb
+++ b/chef/cookbooks/nova/recipes/config.rb
@@ -39,6 +39,19 @@ package "nova-common" do
   action :install
 end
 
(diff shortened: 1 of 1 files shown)


aplanas:
[P2 Bug 1103383 - L3: nova-placement-api self-signed cert failed](https://trello.com/c/HAacvsBB/1048-p2-bug-1103383-l3-nova-placement-api-self-signed-cert-failed)

PR #1731, opened 2018-08-07
mysql: ha galera needs op monitor for slaves

AbelNavarro:
In a Galera HA environment the pacemaker primitive op monitor is only
set for master mode. This is not a problem because all galera resources
are in master mode.

The problem comes when changing the pacemaker primitive and the masters
can become slaves, i.e. when shrinking the cluster. In such case the
slaves never get promoted to masters because they are not monitored.
Adding the monitor operation for slaves solves this situation.

### DIFF
--- a/chef/cookbooks/mysql/attributes/server.rb
+++ b/chef/cookbooks/mysql/attributes/server.rb
@@ -31,8 +31,9 @@ default[:mysql][:ha][:ports][:admin_port] = 3306
 
 # Default operation setting for the galera resource
 # in pacemamker
-default[:mysql][:ha][:op][:monitor][:interval] = "20s"
-default[:mysql][:ha][:op][:monitor][:role]     = "Master"
+default[:mysql][:ha][:op][:monitor] = [
+  { interval: "23s" }, { interval: "20s", role: "Master" }
+]
 
 # If needed we can enhance this to set the mariadb version
 # depeding on "platform" and "platform_version". But currently


AbelNavarro:

chef/cookbooks/mysql/attributes/server.rb
```diff
@@ -31,8 +31,9 @@
 
 # Default operation setting for the galera resource
 # in pacemamker
-default[:mysql][:ha][:op][:monitor][:interval] = "20s"
-default[:mysql][:ha][:op][:monitor][:role]     = "Master"
+default[:mysql][:ha][:clustered_op][:monitor] = [
+  { interval: "23s" }, { interval: "20s", role: "Master" }
```
jsuchome: Why 23?

chef/cookbooks/mysql/attributes/server.rb
```diff
@@ -31,8 +31,9 @@
 
 # Default operation setting for the galera resource
 # in pacemamker
-default[:mysql][:ha][:op][:monitor][:interval] = "20s"
-default[:mysql][:ha][:op][:monitor][:role]     = "Master"
+default[:mysql][:ha][:clustered_op][:monitor] = [
```
jsuchome: Can't you just extend `default[:mysql][:ha][:op]` instead of replacing it?

chef/cookbooks/mysql/attributes/server.rb
AbelNavarro: The idea was to differentiate the two roles, but I reckon I just made up that value and can be something else. Maybe the original 20s can stay.

chef/cookbooks/mysql/attributes/server.rb
AbelNavarro: I based this change on a similar PR for rabbitmq (https://github.com/crowbar/crowbar-openstack/pull/1463). In there they also added `clustered_op` replacing `op`. Notice that `op` remained in the old code, but is unused afaik.

chef/cookbooks/mysql/attributes/server.rb
rhafer: Actually you need to have different intervals for each monitor operations other pacemaker can't distinguish them. So it can't be 20 for both.

AbelNavarro: I just followed the pattern with other monitors, but thanks @rhafer for the explanation.

chef/cookbooks/mysql/attributes/server.rb
rhafer: Hm, this is actually wrong. `[:mysql][:ha][:op]` is used here: https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/mysql/recipes/ha_galera.rb#L187

The cluster_op thing is really rabbitmq specific since it is using different ops settings depending on the type of HA setup (native clustering vs. active/passive)

AbelNavarro: @rhafer you are right. I'm afraid then I'll have to add some logic to differentiate between arrays and hashes when merging. I did something similar [here](https://github.com/crowbar/crowbar-ha/pull/297/files#diff-a2789b0ee89232d5d92e03b62b98ed2fR140) but I still have to adapt it.

chef/cookbooks/mysql/attributes/server.rb
AbelNavarro: Actually looking at that again, I only renamed the variable in both places so there was no conflict there. I think it will be easier to understand if I keep the original variable name.

About the extra logic I mentioned before, I don't think it is needed. We want the proposal values to override the barclamp ones, and the `primitive_op.merge!` already does that. The problem would be the proposal doesn't have the two roles defined.

ilausuch:
I agree with @Itxaka mention the 23 seconds of interval. Is there any reason to this number? In spite of I suppose it is not important

rhafer:
> I agree with @Itxaka mention the 23 seconds of interval. What is the reason to set this number?

It's not important that is it 23, but is has to be different that the interval of the other monitor operation (which is 20). It's AFAIK a requirement of pacemaker that all monitor ops need to use a different interval.

PR #1734, opened 2018-08-08
[4.0] Synchronize SSL in the cluster (bsc#1081518)

aplanas:
When Horizon in deployed via a autogenerated self certificate, the
current code will generate a different one of each node in the
cluster.

This patch will generate a single certificate in the founder, and
share it with the rest of the cluster via `rsync`.

(cherry picked from commit b264c759f862d4b81e0a67095032e156965a17e3)

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -494,16 +494,47 @@ end
 
 include_recipe "horizon::ha" if ha_enabled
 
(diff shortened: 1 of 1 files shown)


aplanas:
[P2: Bug 1081518 - Loop accepting certificates when using self-signed certificates](https://trello.com/c/mUUkRrpJ/1055-p2-bug-1081518-loop-accepting-certificates-when-using-self-signed-certificates)

aplanas:
Tested by QA

PR #1736, opened 2018-08-17
Heat: Make non-founder HA nodes do less work

matthewoliver:
There is an intermittent update and HA ci failure that originates
from the heat::server recipe. It has been observed in manual upgrade
tests and in ci.

The problem goes something like this. The founder set's a wait pacemaker
sync mark:

    crowbar_pacemaker_sync_mark "wait-heat_register" if ha_enabled

And then goes off and registers heat with keystone. The non-founders
happily sit and wait for the founder to finish and set the create mark.
When the non-founders finally start, none of the register jobs that the
founder has just completed has "only_if's" attached so the non-founders
happily go and re-run them.
The problem is, these register jobs can take time. In the failures I've
seen the `register heat domain` block can take up to 40-50 seconds. This places
the non-founders well behind the founder.

The intermittent failure happens when we get later in the chef run, where any
more delays on one of the non-founders can cause later sync marks to
timeout. Which usually fails in the ceilometer recipe where the founder
times out after waiting a minute for the other nodes to catch up,
resulting in:

    FATAL: RuntimeError: crowbar-pacemaker_sync_mark[sync-ceilometer_central_before_ha] (ceilometer::central_ha line 19) had an error: \
    RuntimeError: Some cluster nodes didn't set ceilometer_central_before_ha!

This patch goes and adds a:

to each keystone register block between the wait and create so once the
non-founders start again they don't go re-do all that the founder just
completed.

### DIFF
--- a/chef/cookbooks/heat/recipes/server.rb
+++ b/chef/cookbooks/heat/recipes/server.rb
@@ -130,6 +130,7 @@ keystone_register "heat wakeup keystone" do
   port keystone_settings["admin_port"]
   auth register_auth_hash
   action :wakeup
+  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
 keystone_register "register heat user" do
@@ -142,6 +143,7 @@ keystone_register "register heat user" do
   user_password keystone_settings["service_password"]
   project_name keystone_settings["service_tenant"]
   action :add_user
+  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
 keystone_register "give heat user access" do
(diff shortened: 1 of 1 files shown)


matthewoliver:
The option was to simply add another sync mark to make sure the non-founder and founders where back inline. But that seemed more of a hack then stopping the non-founders from reregistering everything with keystone again.

I'm still running this in a mkcloud, but wanted to get the code up. Thus the WIP. Plus to make sure this is the best way to do this in chef/ruby by getting eyes on it.

matthewoliver:
Lol, thanks @aspiers thanks for the corrections. Of course yeah the non-founders just stop at wait, the founder doesn't fire off anything, that makes sense. If I need to get around to pushing another patchset I'll amend :)

What you mean 'set' doesn't have ownership or it isn't read "the founder set is a wait pacemaker mark" :P 
Thanks, goes to show one should re-read the commit message and not auto filter out punctuation in their (was so tempted to use 'there' here) head.

PR #1737, opened 2018-08-21
Do not share ports Apache2 and HAProxy (bsc#1105086)

aplanas:
If HAProxy is serving :80 or :443 (because Horizon), we do not want
that Apache2 can also be listening on those ports.

HAProxy can be configured to resuse sockets, so that means that the
tuple IP:Port can be reused for different services, produncing bugs
like bsc#1105086, where the same IPv4 and the same port (:80) is
being listening connections from Apache and for HAProxy.

This patch remove the default :listen_ports from Apache if HAProxy
is deployed for Horizon, and reapply the Apache2 recipe from Crownar
Core.

### DIFF
--- a/chef/cookbooks/horizon/recipes/ha.rb
+++ b/chef/cookbooks/horizon/recipes/ha.rb
@@ -33,6 +33,15 @@ if node[:horizon][:apache][:ssl]
   end.run_action(:create)
 end
 
+# Once HAProxy is taking care of :80 and :443 we need to remove this
+# from Apache realm.  This requires update the node information from
+# Apache, and maybe the listen.conf file
+if node[:apache][:listen_ports].include?("80") || node[:apache][:listen_ports].include?("443")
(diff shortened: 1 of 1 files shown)


aplanas:
[P2 Bug 1105086: L3: haproxy failing to direct to horizon on admin and public (tcs.com)](https://trello.com/c/DEbZUGe8/1100-p2-bug-1105086-l3-haproxy-failing-to-direct-to-horizon-on-admin-and-public-tcscom)

aplanas:
Backpors in https://github.com/crowbar/crowbar-openstack/pull/1738 and in https://github.com/crowbar/crowbar-openstack/pull/1739

aplanas:
@aojea haproxy is attached to the VIP only, but also apache, as is using *:80 in the default configuration. This patch remove this *:80, detaching it from all the interfaces (including the VIP)

PR #1738, opened 2018-08-21
Do not share ports Apache2 and HAProxy (bsc#1105086)

aplanas:
If HAProxy is serving :80 or :443 (because Horizon), we do not want
that Apache2 can also be listening on those ports.

HAProxy can be configured to resuse sockets, so that means that the
tuple IP:Port can be reused for different services, produncing bugs
like bsc#1105086, where the same IPv4 and the same port (:80) is
being listening connections from Apache and for HAProxy.

This patch remove the default :listen_ports from Apache if HAProxy
is deployed for Horizon, and reapply the Apache2 recipe from Crownar
Core.

(cherry picked from commit 55ce3c5a6bff726a5148018915627028cae4f3bb)

### DIFF
--- a/chef/cookbooks/horizon/recipes/ha.rb
+++ b/chef/cookbooks/horizon/recipes/ha.rb
@@ -33,6 +33,15 @@ if node[:horizon][:apache][:ssl]
   end.run_action(:create)
 end
 
+# Once HAProxy is taking care of :80 and :443 we need to remove this
+# from Apache realm.  This requires update the node information from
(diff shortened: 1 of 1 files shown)


aplanas:
[P2 Bug 1105086: L3: haproxy failing to direct to horizon on admin and public (tcs.com)](https://trello.com/c/DEbZUGe8/1100-p2-bug-1105086-l3-haproxy-failing-to-direct-to-horizon-on-admin-and-public-tcscom)

PR #1746, opened 2018-08-27
[4.0] split database-server role into backend specific roles

jsuchome:
New PR created after the revert of https://github.com/crowbar/crowbar-openstack/pull/1713

(original is https://github.com/crowbar/crowbar-openstack/pull/1691)

https://github.com/SUSE-Cloud/automation/pull/2600 is required for correct roles assignement in HA setup

### DIFF
--- a/chef/cookbooks/database/recipes/role_mysql_server.rb
+++ b/chef/cookbooks/database/recipes/role_mysql_server.rb
@@ -1,14 +1,11 @@
 #
-# Cookbook Name:: database
-# Recipe:: server
-#
-# Copyright 2012, SUSE Linux Products GmbH
+# Copyright 2018, SUSE LINUX GmbH
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     http://www.apache.org/licenses/LICENSE-2.0
+#   http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
@@ -16,9 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
-backend = node[:database][:sql_engine]
-
-Chef::Log.info("Running database::server for #{backend}")
-
-include_recipe "#{backend}::server"
+if CrowbarRoleRecipe.node_state_valid_for_role?(node, "database", "mysql-server")
+  include_recipe "mysql::server"
+end
--- a/chef/data_bags/crowbar/migrate/database/109_separate_db_roles.rb
+++ b/chef/data_bags/crowbar/migrate/database/109_separate_db_roles.rb
@@ -1,0 +1,47 @@
+def upgrade(ta, td, a, d)
+  d["element_states"] = td["element_states"]
+  d["element_order"] = td["element_order"]
+
+  if a["sql_engine"] == "mysql"
+    d["elements"]["mysql-server"] = d["elements"]["database-server"]
+    d["elements"]["atabase-server"] = []
+    if d.fetch("elements_expanded", {}).key? "database-server"
+      d["elements_expanded"]["mysql-server"] = d["elements_expanded"]["database-server"]
+      d["elements_expanded"].delete("database-server")
+    end
+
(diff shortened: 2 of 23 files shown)


rhafer:
It seems to be a bug in the schema migration.

chef/data_bags/crowbar/migrate/database/109_separate_db_roles.rb
```diff
@@ -0,0 +1,62 @@
+def upgrade(ta, td, a, d)
+  db_engine = a["sql_engine"]
+
+  # 'ha' hash needs to be moved under 'postgresql' to keep it consistent with mysql
+  if db_engine == "postgresql"
+    a["postgresql"]["ha"] = a["ha"]
+  else
+    a["postgresql"]["ha"] = ta["postgresql"]["ha"]
```
rhafer: In the `mysql` case we need to move parts of the `database.ha` attributes (at least `database.ha.enabled` is needed) to `database.mysql.ha` but make sure to not entirely overwrite what is already there.

jsuchome: Hm, strange, I thought we've played with it last time and this was the working version...

jsuchome: @rhafer I think the problem is that "enabled" is not part of the schema, thus it's gonna fail with
`a["mysql"]["ha"]["enabled"] = ta["database"]["ha"]["enabled"]`

But shouldn't it be set to correct value when you reapplied database barclamp?

skazi0: `ha/enabled` is set by `prepare_role_for_ha` which will be run on proposal (re)apply. Do we want to support migration without this?

skazi0: iirc my idea was to do `a["mysql"]["ha"]["enabled"] = a["ha"]["enabled"] if db_engine == "mysql"`

skazi0: or even check for presence of `ha/enabled` to support "never applied" proposals as well.

jsuchome: Yeah, here starts the old thread https://github.com/crowbar/crowbar-openstack/pull/1691#discussion_r203997377

jsuchome: `a["mysql"]["ha"]["enabled"] = ta["database"]["ha"]["enabled"] if a["mysql"]["ha"].key? "enabled"`

rhafer: Hm, that can't really work IMO. `ta["database"]["ha"]["enabled"]` should actually always throw an exception. As `ta` contains the attributes tree from below the `database` key.

I think you really just want `a["mysql"]["ha"]["enabled"] = true if  a["ha"]["enabled"]`

If `a["ha"]["enabled"]` is not present is will evaluate to `nil` and `a["mysql"]["ha"]["enabled"]` will not be set.

BTW, regaring your other comment. If `[ha][enabled]` is not part of the schema it will also never be part of the template (`ta`), but that is actually fine.

rhafer: > But shouldn't it be set to correct value when you reapplied database barclamp?

Yes. It should. I didn't try it though. And  I think we can't require the user to reapply the database barclamp after installing this as a maintenance update IMO.

jsuchome: You're right

chef/cookbooks/database/recipes/role_mysql_server.rb
```diff
@@ -1,24 +1,18 @@
...
 # limitations under the License.
 #
-
-backend = node[:database][:sql_engine]
-
-Chef::Log.info("Running database::server for #{backend}")
-
-include_recipe "#{backend}::server"
+if CrowbarRoleRecipe.node_state_valid_for_role?(node, "database", "mysql-server")
+  include_recipe "mysql::server"
```
jhesketh: (nit) Should probably also have a log line like the postgres one:
`Chef::Log.info("Running mysql::server for MySQL")`

PR #1747, opened 2018-08-29
rabbitmq: allow tuning tcpnodelay and keepalive

dirkmueller:
nodelay is already on by default and keepalive is off by default,
but some env might want to be able to configure it.


dirkmueller:
@nicolasbock all options are rendered into the configuration template (it iterates over the set options)

nicolasbock:
@dirkmueller Sorry I didn't see that. Then it looks great! :smile:

PR #1749, opened 2018-08-30
[neutron:cisco-aci] ACI role - Backport from #1699

mmnelemane:
So far, the implementation assumed that only compute nodes
would be attached to the ACI leaf ports and the controller
nodes are independently attached to the network not connected
to the ACI fabric itself. The only form of communication from
the controller to the ACI is through the agents. However,
recent customer POCs have changed our assumption since these
setups expect all or most of the nodes attached the ACI leaf
as the ACI fabric forms the central network infrastructure
of the entire deployment. In such scenarios, we need a mechanism
where we can know which of the nodes are attached to the ACI
leaf and which nodes are not. The agents should be enabled
and run on each node attached to the ACI leaf ports.
This commit tries to achieve exactly this result, so that
future deployments can flexibly allow any node attached to
the fabric.

### DIFF
--- a/chef/cookbooks/heat/recipes/server.rb
+++ b/chef/cookbooks/heat/recipes/server.rb
@@ -130,6 +130,7 @@ keystone_register "heat wakeup keystone" do
   port keystone_settings["admin_port"]
   auth register_auth_hash
   action :wakeup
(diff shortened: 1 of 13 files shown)


mmnelemane:
Replaced by #1750 for the correct branch.

PR #1754, opened 2018-09-03
Correct configuration of DRBD resources in 3-node cluster

jsuchome:
We have to allow user to add a new node into 2-node DRBD cluster, so that such cluster can hold (3-node) galera configuration later.

But as (our version of) DRBD needs exactly 2 nodes, there are some tweaks necessary so that the extra node does not have DRBD configured, although it is part of the pacemaker cluster.

This part handles necessary changes for DRBD-enabled services (postgresql, rabbitmq)

Requires https://github.com/crowbar/crowbar-ha/pull/327

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/ha.rb
+++ b/chef/cookbooks/rabbitmq/recipes/ha.rb
@@ -33,7 +33,10 @@ end
 fs_params = {}
 fs_params["directory"] = "/var/lib/rabbitmq"
 if node[:rabbitmq][:ha][:storage][:mode] == "drbd"
-  include_recipe "crowbar-pacemaker::drbd"
+
+  if CrowbarPacemakerHelper.drbd_node?(node)
+    include_recipe "crowbar-pacemaker::drbd"
+  end
 
   crowbar_pacemaker_drbd drbd_resource do
     size "#{node[:rabbitmq][:ha][:storage][:drbd][:size]}G"
@@ -133,7 +136,7 @@ if node[:rabbitmq][:ha][:storage][:mode] == "drbd"
   end
   storage_transaction_objects << "pacemaker_ms[#{ms_name}]"
 
-  ms_location_name = openstack_pacemaker_controller_location_ignoring_upgrade_for ms_name
+  ms_location_name = openstack_pacemaker_drbd_controller_only_location_for ms_name
   storage_transaction_objects << "pacemaker_location[#{ms_location_name}]"
 end
 
@@ -146,7 +149,12 @@ pacemaker_primitive fs_primitive do
 end
 storage_transaction_objects << "pacemaker_primitive[#{fs_primitive}]"
 
-fs_location_name = openstack_pacemaker_controller_only_location_for fs_primitive
+fs_location_name = if node[:rabbitmq][:ha][:storage][:mode] == "drbd"
+  openstack_pacemaker_drbd_controller_only_location_for fs_primitive
+else
+  openstack_pacemaker_controller_only_location_for fs_primitive
+end
+
 storage_transaction_objects << "pacemaker_location[#{fs_location_name}]"
 
 if node[:rabbitmq][:ha][:storage][:mode] == "drbd"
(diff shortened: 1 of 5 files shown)


crowbar_framework/lib/openstack/ha.rb
```diff
@@ -54,5 +54,26 @@ def self.set_controller_role(nodes)
     def self.set_compute_role(nodes)
       set_role(nodes, "compute")
     end
+
+    def self.set_drbd_controller_role(nodes)
```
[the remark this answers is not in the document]
jsuchome: it's in line with other methods in the file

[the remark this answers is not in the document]
s-t-e-v-e-n-k: The Hound can not be argued or bargained with.

crowbar_framework/lib/openstack/ha.rb
```diff
@@ -54,5 +54,26 @@ def self.set_controller_role(nodes)
...
+      nodes.each do |nodename|
+        save_it = false
+        node = NodeObject.find_node_by_name nodename
+
+        drbd_controller = false
+        node[:pacemaker] ||= {}
+
+        if node[:pacemaker][:drbd] && node[:pacemaker][:drbd][:nodes]
```
s-t-e-v-e-n-k: This should also be node[:drbd][:nodes].

crowbar_framework/app/models/database_service.rb
```diff
@@ -180,7 +180,10 @@ def apply_role_pre_chef_call(old_role, role, all_nodes)
       database_elements, database_nodes, database_ha_enabled = role_expand_elements(role, db_role)
       db_enabled[engine]["enabled"] = true unless database_nodes.empty?
       db_enabled[engine]["ha"] = database_ha_enabled
-      Openstack::HA.set_controller_role(database_nodes) if database_ha_enabled
+      if database_ha_enabled
+        Openstack::HA.set_controller_role(database_nodes)
+        Openstack::HA.set_drbd_controller_role(database_nodes) if engine == "postgresql"
```
jsuchome: Hm, I guess we need to call this from pacemaker mode rather than database one

jsuchome: The problem is, these attributes may not be set if I only modify existing pacemaker proposal (by adding new node). Code from database model is not executed in such case and drbd-controller attributes are not correctly set...

jsuchome:
So, second commit (database: Set pacemaker attribute to nodes in DRBD cluster ) should not be needed and its functionality should be replaced by https://github.com/crowbar/crowbar-ha/pull/327/commits/1e9c713091876a487e0a9974c9c43ce06a6a1b71 in crowbar-ha

chef/cookbooks/rabbitmq/recipes/ha.rb
```diff
@@ -146,7 +149,12 @@
 end
 storage_transaction_objects << "pacemaker_primitive[#{fs_primitive}]"
 
-fs_location_name = openstack_pacemaker_controller_only_location_for fs_primitive
+fs_location_name = if node[:rabbitmq][:ha][:storage][:mode] == "drbd"
+  openstack_pacemaker_drbd_controller_only_location_for fs_primitive
+else
+  openstack_pacemaker_controller_only_location_for fs_primitive
+end
```
rhafer: I think this change is also not really needed, because of the existing colocations constraints.

chef/cookbooks/rabbitmq/recipes/ha.rb
```diff
@@ -318,7 +326,7 @@
     service_transaction_objects << "pacemaker_location[#{public_vip_location_name}]"
   end
 
-  location_name = openstack_pacemaker_controller_only_location_for service_name
+  location_name = openstack_pacemaker_drbd_controller_only_location_for service_name
```
rhafer: Same as for postgresql here.

crowbar_framework/app/models/database_service.rb
rhafer: Yes. This has to happen in the pacemaker barclamp. I guess depending on value of the `[drbd][enabled]` attribute. (It's a bit unfortunate that the `set_role` method is part of crowbar-openstack, but I am not sure if that really matters)

jsuchome: I'm deleting this commit, and crowbar-ha code has to take care

jsuchome:
@rhafer I guess you are right about those collocations; I was not sure which should I make mandatory so I assume it does not hurt to have the constrain with each location...

rhafer:
> EDIT: I added them back, seems to work better

I am still not entirely sure why they are needed. But if it keeps the failed probe operations for rabbitmq and postgresql away I am fine with that.

PR #1756, opened 2018-09-05
trove: Remove

dirkmueller:


rsalevsky:
Can you please add why we are removing it?

dirkmueller:
@rsalevsky what would be a good reason on why we're removing it?

rsalevsky:
Need a rebase but the commit message is now good enough for me. ;)

dirkmueller:
Related to #1795

PR #1758, opened 2018-09-05
keystone: Always try to rsync keys to new nodes 

skazi0:
This is forward port of #1741 with modifications to make it compatible with #1492 and #1240


jsuchome:
@rhafer @matthewoliver review please?

PR #1760, opened 2018-09-06
database: keep the sst passwort in the proposal

dirkmueller:
When you enable/disable HA on the database proposal in your
existing deployment, the sstuser_password wasn't stored
and on reenable was reset to "random_password", which broke
IST/SST of the existing cluster, so an inconsistent cluster was
not recoverable without manual intervention. Always keep
the password that was previously set, even if we switch to
non-HA.


dirkmueller:
need to figure out how to migrate the data back into the proposal

rhafer:
> need to figure out how to migrate the data back into the proposal

@dirkmueller I've tried adding a migration there. It's only briefly tested. But at least it survived an installation on an already deployed galera system and migrated the (mysql) password into the proposal.

chef/data_bags/crowbar/template-database.json
```diff
@@ -57,6 +60,9 @@
           "log_filename": "postgresql.log-%Y%m%d%H%M",
           "log_truncate_on_rotation": false,
           "log_min_duration_statement": -1
+        },
+        "password": {
+          "postgres": ""
```
jsuchome: Why do we care about saving postgres password when postgresql cannot be deployed with SOC8?

rhafer: I think the goal is to backport this down to cloud 7.

rhafer (commented):
I am pretty sure my review isn't needed here anymore :-)

PR #1762, opened 2018-09-07
Neutron: disable metadata agent service if not necessary

ilausuch:
In order to free some connections to rabbitmq the neutron
metadata agent is disabled in computer nodes if the
force_metadata flag is set to true. The pacemaker
primitive is removed as well if exists.

### DIFF
--- a/chef/cookbooks/neutron/definitions/neutron_metadata.rb
+++ b/chef/cookbooks/neutron/definitions/neutron_metadata.rb
@@ -87,13 +87,19 @@ define :neutron_metadata,
     use_crowbar_pacemaker_service = \
       (neutron_network_ha && node[:pacemaker][:clone_stateless_services]) || nova_compute_ha_enabled
 
+    enable_metadata = node.roles.include?("neutron-network") || !node[:neutron][:metadata][:force]
+
     # In case of Cisco ACI driver, supervisord takes care of starting up
     # the metadata agent.
     service node[:neutron][:platform][:metadata_agent_name] do
-      action [:enable, :start]
-      subscribes :restart, resources(template: node[:neutron][:config_file])
-      subscribes :restart, resources(template: node[:neutron][:metadata_agent_config_file])
-      subscribes :restart, resources(file: "/etc/neutron/metadata_agent.ini")
+      if enable_metadata
+        action [:enable, :start]
+        subscribes :restart, resources(template: node[:neutron][:config_file])
+        subscribes :restart, resources(template: node[:neutron][:metadata_agent_config_file])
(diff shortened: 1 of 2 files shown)


ilausuch:
[Disable Service on compute nodes](https://trello.com/c/pIBnFo2W/296-disable-service-on-compute-nodes)

dirkmueller:
I'm confused about this.. isn't the requirement to always disable it on *only compute nodes* but keep it running on the network nodes?

ilausuch:
Las changes is to disable agent the metadata agent **only** in the computer nodes.

chef/cookbooks/neutron/definitions/neutron_metadata.rb
```diff
@@ -87,10 +87,18 @@
     use_crowbar_pacemaker_service = \
       (neutron_network_ha && node[:pacemaker][:clone_stateless_services]) || nova_compute_ha_enabled
 
+    is_compute = node.roles.any? { |role| /^nova-compute-/ =~ role }
+    enable_metadata = (neutron[:neutron][:use_dvr] &&
+      (!node[:neutron][:metadata][:force] || !is_compute)) || node.roles.include?("neutron-network")
```
AbelNavarro: shouldn't it be:

```
(neutron[:neutron][:use_dvr] &&
      (node[:neutron][:metadata][:force] || is_compute)) || node.roles.include?("neutron-network")
```

(removed negations in the OR)

ilausuch: @AbelNavarro I think it's correct.

We want to enable metadata agent in case some of these conditions is true:
- This node is included in network role
- Uses DVR=true and force_metadata=false
- Uses DVR=true and force_metadata=true and is a compute

May be it can be simplified considering that if DVR=true and force_metadata=true only have to enable metadata in network nodes 

(neutron[:neutron][:use_dvr] && !node[:neutron][:metadata][:force]) || node.roles.include?("neutron-network")

In this case will be apply to network-nodes or if DVR=true and force_metadata=false
I am thinking that this last option is good.

What do you think?

ilausuch: @AbelNavarro The condition you mentioned, the problem is that could be DVR=true, force_metada=true and then the metadata agent over computer nodes would be enabled. This is what we want to avoid

ilausuch: Also we can be simplify removing the dvr check because is checked already here https://github.com/crowbar/crowbar-openstack/blob/d6fefd9b1488578c8b5ff9b9592afd2d948dc846/chef/cookbooks/neutron/recipes/common_agent.rb#L362

PR #1763, opened 2018-09-07
database: Prepare mysql database for data migration

skazi0:
Before data could be migrated from PostgreSQL databases and
users for all services need to be created in target MariaDB.
Also table structures need to be deployed (using OpenStack
manage commands).

New script (prepare-mariadb) was added to perform the whole
process. Role specific data was extracted from existing
pg2mariadb_preparation recipe so that it can be used in both
recipe and the prepare script.

The script looks up nodes which host services using the
database and prepares each of them.

First step is execution of pg2mariadb_preparation chef recipe
which creates all needed databases and users. It also writes
set of config override files and per-role scripts which are
later used to execute dbsync commands.
The recipe is executed by temporary modification of node-role
run_list and running chef-client on the target node.

Second step is execution of all dbsync scripts on the target
node.

The script should be executed on the admin server.
Stdout and stderr of all steps are written to log files
under /var/log/crowbar/db-prepare*.log

See https://jira.suse.de/browse/SCRD-3886


jsuchome:
Could we have `bin/prepare-mariadb` with executable flag?

bin/prepare-mariadb
```diff
@@ -0,0 +1,148 @@
...
+end
+
+def prepare_node(node, roles)
+    log "Preparing node #{node.name}"
+    log "Adding #{RECIPE} to run_list"
+    add_recipe node
+    log "Running chef-client on #{node.name}..."
+    res = run_ssh_cmd(node, "chef-client", "30m")
+    log "Run time: #{res[:run_time]}s"
+    write_log("chef-client", res[:stdout_and_stderr])
```
jsuchome: Please print the information about the log file right after the action has been executed:

```
Running chef-client on d52-54-77-77-01-01.vo6.cloud.suse.de...
Run time: 249.667735281s
Stdout/Stderr written to: /var/log/crowbar/db-prepare.chef-client.log```

After that "Running chef-client" message user has no idea if anything is being done.

File: chef/cookbooks/database/libraries/crowbar.rb [hunk 2]
```diff
@@ -24,4 +24,82 @@ def self.get_listen_address(node, sql_engine = node[:database][:sql_engine])
...
+    all_meta[role]
+  end
+
+  def self.override_conf_path(role)
+    "/etc/pg2mysql/#{role}.mariadb-conf.d/"
+  end
+
+  private
+
+  def self.all_meta
```
jsuchome: Could we have some more self-explanatory method name here?

bin/prepare-mariadb
```diff
@@ -0,0 +1,148 @@
...
+    unless res[:exit_code].zero?
+      log "Chef-client failed with code: #{res[:exit_code]}"
+      return
+    end
+    log "Processing db_sync commands on #{node.name}"
+    roles.each do |role|
+      db_override_conf = CrowbarDatabaseHelper.override_conf_path(role)
+      role_meta = CrowbarDatabaseHelper.role_meta(role)
+      cmds = role_meta['db_sync_cmd']
+      cmds = [cmds] unless cmds.kind_of?(Array)
```
jsuchome: What about defining 'db_sync_cmd' as Array direcly? (in `all_meta`)

chef/cookbooks/database/libraries/crowbar.rb
```diff
@@ -24,4 +24,82 @@ def self.get_listen_address(node, sql_engine = node[:database][:sql_engine])
...
+        "db_sync_cmd" => "manila-manage --config-dir /etc/manila/manila.conf.d/ --config-dir <%=db_override_conf%> db sync"
+      },
+      "neutron-server" => {
+        "barclamp" => "neutron",
+        "db_sync_cmd" => "neutron-db-manage --config-dir /etc/neutron/neutron.conf.d/ --config-dir <%=db_override_conf%> upgrade head"
+      },
+      "nova-controller" => {
+        "barclamp" => "nova",
+        "db_sync_cmd" => [
+          "nova-manage --config-dir /etc/nova/nova.conf.d/ --config-dir <%=db_override_conf%> db sync",
```
jsuchome: Please change the order - api_db sync must be run first

skazi0: done

chef/cookbooks/database/libraries/crowbar.rb
skazi0: Generic data calls for generic name :o)
Any better idea? Note that the method is private.

bin/prepare-mariadb
jsuchome: Well, OK then

chef/cookbooks/database/recipes/pg2mariadb_preparation.rb
```diff
@@ -0,0 +1,98 @@
...
+  Chef::Log.info("creating database #{db["database"]}")
+  Chef::Log.info("creating database user #{db["user"]} with password #{db["password"]}")
+  Chef::Log.info("db settings: #{db_settings.inspect}")
+
+  database "create #{db[:database]} database (pg2my)" do
+    connection db_settings[:connection]
+    database_name db[:database]
+    provider db_settings[:provider]
+    action :create
+#    only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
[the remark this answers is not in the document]
jsuchome: These commented lines (also below) are not needed, right?

chef/cookbooks/database/libraries/crowbar.rb
jsuchome: `openstack_database_data` ?

skazi0: This is a member in `CrowbarDatabaseHelper` which by the way is a misnomer as it refers to OpenStack not Crowbar DB. So if we strip the redundancy only "data" will be left :o)

skazi0: renamed to `migration_data`.

bin/prepare-mariadb
```diff
@@ -0,0 +1,148 @@
...
+      idx=nil
+      cmds.each do |cmd|
+        cmd = ERB.new(cmd).result(binding)
+        log "Running #{cmd} for role #{role}"
+        res = run_ssh_cmd(node, cmd, "5m")
+        log "Run time: #{res[:run_time]}s"
+        write_log("role-#{role}", res[:stdout_and_stderr], idx)
```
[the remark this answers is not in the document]
skazi0: You mean something like: "mysql detected at: ...."

chef/cookbooks/database/recipes/pg2mariadb_preparation.rb
```diff
@@ -0,0 +1,119 @@
...
+    provider db_settings[:user_provider]
+    require_ssl db_settings[:connection][:ssl][:enabled]
+    action :grant
+  end
+end
+
+commands.each do |command|
+  Chef::Log.info("running dbsync command: #{command[:cmd]} for role #{command[:role]}")
+
+  execute "dbsync-role-#{command[:role]}"
```
jsuchome: `do` is missing at the end

bin/prepare-mariadb
```diff
@@ -0,0 +1,144 @@
...
+  outfile = File.join(LOGDIR, "db-prepare.#{part}.log")
+  open(outfile, "w") { |f| f << stdout_and_stderr }
+  log "Stdout/Stderr written to: #{outfile}"
+end
+
+def prepare_node(node, roles)
+    log "Preparing node #{node.name}"
+    log "Adding #{RECIPE} to run_list"
+    add_recipe node
+    log "Running chef-client on #{node.name}..."
```
jsuchome: This first run can take a lot of time (`Run time: 594.473499744s`) and there's no progress anywhere.

Could the ssh command tee the output to some file at the remote node? And here we would print something like "Progress is being written to X at Y."

bin/prepare-mariadb
```diff
@@ -0,0 +1,144 @@
...
+    log "Processing db_sync commands on #{node.name}"
+    roles.each do |role|
+      cmd = "/etc/pg2mysql/scripts/#{role}-db_sync.sh"
+      log "Running db_sync script #{cmd} for role #{role}"
+      res = run_ssh_cmd(node, cmd, "5m")
+      log "ERROR: Failed with code: #{res[:exit_code]}" unless res[:exit_code].zero?
+      write_log("role-#{role}", res[:stdout_and_stderr])
+      log "Run time: #{res[:run_time]}s"
+    end
+    log "Prepare completed for #{node.name}"
```
jsuchome: Please make the log output more user friendly. Something like "MariaDB tables were successfully prepared. You can now continue with next step of the migration."
And btw, the node is not important information here, right?

jsuchome: Also, could you please list to the log:

- databases that need to be migrated
- and/or connection strings that are used for these databases?

Basically the info written in `999-db.conf` files

chef/cookbooks/database/libraries/crowbar.rb
```diff
@@ -24,4 +24,96 @@ def self.get_listen_address(node, sql_engine = node[:database][:sql_engine])
...
+        # this doesn't work because of a bug in barbican-manage handling of oslo_config
+        # "db_sync_cmd" => "barbican-manage --config-dir /etc/barbican/barbican.conf.d/ " \
+        #   "--config-dir <%=@db_override_conf%> db upgrade"
+        "db_sync_cmd" => "barbican-manage db upgrade --db-url <%=@db_conf_sections['DEFAULT']%>"
+      },
+      "trove-server" => {
+        "barclamp" => "trove",
+        "db_sync_cmd" => "trove-manage --config-dir /etc/trove/trove.conf.d/ " \
+          "--config-dir <%=@db_override_conf%> db_sync"
+      }
```
jsuchome: Could you add manila (`manila-server` role), magnum (magnum-server) and sahara (sahara-server)?

Seems that commands are 

- `sahara-db-manage upgrade head`
- `manila-manage db sync`
- `magnum-db-manage upgrade`

bin/prepare-mariadb
skazi0: That's a great idea! Better than writing logs only after the work is done. Should we keep both or remove the log writing from admin server>

bin/prepare-mariadb
skazi0: There can be more nodes if you have multiple clusters. We can add some "ending words" at the end of the script.

chef/cookbooks/database/libraries/crowbar.rb
skazi0: manila was there already. magnum and sahara added.

bin/prepare-mariadb
skazi0: done

jsuchome:
please rebase after merging https://github.com/crowbar/crowbar-openstack/pull/1771

skazi0:
rebased

chef/cookbooks/database/templates/default/mariadb-databases.txt.erb
```diff
@@ -0,0 +1,3 @@
+<% @db_connections.each do |conn| -%>
+<%= conn["role"] -%>|<%= conn["section"] -%>|<%= conn["url"] %>
```
jsuchome: Hm, instead of `section`, could we have the name of the database?

`nova-controller|database|mysql+pymysql://nova:O62F1wNospCz@192.168.243.87/nova?charset=utf8`

Also, maybe the role is not important as well, so maybe rather something like
```
nova|mysql+pymysql://nova:O62F1wNospCz@192.168.243.87/nova?charset=utf8
nova_api|mysql+pymysql://nova_api:0S3BvWQEmtcw@192.168.243.87/nova_api?charset=utf8
```

I'm not very sure about that separator ... do you have reason for `|`?

chef/cookbooks/database/recipes/pg2mariadb_preparation.rb
```diff
@@ -0,0 +1,145 @@
...
+  end
+end
+
+template "/etc/pg2mysql/databases.txt" do
+  source "mariadb-databases.txt.erb"
+  mode 0640
+  owner "root"
+  group "root"
+  variables(
+    db_connections: db_connections
```
jsuchome: Could we pass also the original (postgresql) DB connections in there? (And adapt the output accordingly?)

bin/prepare-mariadb
```diff
@@ -0,0 +1,139 @@
...
+    log "Running chef-client on #{node.name}..."
+    res = run_ssh_cmd(node, "chef-client", "chef-client", "30m")
+    log "Run time: #{res[:run_time]}s"
+    log "Removing #{RECIPE} from run_list"
+    remove_recipe node
+    unless res[:exit_code].zero?
+      log "ERROR: Chef-client failed with code: #{res[:exit_code]}"
+      return
+    end
+    log "Summary of used databases: /etc/pg2mysql/databases.txt on #{node.name}"
```
jsuchome: Could you `scp` this file to the place where the script is called?

chef/cookbooks/database/templates/default/mariadb-databases.txt.erb
skazi0: Regarding separator I was trying to avoid using something which can exist in the URL (especially the password field).

bin/prepare-mariadb
skazi0: There will be multiple databases.txt if multiple nodes are involved. What then?

chef/cookbooks/database/templates/default/mariadb-databases.txt.erb
skazi0: Switched to <tab>... should also be safe.

chef/cookbooks/database/recipes/pg2mariadb_preparation.rb
skazi0: done

bin/prepare-mariadb
jsuchome: Well, add the node name to the file name, or merge all databases.txt files into one. 
I think after all we do not care about node data in this, in the end there's just one database per component.

skazi0: I've just added collecting of the indexes and saving merged file on the admin server (or wherever the script is called from).

bin/prepare-mariadb
```diff
@@ -0,0 +1,145 @@
...
+    summary = run_ssh_cmd(node, "cat /etc/pg2mysql/databases.txt")[:stdout_and_stderr].lines
+    log "Prepare completed for #{node.name}"
+    summary
+end
+
+def main
+  chef_init
+  if mysql_node.nil?
+    raise "MySQL server not found. Please assign mysql-server role "
+      "to some node or cluster and re-apply database proposal."
```
jsuchome: Could you just use log + exit instead of raise? That makes it similar to other kinds of output, and looks less ugly.

chef/cookbooks/database/templates/default/mariadb-databases.txt.erb
```diff
@@ -0,0 +1,3 @@
+<% @databases.each do |dbdata| -%>
+<%= dbdata[:db][:database] -%>	<%= dbdata[:psql_url] -%>	<%= dbdata[:url] %>
+<% end %>
```
rhafer: How about using some standard file format for this? (yaml, json, ini?) We could just use that as an input file for the migration tool.

jsuchome: I vote for yaml. This seems reasonably useful for the parser:
```
keystone:
  source: postgresql://keystone:bA3o7ydd5CWa@192.168.243.86/keystone
  target:
cinder:
  source: postgresql://cinder:idRll2gJPodv@192.168.243.86/cinder
  target:
```

bin/prepare-mariadb
```diff
@@ -0,0 +1,146 @@
...
+  nodes_and_roles
+end
+
+def node_role(node)
+    Chef::Role.load("crowbar-"+node.name.gsub(".", "_"))
+end
+
+def add_recipe(node)
+  role = node_role(node)
+  role.run_list << RECIPE
```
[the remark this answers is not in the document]
skazi0: Chef already does that.

bin/prepare-mariadb
```diff
@@ -0,0 +1,146 @@
...
+  Chef::Config.client_key "/opt/dell/crowbar_framework/config/client.pem"
+  Chef::Config.chef_server_url "http://localhost:4000"
+end
+
+def node_for_role(role)
+  nodes = []
+  Chef::Search::Query.new.search "node", "roles:#{role}" do |n|
+    nodes << n
+  end
+  nodes.sort_by! { |n| n.name }
```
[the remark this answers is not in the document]
rhafer: Ah, ok. Makes sense.

bin/prepare-mariadb
```diff
@@ -0,0 +1,146 @@
...
+def main
+  chef_init
+  if mysql_node.nil?
+    log "ERROR: MySQL server not found. Please assign mysql-server role "
+      "to some node or cluster and re-apply database proposal."
+    return
+  end
+  summary = Set.new
+  selected_nodes.values.each do |node_data|
+    summary |= prepare_node(node_data[:node], node_data[:roles])
```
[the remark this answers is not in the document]
skazi0: Added together with exit codes. For now the script continues if some dbsync fails. Do you think it should stop?

chef/cookbooks/database/templates/default/mariadb-databases.txt.erb
skazi0: I'll repeat what I said offline: I wanted to have this format easy for shell scripts (grep, awk, etc). If the main usecase here is for psql2mysql then I'm fine with yaml.

skazi0:
@rhafer regarding execution of dbsync commands from the script rather than the recipe: you're right that with remote logs the benefits are smaller but I still think that it's easier to see the progress/errors of the dbsync commands directly than in the chef-client log. When logging was local the benefit was way more clear. Now I'm not going to push too hard for that. If you think that it is not worth it then I'll move it to the recipe.

bin/prepare-mariadb
```diff
@@ -0,0 +1,138 @@
...
+  chef_init
+  first_mysql_node = mysql_node
+  if first_mysql_node.nil?
+    log "ERROR: MySQL server not found. Please assign mysql-server role "
+      "to some node or cluster and re-apply database proposal."
+    return -1
+  end
+  selected_nodes(first_mysql_node).values.each do |node_data|
+    node_ret = prepare_node(node_data[:node], node_data[:roles])
+    ret = node_ret if ret.zero? && !node_ret.zero?:
```
jsuchome: extra `:` at the end of line

bin/prepare-mariadb
```diff
@@ -0,0 +1,138 @@
...
+
+def mysql_node
+    node_for_role "mysql-server"
+end
+
+# Select nodes which represent all services which use database
+# they might be standalone nodes or cluster members. In most simple
+# case there will be only one node used for all services.
+def selected_nodes(first_mysql_node)
+  nodes_and_roles = {first_mysql_node.name => "mysql-server"}
```
jsuchome: Shouldn't this be `nodes_and_roles = {
  node: first_mysql_node,
  roles: ["mysql-server"]
}` ?

bin/prepare-mariadb
skazi0: old python leftovers :o)

bin/prepare-mariadb
```diff
@@ -0,0 +1,143 @@
...
+  first_mysql_node = mysql_node
+  if first_mysql_node.nil?
+    log "ERROR: MySQL server not found. Please assign mysql-server role " \
+      "to some node or cluster and re-apply database proposal."
+    return -1
+  end
+  selected_nodes(first_mysql_node).values.each do |node_data|
+    node_ret = prepare_node(node_data[:node], node_data[:roles])
+    ret = node_ret if ret.zero? && !node_ret.zero?
+    log "Summary of used databases: /etc/pg2mysql/databases.txt on " \
```
jsuchome: Please fix the file name in the log message

PR #1770, opened 2018-09-12
[WIP] neutron: [cisco-aci]: Enabling unified mode for Cisco ACI 

mmnelemane:
Cisco ACI integration with OpenStack has enabled support for deployment of a single ML2 mode for both type of driving mechanisms: ML2 and GBP. The earlier method of specifying one of the drivers in a deployment is not valid anymore and the individual ML2 driver is not supported after OpenStack Ocata. The driver is now replaced by ACI Integration Module which allows use of both ML2 and GBP modes of operation. The below commit installs and configure the AIM driver and the config files and does the corresponding changes in the neutron cookbook and Crowbar UI options. The changes are only tested so far until the changes are applied on the respective nodes and the configurations are done. The VMs are launched either as a normal Nova instance or as a Group Member using GBP. However, the testing can only be concluded if there is reliable L2 and L3 connectivity between VMs and this has not yet been established. The unified mode is relevant for Cloud 7, Cloud 8 and Cloud 9 and needs to be tested on all three platforms. Hence this is not yet ready for review or merge until there is a confirmed test on connectivity.


dirkmueller:
@mmnelemane rebase?

chef/cookbooks/neutron/recipes/cisco_apic_support.rb
```diff
@@ -49,11 +55,39 @@
...
+    policy_drivers: "aim_mapping",
+    extension_drivers: "aim_extension,proxy_group",
+    default_ip_pool: "192.168.0.0/16",
+    optimized_dhcp: node[:neutron][:apic][:optimized_dhcp],
+    optimized_metadata: node[:neutron][:apic][:optimized_metadata]
+  )
+  notifies :restart, "service[#{node[:neutron][:platform][:service_name]}]"
+end
+
+template "/etc/aim/aim.conf" do
```
dirkmueller: whcih package creates that directory?

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -86,20 +83,24 @@
...
-  elsif neutron[:neutron][:ml2_mechanism_drivers].include?("apic_gbp")
-    service_plugins = ["group_policy", "servicechain", "apic_gbp_l3"]
+  if neutron[:neutron][:ml2_mechanism_drivers].include?("apic_aim")
+    service_plugins.unshift("apic_aim_l3", "group_policy", "ncp")
+  else
+    service_plugins.unshift("neutron.services.l3_router.l3_router_plugin.L3RouterPlugin")
   end
 end
+
+core_plugins = use_apic ? "ml2plus" : neutron[:neutron][:networking_plugin]
```
dirkmueller: iirc the ml2 plugin is a config option right?

chef/cookbooks/neutron/recipes/cisco_apic_support.rb
mmnelemane: python-aci-integration-module does that.

chef/cookbooks/neutron/recipes/common_config.rb
mmnelemane: "apic_aim" is the ml2 mechanism driver which is a config option and this expects the "core_plugins" to be set to "ml2plus". Hence this line.

chef/data_bags/crowbar/migrate/neutron/304_apic_phys_vm_domains.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(template_attributes, template_deployment, attributes, deployment)
+  attributes["apic"]["phys_domain"] = template_attributes["apic"]["phys_domain"]
+  attributes["apic"]["vm_domains"] = template_attributes["apic"]["vm_domains"]
```
rhafer: As you plan to backport this to 7 and 8 you need to make the migrations save for backport, by only setting attributes to the template value if the attribute doesn't have a value assigned yet. Usually this would look like:
`attributes["apic"]["phys_domain"] = template_attributes["apic"]["phys_domain"] unless attributes["apic"]["phys_domain"]`

The same is true for the 305 migration

chef/data_bags/crowbar/migrate/neutron/304_apic_phys_vm_domains.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(template_attributes, template_deployment, attributes, deployment)
+  attributes["apic"]["phys_domain"] = template_attributes["apic"]["phys_domain"]
```
rhafer: As you plan to backport this to 7 and 8 you need to make the migrations save for backport, by only setting attributes to the template value if the attribute doesn't have a value assigned yet. Usually this would look like:
`attributes["apic"]["phys_domain"] = template_attributes["apic"]["phys_domain"] unless attributes["apic"]["phys_domain"]`

The same is true for the 305 migration

chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -125,12 +132,18 @@
                    "ruby2.1-rubygem-faraday"],
     cisco_pkgs: ["python-networking-cisco"],
     cisco_apic_pkgs: ["python-apicapi",
-                      "python-neutron-ml2-driver-apic"],
-    cisco_apic_gbp_pkgs: ["openstack-neutron-gbp",
-                          "python-gbpclient"],
+                      "python-acitoolkit",
+                      "python-aci-integration-module",
+                      "openstack-neutron-gbp",
```
dirkmueller: so why do we need the *-gbp packages if you remove support for cisco_gbp? what else depends on it?

PR #1771, opened 2018-09-13
helpers: Pass sql_engine into get_listen_address

rhafer:
When fetch_database_settings is called by one of the database specific
cookbooks it is supposed to lookup the settings for the respective
backend instead of for the currently selected default backend.


jsuchome:
Gating fails because of broken mkcloud deployment, I'm trying to fix it with https://github.com/SUSE-Cloud/automation/pull/2737

PR #1772, opened 2018-09-18
[4.0] keystone: Reorganize HA/fernet code

skazi0:
Moved most of the fernet setup to ha.rb. Some redundancy was added but
it is now easier to work with the (more-complex) HA case.

The main goal was to reorder step so that fernet keys are configured and
distributed to cluster nodes before apache is (re)started. This way, there
is no time when apache would be configured to use fernet keys but no actual
keys would be present. The problem was detected in cluster extension scenario
where new node was added to an already existing cluster.

### DIFF
--- a/chef/cookbooks/keystone/providers/fernet.rb
+++ b/chef/cookbooks/keystone/providers/fernet.rb
@@ -1,0 +1,21 @@
+action :setup do
+  execute "keystone-manage fernet_setup" do
+    command "keystone-manage fernet_setup \
+      --keystone-user #{node[:keystone][:user]} \
+      --keystone-group #{node[:keystone][:group]}"
+    action :run
+  end
+end
+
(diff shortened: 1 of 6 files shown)


chef/cookbooks/keystone/providers/fernet.rb
```diff
@@ -0,0 +1,22 @@
+property :rsync_command, String, default: ""
```
jsuchome: Doesn't this go into a resource?

skazi0: Yeah, that's a benefit of using software which is so old that the docs are already removed from the official site :o)

jsuchome:
https://github.com/crowbar/crowbar-openstack/pull/1787 was merged, you can rebase

skazi0:
rebased and added some more details in commit messages

PR #1774, opened 2018-09-20
[5.0] mysql: ha galera needs op monitor for slaves

AbelNavarro:
Backport from: https://github.com/crowbar/crowbar-openstack/pull/1731

In a Galera HA environment the pacemaker primitive op monitor is only
set for master mode. This is not a problem because all galera resources
are in master mode.

The problem comes when changing the pacemaker primitive and the masters
can become slaves, i.e. when shrinking the cluster. In such case the
slaves never get promoted to masters because they are not monitored.
Adding the monitor operation for slaves solves this situation.

(cherry picked from commit 67fa21f8b086c862163ac044a788226ce159edcd)


AbelNavarro:

PR #1783, opened 2018-09-24
Make notifications optional

dirkmueller:


dirkmueller:
@nicolasbock there are multiple possible consumer implementations (ceilometer being an example, but also monasca or some entire 3rd party solution). what I can imagine is that we do a dependency in the ceilometer barclamp that this setting must be enabled.

AbelNavarro:
If Adam were looking at this PR he would ask for a PR description. Don't make me call Adam.

dirkmueller:
@Itxaka all the good information is in the git commit message, not in the PR summary, which is what @aspiers usually asks for ;)

aspiers:
Actually @Itxaka is right; I normally ask for both - if there are multiple commits then a TL;DR summary in the PR is how to be nice and friendly to reviewers.  Hopefully this is obvious and non-contentious.  Thanks for the reminder - I've added that as a TODO in https://github.com/SUSE/cloud-specs/pull/177

PR #1789, opened 2018-09-25
[4.0] make notifications optional

dirkmueller:


vuntz:
Maybe a stupid question, but: do we only have cinder, glance, magnum, neutron and nova that can send notifications? I thought keystone had notifications now, and I could imagine heat has this too.

Itxaka:
> Maybe a stupid question, but: do we only have cinder, glance, magnum, neutron and nova that can send notifications? I thought keystone had notifications now, and I could imagine heat has this too.

There is no stupid questions, only stupid people that dont ask questions :P

I believe keystone *does not* use rabbit at all, even if its configured.

@dirkmueller Im assuming that the default values for `oslo_messaging_notifications.driver` are set to `noop` then? then you have to actually enabled them to receive notifications. Just to make sure that we are not getting the default value and its different than that.

dirkmueller:
@Itxaka correct, in my observation the default is noop and its not setting anything. in my testing even with a full tempest run (all services but  ceilometer deployed) no notifications were triggered by any service with this change in place

PR #1781, opened 2018-09-24
Remove .mention-bot config

rsalevsky:
We don't use this bot since a long time so let's remove the config.

PR #1790, opened 2018-09-25
kick off an experiment with CODEOWNERS

aspiers:
This is an experiment which could supersede the mentionbot which was just removed in #1781.  It will cause the mentioned users to be automatically notified when pull requests are proposed which modify files matching the specified patterns, as described in

- https://help.github.com/articles/about-codeowners/

In our context, it does not mean anything about real ownership (yet), only notification.

Users and file patterns have been picked without any serious thought, and can of course easily be improved later.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,23 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners would be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @global-owner1 and @global-owner2 will be requested for
+# review when someone opens a pull request.
+#*       @global-owner1 @global-owner2
+
+# Order is important; the last matching pattern takes the most
+# precedence. When someone opens a pull request that only
+# modifies JS files, only @js-owner and not the global
+# owner(s) will be requested for a review.
+#*.js    @js-owner
+
+**/*cinder*  @toabctl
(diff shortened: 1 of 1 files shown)


rsalevsky:
I'm happy to retry this approach but my feeling form the old trial was that as soon as people moved to different tasks which are not focused around crowbar they usually stoppend reviewing changes.

aspiers:
[**@rsalevsky**](https://github.com/rsalevsky) commented on [25 Sep 2018, 15:43 BST](https://github.com/crowbar/crowbar-openstack/pull/1790#issuecomment-424372221 "2018-09-25T14:43:13Z - Replied by Github Reply Comments"):
> I'm happy to retry this approach

We never tried it before AFAIK.  We only tried mentionbot, but that broke and I didn't get round to fixing it.

>but my feeling form the old trial was that as soon as people moved to different tasks which are not focused around crowbar they usually stoppend reviewing changes.

Then they can update the file when they shift focus.  This is no different to the new processes we are using for establishing squads and then assigning people to them.  We are a much bigger team now, therefore we need more structure in the way we decide who works on what.  The problems are already [clearly documented](https://trello.com/c/UslitgVD/1191-decide-maintainership-specialisation-strategy-codereview) so please read them and provide feedback on the proposed solutions.

aspiers:
Ignoring CI failure since it's obviously unrelated.

PR #1507, opened 2018-01-05
[POC] nova: move flavor creation to converge phase

Itxaka:
just a test

PR #1791, opened 2018-09-25
nova: Check flavor presence right before the attempt for creation

jsuchome:
Sometimes it seems nova-api is starting too long and generating list
of flavors in advance returns an empty list. The attempt for creating
such flavor that is run little bit later fails because the flavor
already exists.


jsuchome:
I did not. I can check but honestly, I do not believe it

jsuchome:
backporting https://github.com/crowbar/crowbar-openstack/pull/1507 seems like better idea

PR #1795, opened 2018-09-27
trove: Remove support

rsalevsky:
This commit removes trove support as it was deprecated in Cloud 8.


dirkmueller:
as discussed: we'd have to do that on stable/5.0-pike so that it is being removed prior upgrade, as the code is removed for cloud9.

PR #1800, opened 2018-09-28
nova: Increase retries for listing flavors

rhafer:
The command getting the list of flavors is executed pretty close to
starting the nova ha resource, which can take bit more time (especially
on slow clusters). So give the flavor list command a few more tries
before erroring out.

Note: this is only an issue in setups where the stateless resources are still
managed by pacemaker (clone-stateless-service=true). In the alternative
setup (managed by systemd) the service are started a bit earlier during
the chef-client run.


rhafer:
This should fix: https://ci.suse.de/job/openstack-mkcloud/135619/parsed_console/ (cloud 7)

PR #1805, opened 2018-09-28
use Keystone v3 params for neutron_l3_ha service

guangyee:
We no longer support Keystone v2.0.


chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -57,20 +57,23 @@
...
+                                                       keystone_settings["internal_url_host"],
+                                                       keystone_settings["service_port"],
+                                                       keystone_settings["api_version"])
 
     # Add configuration file
     insecure_flag = keystone_settings["insecure"] || node[:neutron][:ssl][:insecure]
     default_settings = node[:neutron][:ha][:neutron_l3_ha_service].to_hash
+    # FIXME: need to convey the Keystone version specific parameters here as well.
+    # i.e. OS_USER_DOMAIN_NAME, OS_PROJECT_DOMAIN_NAME, etc if we ever elect to
+    # create them in a non-default domain.
```
cmurphy: These parameters should already be available in `keystone_settings` and it's technically necessary to provide them when using v3 even if it's just 'default'.

PR #1806, opened 2018-09-29
rabbitmq: default to openstack/openstack for new installs

dirkmueller:
The "nova" vhost is from the dark ages before openstack was
more than just nova. Lets use a more sensible default for new
installs. Intentionall no migration for existing installations.


aojea:
What are the implications of this change for updates/upgrades?

Itxaka:
> What are the implications of this change for updates/upgrades?

Should only affect new installs as this sets the default value when creating the barclamp only (or deleting/creating)

PR #1808, opened 2018-10-01
rabbitmq: allow disabling queue mirroring

Itxaka:
Allows to disable queue mirroring on rabbitmq as it should not be
linked to having a rabbitmq cluster

NOTICE: before someone complains, this does *not* require a migration file because we are building cloud9 and will backport this to cloud8/7 so either this gets in on a new install of cloud9 or anyone upgrading from cloud8 will already have it migrated due to the backport, so there is no need for a migration file here

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/rabbit.rb
+++ b/chef/cookbooks/rabbitmq/recipes/rabbit.rb
@@ -170,7 +170,8 @@ if cluster_enabled
     quorum = 1
   end
 
-  queue_regex = "^(?!amq.).*"
+  # don't mirror queues that are 'amqp.*' or '*_fanout_*' or `reply_*` in their names
+  queue_regex = "^(?!(amqp.)|(.*_fanout_)|(reply_)).*"
   # policy doesnt need spaces between elements as they will be removed when listing them
   # making it more difficult to check for them
   policy = "{\"ha-mode\":\"exactly\",\"ha-params\":#{quorum}}"
--- a/chef/cookbooks/rabbitmq/recipes/rabbit.rb
+++ b/chef/cookbooks/rabbitmq/recipes/rabbit.rb
@@ -174,7 +174,7 @@ if cluster_enabled
   queue_regex = "^(?!(amqp.)|(.*_fanout_)|(reply_)).*"
   # policy doesnt need spaces between elements as they will be removed when listing them
   # making it more difficult to check for them
-  policy = "{\"ha-mode\":\"exactly\",\"ha-params\":#{quorum}}"
+  policy = "{\"ha-mode\":\"exactly\",\"ha-params\":#{quorum},\"ha-sync-mode\":\"automatic\"}"
   vhost = node[:rabbitmq][:vhost]
   # we need to scape the regex properly so we can use it on the grep command
   queue_regex_escaped = ""
--- a/chef/cookbooks/rabbitmq/recipes/rabbit.rb
+++ b/chef/cookbooks/rabbitmq/recipes/rabbit.rb
@@ -164,12 +164,25 @@ node[:rabbitmq][:users].each do |user|
 end
 
 if cluster_enabled
-  quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+  if node[:rabbitmq][:enable_queue_mirroring]
+    quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+  else
+    quorum = 1
+  end
 
-  set_policy_command = "rabbitmqctl set_policy -p #{node[:rabbitmq][:vhost]} --apply-to queues " \
-      " ha-queues '^(?!amq\.).*' '{\"ha-mode\": \"exactly\", \"ha-params\": #{quorum}}'"
-  check_policy_command = "rabbitmqctl list_policies -p #{node[:rabbitmq][:vhost]} | " \
-      " grep -q '^#{node[:rabbitmq][:vhost]}\\s*ha-queues\\s'"
+  queue_regex = "^(?!amq.).*"
+  # policy doesnt need spaces between elements as they will be removed when listing them
+  # making it more difficult to check for them
+  policy = "{\"ha-mode\":\"exactly\",\"ha-params\":#{quorum}}"
+  vhost = node[:rabbitmq][:vhost]
+  # we need to scape the regex properly so we can use it on the grep command
+  queue_regex_escaped = ""
+  queue_regex.split("").each { |c| queue_regex_escaped << "\\" + c }
+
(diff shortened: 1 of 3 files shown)


Itxaka:
[Allow disabling mirroring queues on rabbit ha cluster](https://trello.com/c/xSjwtSq3/323-allow-disabling-mirroring-queues-on-rabbit-ha-cluster)

chef/cookbooks/rabbitmq/recipes/rabbit.rb
```diff
@@ -164,17 +164,19 @@
 end
 
 if cluster_enabled
-  quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
```
dirkmueller: I think instead of setting quorum only if queue mirroring is enabled, we should set quorum to 1, so that the policy is always et (aka toggling the parameter post initial deployment also updates the policy that is deployed).

dirkmueller: (I assume that ha-mode: exactly, ha-params: 1 is effectively unmirrored, according to my reading of https://www.rabbitmq.com/ha.html

chef/data_bags/crowbar/template-rabbitmq.json
```diff
@@ -60,7 +60,8 @@
         "sndbuf": null,
         "recbuf": null
       },
-      "collect_statistics_interval": 5000
+      "collect_statistics_interval": 5000,
+      "disable_queue_mirroring": false
```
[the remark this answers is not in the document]
Itxaka: sure, will update

chef/cookbooks/rabbitmq/recipes/rabbit.rb
Itxaka: ahhh I see, so we always set the policy but set it to 1 if toggled and calculate the number of real quorum if toggled off, that way it updates without having to create/delete the policy.

Great idea, Im stealing it!

chef/cookbooks/rabbitmq/recipes/rabbit.rb
```diff
@@ -164,7 +164,11 @@
...
+  if node[:rabbitmq][:enable_queue_mirroring]
```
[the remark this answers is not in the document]
Itxaka: I rather ignore this for readability

Itxaka:
@dirkmueller updated!

Itxaka:
will delay the backports updates until this is merged for good measure

Itxaka:
@dirkmueller updated with all reqs.

The only thing I can see its that this will reapply the policy to all queues so it may take a few seconds to work as per the rabbit docs:

```
Policies can change at any time. When a policy definition is updated, its effect on matching exchanges and queues will be reapplied. Usually it happens instantaneously but for very busy queues can take a bit of time (say, a few seconds). 
```

chef/cookbooks/rabbitmq/recipes/rabbit.rb
```diff
@@ -164,12 +164,26 @@
...
+    quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+  else
+    quorum = 1
+  end
+
+  # only mirror queues that dont start with 'amqp' or contain '_fanout_' in their names
+  queue_regex = "^(?!(amqp.)|(.*_fanout_))^.*$"
```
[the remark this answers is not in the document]
Itxaka: just wanted to make sure that I pick everything from the start of the line to the end that doesnt match the other regex. Maybe its not necessary, Im not that good a regex tbf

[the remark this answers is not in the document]
Itxaka: yep, seems that either `^` nor `$` are necessary to catch everything in the line, after the negative look ahead, let me update this

chef/cookbooks/rabbitmq/recipes/rabbit.rb
```diff
@@ -164,12 +164,26 @@
...
 
 if cluster_enabled
-  quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+  if node[:rabbitmq][:enable_queue_mirroring]
+    quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+  else
+    quorum = 1
+  end
+
+  # only mirror queues that dont start with 'amqp' or contain '_fanout_' in their names
```
dirkmueller: why do we mirror fanout queues? they're non-durable by default

Itxaka: I believe this was discussed prior to the creation of the PR and we agreed to mirror those queues but indeed it makes no sense if they are not durable

Itxaka: uh, actually this means exactly that, we wont mirror the `amqp.*` queues or any queues that have `_fanout_`

Maybe the comment is not clear enough.

chef/cookbooks/rabbitmq/recipes/rabbit.rb
```diff
@@ -164,12 +164,26 @@
...
 if cluster_enabled
-  quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+  if node[:rabbitmq][:enable_queue_mirroring]
+    quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+  else
+    quorum = 1
+  end
+
+  # don't mirror queues that start with 'amqp' or contain '_fanout_' in their names
+  queue_regex = "^(?!(amqp.)|(.*_fanout_)).*"
```
dirkmueller: I think we should also include ^reply_ in there (these are non-durable direct topic queues and hence are not useful to be mirrored. 

it would be all so simple if we could "only"c onfigure topic queues but that doesn't seem easy

Itxaka: ahh good point those are the ones that get created for the replies and then ignored until they expire, will update

chef/cookbooks/rabbitmq/recipes/rabbit.rb
```diff
@@ -164,12 +164,26 @@
 end
 
 if cluster_enabled
-  quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+  if node[:rabbitmq][:enable_queue_mirroring]
```
AbelNavarro: Shouldn't we refer to the barclamp data instead of the node data?

` if rabbitmq[:rabbitmq][:enable_queue_mirroring]`

Itxaka: recipe will only be run on rabbit nodes so in this case `rabbit` == `node` :+1:

PR #1811, opened 2018-10-01
Fix neutron database connection pool settings

dirkmueller:
In the current default configuration we were always setting max_pool_size to empty, which let it default to 5, which is far too low for a scale env. 

set the variable so that the raised default from neutron barclamp template is actually in effect.

### DIFF
--- a/chef/data_bags/crowbar/template-neutron.json
+++ b/chef/data_bags/crowbar/template-neutron.json
@@ -101,9 +101,9 @@
         "user": "neutron"
       },
       "sql": {
-        "min_pool_size": 30,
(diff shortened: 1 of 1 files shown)


sayalilunkad:
@dirkmueller Is there a bug for this?

dirkmueller:
@sayalilunkad dunno, the previous change that introduced the bug also was without bugreport.. it was part of #508. I can add a bugreport if you want but it wouldn't have more description than the git commit message

sayalilunkad:
The gating failures do not seem to be related..

PR #1812, opened 2018-10-02
ec2: Remove EC2 support

rsalevsky:
EC2 support is deprecated in Cloud 8 and should be removed in Cloud 9


dirkmueller:
@rsalevsky where is this deprecation documented? I don't see it in https://www.suse.com/releasenotes/x86_64/SUSE-OPENSTACK-CLOUD/8/#Deprecated

sayalilunkad:
Is there a bug for this?

rsalevsky:
@dirkmueller Mmh... It seams we forgot to add it to the release notes but according to my list from Cloud 8 we deprecated it.

chef/data_bags/crowbar/migrate/nova/301_remove_ec2.rb
```diff
@@ -0,0 +1,26 @@
...
+  deployment["elements"]["ec2-api"] &&
+    deployment["elements"].delete("ec2-api")
+  deployment.fetch("elements_expanded", {}).key?("ec2-api") &&
+    deployment["elements_expanded"].delete("ec2-api")
+
+  nodes = NodeObject.find("run_list_map:ec2-api")
+  nodes.each do |node|
+    node.delete_from_run_list("ec2-api")
+    node.save
+  end
```
Itxaka: should we remove the `ec2-api` attributes here as to not leave them lingering around?

dirkmueller:
still want to understand why we're removing this.. :-)

toabctl:
@rsalevsky @dirkmueller is this still relevant? I also don't see a reason why ec2api should be removed. But if we want to get this in Cloud9, please rebase !:)

dirkmueller:
Still think it should remain in cloud 9.. we may be should make sure it's not deployed by default (Haven't checked)

PR #1817, opened 2018-10-03
rabbit: Change failover strategy to shuffle

dirkmueller:
The default, round-robin, leads to a predictable reconnect order
as we're listing the services always in the same order in the config
file. This causes queues to be be less balanced, e.g. the first
node is always servicing more load than the other nodes in the rabbitmq
cluster. with a shuffle strategy this should become more balanced over
time.


dirkmueller:
@Itxaka explained to me that this might not be such a useful idea.

Itxaka:
So the problem with this is the pacemaker agent and how it brings the cluster up.
Mainly it just brings up the "master" first as the "slaves" need to join the cluster, so it needs an active and running node first. The main problem is that bringing up a rabbitmq cluster takes a while and after its up there is still even more checks to see if the apps are running and the monitor actions return 8 (ok for masters)

So all the clients end up connecting to the first node up as its the only one available.

So Im not sure this would work as expected, but there is no harm in testing it out.

dirkmueller:
not worth it

PR #1819, opened 2018-10-04
[4.0] rabbitmq: active the haproxy balancing

ilausuch:

The balancer gets the default client ports specified in the proposal
and each rabbitmq node uses (5630 and 5631)

(cherry picked from commit 1cfe290e6f18cf896145bf25fb59e52eb8dba16c)
Backported from #1571

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/default.rb
+++ b/chef/cookbooks/rabbitmq/recipes/default.rb
@@ -22,6 +22,16 @@ ha_enabled = node[:rabbitmq][:ha][:enabled]
 # we only do cluster if we do HA
 cluster_enabled = node[:rabbitmq][:cluster] && ha_enabled
 quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+crm_resource_stop_cmd = "force-stop"
+crm_resource_start_cmd = "force-start"
+
+if node[:rabbitmq][:ha][:haproxy_enabled] && cluster_enabled
+  rabbitmq_port = node[:rabbitmq][:ha][:ports][:api]
+  rabbitmq_ssl_port = node[:rabbitmq][:ha][:ports][:api_ssl]
+else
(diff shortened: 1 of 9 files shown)


Itxaka:
codewise looks good, testing it....

ilausuch:
[RabbitMQ connections rebalancing](https://trello.com/c/bMDQwg0D/246-rabbitmq-connections-rebalancing)

ilausuch:
@aojea In master is tested and works, now we are testing here.

About if it is a good idea or not. I agree with you, that is to add an extra layer. I tried to reproduce the problems with closed connections by haproxy and I couldn't reproduced with the appropriate haproxy configuration (and this is the our by default configuration). Of course I would like to do more tests.
Additionally, at this moment the clients already have implemented rotatory access to each rabbit client. 

And I would like to do read more documentation and prepare some extra tests about that to compare the efficiency of balancing compared to haproxy.

aojea:
I really think we should use whatever everybody is using and testing, you can check that the community of devs and ops moved away from haproxy in front of rabbitmq in 2015
https://bugs.launchpad.net/fuel/+bug/1465263
```
Davanum Srinivas (DIMS) (dims-v) wrote on 2015-07-08: | #3
upstream oslo.messaging team and ops folks do *NOT* recommend using HAProxy. rabbit_hosts is what everyone is using and testing with. (Example - http://markmail.org/message/dzyokzmtu2khdish)
```

chef/cookbooks/rabbitmq/recipes/default.rb
```diff
@@ -157,7 +169,12 @@
 end
 
 service "rabbitmq-server" do
-  supports restart: true, start: true, stop: true, status: true
+  supports restart: true,
```
Itxaka: If we want this to be automatically restarted, this needs a different service name for ha_enabled as the pacemaker resource is called `rabbitmq` instead!

dirkmueller:
master PR was closed unmerged

PR #1821, opened 2018-10-04
nova: Skip retrieving of flavor list when nova-api is down

jsuchome:
It is possible that nova-api gets (re)started only at the very end
of chef-client run, and running nova flavor-list would not get any
results before.

### DIFF
--- a/chef/cookbooks/nova/recipes/flavors.rb
+++ b/chef/cookbooks/nova/recipes/flavors.rb
@@ -104,6 +104,16 @@ end
 
 ruby_block "Get current flavors" do
   block do
+    cmd = if node[:nova][:ha][:enabled]
+      "crm resource show nova-api 2> /dev/null | grep is\ running"
+    else
+      "systemctl status openstack-nova-api"
+    end
+    unless ::Kernel.system(cmd)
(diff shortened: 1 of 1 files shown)


chef/cookbooks/nova/recipes/flavors.rb
```diff
@@ -104,6 +104,16 @@
...
 ruby_block "Get current flavors" do
   block do
+    cmd = if node[:nova][:ha][:enabled]
+      "crm resource show nova-api 2> /dev/null | grep is\ running"
+    else
+      "systemctl status openstack-nova-api"
+    end
+    unless ::Kernel.system(cmd)
+      Chef::Log.info("nova-api does not seem to be running, skipping flavor list")
+      return
```
Itxaka: This may cause to have a nonexistant `node.run_state["flavorlist"]` so the resources creation may not work as expected or they may fail. You would probably fix that by creating an empty list for it instead

jsuchome: Yeah. I was thinking of putting some special value into `node.run_state["flavorlist"]`

But turns out, this fix is probably not needed at all

PR #1822, opened 2018-10-05
[5.0] Neutron force metadata disable metadata agent fix

ilausuch:
In commit 8e8f8e873ab9cb94ca1042529f289ca017ceff3d is introduced
an option to disable metadata agent in computer nodes.

In some cases, the condition to check if metadata is enabled
fails because these computers don't have the information of

This commit fixes this problem

(cherry picked from commit fa92cfd)
Backported from #1825

### DIFF
--- a/chef/cookbooks/neutron/definitions/neutron_metadata.rb
+++ b/chef/cookbooks/neutron/definitions/neutron_metadata.rb
@@ -87,7 +87,7 @@ define :neutron_metadata,
     use_crowbar_pacemaker_service = \
       (neutron_network_ha && node[:pacemaker][:clone_stateless_services]) || nova_compute_ha_enabled
 
-    enable_metadata = node.roles.include?("neutron-network") || !node[:neutron][:metadata][:force]
+    enable_metadata = node.roles.include?("neutron-network") || !neutron[:neutron][:metadata][:force]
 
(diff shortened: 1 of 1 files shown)


ilausuch:

chef/cookbooks/neutron/definitions/neutron_metadata.rb
```diff
@@ -87,7 +87,8 @@
     use_crowbar_pacemaker_service = \
       (neutron_network_ha && node[:pacemaker][:clone_stateless_services]) || nova_compute_ha_enabled
 
-    enable_metadata = node.roles.include?("neutron-network") || !node[:neutron][:metadata][:force]
+    enable_metadata = node.roles.include?("neutron-network") || node[:neutron].nil? ||
```
Itxaka: You can probably use `node.fetch` to make this more readable like so:

```
enable_metadata = node.roles.include?("neutron-network") || !node.fetch(:neutron, {}).fetch(:metadata, {}).fetch(:force, false)
```

Kind of like its used here: https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/provisioner/recipes/update_nodes.rb#L100

ilausuch: I agree

PR #1824, opened 2018-10-05
[4.0] Neutron force metadata disable metadata agent fix 4

ilausuch:
In commit 098b3a478bd9e78bb8c4bf112afd46f2f568846a is introduced
and option to dissable metadata agent in computer nodes.

In some cases, the condition to check if metadata has to enable
fails because these computers doesn't have the information of

This commit fix this problem

(cherry picked from commit 2b63ecd5bdb7e24cab5b8590f67668d7a35a3745)
Backported from #1822


ilausuch:

PR #1825, opened 2018-10-05
Neutron force metadata disable metadata agent fix

ilausuch:
In commit 803e973e1b2db8df2f1eafef41340a5dc58d94d9 is introduced
an option to disable metadata agent in computer nodes.

In some cases, the condition to check if metadata is enabled
fails because these computers don't have the information of

This commit fixes this problem

(cherry picked from commit 06b538ac16ec5bf0edf475e9909eeacf519d8dcf)
Backport from #1822

### DIFF
--- a/chef/cookbooks/neutron/definitions/neutron_metadata.rb
+++ b/chef/cookbooks/neutron/definitions/neutron_metadata.rb
@@ -87,7 +87,7 @@ define :neutron_metadata,
     use_crowbar_pacemaker_service = \
       (neutron_network_ha && node[:pacemaker][:clone_stateless_services]) || nova_compute_ha_enabled
 
-    enable_metadata = node.roles.include?("neutron-network") || !node[:neutron][:metadata][:force]
+    enable_metadata = node.roles.include?("neutron-network") || !neutron[:neutron][:metadata][:force]
 
     # In case of Cisco ACI driver, supervisord takes care of starting up
     # the metadata agent.


ilausuch:

chef/cookbooks/neutron/definitions/neutron_metadata.rb
```diff
@@ -87,7 +87,8 @@
     use_crowbar_pacemaker_service = \
       (neutron_network_ha && node[:pacemaker][:clone_stateless_services]) || nova_compute_ha_enabled
 
-    enable_metadata = node.roles.include?("neutron-network") || !node[:neutron][:metadata][:force]
+    enable_metadata = node.roles.include?("neutron-network") ||
+      !node.fetch(:neutron, {}).fetch(:metadata, {}).fetch(:force, false)
```
dirkmueller: well.. if the node doesn't have the data its the wrong node, how about using "neutron" instead of node here?

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -182,7 +182,15 @@
...
 
-  enable_metadata = node.roles.include?("neutron-network") || !node[:neutron][:metadata][:force]
+  if node.roles.include?("neutron-network")
+    enable_metadata = true
+  else
+    neutrons = node_search_with_cache("roles:neutron-network")
+    neutron = neutrons.first || \
+      raise("Neutron instance '#{nova[:nova][:neutron_instance]}' for nova not found")
+
+    enable_metadata = !neutron.fetch(:neutron, {}).fetch(:metadata, {}).fetch(:force, false)
```
dirkmueller: can you remove the .fetch() part now again as it would hide an issue ?

chef/cookbooks/neutron/definitions/neutron_metadata.rb
```diff
@@ -87,7 +87,15 @@
...
 
-    enable_metadata = node.roles.include?("neutron-network") || !node[:neutron][:metadata][:force]
+    if node.roles.include?("neutron-network")
+      enable_metadata = true
+    else
+      neutrons = node_search_with_cache("roles:neutron-network")
+      neutron = neutrons.first || \
+        raise("Neutron instance '#{nova[:nova][:neutron_instance]}' for nova not found")
+
+      enable_metadata = !neutron.fetch(:neutron, {}).fetch(:metadata, {}).fetch(:force, false)
```
dirkmueller: can you remove the .fetch() part now again as it would hide an issue ?

ilausuch: Off course.

ilausuch: But, @dirkmueller does it could cause a problem if this attribute is not defined yet? because is a new attribute.For instance, If  you created the proposal before that, and you run chef-client could generate problems?

chef/cookbooks/neutron/definitions/neutron_metadata.rb
```diff
@@ -87,7 +87,15 @@
     use_crowbar_pacemaker_service = \
       (neutron_network_ha && node[:pacemaker][:clone_stateless_services]) || nova_compute_ha_enabled
 
-    enable_metadata = node.roles.include?("neutron-network") || !node[:neutron][:metadata][:force]
+    if node.roles.include?("neutron-network")
+      enable_metadata = true
+    else
+      neutrons = node_search_with_cache("roles:neutron-network")
```
dirkmueller: no, the LWRP should not do a lookup on its own. it is fetching "neutron" node from line 24 in this recipe. please just use that instead of overwriting it. 

the whole block above merely needs to be changed from

enable_metadata = node.roles.include?("neutron-network") || !node[:neutron][:metadata][:force]

to

enable_metadata = node.roles.include?("neutron-network") || !neutron[:neutron][:metadata][:force]

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -182,7 +182,15 @@
     l3_agent_clone = "cl-#{l3_agent_primitive}"
   end
 
-  enable_metadata = node.roles.include?("neutron-network") || !node[:neutron][:metadata][:force]
```
dirkmueller: this change should be unnecessary, "node" is always a neutron related role here.

Itxaka: IIRC @ilausuch found some issue with this part where it was not finding the proper keys in the node. Unfortunately I dont know more about that

PR #1826, opened 2018-10-06
neutron: disable metadata proxy when metadata is forced

dirkmueller:
If 'force_metadata' is configured as 'True' then disable the metadata
proxies in all nodes doe not have dhcp namespaces. They are not needed
in l3 agents.

And then 'enable_isolated_metadata' is neglected. So there is no need to
configure it, since 'force_metadata' is a superset and configured that
alone is sufficient to get the metadata from the dhcp server.

### DIFF
--- a/chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/dhcp_agent.ini.erb
@@ -2,8 +2,13 @@
 interface_driver = <%= @interface_driver %>
 resync_interval = <%= @resync_interval %>
 dhcp_driver = <%= @dhcp_driver %>
+<% if @force_metadata -%>
+enable_isolated_metadata = False
+enable_metadata_network = False
+<% else -%>
 enable_isolated_metadata = <%= @enable_isolated_metadata %>
 enable_metadata_network = <%= @enable_metadata_network %>
+<% end -%>
 force_metadata = <%= @force_metadata %>
 dns_domain = <%= @dns_domain %>
 <% if @nameservers -%>
--- a/chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
@@ -3,6 +3,9 @@ interface_driver = <%= @interface_driver %>
 <% if @dvr_enabled -%>
 agent_mode = <%= @dvr_mode %>
 <% end -%>
+<% if node[:neutron][:metadata][:force] -%>
+enable_metadata_proxy = False
+<% end -%>
 metadata_port = <%= @metadata_port %>
 handle_internal_only_routers = <%= @handle_internal_only_routers %>
 periodic_interval = <%= @periodic_interval %>


dirkmueller:
A discussion with Swami revealed that we configure the l3 agent incorrectly and need to disable the metadata proxy.

swami-dvr:
@ilausuch I thought you were using 'forced_metadata' option in the dhcp_agent.ini.
With my testing I had just the 'force_metadata=True' enabled in dhcp_ini and 'enabled_metadata_proxy = False' in l3_agent.ini.

I would prefer the 'force_metadata' instead of 'enabled_isolated_metadata' the reason being the 'enabled_isolated_metada' if enabled will add routes to dhcp namespace only when the subnet is not connected to a router.
But in the case of the 'force_metadata' it would always use the dhcp namespace and add routes to the dhcp namespace.

I agree that we should be running the 'metadata-proxy' service on nodes that run the dhcp service.

swami-dvr:
@ilausuch Also you can get rid of 'enable_metadata_network = True'.

# Allows for serving metadata requests coming from a dedicated metadata access
# network whose CIDR is 169.254.169.254/16 (or larger prefix), and is connected
# to a Neutron router from which the VMs send metadata:1 request. In this case
# DHCP Option 121 will not be injected in VMs, as they will be able to reach
# 169.254.169.254 through a router. This option requires
# enable_isolated_metadata = True. (boolean value)
#enable_metadata_network = false

This is only required if you configure 'enable_isolated_metadata'. But if you just configure 'force_metadata=True',  'enable_metadata_network' may not be required.

ilausuch:
@swami-dvr I understand that this configuration is valid
- enable_isolated_metadata = True
- enable_metadata_network = False
- force_metadata = true

Because force_metadata=true doesn't require enable_metadata_network
isn't it?

swami-dvr:
> @swami-dvr I understand that this configuration is valid
> 
> * enable_isolated_metadata = True
> * enable_metadata_network = False
> * force_metadata = true
> 
> Because you mentioned force_metadata=true doesn't require enable_metadata_network
> isn't it?
> But can be enable_isolated_metadata and force_metadata activated together?

Hi @ilausuch 
If 'force_metadata' is configured as 'True', 'enable_isolated_metadata' is neglected. So there is no need to configure it, since 'force_metadata' is a superset and configured that alone is sufficient to get the metadata from the dhcp server. 

For the purpose of clarity I am posting here the 'comment' from the 'dhcp_agent.ini' file for the 'enable_isolated_metadata'.

"The DHCP server can assist with providing metadata support on isolated
 networks. Setting this value to True will cause the DHCP server to append
 specific host routes to the DHCP request. The metadata service will only be
 activated when the subnet does not contain any router port. The guest
 instance must be configured to request host routes via DHCP (Option 121).
 **This option doesn't have any effect when force_metadata is set to True.**
 (boolean value)
 enable_isolated_metadata = false"

Hope this helps.

ilausuch:
@swami-dvr Thanks. PR modified as spoken

swami-dvr:
Looks good to me. +1

swami-dvr:
I just noticed a 'typo' in the commit message.

'nodes doe not'  ==> nodes that does not have

swami-dvr:
A small nit. Typo in the commit message.

nodes doe not ==> nodes that does not

PR #1829, opened 2018-10-09
horizon: stop installing trove plugin

dirkmueller:
This has been deprecated in Cloud8 and removed in cloud9


dirkmueller:
@comurphy maybe, although this is still in sync with the upstream template iirc

PR #1831, opened 2018-10-10
[5.0] database: Increase galera write set limits

rhafer:
For some reason the default configuration shipped with the
mariadb-galera package configures non-default limits for wsrep_max_ws_rows
and wsrep_max_ws_size. This commit changes the limits back to the
maximum supported values (which also happen to be the defaults used by
MariaDB if no configuration is present).
This is to reduce the likelyhood of failing DB transactions (with e.g.
"wsrep_max_ws_rows exceeded") when e.g.  ceilometer-expirer needs to
DELETE a large set of database rows.

(cherry picked from commit 8a2877cd9c3084670c6dcc4230661cd31405dffb)


rhafer:
5.0 backport of https://github.com/crowbar/crowbar-openstack/pull/1830

PR #1833, opened 2018-10-10
rabbitmq: Fix address handling in rabbitmq (aka "glance timeout")

skazi0:
Storing the address in node attributes in rabbitmq cookbook and reading them in other cookbooks (via CrowbarOpenStackHelper) can cause problems when recipe tries to use address from new node before it sets it in the node.

Replacing this with direct address lookup based on Network barclamp attributes should solve this problem.

In addition, similar change was done for `addresses` and `management_address` but these were only used internally by rabbitmq cookbook so this is mostly "refactoring".


chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -244,7 +245,7 @@ def self.rabbitmq_settings(node, barclamp)
             end
             url = "#{rabbit[:rabbitmq][:user]}:"
             url << "#{rabbit[:rabbitmq][:password]}@"
-            url << "#{rabbit[:rabbitmq][:address]}:#{port}"
+            url << "#{CrowbarRabbitmqHelper.get_listen_address(rabbit)}:#{port}"
```
jsuchome: Didn't you set the value of `address` earlier in this recipe?

skazi0: Ha! This code is a bit misleading. Note that `rabbit` here is a loop variable. Maybe I should rename one of those `rabbit`s?

jsuchome: Oh, I should have look closer :-)
Well, I think renaming wouldn't change it, so it's fine as it is I guess

PR #1839, opened 2018-10-15
[4.0] upgrade: Remove chef-client timeout from prepare-mariadb 

skazi0:
There is no need for this timeout since we simplified the script to do everything in
one ssh call. In addition, timeout command only kills the ssh client and leaves
chef-client running remotely. This gives the user false impression that prepare is
already done but also returns false-error result (exit status 124 from timeout reported
as chef-client exit code).


jsuchome:
Please add an "upgrade: " prefix to your commit so Rick can find it

PR #1844, opened 2018-10-17
[5.0] neutron: disable metadata proxy when metadata is forced

ilausuch:
If 'force_metadata' is configured as 'True' then disable the metadata
proxies in all nodes doe not have dhcp namespaces. They are not needed
in l3 agents.

And then 'enable_isolated_metadata' is neglected. So there is no need to
configure it, since 'force_metadata' is a superset and configured that
alone is sufficient to get the metadata from the dhcp server.

(cherry picked from commit e2b408e571e302cf2574a5b9517ca9ca0220becd)
Backport from #1826


ilausuch:

swami-dvr:
Typo in commit message.
nodes doe not ==> nodes that does not

PR #1848, opened 2018-10-17
nova: Only emit unversioned notfications

rhafer:
There is currently nothing consuming the version notifications. So
they're piling up in rabbitmq


rhafer:
> Not yet. This is untested.

I've just tested this. And I can confirm that without this patch the version_notificatios.info queue is growing (when notifcations are enabled) when e.g. launching an instance and that this stops after applying the patch)

PR #1851, opened 2018-10-17
[5.0] rabbitmq: Fix address handling in rabbitmq (aka "glance timeout")

skazi0:
PR #1833 removed some attributes from rabbitmq nodes and replaced them with direct value lookups. This was a fix for race condition leading to empty addresses in rabbitmq connection strings. The change needs to be forward ported to later branches because otherwise the attributes (and the original problem) would be re-added in new systems and during the upgrade.

forward port of #1833

### DIFF
--- a/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
+++ b/chef/cookbooks/crowbar-openstack/libraries/helpers.rb
@@ -170,6 +170,8 @@ class CrowbarOpenStackHelper
       else
         rabbit = rabbits.first
 
+        address = CrowbarRabbitmqHelper.get_listen_address(rabbit)
+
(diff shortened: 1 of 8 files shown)


skazi0:
@jsuchome I wanted to have https://github.com/crowbar/crowbar-openstack/pull/1857 merged first for better cherry-picking of my changes between 5.0 and master.

jsuchome:
I see. Well now it looks you can start merging...

skazi0:
@jsuchome we will need https://github.com/crowbar/crowbar-openstack/pull/1861 before I get a clean cherry-pick. And one more change in 4.0 to make things consistent.

jsuchome:
@rsalevsky we're gonna need this to release with the upgrade packages

PR #1856, opened 2018-10-19
[5.0] ceilometer: add validation that notifications are enabled client side

dirkmueller:
Telemetry collection is based on clients configured to send
notifications. With the change to make the client notification sending
optional, we now should warn the user that the notifications have to
be enabled for Ceilometer/Telemetry to work.


dirkmueller:
backport of #1855

PR #1860, opened 2018-10-22
ironic: Use ironic network for provisioning (bsc#1108398)

skazi0:
Switched to using "ironic" network IP in api_url which is passed to
ironic-python-agent running on the node during deployment and/or cleanup.
This should fix problems when IPA can't access ironic API because ironic
network is the only one attached to the node and there is no routed
access to public network via ironic network.


toabctl:
@skazi0 could you explain in the commit msg why this is needed please?

toabctl:
> @toabctl better?

yes. Thanks!

toabctl:
Backprort for stable-5.0 (aka SOC8) is https://github.com/crowbar/crowbar-openstack/pull/1924

PR #1857, opened 2018-10-19
[5.0] rabbitmq: configure notification settings also in clustered mode

dirkmueller:
Accidentally when forward porting the patch the setting was
only set for the native clustered mode because the hash is
being assembled in two locations. Break out the common part
and merge it so that we follow DRY principles.

Backport of #1854 
(cherry picked from commit a06bbb3663f311ac47b85e0fea64d6973c221d12)

PR #1862, opened 2018-10-23
[4.0] rabbitmq: configure notification settings also in clustered mode

skazi0:
Accidentally when forward porting the patch the setting was
only set for the native clustered mode because the hash is
being assembled in two locations. Break out the common part
and merge it so that we follow DRY principles.

(based on commit a06bbb3663f311ac47b85e0fea64d6973c221d12)

Backport of https://github.com/crowbar/crowbar-openstack/pull/1857 with required modifications


skazi0:
duplicate of https://github.com/crowbar/crowbar-openstack/pull/1862

PR #1863, opened 2018-10-24
monasca: add :start subscribes for monasca-agent

jgrassler:
This fixes a race condition where monasca-agent might get started before its
configuration files are ready.

### DIFF
--- a/chef/cookbooks/monasca/recipes/agent.rb
+++ b/chef/cookbooks/monasca/recipes/agent.rb
@@ -144,6 +144,8 @@ service agent_settings[:agent_service_name] do
   # provider Chef::Provider::CrowbarPacemakerService if ha_enabled
   subscribes :restart, resources(template: "/etc/monasca/agent/agent.yaml")
   subscribes :restart, resources(template: "/etc/monasca/agent/supervisor.conf")
+  subscribes :start, resources(template: "/etc/monasca/agent/agent.yaml")
+  subscribes :start, resources(template: "/etc/monasca/agent/supervisor.conf")
 end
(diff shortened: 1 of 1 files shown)


jgrassler:
Note: this is likely to require additional work (with this change in place monasca-agent on the Monasca node simply ends up not getting started without breaking the proposal). I'll submit a followup for taking care of that problem once I've got it debugged properly, but I'd like to see this change land for now because it will unbreak barclamp deployment).

chef/cookbooks/monasca/recipes/agent.rb
```diff
@@ -144,6 +144,8 @@
   # provider Chef::Provider::CrowbarPacemakerService if ha_enabled
   subscribes :restart, resources(template: "/etc/monasca/agent/agent.yaml")
   subscribes :restart, resources(template: "/etc/monasca/agent/supervisor.conf")
+  subscribes :start, resources(template: "/etc/monasca/agent/agent.yaml")
+  subscribes :start, resources(template: "/etc/monasca/agent/supervisor.conf")
```
dirkmueller: so I'm confused about that. shouldn't we have action changed to not do :start here?

the easiest way to do that would be to just move the service resource to after the creation of the template block.

jgrassler:
Just had a chat with Witek from Upstream and we can get rid of this portion of the code entirely since upstream provides a systemd unit file (without supervisor) as of stable/rocky:

https://github.com/openstack/monasca-agent/commit/a8a2bb845b1129d24683d462c5d1ab08955a2840

I'll fix this upstream in rpm-packaging and create a fresh pull request for removing the supervisor stuff (closing this one).

PR #1865, opened 2018-10-25

ilausuch:
This script blocks the connection to the rabbitmq cluster in case the number of nodes decay below the half
of the total. In this case the remaining rabbit nodes won't accept new connections until quorum is reached.

It takes advantage of the ClusterMon agent that notifies when a rabbitmq has failed or has restored.

The reasoning behind this patch is to allow waiting for more than one node in the cluster to be ready,
which is useful for two reasons:

 - clients would connect to more than one node
 - master for queues would be spread more evenly

Currently most of the connections and master queues are created on the master node, as its the first
to come up when the cluster is restarted/new master promoted.

co-authored-by: Ivan Lausuch <ilausuch@suse.com>
(cherry picked from commit 0d26f8b119e1a3b0f6df75e554001956504d19c6)
Forward port from #1681

### DIFF
--- a/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
+++ b/chef/cookbooks/rabbitmq/recipes/ha_cluster.rb
@@ -159,3 +159,141 @@ pacemaker_transaction "rabbitmq service" do
 end
 
 crowbar_pacemaker_sync_mark "create-rabbitmq_ha_resources"
+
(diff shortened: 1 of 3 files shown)


Itxaka:

Itxaka:
I cant approve this as I was involved in the code wrinting but the 5.0 version was merged so this shouldbe too @rhafer @jsuchome

PR #1867, opened 2018-10-25
neutron: disable metering agent if no ceilometer

Itxaka:
Cuirrently we are enabling and running the metering agent at all
times in the network nodes for no reason.

Instead this patch makes it so its only deployed in the case of having
the ceilometer-agent role, which indicates that we are gathering
metrics and we want to run it. Otherwise its just wasted resources

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents.rb
@@ -15,9 +15,10 @@
 #
 
 include_recipe "neutron::common_agent"
+ceilometer_agent_enabled = node.roles.include? "ceilometer-agent"
 
 package node[:neutron][:platform][:dhcp_agent_pkg]
-package node[:neutron][:platform][:metering_agent_pkg]
+package node[:neutron][:platform][:metering_agent_pkg] if ceilometer_agent_enabled
 
 if node[:neutron][:use_lbaas]
   if node[:neutron][:lbaasv2_driver] == "f5" &&
@@ -119,6 +120,7 @@ template "/etc/neutron/metering_agent.ini" do
     debug: node[:neutron][:debug],
     interface_driver: interface_driver,
   )
+  only_if { ceilometer_agent_enabled }
 end
 
 # Delete pre-existing configuration file.
@@ -223,9 +225,11 @@ service node[:neutron][:platform][:metering_agent_name] do
   subscribes :restart, resources(template: node[:neutron][:config_file])
   subscribes :restart, resources("template[/etc/neutron/metering_agent.ini]")
   provider Chef::Provider::CrowbarPacemakerService if use_crowbar_pacemaker_service
+  only_if { ceilometer_agent_enabled }
 end
 utils_systemd_service_restart node[:neutron][:platform][:metering_agent_name] do
   action use_crowbar_pacemaker_service ? :disable : :enable
+  only_if { ceilometer_agent_enabled }
 end
 
 if node[:neutron][:use_lbaas] &&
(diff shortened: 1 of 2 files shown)


chef/cookbooks/neutron/recipes/network_agents.rb
```diff
@@ -223,9 +225,11 @@
   subscribes :restart, resources(template: node[:neutron][:config_file])
   subscribes :restart, resources("template[/etc/neutron/metering_agent.ini]")
   provider Chef::Provider::CrowbarPacemakerService if use_crowbar_pacemaker_service
+  only_if { ceilometer_agent_enabled }
```
ilausuch: I think is necessary to add the disable and stop actions in case of this is disabled without the subscribes. This is important if you have enabled once the feature and you want to disable it. I checked that in this case when the config files change authomatically the service will stand up. This is the solution I did for the neutron metadata agent service:  https://github.com/crowbar/crowbar-openstack/blob/ff63317590edf6cf463a797ed54128bb0a282c2c/chef/cookbooks/neutron/definitions/neutron_metadata.rb#L94

Itxaka: Good point, I did this only for new installs but if this requires a backport it needs to deal with enabling/disabling

Althougth even if you install then uninstall ceilometer, it wont really be uninstalled so is there a point in doing that? The barclamp will get deleted yes, but the service will still be running in the background so.....

That is, taking into consideration that for this to be enabled/disabled it expects the barclamp to be there, so its different from a enable/disable toggle. Not sure if we need to cover that case of someone disabling ceilometer afterwards, as I dont think we support that path at all...

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -80,8 +80,12 @@
 nova_config = Barclamp::Config.load("openstack", "nova")
 nova_insecure = CrowbarOpenStackHelper.insecure(nova_config) || keystone_settings["insecure"]
 
-service_plugins = ["neutron.services.metering.metering_plugin.MeteringPlugin",
-                   "neutron_fwaas.services.firewall.fwaas_plugin.FirewallPlugin"]
+service_plugins = ["neutron_fwaas.services.firewall.fwaas_plugin.FirewallPlugin"]
+
+if node.roles.include? "ceilometer_agent"
```
dirkmueller: is there a typo here? I would have expected ceilometer-agent

Itxaka: yes there is I thought I fixed this....

Itxaka:
Needed a push with latest changes, should be ok now.

Itxaka:
Umm, test part in jenkins mkcloud fails because metering is disabled:
```
tempest.api.network.test_extensions.ExtensionsTestJSON.test_list_show_extensions
testtools.matchers._impl.MismatchError: 'metering' not in [....
```

Maybe I should actually let the config untouched for the test not to fail?

Itxaka:
oh, this is nice, its hardcoded in the tempest test, so we cannot avoid it. https://github.com/openstack/tempest/blob/19.0.0/tempest/api/network/test_extensions.py#L36

Itxaka:
> are we sure that its only needed for ceilometer? we discussed this at the openstack summit last week and I remember a comment that something else is using this as well (of course I forgot what..)

@dirkmueller Im not sure, I deployed and tested and it seems to work as expected. Maybe I should install *all* the barclamps and test all services to see if it works? I dont have that much knowledge in neutron but I would expect the metering agent to not be a dependency of anything other that what process metrics, and even that they should not fail if there is no metering agent.


In any case, I can change this to be a toggle instead of automatic I guess?

dirkmueller:
@Itxaka I checked with codesearch, only ceilometer listens on the metering queue.. so that looks good to me.

PR #1869, opened 2018-10-26
Set number of neutron rpc workers to cpu number

bbobrov:
Having only 1 worker turned out to be insufficient for seriuos load and
we are increasing it by recommendation from the networking team.


chef/data_bags/crowbar/template-neutron.schema
```diff
@@ -16,7 +16,6 @@
                     "keystone_instance": { "type": "str", "required": true },
                     "create_default_networks": { "type": "bool", "required": true },
                     "dns_domain": { "type": "str", "required": true },
-                    "rpc_workers": { "type": "int", "required": true },
```
dirkmueller: removing a required attribute requires a schema migration

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -12,7 +12,7 @@ ipam_driver = <%= @ipam_driver %>
 global_physnet_mtu = <%= @mtu_value %>
 use_ssl = <%= @ssl_enabled ? "True" : "False" %>
 api_workers = <%= [node["cpu"]["total"], 2, 4].sort[1] %>
-rpc_workers = <%= @rpc_workers %>
+rpc_workers = <%= node["cpu"]["total"] %>
```
dirkmueller: we should only do that for ml2 plugin - and IMHO nproc/2 is a better default (with some limits, e.g. more than 8 imho is probably counter productive)

bbobrov: Do you mean for the whole ml2 plugin or only for l3 agents?
If for ml2, which config value would affect only agents from ml2?

dirkmueller: I remember that there were soem core plugins having issues with multiple workers. we know that the ml2 core plugin is welltested regarding that. hence I suggested to do it only when the ml2 core plugin is selected (Unless we only have ml2 as coreplugin.. there used to be a vmware plugin that was cranky..)

dirkmueller: might not be a relevant issue anymore though. so its a tradeoff discussion  (risk vs balances)

PR #1870, opened 2018-10-26
Fix migrations

Itxaka:
There were a number of duplicated numbering migrations in several barclamps.
Fix them by moving them to new numbers


Itxaka:
@dirkmueller fixed!!

PR #1872, opened 2018-10-31
[5.0] Fix migrations

Itxaka:
There were a number of duplicated numbering migrations in several barclamps.
Fix them by moving them to new numbers

(cherry picked from commit 963c713c1dffe2a0f07ac4b0223e9c3efcc5824c)

Backport-of: https://github.com/crowbar/crowbar-openstack/pull/1870


Itxaka:
ummm, seems that testbuild is failing because it cannot apply a "just rename" patch properly??

```
14:41:21 [  315s] + echo 'Patch #54 (prtest.patch):'
14:41:21 [  315s] Patch #54 (prtest.patch):
14:41:21 [  315s] + /usr/bin/cat /home/abuild/rpmbuild/SOURCES/prtest.patch
14:41:21 [  315s] + /usr/bin/patch -p0 --fuzz=2
14:41:21 [  315s] Cannot rename file without two valid file names
14:41:21 [  315s] Cannot rename file without two valid file names
14:41:21 [  315s] Cannot rename file without two valid file names
14:41:21 [  315s] error: Bad exit status from /var/tmp/rpm-tmp.UGMLb3 (%prep)
```

Itxaka:
added a empty test file to check if the problem comes from that issue....

Itxaka:
wow, so it does require a change to a file in order to be able to apply the patch properly, cant do it if there is only renames :o

PR #1880, opened 2018-11-05
cinder: wait for volume group to exist

jgrassler:
Occasionally volume groups take a while to appear after `vgcreate`. If
openstack-cinder-volume is started before that, it will exit 1 because it
cannot find them. This commit waits 60s for the volume groups to be visible.

### DIFF
--- a/chef/cookbooks/cinder/recipes/volume.rb
+++ b/chef/cookbooks/cinder/recipes/volume.rb
@@ -57,8 +57,13 @@ def make_loopback_volume(backend_id, volume)
 
   return if volume_exists(volname)
 
+  # Make our volume group. Wait for it to be created to prevent race conditions
+  # upon cinder-volume startup.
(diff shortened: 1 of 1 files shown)


chef/cookbooks/cinder/recipes/volume.rb
```diff
@@ -57,8 +57,13 @@ def make_loopback_volume(backend_id, volume)
 
   return if volume_exists(volname)
 
+  # Make our volume group. Wait for it to be created to prevent race conditions
+  # upon cinder-volume startup.
   bash "Create volume group #{volname}" do
```
Itxaka: You could use a simple `retries` and `retry_delay` to do the same, like setting 2 retries with a delay of 30 seconds between them instead of doing it in bash :+1:

jgrassler:
Not needed anymore - the problem was a different one anyway. Closing.

PR #1881, opened 2018-11-05
add tempest plugins

guangyee:
Add the tempest plugins for services we care about at the moment.

barbican, heat, and monasca are being added in #1611.

### DIFF
--- a/chef/cookbooks/tempest/templates/default/tempest.conf.erb
+++ b/chef/cookbooks/tempest/templates/default/tempest.conf.erb
@@ -165,6 +165,9 @@ sahara = <%= @enabled_services.include? "data-processing" %>
 trove = <%= @enabled_services.include? "database" %>
 manila = <%= @enabled_services.include? "sharev2" %>
 magnum = <%= @enabled_services.include? "container-infra" %>
+ironic = <%= @enabled_services.include? "baremetal" %>
+ironic_inspector = <%= @enabled_services.include? "baremetal" %>
+designate = <%= @enabled_services.include? "dns" %>
 
 [share]
 region = <%= @keystone_settings['endpoint_region'] %>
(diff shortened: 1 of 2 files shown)


chef/cookbooks/tempest/templates/default/tempest.conf.erb
```diff
@@ -162,6 +162,9 @@ sahara = <%= @enabled_services.include? "data-processing" %>
 trove = <%= @enabled_services.include? "database" %>
 manila = <%= @enabled_services.include? "sharev2" %>
 magnum = <%= @enabled_services.include? "container-infra" %>
+barbican = <%= @enabled_services.include? "key-manager" %>
+designate = <%= @enabled_services.include? "dns" %>
```
kbaikov: Do we have designate in crowbar?

guangyee: Good question! I can't seem to find it. I'll remove designate-tempest-plugin from the list.

guangyee:
@dirkmueller thanks for the information. Looks like #1611 only has barbican-tempest-plugin change so far. Please advice on the path forward. We can either

1. abandon this one and merge the changes into #1611, or
2. remove the barbican change from this one

chef/cookbooks/tempest/templates/default/tempest.conf.erb
dirkmueller: we have an open PR to add designate, so I'd be okay with keeping it in.

dirkmueller:
@guangyee are we looking at the same thing? its adding heat, barbican and monasca. the other tempest plugins weren't added to the product back then (might have changed meanwhile).

chef/cookbooks/tempest/templates/default/tempest.conf.erb
guangyee: @dirkmueller done. I've added back designate.

PR #1885, opened 2018-11-06
adapt options names for fence_compute (SCRD-5553)

sjamgade:
these were changed
https://github.com/ClusterLabs/fence-agents/commit/eae9d029b7073e7eb8c7ba4df9ec19b755a8f603

### DIFF
--- a/chef/cookbooks/nova/recipes/compute_ha.rb
+++ b/chef/cookbooks/nova/recipes/compute_ha.rb
@@ -274,16 +274,16 @@ pacemaker_primitive fence_primitive do
   agent "stonith:fence_compute"
   params ({
     "pcmk_host_map"  => hostmap,
-    "auth-url"       => internal_auth_url_v2,
+    "auth_url"       => internal_auth_url_v2,
     # "region-name"    => keystone_settings["endpoint_region"],
-    "endpoint-type"  => "internalURL",
+    "endpoint_type"  => "internalURL",
     "login"          => keystone_settings["admin_user"],
     "passwd"         => keystone_settings["admin_password"],
-    "tenant-name"    => keystone_settings["admin_tenant"],
+    "tenant_name"    => keystone_settings["admin_tenant"],
     # "insecure"       => keystone_settings["insecure"] || nova[:nova][:ssl][:insecure],
     "domain"         => node[:domain],
-    "no-shared-storage" => no_shared_storage,
-    "record-only"    => "1",
+    "no_shared_storage" => no_shared_storage,
+    "record_only"    => "1",
     "verbose"        => "1",
     "debug"          => "/var/log/nova/fence_compute.log"
   })


cmurphy:
That commit was over a year ago, have we been having this problem for a year? Or why is it happening now?

sjamgade:
nice, thanks! 
I didnot know that patches were being maintained locally. Will dig more

aspiers:
[**@cmurphy**](https://github.com/cmurphy) commented on [6 Nov 2018, 12:02 GMT](https://github.com/crowbar/crowbar-openstack/pull/1885#issuecomment-436229205 "2018-11-06T12:02:08Z - Replied by Github Reply Comments"):
> That commit was over a year ago, have we been having this problem for a year? Or why is it happening now?

The HA team has confirmed in IRC and [on bugzilla](https://bugzilla.suse.com/show_bug.cgi?id=1111579#c48) that this breakage was accidentally introduced by a maintenance update.  See [bsc#1111579 (Partner-L3: Nova barclamp fails [crm configure nova compute (non-remote bits)])](https://bugzilla.suse.com/show_bug.cgi?id=1111579) for the latest.

jsuchome:
If there's an active development here and you're trying to make it work for recent `fence-agents`, don't forget about required value for `plug`:
```
ERROR: 8: fence-nova: required parameter plug not defined
```

aspiers:
[**@jsuchome**](https://github.com/jsuchome) commented on [9 Nov 2018, 13:11 GMT](https://github.com/crowbar/crowbar-openstack/pull/1885#issuecomment-437355189 "2018-11-09T13:11:31Z - Replied by Github Reply Comments"):
> If there's an active development here and you're trying to make it work for recent `fence-agents`, don't forget about required value for `plug`:
> ```
> ERROR: 8: fence-nova: required parameter plug not defined
> ```

My brain is fried but IIUC `plug` is provided by Pacemaker internals not by configuring it, because the value represents a host which gets determined by pengine.  crm is getting confused when figuring it because the metadata returned by the agent doesn't meet crm's expectations (`plug` vs. `port`).  The fix is in https://bugzilla.suse.com/show_bug.cgi?id=1111579#c38

jsuchome:
Oh. and you must be right. 

I did not notice that the patch actually solves the "plug" problem, somehow I thought it is still present after applying the patch you mention.

I should not comment on the bugs/PR's when I read through them only briefly :-(

aspiers:
[**@jsuchome**](https://github.com/jsuchome) commented on [12 Nov 2018, 09:12 GMT](https://github.com/crowbar/crowbar-openstack/pull/1885#issuecomment-437808005 "2018-11-12T09:12:48Z - Replied by Github Reply Comments"):
> I should not comment on the bugs/PR's when I read through them only briefly :-(

I think it's fine to do this, because sometimes a random thought from an observer with a fresh pair of eyes can be really useful.  Sure there is a risk of saying something incorrect, which is why I often put big disclaimers like "I haven't been following this closely, but ..." or "sorry, haven't had time to review this properly but ..." in my comments ;-)

sjamgade:
this is no longer required

PR #1887, opened 2018-11-06
Neutron: remove deprecated options

ilausuch:
Removed deprecated options from the configuration

sql/min_pool_size
ovs/ovsdb_interface


ilausuch:
Will be integrated in #1892

PR #1889, opened 2018-11-07
Fix "enable_metadata_proxy" setting for DVR setups

rhafer:
With DVR the l3-agent needs to be configured on the compute nodes as
well. As the computes don't have the neutron proposal assigned we need
to read the [:metadata][:force] attribute from a neutron-node.

### DIFF
--- a/chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
+++ b/chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
@@ -3,7 +3,7 @@ interface_driver = <%= @interface_driver %>
 <% if @dvr_enabled -%>
 agent_mode = <%= @dvr_mode %>
 <% end -%>
-<% if node[:neutron][:metadata][:force] -%>
+<% if @force_metadata -%>
 enable_metadata_proxy = False
 <% end -%>
 metadata_port = <%= @metadata_port %>
--- a/chef/cookbooks/neutron/recipes/common_agent.rb
+++ b/chef/cookbooks/neutron/recipes/common_agent.rb
@@ -334,6 +334,7 @@ if neutron[:neutron][:networking_plugin] == "ml2"
         interface_driver: interface_driver,
         handle_internal_only_routers: "True",
         metadata_port: 9697,
+        force_metadata: neutron[:neutron][:metadata][:force],
         periodic_interval: 40,
         periodic_fuzzy_delay: 5,
         dvr_enabled: neutron[:neutron][:use_dvr],


rhafer:
Seems this was introduced as part of: e2b408e571e302c

chef/cookbooks/neutron/templates/default/l3_agent.ini.erb
```diff
@@ -3,7 +3,7 @@ interface_driver = <%= @interface_driver %>
 <% if @dvr_enabled -%>
 agent_mode = <%= @dvr_mode %>
 <% end -%>
-<% if node[:neutron][:metadata][:force] -%>
+<% if @force_metadata -%>
```
ilausuch: Should it be inside of dvr_enabled condition? Because is a requisite in case of DVR. Do you agree?

rhafer: @ilausuch Hm, what is a requisite in case of DVR? I must admit that I didn't get the complete idea behind the orginal fix (https://github.com/crowbar/crowbar-openstack/pull/1826). I was just trying to address a pretty obvious issue with an `undefined method `[]' for nil:NilClass` exception thrown on compute nodes in DVR setups. And this is basically the minimal fix for that without changing any logic in the code.

Are you saying that the whole "enable_metadata_proxy" setting is only relevant to DVR configurations?

Itxaka:
This is weird, so this seems to fix it indeed but only because of ruby shenanigans and erb not complaining if that `@force_metadata` var doesnt exists.

Seems ok to me thought.

rhafer:
> This is weird, so this seems to fix it indeed but only because of ruby shenanigans and erb not complaining if that `@force_metadata` var doesnt exists.

Hm? `@force_metadata` exists. It's passed in as template var (see: https://github.com/crowbar/crowbar-openstack/pull/1889/files#diff-91474df28ca9d038c83a65ad4c220591R337)

Itxaka:
> Hm? @force_metadata exists. It's passed in as template var (see: crowbar/crowbar-openstack/pull/1889/files#diff-91474df28ca9d038c83a65ad4c220591R337)

Shouldnt then it fail because it fails to find the var in the common_config.rb file?

I thougth that maybe l3 agent is not running the common_config recipe so it will fail, but doesnt seem like that.


Oooh, I see, this is because we don not set `node[:neutron][:metadata][:force]` as a default attribute in the attributes.rb file! That should be the proper fix instead, so we have a default value for that, isnt it?

rhafer:
> Shouldnt then it fail because it fails to find the var in the common_config.rb file?

The  common_config.rb  file is accessing the var via `neutron[....]` instead of `node[...]`. `neutron` points to one of the neutron nodes instead of the node that the recipe is currently running on. (Might it be that you overlooked the "node" vs "neutron" change?

[..]

> Oooh, I see, this is because we don not set `node[:neutron][:metadata][:force]` as a default attribute in the attributes.rb file! That should be the proper fix instead, so we have a default value for that, isnt it?

In that case we'd be using the wrong value on compute nodes that use DVR. We really need to take what  ever is in the neutron proposal. And we get that by taking the value from any neutron node.

Itxaka:
> (Might it be that you overlooked the "node" vs "neutron" change?

wooops. I just made this more difficult for myself than it really is. No wonder I was puzzled!

PR #1892, opened 2018-11-07
neutron: new configuration variables

ilausuch:
Removed deprecated options from the configuration

From neutron.conf
- sql/min_pool_size
Minimum number of SQL connections to keep open in a pool.

From ovs openvswitch_agent.ini
- ovs/ovsdb_interface
The interface for interacting with the OVSDB

Added additional configuration variables to neutron.conf:

- report_interval
Seconds between nodes reporting state to server; should be less
than agent_down_time, best if it is half or less than agent_down_time.

- agent_down_time
Seconds to regard the agent is down; should be at least twice
report_interval, to be sure the agent is down for good.

- sql/connection_recycle_time
Connections which have been present in the connection pool longer
than this number of seconds will be replaced with a new one the
next time they are checked out from the pool.

- sql/connection_parameters
Optional URL parameters to append onto the connection URL at connect
time; specify as param1=value1&param2=value2&…

- nova/split_loggers
Log requests to multiple loggers.

Added additional configuration variables to openvswitch_agent.conf:

- network_log/rate_limit
Maximum packets logging per second.

- burst_limit
Maximum number of packets per rate_limit.

- local_output_log_base
Output logfile path on agent side, default syslog file.

Added additional configuration variables to dhcp_agent conf:

- dhcp_renewal_time
DHCP renewal time T1 (in seconds). If set to 0, it will default to half of the lease time.

- dhcp_rebinding_time
DHCP rebinding time T2 (in seconds). If set to 0, it will default to 7/8 of the lease time.

- ovs/ovsdb_debug
Enable OVSDB debug logs

- ovs/ovsdb_timeout
Timeout in seconds for ovsdb commands. If the timeout expires, ovsdb commands will fail with ALARMCLOCK error.

- ovs/bridge_mac_table_size
The maximum number of MAC addresses to learn on a bridge managed by the
Neutron OVS agent. Values outside a reasonable range (10 to 1,000,000)
might be overridden by Open vSwitch according to the documentation.

### DIFF
--- a/chef/cookbooks/neutron/attributes/default.rb
+++ b/chef/cookbooks/neutron/attributes/default.rb
@@ -54,7 +54,6 @@ default[:neutron][:api][:protocol] = "http"
 default[:neutron][:api][:service_port] = "9696"
 default[:neutron][:api][:service_host] = "0.0.0.0"
 
-default[:neutron][:sql][:min_pool_size] = 30
 default[:neutron][:sql][:max_pool_size] = 60
 default[:neutron][:sql][:max_pool_overflow] = 10
 default[:neutron][:sql][:pool_timeout] = 30
--- a/chef/cookbooks/neutron/templates/default/neutron.conf.erb
+++ b/chef/cookbooks/neutron/templates/default/neutron.conf.erb
@@ -28,6 +28,8 @@ wsgi_keep_alive = false
 <% unless @default_log_levels.length.zero? -%>
 default_log_levels = <%= @default_log_levels.join(", ") %>
 <% end -%>
+report_interval = <%= @report_interval %>
+agent_down_time = <%= @agent_down_time %>
 
 [agent]
 root_helper = sudo neutron-rootwrap /etc/neutron/rootwrap.conf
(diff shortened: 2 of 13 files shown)


chef/cookbooks/neutron/attributes/default.rb
```diff
@@ -54,7 +54,6 @@
 default[:neutron][:api][:service_port] = "9696"
 default[:neutron][:api][:service_host] = "0.0.0.0"
 
-default[:neutron][:sql][:min_pool_size] = 30
 default[:neutron][:sql][:max_pool_size] = 60
```
jgrassler: Does this option still exist? It strikes me as a bit odd that only `min_pool_size` is deprecated. Given that it's in the `[sql]` section, the deprecation is probably not originating from Neutron itself anyway...

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -86,6 +89,7 @@ project_name = <%= @keystone_settings['service_tenant'] %>
 username = <%= @keystone_settings['service_user'] %>
 project_domain_name = <%= @keystone_settings['default_user_domain'] %>
 user_domain_name = <%= @keystone_settings['default_user_domain'] %>
+split_loggers = <%= @nova_split_loggersf ? "True" : "False" %>
```
jgrassler: Typo: should be `nova_split_loggers`, not `split_loggersf`.

chef/cookbooks/neutron/attributes/default.rb
ilausuch: Yes only the minimum configuration is deprecated
Reason: The option to set the minimum pool size is not supported by sqlalchemy.

https://docs.openstack.org/neutron/rocky/configuration/neutron.html

chef/cookbooks/neutron/templates/default/neutron.conf.erb
ilausuch: Good one. Solved

ilausuch:
Closed because the deprecations and new features are going to be in different PR. At this moment only deprecated settings are implemented in #1916

PR #1893, opened 2018-11-08
nova: Remove deprecated vnc keymap configuration (SCRD-5787)

kwu83tw:
- get rid of keymap in [vnc]

### DIFF
--- a/chef/cookbooks/nova/templates/default/nova.conf.erb
+++ b/chef/cookbooks/nova/templates/default/nova.conf.erb
@@ -311,11 +311,11 @@ cluster_name = "<%= cluster %>"
   <% end %>
 datastore_regex = <%= node[:nova][:vcenter][:datastore] %>
 vlan_interface = <%= node[:nova][:vcenter][:interface] %>
+vnc_keymap = <%= node[:nova][:vcenter][:vnc_keymap] %>
 <% end %>
 
 [vnc]
 enabled = <%= @vnc_enabled ? "True" : "False" %>
-keymap = <%= node[:nova][:vnc_keymap] %>
 server_listen = "0.0.0.0"
 server_proxyclient_address = <%= node[:nova][:my_ip] %>
 novncproxy_host = <%= @bind_host %>
--- a/chef/data_bags/crowbar/migrate/nova/303_update_vnc_keymap.rb
+++ b/chef/data_bags/crowbar/migrate/nova/303_update_vnc_keymap.rb
@@ -1,0 +1,11 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["vcenter"]["vnc_keymap"] = attrs["vnc_keymap"]
+  attrs.delete("vnc_keymap")
+  return attrs, deployment
+end
+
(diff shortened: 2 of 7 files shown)


kwu83tw:
https://jira.suse.de/browse/SCRD-5923 This is the ticket for documentation tracking purpose.

sjamgade:
the deprecation document talks about adding `[vmware]novnc_keymap`, should we not add it ?

kwu83tw:
> the deprecation document talks about adding `[vmware]novnc_keymap`, should we not add it ?

Sounds like that. Patch updated. Thanks.

This is the PR for doc change as reference https://github.com/SUSE-Cloud/doc-cloud/pull/652. I'll do a follow up with the additional attr introduced in [vmware]

chef/data_bags/crowbar/migrate/nova/303_update_vnc_keymap.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["vcenter"]["vnc_keymap"] = template_attrs["vcenter"]["vnc_keymap"]
```
dirkmueller: it would be good to preserve the existing setting rather than replacing with template default. a.a fetch it from attrs['vnc_keymap'] instead.

crowbar_framework/app/views/barclamp/nova/_edit_attributes.html.haml
```diff
@@ -52,6 +52,7 @@
       = string_field %w(vcenter interface)
       = string_field %w(vcenter ca_file)
       = boolean_field %w(vcenter insecure)
+      = string_field :vnc_keymap
```
dirkmueller: thats not referring to the right field, you moved it inside %w(vcenter vnc_keymap)

sjamgade: this should still be as dirk suggested, because vnc_keymap is still deleted.

chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -311,11 +311,11 @@ cluster_name = "<%= cluster %>"
   <% end %>
 datastore_regex = <%= node[:nova][:vcenter][:datastore] %>
 vlan_interface = <%= node[:nova][:vcenter][:interface] %>
+vnc_keymap = <%= node[:nova][:vnc_keymap] %>
```
sjamgade: iiuc this should be  ```vnc_keymap = <%= node[:nova][:vcenter][:vnc_keymap] %> ``` because vnc_keymap is deleted

toabctl: This is not inline with the commit msg (which only says that you are removing something). Please split the commits or adjust the commit msg

kwu83tw: I went back to reuse the existing value instead of creating new attr under [vcenter]. iiuc it shall retrieve the original value.

kwu83tw: patch updated.

PR #1894, opened 2018-11-08
cinder: enable initial deletion resume offloading

dirkmueller:
With this setting, the resume of deleting volumes on startup
is done in the background, which solves availability issues
with the backend when many volumes are in deleting state.


dirkmueller:
this is not working.

PR #1896, opened 2018-11-15
add 'creator' role to dynamic credential (SCRD-5821)

guangyee:
It is required for creating secrets in Barbican. Otherwise
barbican_tempest_plugin will fail miserably.


jgrassler:
The HA job failed due to being out of pool slots on mkcha, retriggering.

kbaikov:
Tested it manually and this fixed the "Forbidden" error. But triggered 2 other errors which will have their own bug report.

PR #1899, opened 2018-11-19
[4.0] neutron: disable metering agent if no ceilometer

Itxaka:
Currently we are enabling and running the metering agent at all
times in the network nodes for no reason.

Instead this patch makes it so its only deployed in the case of having
the ceilometer-agent role, which indicates that we are gathering
metrics and we want to run it. Otherwise its just wasted resources

(cherry picked from commit 17206688fe668ac5cba0517d4135155c9f5afd3b)

Backport-of: https://github.com/crowbar/crowbar-openstack/pull/1867

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents.rb
@@ -15,9 +15,10 @@
 #
 
 include_recipe "neutron::common_agent"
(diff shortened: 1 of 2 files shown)


Itxaka:
[Crowbar patch for openstack-neutron-metering-agent.service](https://trello.com/c/vkFxh2eU/326-crowbar-patch-for-openstack-neutron-metering-agentservice)

PR #1900, opened 2018-11-19
Horizon policy fix (bsc#1111421)

sjamgade:
the policy file names are part of the settings files and the files are no longer there, so removing the config

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -84,6 +84,24 @@ else
   end
 end
 
+# install horizon neutron fwaas plugin if needed
+neutron_fwaas_ui_pkgname =
+  case node[:platform_family]
+  when "suse"
+    "openstack-horizon-plugin-neutron-fwaas-ui"
+  when "rhel"
+    "openstack-neutron-fwaas-ui"
+  end
+
+unless neutron_fwaas_ui_pkgname.nil?
+  unless Barclamp::Config.load("openstack", "neutron").empty?
+    package neutron_fwaas_ui_pkgname do
+      action :install
+      notifies :reload, "service[horizon]"
+    end
+  end
+end
+
 # install horizon manila plugin if needed
 manila_ui_pkgname =
   case node[:platform_family]
--- a/chef/cookbooks/horizon/templates/default/local_settings.py.erb
+++ b/chef/cookbooks/horizon/templates/default/local_settings.py.erb
@@ -102,7 +102,6 @@ POLICY_FILES = {
     'compute': '<%= @policy_file[:compute] %>',
     'volume': '<%= @policy_file[:volume] %>',
     'image': '<%= @policy_file[:image] %>',
-    'orchestration': '<%= @policy_file[:orchestration] %>',
     'network': '<%= @policy_file[:network] %>',
     'neutron-fwaas': '<%= @policy_file[:neutron_fwaas] %>',
 }
--- a/chef/data_bags/crowbar/migrate/horizon/302_remove_heat_policy_file.rb
+++ b/chef/data_bags/crowbar/migrate/horizon/302_remove_heat_policy_file.rb
@@ -1,0 +1,9 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["policy_file"].delete("orchestration")
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["policy_file"]["orchestration"] = template_attrs["policy_file"]["orchestration"]
+  return attrs, deployment
+end
--- a/chef/cookbooks/horizon/attributes/default.rb
+++ b/chef/cookbooks/horizon/attributes/default.rb
@@ -29,7 +29,6 @@ default[:horizon][:policy_file][:identity] = "keystone_policy.json"
 default[:horizon][:policy_file][:compute] = "nova_policy.json"
 default[:horizon][:policy_file][:volume] = "cinder_policy.json"
 default[:horizon][:policy_file][:image] = "glance_policy.json"
-default[:horizon][:policy_file][:orchestration] = "heat_policy.json"
 default[:horizon][:policy_file][:network] = "neutron_policy.json"
 default[:horizon][:policy_file][:neutron_fwaas] = "neutron-fwaas-policy.json"
 
--- a/chef/data_bags/crowbar/template-horizon.json
+++ b/chef/data_bags/crowbar/template-horizon.json
@@ -27,7 +27,6 @@
         "compute": "nova_policy.json",
         "volume": "cinder_policy.json",
         "image": "glance_policy.json",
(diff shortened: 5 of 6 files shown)


chef/data_bags/crowbar/migrate/horizon/301_remove_heat_policy_file.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  a[policy_file].delete("orchestration")
```
toabctl: Where does `a` come from? Please use the usual variable names that are used in 99% of the migrations (`ta, td, a, d`)

chef/data_bags/crowbar/migrate/horizon/301_remove_heat_policy_file.rb
```diff
@@ -0,0 +1,9 @@
...
+  return a, d
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  a["policy_file"]["orchestration"] = ta["policy_file"]["orchestration"]
```
toabctl: same here

chef/data_bags/crowbar/migrate/horizon/301_remove_heat_policy_file.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  a[policy_file].delete("orchestration")
+  return a, d
```
toabctl: same here

chef/data_bags/crowbar/migrate/horizon/301_remove_heat_policy_file.rb
```diff
@@ -0,0 +1,9 @@
...
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  a["policy_file"]["orchestration"] = ta["policy_file"]["orchestration"]
+  return a, d
```
toabctl: same here

chef/data_bags/crowbar/migrate/horizon/301_remove_heat_policy_file.rb
sjamgade: done

chef/data_bags/crowbar/migrate/horizon/301_remove_heat_policy_file.rb
sjamgade: done

chef/data_bags/crowbar/migrate/horizon/301_remove_heat_policy_file.rb
sjamgade: done

chef/data_bags/crowbar/migrate/horizon/301_remove_heat_policy_file.rb
sjamgade: done

chef/data_bags/crowbar/migrate/horizon/302_remove_heat_policy_file.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs[policy_file].delete("orchestration")
```
cmurphy: `policy_file` should be in quotes like it is below

sjamgade:
So heat-dashboard if and when packaged and installed, will have to add 
```policy_files["orchestration"] = "/path/to/file.yaml" ```

File: chef/data_bags/crowbar/migrate/horizon/302_remove_heat_policy_file.rb (hunk 5 above)
Comment (sjamgade) [reply chef/data_bags/crowbar/migrate/horizon/302_remove_heat_policy_file.rb]: I wonder how this passed the CI, thanks !

File: chef/cookbooks/horizon/templates/default/local_settings.py.erb [hunk 6]
```diff
@@ -98,13 +98,9 @@ POLICY_FILES_PATH = '<%= @policy_file_path %>'
...
 POLICY_FILES = {
-    'identity': '<%= @policy_file[:identity] %>',
-    'compute': '<%= @policy_file[:compute] %>',
-    'volume': '<%= @policy_file[:volume] %>',
-    'image': '<%= @policy_file[:image] %>',
-    'orchestration': '<%= @policy_file[:orchestration] %>',
-    'network': '<%= @policy_file[:network] %>',
-    'neutron-fwaas': '<%= @policy_file[:neutron_fwaas] %>',
+<% @policy_file.each do | service, file| -%>
+    "<%= service %>" : '<%= file %>',
```
toabctl: This is not the same as before in the neutron-fwaas case. Now this would result in `neutron_fwaas: xxx` but before there was no underscore.

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -385,6 +385,9 @@
   # to be in the certfile
 end
 
+# replace _ with - 
+policy_filemap = node["horizon"]["policy_file"]
+policy_filemap['neutron-fwaas'] = policy_filemap.delete ['neutron_fwaas']
```
toabctl: Hm. I like the currently static POLICY_FILES hash more because imo that is easier to read and understand. but I'll not block the merge...

chef/cookbooks/horizon/templates/default/local_settings.py.erb
sjamgade: thanks, I have added a local hash map with the altered keys to avoid altering the node object

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -385,6 +385,9 @@
   # to be in the certfile
 end
 
+# replace _ with -
+policy_files = node["horizon"]["policy_file"].dup
+policy_files["neutron-fwaas"] = policy_files.delete ["neutron_fwaas"]
```
cmurphy: What about fixing the schema instead of doing this workaround?

chef/cookbooks/horizon/recipes/server.rb
cmurphy: I agree with this

sjamgade:
yes I noticed that, thanks for that.
updated the PR to rename it at schema level, but ruby doesnot like `-` in its `symbols` (for attributes\default.rb) so either we go back to the minor workaround, or I just drop the the looping behavior

sjamgade:
moved back to old style of doing things ! 
but looks like the heat dashboard has been packaged !

cmurphy:
So do we not need this anymore now that heat-dashboard is packaged? Or should we keep this but make it conditional on whether the heat barclamp is applied?

sjamgade:
I am in favor of this as and let https://github.com/openstack/heat-dashboard/blob/master/heat_dashboard/local_settings.d/_1699_orchestration_settings.py take care of things. 

this settings will be overridden by that file, so its actually misleading, and that file does more that what is being done here.

PR #1902, opened 2018-11-21
Monasca: fix agent is not ready in each barclamp

ilausuch:
The problem: When the barclamps are preparing the monasca monitor
it cannot find a resource matching openstack-monasca-agent.target
because is not defined.

The solution: Delay the monitoring until the agent is already
defined. To do that the monitor setup is performed also inside of
monasca agent recipe. Additionally the monitor setup has a check
to be sure the monasca master, service and agent are running.


chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -255,6 +255,14 @@ def self.rabbitmq_settings(node, barclamp)
     @rabbitmq_settings[instance]
   end
 
+  def self.no_monasca_server_or_master(node)
```
jgrassler: There's no need to pass a node object to this function, for...

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -255,6 +255,14 @@ def self.rabbitmq_settings(node, barclamp)
...
+    monasca_master = CrowbarUtilsSearch.node_search_with_cache(node, "roles:monasca-master").first
```
jgrassler: ...for you can search for the Monasca node without one as we do here:

https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/monasca/recipes/agent.rb#L24

jgrassler: First option sounds more reasonable, yes.

chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -32,6 +32,14 @@ def fetch_database_connection_string(db_auth, barclamp = @cookbook_name)
     def fetch_rabbitmq_settings(barclamp=@cookbook_name)
       CrowbarOpenStackHelper.rabbitmq_settings(node, barclamp)
     end
+
+    def no_monasca_server_or_master
+      monasca_master = node_search_with_cache("roles:monasca-master").first
+      return true if monasca_master.nil?
```
cmurphy: (nitpick) you could replace this with `return monasca_master.nil?`

ilausuch: Done

ilausuch: I put it in the second one

ilausuch: @cmurphy you were wright. I change it in my last commit

PR #1906, opened 2018-11-23
[5.0] upgrade: reduce keystone downtime

rhafer:
This introduces the needed changes in the keystone cookbook for reducing its downtime during the upgrade. More details in: https://github.com/crowbar/crowbar-core/pull/1715

ToDo items:

- [x] Skip clearing the db_sync flag for keystone only when really needed. Currenty this PR breaks upgrades of the "normal" setups

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -129,6 +129,7 @@ if node[:keystone][:frontend] == "uwsgi"
   end
 
 elsif node[:keystone][:frontend] == "apache"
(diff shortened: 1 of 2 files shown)


jsuchome:
+1 to that db_synced fix

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -129,6 +129,7 @@
   end
 
 elsif node[:keystone][:frontend] == "apache"
+  keystone_enabled = !node[:keystone][:disable_vhost]
```
skazi0: do we need some default for this attribute or we handle this with ruby magic (nil => false)?

PR #1908, opened 2018-11-27
openstack: don't install ruby database clients

dirkmueller:
So afaik openstack is written in python, so we only need
to install the database backend for python clients. We
can also skip installation of python-mysql as it is not
supposed to be used anymore.


dirkmueller:
not a good idea atm, thanks for the reviews.

PR #1909, opened 2018-11-27
cinder: add nfs qcow2 option

ilausuch:
Allows cinder to create volumes as QCOW2 files rather than raw files for
nfs volume

- nfs_qcow2_volumes

https://docs.openstack.org/cinder/pike/configuration/tables/conf-changes/cinder.html


chef/data_bags/crowbar/migrate/cinder/301_nfs_new_settings.rb
```diff
@@ -0,0 +1,10 @@
+def upgrade(tattr, tdep, attr, dep)
+  entry = "nfs_qcow2_volumes"
+  attr["volume_defaults"]["nfs"][entry] = tattr["volume_defaults"]["nfs"][entry]
+  return attr, dep
```
toabctl: What about existing nfs backends? Have you tested the case where you already have an NFS backend and then apply the migration?
You are only changing the default vaiue but `nfs_qcow2_volumes` is a required field. So I think you need to iterate over all available `nfs` backends and set `nfs_qcow2_volumes=false`
See `015_rename_vmware_volume_folder.rb` for a loop example

ilausuch: This Is not tested yet. Thanks for the info.

ilausuch: I added two checks
- One in the migration file. But I don't know if it is necessary
- The second is in crowbar part. And checks that you cannot change the value of nfs_qcow2_volumes if you already has defined a nfs backend

PR #1910, opened 2018-11-27
Adde SES integration barclamp

hemna:
This patch adds the SUSE Enterprise Storage barclamp.
The SES integration consumes the SES openstack runner's
config.yml, and populates the SES barclamp's information about
the SES/ceph cluster.  The SES integration also updates the
cinder, glance and nova barclamps to automatically consume the
SES/ceph external cluster options.   The patch automatically creates
an /etc/ceph/ceph.conf on all service nodes as well as the required
keyring files, and then configures cinder, glance, nova to work
with the external SES/ceph cluster.


chef/roles/ses-controller.rb
```diff
@@ -0,0 +1,5 @@
+name "ses-controller"
```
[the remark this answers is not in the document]
hemna: every single file in this directory has the same exact naming scheme.

crowbar_framework/app/views/barclamp/ses/_edit_attributes.html.haml
```diff
@@ -0,0 +1,65 @@
...
+    %script#yaml{ :type => "text/javascript", :src => "/assets/barclamps/ses/yaml.min.js" }
+
+    %fieldset
+      %legend
+        = t(".yaml_header")
+
+      = text_field %w(yaml_file)
+
+      %div.form-group.pull-left
+        %input(id='parse_yaml' class="form-control" type="button" value="Parse Yaml")
```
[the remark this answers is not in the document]
hemna: can't use - in names of ids in html, due to javascript not allowing - in var names.

chef/cookbooks/nova/recipes/ceph_post_libvirt.rb
```diff
@@ -0,0 +1,186 @@
...
+          # lazy xml parsing
+          re_match = %r[<usage type='ceph'>.*<name>(.*)</name>]m.match(secret_xml)
+          next if re_match.nil?
+          secret_usage = re_match[1]
+          undefine = false
+
+          if secret_uuid == rbd_uuid
+            undefine = true if secret_usage != "crowbar-#{rbd_uuid} secret"
+          else
+            undefine = true if secret_usage == "crowbar-#{rbd_uuid} secret"
```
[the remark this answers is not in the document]
hemna: no

chef/cookbooks/nova/recipes/ceph_post_libvirt.rb
```diff
@@ -0,0 +1,186 @@
...
+          end
+
+          if undefine
```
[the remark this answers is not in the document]
hemna: no

crowbar_framework/app/assets/javascripts/barclamps/ses/yaml.min.js
```diff
@@ -0,0 +1 @@
...null}}else{o=e;r=o("fs");if(n!=null){return r.readFile(t,function(e,t){if(e){return n(null)}else{return n(String(t))}})}else{i=r.readFileSync(t);if(i!=null){return String(i)}return null}}};return t}();t.exports=r},{"./Pattern":8}],11:[function(e,t,n){var i,r,s,l;r=e("./Parser");i=e("./Dumper");s=e("./Utils");l=function(){function e(){}e.parse=function(e,t,n){if(t==null){t=false}if(n==null){n=null}return(new r).parse(e,t,n)};e.parseFile=function(e,t,n,i){var r;if(t==null){t=null}if(n==null){n=false}if(i==null){i=null}if(t!=null){return s.getStringFromFile(e,function(e){return function(r){var s;s=null;if(r!=null){s=e.parse(r,n,i)}t(s)}}(this))}else{r=s.getStringFromFile(e);if(r!=null){return this.parse(r,n,i)}return null}};e.dump=function(e,t,n,r,s){var l;if(t==null){t=2}if(n==null){n=4}if(r==null){r=false}if(s==null){s=null}l=new i;l.indentation=n;return l.dump(e,t,0,r,s)};e.stringify=function(e,t,n,i,r){return this.dump(e,t,n,i,r)};e.load=function(e,t,n,i){return this.parseFile(e,t,n,i)};return e}();if(typeof window!=="undefined"&&window!==null){window.YAML=l}if(typeof window==="undefined"||window===null){this.YAML=l}t.exports=l},{"./Dumper":1,"./Parser":7,"./Utils":10}]},{},[11]);
```
rhafer: I don't think we want minified/compressed javascript in our git repositories. It's virtually impossible to debug if anything is wrong with this code.

We have rake tasks (assets:precompile) which are called as part of the package building which handle compression/minification.

rhafer: Also I think if we really need to add more 3rd party javascript code, it is probably better to add that via a separate commit, with a clear description in the commit message why it is needed and where it's coming from.

crowbar_framework/app/assets/javascripts/barclamps/ses/LICENSE
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
```
rhafer: Shouldn't this be part of the file to which this license applies? Otherwise it's not exactly obvious.

crowbar_framework/app/models/cinder_service.rb
```diff
@@ -35,7 +35,7 @@ def role_constraints
           "cluster" => true,
           "admin" => false,
           "exclude_platform" => {
-            "suse" => "< 12.4",
+            "suse" => "< 12.3",
```
rhafer: Why? This PR is against master (Cloud 9). Cloud 9 will use SLE12-SP4 AFAIK. (this applies to all the other instances where 12.4 is replaced with 12.3)

crowbar_framework/app/models/ses_service.rb
```diff
@@ -0,0 +1,45 @@
...
+    # it.
+    def self.allow_multiple_proposals?
+      false
+    end
+  end
+
+  def create_proposal
+    base = super
+
+    secret_uuid = `uuidgen`.strip
```
rhafer: I know we're doing this in a lot of other place, but I think we shouldn't add more of this.  Ruby has a method for generating uuid since quite a while:

```
require 'securerandom'
...

base["attributes"][@bc_name]["secret_uuid"] = SecureRandom.uuid
```

crowbar_framework/app/models/cinder_service.rb
```diff
@@ -84,6 +84,30 @@ def create_proposal
     base["attributes"][@bc_name][:db][:password] = random_password
     base["attributes"][@bc_name]["keymgr_fixed_key"] = random_password
 
+    ses_proposal = Proposal.find_by(barclamp: "ses")
+    if ses_proposal.nil?
+      @logger.debug("Cinder create_proposal: did NOT find ses barclamp")
+    else
+      @logger.debug("Cinder create_proposal: FOUND ses barclamp")
+      cinder_ses = ses_proposal["attributes"]["ses"]["cinder"]
```
rhafer: Is it guaranteed that the input yaml file always has settings for all openstack services? Or is it possible to just enable ceph of a subset of services? In the latter case we would need additional checks in all the other proposals.

crowbar_framework/app/assets/javascripts/barclamps/ses/yaml.min.js
hemna: The entire purpose of this barclamp is to support the phase 2 SES support as described in the spec here: https://jira.suse.de/browse/SCPM-63

  SES integration was already added to the ardana codebase for phase 2.  Phase 2 support is to consume the config yaml that SES generates and use that data to configure cinder/glance/nova for integration with the SES/ceph cluster.  I originally proposed that the changes made to the ses barclamp get automatically change the other proposals, but that was rejected.  

I can get the non minified yaml.js, and include it here, but it belongs with this patch because it's only used by this patch.  It doesn't make much sense to add it as a standalone request.

crowbar_framework/app/models/cinder_service.rb
hemna: will fix

crowbar_framework/app/models/cinder_service.rb
hemna: The input yaml comes from the SES openstack integration runner and will always have the same components in it.  It will always have the users, pools keyrings for cinder, glance, nova and radosgw urls.

crowbar_framework/app/assets/javascripts/barclamps/ses/yaml.js
```diff
@@ -0,0 +1,1893 @@
+(function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
```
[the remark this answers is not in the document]
hemna: This is the raw yaml.js from the upstream 3rd party.  Not changing the contents of this.

crowbar_framework/app/views/barclamp/ses/_edit_attributes.html.haml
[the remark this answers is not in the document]
hemna: id's shouldn't contain a - in them because they translate to JS variable names that can't have a -.

toabctl:
> It the above somewhat correct? Or did I get it completely wrong?

Good questions. @hemna could you answer theses please?

toabctl:
Also this needs a rebase

hemna:
> > It the above somewhat correct? Or did I get it completely wrong?
> 
> Good questions. @hemna could you answer theses please?

I did answer it here https://github.com/crowbar/crowbar-openstack/pull/1910#discussion_r238261785

toabctl:
> @hemna could you please rebase and squash so the commits make somehow sense?

@hemna could you please rebase and squash so the commits make somehow sense?

PR #1912, opened 2018-11-28
cinder: remove some deprecated drivers from cinder.conf

ilausuch:
Some drivers has removed from cinder.conf because of deprecation

From https://docs.openstack.org/releasenotes/cinder/rocky.html
- HGST
- Disco 

From https://docs.openstack.org/releasenotes/cinder/pike.html
- Blockbridge
- Itachi Block Storage Driver (HBSD)
- Coho
- Infortrend
- Tegile

### DIFF
--- a/chef/cookbooks/cinder/templates/default/cinder.conf.erb
+++ b/chef/cookbooks/cinder/templates/default/cinder.conf.erb
@@ -59,15 +59,11 @@ strict_ssh_host_key_policy = <%= @strict_ssh_host_key_policy ? 'true' : 'false' 
 [<%= backend_id %>]
 volume_backend_name = <%= volume['backend_name'] %>
 
-  <% if volume['backend_driver'] == 'blockbridge' -%><% end -%>
-  <% if volume['backend_driver'] == 'coho' -%><% end -%>
-
   <% if volume['backend_driver'] == 'datera' -%>
     volume_driver = cinder.volume.drivers.datera.datera_iscsi.DateraDriver
   <% end -%>
 
   <% if volume['backend_driver'] == 'dell' -%><% end -%>
-  <% if volume['backend_driver'] == 'disco' -%><% end -%>
   <% if volume['backend_driver'] == 'drbdmanage' -%><% end -%>
 
   <% if volume['backend_driver'] == 'emc' -%>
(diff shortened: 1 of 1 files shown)


chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -59,15 +59,14 @@ strict_ssh_host_key_policy = <%= @strict_ssh_host_key_policy ? 'true' : 'false'
...
-  <% if volume['backend_driver'] == 'coho' -%><% end -%>
-
   <% if volume['backend_driver'] == 'datera' -%>
     volume_driver = cinder.volume.drivers.datera.datera_iscsi.DateraDriver
   <% end -%>
 
   <% if volume['backend_driver'] == 'dell' -%><% end -%>
-  <% if volume['backend_driver'] == 'disco' -%><% end -%>
+  <% if volume['backend_driver'] == 'disco' -%>
+enable_unsupported_driver = True
```
toabctl: - Why are you setting `enable_unsupported_driver` here but not for others like `blockbridge` or `coho`? 
- In general I would not set `enable_unsupported_driver` anywhere

ilausuch: Yes, you are right, this not have consistency in this way. I maintained them in case of this drivers are still used, but you are right, they are deprecated. So, let's remove them

chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -133,19 +127,13 @@ hitachi_add_chap_user = <%= volume['hitachi']['hitachi_add_chap_user'] %>
 hitachi_auth_method = <%= volume['hitachi']['hitachi_auth_method'] %>
 hitachi_auth_user = <%= volume['hitachi']['hitachi_auth_user'] %>
 hitachi_auth_password = <%= volume['hitachi']['hitachi_auth_password'] %>
-    <% if volume['hitachi']['hitachi_storage_protocol'] == 'iscsi' -%>
```
toabctl: The hitachi drivers are now out-of-tree[1]. @dirkmueller @vuntz do we still need support here? Is somebody using hitachi as backend?

[1] http://lists.openstack.org/pipermail/openstack/2017-March/018812.html

jgrassler: I just had a chat with @vuntz and it's ok to drop the driver if it's external. We just need to make sure we document it in the release notes.

PR #1913, opened 2018-11-28
nova: Register compute node in nova cell

toabctl:
We need to register compute nodes in nova cells. Otherwise the compute
node is not ready for scheduling VMs on it.

The current setup is racy because we rely on the
[scheduler]"discover_hosts_in_cells_interval" variable (which is set
to 600s currently). So if you add a new compute node, you have to wait
up to 600s until the node is usable.

This problem occured in the CI and also on local deployments. The
nova-conductor.log look like:

nova-conductor.log:2018-11-28 15:26:52.096 29009 ERROR
  nova.conductor.manager No host-to-cell mapping found for selected
  host d52-54-77-77-01-02. Setup is incomplete.: HostMappingNotFound:
  Host 'd52-54-77-77-01-02' is not mapped to any cell

and a bit later the nova-scheduler.log indicates that the host got
discovered:

nova-scheduler.log:2018-11-28 15:27:30.816 28987 INFO
  nova.scheduler.manager Discovered 1 new hosts: cell1:d52-54-77-77-01-02

### DIFF
--- a/chef/cookbooks/nova/recipes/compute.rb
+++ b/chef/cookbooks/nova/recipes/compute.rb
@@ -461,3 +461,5 @@ if node[:nova][:resource_limits] && node[:nova][:resource_limits][service]
     action action
   end
 end
+
+include_recipe "nova::compute_register_cell"
--- a/chef/cookbooks/nova/recipes/compute_register_cell.rb
+++ b/chef/cookbooks/nova/recipes/compute_register_cell.rb
@@ -1,0 +1,52 @@
+#
+# Cookbook Name:: nova
+# Recipe:: compute_register_cell
+#
+# Copyright 2018, SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+return if node[:nova][:cellv2_discover_hosts_called]
+
+db_settings = fetch_database_settings
+include_recipe "database::client"
+include_recipe "#{db_settings[:backend_name]}::client"
+include_recipe "#{db_settings[:backend_name]}::python-client"
+
+api_database_connection = fetch_database_connection_string(node[:nova][:api_db])
+
+# create a tmp file with the api-database credentials, register the
+# host and delete the creds
+bash "nova-manage discover_hosts" do
+  user node[:nova][:user]
+  code <<-EOH
+    tmpfile=$(mktemp /tmp/nova-discover-hosts.XXXXXX.conf)
+    chmod 600 $tmpfile
+    echo "[api_database]" >> $tmpfile
+    echo "connection = #{api_database_connection}" >> $tmpfile
+    nova-manage --config-file=$tmpfile cell_v2 discover_hosts --verbose
+    rm -f "$tmpfile"
+    EOH
+end
+
+# We want to keep a note that we've done discover_hosts, so we don't do it again.
+# If we were doing that outside a ruby_block, we would add the note in the
+# compile phase, before the actual discover_hosts is done (which is wrong, since it
+# could possibly not be reached in case of errors).
+ruby_block "mark node for cell_v2 discover_hosts" do
+  block do
+    node.set[:nova][:cellv2_discover_hosts_called] = true
+    node.save
+  end
+end


jsuchome:
Do you really want to go with _discoverd_ ? (I admit it is consistent in your change)

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -461,3 +461,21 @@
     action action
   end
 end
+
+# register the compute in a cell
+execute "nova-manage discover_hosts" do
+  command "nova-manage cell_v2 discover_hosts --verbose"
+  only_if !node[:nova][:cellv2_discoverd_hosts_called]
```
jgrassler: Shouldn't this be "discover_hosts_called" rather than "discoverd_hosts_called"? Looks like a typo to me - it's the same everywhere though.

chef/cookbooks/nova/recipes/compute.rb
```diff
@@ -461,3 +461,21 @@
     action action
   end
 end
+
+# register the compute in a cell
+execute "nova-manage discover_hosts" do
```
jgrassler: Shouldn't this run on a controller rather than on a compute node?

toabctl: There is no way to wait on the controller until all computes are done.

chef/cookbooks/nova/recipes/compute.rb
toabctl: Done

toabctl:
> Do you really want to go with _discoverd_ ? (I admit it is consistent in your change)

Fixed

chef/cookbooks/nova/recipes/compute_register_cell.rb
```diff
@@ -0,0 +1,50 @@
...
+include_recipe "#{db_settings[:backend_name]}::client"
+include_recipe "#{db_settings[:backend_name]}::python-client"
+
+api_database_connection = fetch_database_connection_string(node[:nova][:api_db])
+
+# create a tmp file with the api-database credentials, register the
+# host and delete the creds
+bash "nova-manage discover_hosts" do
+  code <<-EOH
+    tmpfile=$(mktemp /tmp/nova-discover-hosts.XXXXXX.conf)
```
jgrassler: Hmm. That's one way (and less complex than the sync mark approach I mused about earlier) to ensure we don't keep the credentials on the compute node forever. I guess I can live with that :-)

sjamgade: https://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html#method-c-open

ruby has a nice tempfile package which make sure clean up is done.
which might be the case if nova-manage fails to obey

chef/cookbooks/nova/recipes/config.rb
```diff
@@ -62,12 +62,14 @@
...
   database_connection = fetch_database_connection_string(node[:nova][:db])
   placement_database_connection = fetch_database_connection_string(node[:nova][:placement_db])
-  api_database_connection = fetch_database_connection_string(node[:nova][:api_db])
 else
   database_connection = nil
-  api_database_connection = nil
 end
 
+ # we need the api_database connection on the computes to be able to call
+ # nova-manage to register a new compute node in the nova cellv2
```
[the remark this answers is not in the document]
dirkmueller: @toabctl could you fix that?

chef/cookbooks/nova/recipes/compute_register_cell.rb
```diff
@@ -0,0 +1,50 @@
...
+    echo "[api_database]" >> $tmpfile
+    echo "connection = #{api_database_connection}" >> $tmpfile
+    nova-manage --config-file=$tmpfile cell_v2 discover_hosts --verbose
+    rm "$tmpfile"
```
dirkmueller: potentially we need -f here?

chef/cookbooks/nova/recipes/compute_register_cell.rb
```diff
@@ -0,0 +1,50 @@
...
+db_settings = fetch_database_settings
+include_recipe "database::client"
+include_recipe "#{db_settings[:backend_name]}::client"
+include_recipe "#{db_settings[:backend_name]}::python-client"
+
+api_database_connection = fetch_database_connection_string(node[:nova][:api_db])
+
+# create a tmp file with the api-database credentials, register the
+# host and delete the creds
+bash "nova-manage discover_hosts" do
```
dirkmueller: should this really be run as root? I think it would be safer to run it as the nova user

chef/cookbooks/nova/recipes/compute_register_cell.rb
```diff
@@ -0,0 +1,50 @@
...
+  code <<-EOH
+    tmpfile=$(mktemp /tmp/nova-discover-hosts.XXXXXX.conf)
+    echo "[api_database]" >> $tmpfile
+    echo "connection = #{api_database_connection}" >> $tmpfile
```
cmurphy: There are no permissions set on this file so by default it will be world readable, which is not good for password files, even if it's only going to exist for a short time.

chef/cookbooks/nova/recipes/config.rb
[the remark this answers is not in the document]
toabctl: @dirkmueller This is outdated and was already fixed

chef/cookbooks/nova/recipes/compute_register_cell.rb
toabctl: done

chef/cookbooks/nova/recipes/compute_register_cell.rb
toabctl: done

chef/cookbooks/nova/recipes/compute_register_cell.rb
toabctl: done

sjamgade:
@toabctl is there something obviously wrong with `tempfile` gem.  It does everything that is being done in that script. And chef itself uses it for writing templates.

toabctl:
> @toabctl is there something obviously wrong with `tempfile` gem. It does everything that is being done in that script. And chef itself uses it for writing templates.

@sjamgade  I would have to use a `ruby_block` to create the template and inside of the block, call a shell command (`nova-manage`).

PR #1914, opened 2018-11-29
galera: Use monitoring user for observing wsrep state

dirkmueller:
Using the root user with empty password has the downside
that once the root user no longer has an empty password, it
can no longer observe the state. The monitoring user
is more useful for that as it is always available, even
when the cluster was already bootstrapped before. This
is important in case you want to add new nodes to an existing
cluster.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -243,7 +243,7 @@ ruby_block "wait galera bootstrap" do
   block do
     require "timeout"
     begin
-      cmd = "mysql -u '' -N -B " \
+      cmd = "mysql -u 'monitoring' -N -B " \
         "-e \"SHOW STATUS WHERE Variable_name='wsrep_local_state_comment';\" | cut -f 2"
       sync_state = ""
       Timeout.timeout(seconds) do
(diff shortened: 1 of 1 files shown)


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -23,24 +23,23 @@
...
 
-unless node[:database][:galera_bootstrapped]
-  directory "/var/run/mysql/" do
-    owner "mysql"
-    group "root"
-    mode "0755"
-    action :create
-  end
-
-  execute "mysql_install_db" do
```
rhafer: AFAIK this will create the "/var/lib/mysql/" directory. MySQL will not start on the non-founder nodes without that. So I think you at least need to create the directory here. Or just leave out the 2nd commit.

dirkmueller: Leaving out the 2nd commit is not an option, because mysql_install_db fails if the directory is already existing and non-empty (which can happen if it was meanwhile populated via  a previous start of the pacemaker resource doing a SST). This was the reason for reworking this code in the first place. 

Creating the directory manually seems to circumvent the issue though.

dirkmueller:
@rhafer the code has been tested on crowbar.qa2.cloud.suse.de and on engcloud.prv.suse.net (in production)

PR #1916, opened 2018-11-30
neutron: removed deprecation settings

ilausuch:
Removed deprecated settings from the configuration

- sql/min_pool_size
Minimum number of SQL connections to keep open in a pool.
https://github.com/openstack/oslo.db/blob/ab1c2b9076dc4fef9e3b7d880a69099172a5bf27/oslo_db/options.py#L70

- ovs/ovsdb_interface
The interface for interacting with the OVSDB
https://docs.openstack.org/releasenotes/neutron/queens.html

### DIFF
--- a/chef/data_bags/crowbar/migrate/neutron/303_remove_rocky_deprecations.rb
+++ b/chef/data_bags/crowbar/migrate/neutron/303_remove_rocky_deprecations.rb
@@ -1,0 +1,11 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["sql"].delete("min_pool_size")
+  attrs["ovs"].delete("ovsdb_interface")
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["sql"]["min_pool_size"] = template_attrs["sql"]["min_pool_size"]
+  attrs["ovs"]["ovsdb_interface"] = template_attrs["ovs"]["ovsdb_interface"]
+  return attrs, deployment
+end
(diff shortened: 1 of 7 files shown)


chef/data_bags/crowbar/migrate/neutron/303_remove_rocky_deprecations.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(tattr, tdep, attr, dep)
+  attr["sql"].delete("min_pool_size")
+  attr["ovs"].delete("ovsdb_interface")
+  return attr, dep
+end
+
+def downgrade(tattr, tdep, attr, dep)
+  attr["sql"]["verbose"] = tattr["sql"]["verbose"]
```
cmurphy: Looks like a copy paste error? should be min_pool_size not verbose

chef/data_bags/crowbar/migrate/neutron/303_remove_rocky_deprecations.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(tattr, tdep, attr, dep)
```
cmurphy: (nitpick) would prefer if the variable names were consistent with the noop template https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/migrate/neutron/300_noop.rb

ilausuch: Moving it to the standards, Thank you @cmurphy

chef/data_bags/crowbar/migrate/neutron/303_remove_rocky_deprecations.rb
ilausuch: Yes, indeed

PR #1917, opened 2018-11-30
cinder: removed live migrations code

ilausuch:
Due to a recent change in cinder, is considered a failure
if the command cinder-manage db online_data_migrations
exits code to 2

docs.openstack.org/releasenotes/cinder/rocky.html

The current tests include live migration and it generates
a failure (error code 2) in ci. So to avoid this problem in a fresh
installation all the migration code is removed. But it should
take in consideration when the cloud is upgraded.

### DIFF
--- a/chef/cookbooks/cinder/recipes/sql.rb
+++ b/chef/cookbooks/cinder/recipes/sql.rb
@@ -61,36 +61,6 @@ database_user "grant database access for cinder database user" do
     only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
-# Some live migrations need to happen before revision 97, so we first need to
-# sync all revisions up to 96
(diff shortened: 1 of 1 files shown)


jsuchome:
Hm, I'm not sure if for master code we should actually have split those cinder-manage migrate calls to the specific versions (`< 97` vs `=96`). I think we did it mostly for the upgrade. For normal deployment, I think the code could be reduced. @matthewoliver what do you think?

ilausuch:
In previous commit "ignore_failure false" was set and it generates a problem in the CI because the command "cinder-manage db online_data_migrations" fails with exit code 2 considered a failure (docs.openstack.org/releasenotes/cinder/rocky.html)

Now all the migration code is removed in fresh installations. But it should be considered during the upgrade process. Do you agree?

PR #1918, opened 2018-11-30
 horizon: remove option OPENSTACK_TOKEN_HASH_ENABLED (SCRD-5906) 

sjamgade:
this settings was dropped in rocky

https://docs.openstack.org/releasenotes/horizon/rocky.html#upgrade-notes


chef/data_bags/crowbar/migrate/horizon/301_remove_token_hash_enanled_attribute.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  a.delete("token_hash_enabled")
```
toabctl: where does `a` come from? Needs to be `attrs.delete()`

sjamgade: thanks !

PR #1919, opened 2018-12-03
keystone: Fix SSL

cmurphy:
SSL is broken in all kinds of ways. This pull request fixes each of them.


chef/cookbooks/keystone/providers/register.rb
```diff
@@ -317,6 +316,7 @@
       end
     end
     ["public", "internal", "admin"].each do |interface|
+      path = "/v3/endpoints"
```
rhafer: Instead of overwriting `path` here, I think it would be cleaner to rename it in the assignment on line 333 like:

`fullpath = "#{path}/#{endpoints[interface]["id"]}"`

And change the `_update_item` call accordingly.

cmurphy: :+1:

chef/cookbooks/keystone/recipes/update_endpoint.rb
```diff
@@ -0,0 +1,122 @@
...
+  endpoint_insecure = node[:keystone][:endpoint][:insecure]
+  # In order to update keystone's endpoints we need the old internal endpoint.
+  endpoint_port = node[:keystone][:endpoint][:port]
+  endpoint_needs_update =
+    node[:keystone][:bootstrap] &&
+    # Do not try to update keystone endpoint during upgrade, when keystone is not
+    # running yet ("done_os_upgrade" is present when first chef-client run is
+    # executed at the end of upgrade)
+    node["crowbar_upgrade_step"] != "done_os_upgrade" &&
+    node[:keystone].key?(:endpoint) &&
```
rhafer: This is not needed as we are already in the `if node[:keystone].key?(:endpoint)` branch

chef/cookbooks/keystone/recipes/update_endpoint.rb
```diff
@@ -0,0 +1,122 @@
...
+    (
+      endpoint_protocol != node[:keystone][:api][:protocol] ||
+      endpoint_insecure != node[:keystone][:ssl][:insecure] ||
+      endpoint_port != node[:keystone][:api][:admin_port]
+    )
```
rhafer: It might be just me but isn't the whole "endpoint_needs_update" condition a little to long to be in a single statement? :-)

It might be better digestible if split into multiple statements. Something like:

```
endpoint_changed = endpoint_protocol != node[:keystone][:api][:protocol] ||
  endpoint_insecure != node[:keystone][:ssl][:insecure] ||
  endpoint_port != node[:keystone][:api][:admin_port]

endpoint_needs_update = endpoint_changed && 
  node[:keystone][:bootstrap] &&
  node["crowbar_upgrade_step"] != "done_os_upgrade"
```
I know this is basically a copy from the old code, but might still be worth fixing.

cmurphy: Agreed, can do

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -153,9 +153,7 @@
     ssl_enable node[:keystone][:api][:protocol] == "https"
     ssl_certfile node[:keystone][:ssl][:certfile]
     ssl_keyfile node[:keystone][:ssl][:keyfile]
-    if node[:keystone][:ssl][:cert_required]
-      ssl_cacert node[:keystone][:ssl][:ca_certs]
-    end
```
cmurphy: This actually causes an apache syntax error when crowbar generates self-signed certs because the default CA file doesn't exist.

chef/cookbooks/cinder/recipes/common.rb
```diff
@@ -84,7 +84,7 @@
 
 node.save if dirty
 
-if node[:cinder][:api][:protocol] == "https"
+if node[:cinder][:api][:protocol] == "https" && node["roles"].include?("cinder-controller")
```
dirkmueller: can we move this code block to one of the recipes only executed on controller or is that an issue for some reason?

cmurphy:
@dirkmueller proposed in https://github.com/crowbar/crowbar-openstack/pull/1931 for master. The broken code isn't in stable/5.0-pike for some reason.

chef/cookbooks/cinder/recipes/common.rb
cmurphy: Done

cmurphy: I can't see any reason it would be an issue but it's unclear why all this stuff was crammed into cinder::common in the first place

cmurphy:
https://github.com/SUSE-Cloud/automation/pull/3007 demonstrates the bug this fixes (https://ci.suse.de/job/openstack-mkcloud/147974/parsed_console/) and that this pull request corrects the problem (https://ci.suse.de/job/openstack-mkcloud/148068/console)

chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -138,6 +138,7 @@ pool_timeout = <%= node[:nova][:db][:pool_timeout] %>
 host = <%= @glance_server_host %>
 port = <%= @glance_server_port %>
 api_servers = <%= "#{@glance_server_protocol}://#{@glance_server_host}:#{@glance_server_port}" %>
+endpoint_type = internal
```
bmwiedemann: Wouldnt this change break the usual case, where you have valid SSL certs only for the public endpoint?
Because the internal mac-addr-based names were so unpredictable, those often were not covered by the cert. Or can you override the internal name now?

cmurphy: Users will have to have certs that are valid for both the internal and public endpoints or not use ssl on the internal endpoint, we should only ever expect services to communicate with each other on an internal interface. We already set this for other services.

PR #1920, opened 2018-12-03
Glance: deprecate glance registry service.(SCRD-4087)

kwu83tw:
Deprecated in Queens. Removed glance registry configurations.
https://docs.openstack.org/releasenotes/glance/queens.html#deprecation-notes


sjamgade:
we need to get rid of glance-cache also ? the cache prefetches the images via registry http://git.openstack.org/cgit/openstack/glance/tree/glance/image_cache/prefetcher.py?h=stable/rocky#n37

kwu83tw:
> we need to get rid of glance-cache also ? the cache prefetches the images via registry http://git.openstack.org/cgit/openstack/glance/tree/glance/image_cache/prefetcher.py?h=stable/rocky#n37

According to https://github.com/openstack/glance/blob/master/glance/image_cache/__init__.py#L347, *prefetcher* is the one to perform the caching action. However, iiuc we're not using *glance image cache* functionality since we don't enable any option under glance-cache.conf to get *prefetcher* working. Hence, the *prefetcher* is using registry v1 api and *enable_v2_registry* is true since Jan, 2018(8be169b94f69e0d48300988c214feb9f207e8dc6) in our current setup. It looks like it won't bring much impact on removing glance-cache.

chef/data_bags/crowbar/migrate/glance/301_remove_registry.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs.delete("registry")
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["registry"]["bind_port"] = 9191
```
cmurphy: Instead of this you could to

```
attrs["registry"] = template_attrs["registry"]

File: chef/data_bags/crowbar/template-glance.json [hunk 2]
```diff
@@ -18,9 +18,6 @@
         "bind_open_address": true,
         "bind_port": 9292
       },
-      "registry": {
```
cmurphy: Could also fix the description on line 3

chef/data_bags/crowbar/migrate/glance/301_remove_registry.rb
kwu83tw: Will that still be valid even though I've removed "registry" in template-glance.json? Isn't "template_attrs" value retrieving from the databag JSON?

PR #1921, opened 2018-12-04
tempest: configure microversions

jgrassler:
Some Tempest tests have microversion filters these days. For
these to work a minimum and maximum microversion needs to be
defined for services that do implement microversioning. This
commit adds these two values for all services that currently
implement microversioned testing.

### DIFF
--- a/chef/cookbooks/tempest/templates/default/tempest.conf.erb
+++ b/chef/cookbooks/tempest/templates/default/tempest.conf.erb
@@ -25,6 +25,7 @@ aws_access = <%= @ec2_access %>
 
 [baremetal]
 endpoint_type = internalURL
+max_microversion = latest
 
 [compute]
 image_ref = <%= img_id %>
@@ -35,6 +36,7 @@ fixed_network_name = fixed
 region = <%= @keystone_settings['endpoint_region'] %>
 endpoint_type = internalURL
 min_compute_nodes = <%= @use_livemigration ? 2 : 1 %>
+max_microversion = latest
 
 [compute-feature-enabled]
 resize = <%= @use_resize %>
@@ -181,6 +183,7 @@ image_with_share_tools = <%= @manila_settings['image_with_share_tools'] %>
 image_username = <%= @manila_settings['image_username'] %>
 image_password = <%= @manila_settings['image_password'] %>
 default_share_type_name = <%= @manila_settings['default_share_type_name'] %>
+max_microversion = latest
 
 [telemetry]
 endpoint_type = internalURL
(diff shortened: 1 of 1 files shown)


chef/cookbooks/tempest/templates/default/tempest.conf.erb
```diff
@@ -25,6 +25,8 @@ aws_access = <%= @ec2_access %>
 
 [baremetal]
 endpoint_type = internalURL
+min_microversion = 1.0
```
cmurphy: Are you sure a min needs to be set? Looking at http://git.openstack.org/cgit/openstack/tempest/tree/tempest/lib/common/api_version_request.py it looks like it would be enough to set min_microversion = None or leave it unset, and just set max_microversion = latest. The example in http://git.openstack.org/cgit/openstack/tempest/tree/doc/source/microversion_testing.rst gives that impression too.

kbaikov: Do you specifically mean baremetal? Because i tested compute and the following did not work: None, 0.1, 1.1. Tests intelligently complained about needing 2.1 as minimum.

cmurphy: This is for ironic. It looks like the minimum is 1.1 not 1.0 http://git.openstack.org/cgit/openstack/ironic/tree/ironic/api/controllers/v1/versions.py?h=stable/rocky#n30

jgrassler: Hmm...I missed the "not supported since Juno" bit (and `versions.py` for that matter). I used 1.0 since that was the earliest version there was a test around for, this one to be precise: http://git.openstack.org/cgit/openstack/ironic/tree/ironic/tests/unit/api/test_hooks.py?h=stable/rocky#n67

Do you want me to change it to 1.1?

kbaikov: Oh, let me try running ironic tests only and see if it complains.

kbaikov: Ah, i don't have ironic tests yet :)
But they will be broken based on you link i guess.

cmurphy: [**@jgrassler**](https://github.com/jgrassler) commented on [Dec 5, 2018, 4:34 PM GMT+1](https://github.com/crowbar/crowbar-openstack/pull/1921#discussion_r239114561 "2018-12-05T15:34:25Z - Replied by Github Reply Comments"):

> 
> Do you want me to change it to 1.1?

No, I'd rather it be unset or `= None` if that is valid, which I think it.

kbaikov: So i retested with None and i get this:     `tempest.lib.exceptions.InvalidAPIVersionString: API Version String None is of invalid format. Must be of format MajorNum.MinorNum or string 'latest'.`



However if i delete min version line completely and leave only max_microversion = latest then the test works

jgrassler: Ok, I removed min_microversion across the board now.

PR #1922, opened 2018-12-06
rabbitmq: raise credit flow settings

dirkmueller:
We don't want to apply flow control to mirroring and the
defaults seem to be somewhat low for certain production
cases. Lets raise them.


chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
```diff
@@ -36,6 +36,8 @@
 <% if @cluster_enabled -%>
    {cluster_partition_handling, <%= @cluster_partition_handling %>},
    {queue_master_locator, <<"min-masters">>},
+   {mirroring_flow_control,false},
```
cmurphy: I can't find any good documentation on this, have a link?

All I can find is this: https://github.com/rabbitmq/rabbitmq-server/blob/v3.8.0-beta.1/docs/rabbitmq.config.example#L340-L345

Which is not very comforting. Can you explain more why it is okay to disable this?

chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
```diff
@@ -36,6 +36,8 @@
...
+   {credit_flow_default_credit,{800,400}},
```
cmurphy: Maybe this should be configurable instead of hardcoded? Reading this: 
https://www.rabbitmq.com/blog/2015/10/06/new-credit-flow-settings-on-rabbitmq-3-5-5/
They recommend 

> increasing these values just to see what happens can have adverse effects on how RabbitMQ is able to respond to message bursts, affecting the internal strategies that RabbitMQ has in order to deal with memory pressure. Handle with care.

Not all users have the same memory allocations on nodes so raising this so high might be dangerous for some.

aojea:
@Gsantomaggio can you give us some clarification about the `mirroring_flow_control` parameter?

chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
Gsantomaggio: Tuning the credit flow depends on the relation between channels and connections,

Here is an issue where we worked about the credit_flow tuning: 
https://github.com/rabbitmq/rabbitmq-server/issues/964 

In particular this comment:
https://github.com/rabbitmq/rabbitmq-server/issues/964#issuecomment-253507209 

Increase this value could be dangerous, it depends how many connections you have and how many channels for a connection you have.

Is it possible to have the output from:

```
rabbitmqctl list_connections pid name state channels
```

Gsantomaggio: `rabbitmqctl list_connections pid name state channels`  if you don't have an output, 
it would be useful to know how many channels there are for connection, more or less.

chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
dirkmueller: @cmurphy I don't have an explanation, I just wanted to see how the CI likes it.

chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
Gsantomaggio: @dirkmueller yes, these parameters should be changed/tested directly on the installation

Gsantomaggio:
I think we can close this PR

dirkmueller:
@Gsantomaggio could you elaborate? there isn't a supported way other than changing the template to modify them per installation afaik. are you saying there is a way to add local overrides ? if so, could you explain how that would work?

Gsantomaggio:
Credit Flow tuning should be done directly to the installation. At this moment I have no idea about the deploy environment and the system numbers ( throughput, number of connections, number of the queues etc).

There is a way to change the rabbitmq environment variables using `eval`, for example:
if you have:
```
rabbitmqctl environment | grep flow

          {credit_flow_default_credit,{400,200}},
```  

you can do that:

``` 
 scripts git:(v3.6.x) ✗ . ./rabbitmqctl eval "application:set_env(rabbit, credit_flow_default_credit,{500, 300})."
ok
➜  scripts git:(v3.6.x) ✗ ./rabbitmqctl environment | grep flow

          {credit_flow_default_credit,{500,300}},
``` 

Note: This is not persistent. 
You should not change the parameters in this way, the config file is the best way, but it is useful to tests.


`{credit_flow_default_credit,{800,400}},` are  not a big values, but we should try them before change. 
Also depends on what kind of the problem are you trying to solve.

dirkmueller:
@Gsantomaggio so we need to enable some way for users to override the defaults in a persistent fashion. this PR is a start of that.

Gsantomaggio:
We can merge it  ( after internal conversation )

PR #1926, opened 2018-12-07
Drop "clone_stateless_services" feature

rhafer:
With 5.0 we switch the default of "clone_stateless_services" to false to
have stateless services be managed (restart on failure) by systemd
instead of creating pacemaker resources. When upgrading, existing clouds
will also switch to this configuraton. This patch removes the use of the
"clone_stateless_services" attribute and makes systemd approach the only
supported setup.


toabctl:
@rhafer needs rebase.

chef/cookbooks/cinder/recipes/api.rb
```diff
@@ -159,10 +159,8 @@
 
 crowbar_pacemaker_sync_mark "create-cinder_register"
 
-use_crowbar_pacemaker_service = ha_enabled && node[:pacemaker][:clone_stateless_services]
-
 cinder_service "api" do
-  use_pacemaker_provider use_crowbar_pacemaker_service
+  use_pacemaker_provider false
```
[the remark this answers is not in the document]
rhafer: Done

rhafer:
rebased

chef/cookbooks/horizon/recipes/ha.rb
```diff
@@ -48,7 +48,3 @@
 crowbar_pacemaker_sync_mark "sync-horizon_before_ha" do
   timeout 150
 end
```
vuntz: You can drop that sync mark.

chef/cookbooks/neutron/definitions/neutron_metadata.rb
```diff
@@ -84,14 +84,11 @@
   end
 
   unless use_cisco_apic_ml2_driver
-    use_crowbar_pacemaker_service = \
-      (neutron_network_ha && node[:pacemaker][:clone_stateless_services]) || nova_compute_ha_enabled
-
```
vuntz: Hrm. Are you sure about this one? I think it's wrong for the compute HA case.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -128,117 +128,8 @@
...
-    # note that this will also automatically start the resources
-    action :commit_new
-    only_if { CrowbarPacemakerHelper.is_cluster_founder?(node) }
-  end
-else
-  l3_agent_primitive = "neutron-l3-agent"
-  l3_agent_clone = "cl-#{l3_agent_primitive}"
-end
+l3_agent_primitive = "neutron-l3-agent"
+l3_agent_clone = "cl-#{l3_agent_primitive}"
```
vuntz: You could drop that, to be honest. This requires updating the `crowbar_pacemaker_order_only_existing` bit for neutron-ha-tool-service (to not reference l3_agent_clone), but that's about it.

And btw, since you're touching this file, maybe you could drop the "remove old..." bits for neutron-ha-tool -- in Cloud 9, it's not needed anymore.

chef/cookbooks/neutron/definitions/neutron_metadata.rb
rhafer: Yeah, I guess the same is true for l3-agent and the layer2 agents. Should be fixed now.

chef/cookbooks/horizon/recipes/ha.rb
rhafer: Done

chef/cookbooks/neutron/recipes/network_agents_ha.rb
rhafer: Done

PR #1928, opened 2018-12-11
monasca: Prevent deploying monasca-server to the node in pacemaker cluster

jsuchome:
Pacemaker could fence a cluster node for some reason and having
monasca-server running there would mean monasca going down (bsc#1118759)


jsuchome:
@dirkmueller it is in both places actually

PR #1930, opened 2018-12-11
[4.0] galera bootstrap fixes

dirkmueller:


jsuchome:
just one commit is galera related, right?

skazi0:
@cmurphy check Ralf's comment here: https://bugzilla.suse.com/show_bug.cgi?id=1116686#c8 for another reason why database fails recently

PR #1933, opened 2018-12-14
mariadb: Set wsrep_sst_method to mariabackup (bsc#1116686)

rsalevsky:
MariaDB 10.2.16 added a new option innodb_safe_truncate which requies a
the switch to mariabackup which can handle crash-safe rename operations.


rsalevsky:
No backport needed we changed a config option in the package for SOC7 and SOC8. It's too risky to switch in production.

jsuchome:
Ah, OK then

PR #1935, opened 2018-12-14
nova-compute: restart libvirtd on ceph.conf changes

dirkmueller:


toabctl:
A *why* in the commit msg would be useful at least :)

PR #1937, opened 2018-12-16
Fix restart order of nova services for mixed envs

dirkmueller:
We need to start conductor first so that it can be used in arbitration of the RPC version level. Start API last so that API is only available when the rest of the stack is online.


chef/cookbooks/nova/templates/default/nova.conf.erb
```diff
@@ -329,6 +329,11 @@ ssl_cert_file = <%= @ssl_cert_file %>
 ssl_key_file = <%= @ssl_key_file %>
 keep_alive = false
 
+[upgrade_levels]
+# Determine RPC protocol version for nova based on the minimum
+# of all services registered in nova service database.
+compute=auto
```
jsuchome: This is the default, isn't it?

dirkmueller: the default is "None", which reflects "the current version (aka pike/rocky)". we want it to default to the minimum of all agents I think.

PR #1938, opened 2018-12-17
upgrade: Run nova online-db-migrations after the ugprade is complete

jsuchome:


chef/cookbooks/nova/recipes/database.rb
```diff
@@ -159,7 +159,9 @@
   ignore_failure true
   action :run
   only_if do
-    !node[:nova][:db_synced] &&
+    !node[:nova][:online_db_migrations_executed] &&
+      # Do not executed online_data_migrations if the upgrade is incomplete
```
skazi0: "execute"?

PR #1941, opened 2018-12-17
[5.0] nova: use upstream-recommended start order for services

dirkmueller:
Upstream highly recommends to start conductor first and api last.
Adjust our starting order accordingly. See

https://docs.openstack.org/nova/pike/user/upgrade.html
(cherry picked from commit 25de7ab5c4e0c6b6ee3fab84591c9e1723b1c1f1)


jsuchome:
What about the 2nd commit?

dirkmueller:
@jsuchome added it now as well (after testing it some more)

PR #1947, opened 2018-12-18
[5.0] mariadb: Set wsrep_sst_method to mariabackup (bsc#1116686)

rsalevsky:
MariaDB 10.2.16 added a new option innodb_safe_truncate which requies a
the switch to mariabackup which can handle crash-safe rename operations.

(cherry picked from commit cf6018758eca6af67f3f3c2c44c895608a0ae516)

### DIFF
--- a/chef/cookbooks/mysql/templates/default/galera.cnf.erb
+++ b/chef/cookbooks/mysql/templates/default/galera.cnf.erb
@@ -32,5 +32,5 @@ query_cache_type = 0
 expire_logs_days = <%= @expire_logs_days %>
 
 # SST method
-wsrep_sst_method = xtrabackup-v2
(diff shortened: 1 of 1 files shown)


rsalevsky:
It seams disabling innodb_safe_truncate is not enough and just as a fallback we should also do the sst switch.

rhafer:
>  Did you actually check if deploying this config change on a running system does not break the cluster (and that SST still works after that?)

I just have this a brief test. And it seems to work. Even in a mixed setup where some nodes where configured to still use xtrabackup for SST. The node requesting the SST will use what ever tool it has set in wsrep_sst_method.  Which is good as not all nodes will get the config change at the same time.

PR #1950, opened 2018-12-18
neutron: scale up api workers on larger core envs

dirkmueller:
The current max of 4 is by far too low for Cloud8, we need to raise it.
In order to stay the same at 8 cores, try to use like 1/2 of the
available cores so that the rest remains available when nova or cinder
is colocated (which are the others using a lot of CPU)


dirkmueller:
@aojea updated capping to 16.

toabctl:
Backport is https://github.com/crowbar/crowbar-openstack/pull/1952

PR #1951, opened 2018-12-20
upgrade: reduce keystone downtime

rhafer:
This introduces the needed changes in the keystone cookbook for reducing its downtime during the upgrade.  Needs to get merged together with https://github.com/crowbar/crowbar-core/pull/1748

### DIFF
--- a/crowbar_framework/lib/openstack/upgrade.rb
+++ b/crowbar_framework/lib/openstack/upgrade.rb
@@ -33,7 +33,7 @@ module Openstack
       # this point the nodes maybe don't have roles assigned anymore
       components = [
         :ceilometer, :cinder, :glance, :heat,
-        :keystone, :manila, :neutron, :nova
+        :manila, :neutron, :nova
       ]
       NodeObject.all.each do |node|
         save_it = false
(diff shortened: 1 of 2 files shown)


crowbar_framework/lib/openstack/upgrade.rb
```diff
@@ -33,11 +33,24 @@ def self.unset_db_synced
...
       NodeObject.all.each do |node|
         save_it = false
-        components.each do |component|
+        # When the setup was deploying with "clone_stateless_services=false" we
+        # keep keystone running for some longer and don't what the chef-recipes
+        # to trigger the database schema migrations. So we only add keystone here
+        # if clone_stateless_services was true.
+        clone_stateless = if node["pacemaker"]
+          node["pacemaker"]["clone_stateless_services_orig"].nil? ||
+            node["pacemaker"]["clone_stateless_services_orig"]
```
jsuchome: This should not be necessary in SOC9, right?

rhafer: Actually Cloud 8 still has the option to set "clone_stateless_services", so I think we still need it here.

jsuchome: Ah, you are right, so we need to keep all this

PR #1955, opened 2019-01-01
nova: Make sure we use the correct conf dir when running nova-manage

guangyee:
By default, if --conf-dir and --conf-file are not specified,
oslo_config will try to load the conf files from the default
locations. While running nova-manage as a non-priviledged user,
this will result in permission denied error even if we are using
a custom conf file in a non-default location. In this case, we
must create a temp conf dir with a temp conf file in it, and we
must specify both --conf-dir and --conf-file options together
in order to avoid loading anything from the default locations.

This patch also improves supportability and resilency by adding
logging and retry capabilities. If we are unable to create
host mapping for whatever reason, we will preserve the log file
for troubleshooting.

### DIFF
--- a/chef/cookbooks/nova/recipes/compute_register_cell.rb
+++ b/chef/cookbooks/nova/recipes/compute_register_cell.rb
@@ -31,12 +31,31 @@ api_database_connection = fetch_database_connection_string(node[:nova][:api_db])
 bash "nova-manage discover_hosts" do
   user node[:nova][:user]
   code <<-EOH
-    tmpfile=$(mktemp /tmp/nova-discover-hosts.XXXXXX.conf)
+    tmpdir=$(mktemp -d)
+    tmpfile=$(mktemp ${tmpdir}/nova-discover-hosts.XXXXXX.conf)
+    tmplogfile=$(mktemp ${tmpdir}/nova-manage.XXXXXX.log)
     chmod 600 $tmpfile
     echo "[api_database]" >> $tmpfile
     echo "connection = #{api_database_connection}" >> $tmpfile
-    nova-manage --config-file=$tmpfile cell_v2 discover_hosts --verbose
-    rm -f "$tmpfile"
+    for i in `seq 1 10`;
+    do
+	echo "$i try ..." >> $tmplogfile
+        nova-manage --config-file=$tmpfile --config-dir=$tmpdir cell_v2 discover_hosts --verbose >> $tmplogfile 2>&1
+	host_mapping_created=$(grep 'Creating host mapping' $tmplogfile | grep #{node[:hostname]})
+	echo "host_mapping_created=${host_mapping_created}" >> $tmplogfile
+	if [ -z "$host_mapping_created" ]; then
+	    # allow some time for the compute node to be visible to the controller
+	    sleep 10
+	else
+	    # host mapping created, we're done
+	    break
+	fi
+    done
+    if [ ! -z "$host_mapping_created" ]; then
+        # preserve the tmpdir and log files only if cell registration failed in
+    	# order to aide troubleshooting
+        rm -rf "$tmpdir"
+    fi
     EOH
 end
 


toabctl:
> By default, if --conf-dir and --conf-file are not specified,
> oslo_config will try to load the conf files from the default
> locations. While running nova-manage as a non-priviledged user,
> this will result in permission denied error even if we are using
> a custom conf file in a non-default location. 

It is running as nova user and that worked for me:
```
root@d52-54-77-77-01-01:~ # su - nova -c "id; head -1 /etc/nova/nova.conf.d/010-nova.conf"
uid=96(nova) gid=96(nova) groups=96(nova)
[DEFAULT]
```
Did you change any permissions in /etc/nova/ ?

guangyee:
@toabctl No. The permissions hasn't changed. The "nova" group is supposed to be able to read the config files.

root@d52-54-77-77-01-02:/tmp # ls -alrt /etc/nova/nova.conf.d/
total 24
-rw-r----- 1 root nova  122 Jan  1 18:47 010-nova.conf
drwxr-xr-x 4 root root 4096 Jan  2 18:00 ..
-rw-r----- 1 root nova  340 Jan  2 18:00 101-nova-placement.conf
-rw-r----- 1 root nova 4435 Jan  2 18:00 100-nova.conf
drwxr-xr-x 2 root root 4096 Jan  2 18:00 .

I supposed it has something to do with the session. For example, for the newly created "nova" group to take effect, we must utilize it in a brand new session (i.e. log out and log back in), not in the same session. I suspect chef client is doing all that in the same session which causing the problem.

cmurphy:
Will this fix the `Setting instance to ERROR state.: HostMappingNotFound: Host 'd52-54-77-77-01-02' is not mapped to any cell` problem that sometimes causes the manila barclamp to fail? eg https://ci.suse.de/job/openstack-mkcloud/151373/console

guangyee:
> Will this fix the `Setting instance to ERROR state.: HostMappingNotFound: Host 'd52-54-77-77-01-02' is not mapped to any cell` problem that sometimes causes the manila barclamp to fail? eg https://ci.suse.de/job/openstack-mkcloud/151373/console

Yes. This patch will fix the initial mapping.

chef/cookbooks/nova/recipes/compute_register_cell.rb
```diff
@@ -31,12 +31,31 @@
 bash "nova-manage discover_hosts" do
   user node[:nova][:user]
   code <<-EOH
-    tmpfile=$(mktemp /tmp/nova-discover-hosts.XXXXXX.conf)
+    tmpdir=$(mktemp -d)
+    tmpfile=$(mktemp ${tmpdir}/nova-discover-hosts.XXXXXX.conf)
+    tmplogfile=$(mktemp ${tmpdir}/nova-manage.XXXXXX.log)
```
cmurphy: I think nova-manage itself can be configured to log to a file, I'd prefer to use that rather than log to /tmp. Also won't the command output be logged to the chef client log?

chef/cookbooks/nova/recipes/compute_register_cell.rb
```diff
@@ -31,12 +31,31 @@
...
     chmod 600 $tmpfile
     echo "[api_database]" >> $tmpfile
     echo "connection = #{api_database_connection}" >> $tmpfile
-    nova-manage --config-file=$tmpfile cell_v2 discover_hosts --verbose
-    rm -f "$tmpfile"
+    for i in `seq 1 10`;
+    do
+	echo "$i try ..." >> $tmplogfile
+        nova-manage --config-file=$tmpfile --config-dir=$tmpdir cell_v2 discover_hosts --verbose >> $tmplogfile 2>&1
```
cmurphy: Inconsistent indendation

chef/cookbooks/nova/recipes/compute_register_cell.rb
```diff
@@ -31,12 +31,31 @@
...
+	host_mapping_created=$(grep 'Creating host mapping' $tmplogfile | grep #{node[:hostname]})
```
cmurphy: How about just save the stdout of the nova-manage command to a variable and check that, rather than checking a tempfile? If the tempfile is for some reason cleared or removed by another process then this would break. Would be even better if the nova-manage return code could be used.

jgrassler: Agreed. Sync marks would be a lot cleaner. We discussed this a bit today and we converged on roughly the following:

1. Use a sync mark to signal to the recipe on the controller once all compute 
2. Use the timeout/retry facilities of the chef execute resource to retry running `nova-manage`
3. Run `nova-manage` with a `--strict` option (that will cause it to return non-zero if anything fails and we won't need to parse its output).

Do you think that's feasible?

cmurphy: FYI - I'm taking a look at this

cmurphy: I think using sync marks here will never work, because sync marks are only designed to be used on nodes that are grouped in a pacemaker cluster together. The controller and computes are never in a cluster together. We'd have to invent a whole new orchestration mechanism to get the computes and controllers to talk to each other. This is why Tom had the compute nodes register themselves this way.

PR #1957, opened 2019-01-02
db: Raise default connection limit to 2048

dirkmueller:
It seems the previous setting of 1400 is still not enough, with
https://github.com/crowbar/crowbar-openstack/pull/1952 we need
higher limits. 2048 should be safer.

### DIFF
--- a/chef/data_bags/crowbar/template-database.json
+++ b/chef/data_bags/crowbar/template-database.json
@@ -9,7 +9,7 @@
         "slow_query_logging": true,
         "innodb_flush_log_at_trx_commit": 1,
         "innodb_buffer_pool_instances": 1,
-        "max_connections": 1400,
+        "max_connections": 2048,
         "tmp_table_size": 64,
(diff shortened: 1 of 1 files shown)


cmurphy:
Is there a CI problem or bug report that shows the default is too low?

dirkmueller:
@cmurphy there is no CI problem or bug report that I know of. with https://github.com/crowbar/crowbar-openstack/pull/1952 and similar scale changes, it is very easy to exceed the number of connections by default (as seen by an internal cloud8 deployment) in a 3 times 3 cluster environment where services are spread out and machines have more than 24 cores (and scaling of the number of rpc threads etc is happening per core). 

We could be more intelligent and calculate the actual limit from the services configuration that is deployed but that would have some awkward reverse dependency loops between services so I'm not sure I want to go there. I think there is no harm in raising the limit, its just a limit (e.g. it doesn't increase static resource utilisation if its not actually using that many threads).

PR #1958, opened 2019-01-02
[5.0] database: Prevent deploying mysql-server role to monasca node

jsuchome:
Monasca has its own MariaDB instance and we can't have it conflict
with the one deployed by database barclamp. See also bsc#1118759.

(cherry picked from commit f552e748361703d515bd71c7fbec079066befa92)
(cherry picked from commit fd0b95a8bb9b6f649edc95d459513771375ee0ca)


jsuchome:
Yeah, I've cherry-picked from the master which was already cherry-picked from 4.0. But master version was already modified after some conflicts and 5.0 has the same code as master around...

PR #1966, opened 2019-01-04
[5.0] heat: Migrate properties data from the legacy locations in the db

jsuchome:
Although not strictly required, it seems that heat works more
efficiently after such migration.


jsuchome:
> Is there a specific reason why we are handling this different to the nova online migration

Basically I did not want to add another upgrade substep.

jsuchome:
Could be replaced by https://github.com/crowbar/crowbar-core/pull/1779

PR #1969, opened 2019-01-08
manila: Fixup deprecated options (SCRD-6084)

cmurphy:
The `api_insecure` option in the [nova], [neutron], and [cinder] groups
was replaced by just `insecure`[1].

[1] https://docs.openstack.org/releasenotes/manila/rocky.html


cmurphy:
We were already setting `insecure`, removed the duplicated option now.

PR #1975, opened 2019-01-10
keystone: Remove use of SSLCACertificate

cmurphy:
The SSLCACertificate option in Apache is not about specifying the
server's certificate bundle but actually about what certificate chain to
use to validate certificates of client connections. This does not
provide significant value since none of the authentication information
is passed to keystone and the meaning of the parameter has caused
significant user and developer confusion. This change therefore removes
the cert_required and ca_certs parameters in order to clean up the code
and simplify the user interface.

### DIFF
--- a/chef/cookbooks/keystone/attributes/default.rb
+++ b/chef/cookbooks/keystone/attributes/default.rb
@@ -55,7 +55,6 @@ default[:keystone][:ssl][:certfile] = "/etc/keystone/ssl/certs/signing_cert.pem"
 default[:keystone][:ssl][:keyfile] = "/etc/keystone/ssl/private/signing_key.pem"
 default[:keystone][:ssl][:generate_certs] = false
 default[:keystone][:ssl][:insecure] = false
-default[:keystone][:ssl][:ca_certs] = "/etc/keystone/ssl/certs/ca.pem"
 
 default[:keystone][:ldap][:url] = "ldap://localhost"
 default[:keystone][:ldap][:user] = "dc=Manager,dc=example,dc=com"
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -82,8 +82,6 @@ if node[:keystone][:api][:protocol] == "https"
     group node[:keystone][:group]
     fqdn node[:fqdn]
     alt_names ["DNS:#{my_admin_host}", "DNS:#{my_public_host}"]
-    cert_required !node[:keystone][:ssl][:insecure]
-    ca_certs node[:keystone][:ssl][:ca_certs]
   end
 end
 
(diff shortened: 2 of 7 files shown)


cmurphy:
Keystone can support client X.509 authentication, but not without additional configuration and this isn't about that. Apache on its own could require a valid client cert for an extra layer of security, even if it doesn't pass any auth data to keystone. I thought that was the original idea and was trying to restore it but I agree it's not all that useful.

cmurphy:
@dirkmueller done

cmurphy:
> Looks good to me in general. Bonus points for updating the PR description to match the latest changes. (it still mentions restoring "cert_required")

Thanks, fixed.

> The only thing I am wondering about is what should be done with the rest of the services. All of them still know have that "cert_required" attribute. Do they really allow client cert verification and does it make much sense to have it there when keystone doesn't allow it?

The other services still have config options like `[ssl]/ca_cert` which would have been the equivalent of Apache's SSLCACertificate. I don't think we should remove support for that since that's part of the service configuration. If we want keystone to have feature parity with the other services then we should go back to my original change and restore `cert_required` and `ca_certs` so that Apache can verify client requests.

dirkmueller:
I just did a rebase to fix the migration conflict, no other change.

cmurphy:
@dirkmueller do you have an opinion on the question @rhafer raised? I'm not sure we should be removing this functionality from keystone when it's still present in all the other services.

dirkmueller:
@cmurphy thanks for being so diligent and pointing out that I missed this comment, much appreciated.

My current understanding is that the cacert is the setting pointing to the intermediate CS certs bundle that the server need to include in its response to the client. In our autogenerated certs (both crowbar and the mkcloud ones) we have only zero or one length of chain so this setting would be always empty. For a real cert like the suse internal one you can have intermediate ones. 

In Apache those are.to be concatenated and included in the public cert as a full bundle. It is extremely ugly that our crowbar settings expect different behavior for keystone or any other apache backed wervice than for others ehich are not yet behind apache.

I haven't actually souble checked all of that above, so writing this up to see.if I can be proven wrong.

dirkmueller:
If I am correct then I think our stuff is a giant mess and we need to clean it further. Bit then thos patch would need to be updated pointing out that the public cert is actually a full bundle plus cert

cmurphy:
> My current understanding is that the cacert is the setting pointing to the intermediate CA certs bundle that the server need to include in its response to the client. 

The Apache SSLCACertificateFile option points to the file of concatenated certificates uses to authenticate a client request, it's the SSLCertificateFile that is used in a response to the client.

> In our autogenerated certs (both crowbar and the mkcloud ones) we have only zero or one length of chain so this setting would be always empty. 

There would always be at least one certificate in SSLCertificateFile and from a client's perspective it should be functionally equivalent to a chain of certificates as long as the client has also installed that certificate as a trusted certificate. I'm not sure why the number matters here.

> It is extremely ugly that our crowbar settings expect different behavior for keystone or any other apache backed wervice than for others ehich are not yet behind apache.

That's why I'm trying to question assumptions here, I think it would make more sense to align the keystone barclamp with the way the other barclamps are doing this (or supposed to be doing this) than to remove this setting from the keystone barclamp and make it the odd one out.

cmurphy:
I don't plan to continue working on this.

PR #1978, opened 2019-01-22
Update cmurphy's code ownership

cmurphy:
Triggering on the keystone keyword is not very fine-grained because
every service uses keystone_register to create users, projects,
endpoints, and services in keystone. Switch to just managing the
keystone cookbooks and databags.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -16,8 +16,12 @@
 **/*cinder*  @toabctl
 **/*manila*  @toabctl
 
-**/*keystone*  @rhafer @cmurphy
+**/*keystone*  @rhafer
 **/*monasca*  @jgrassler
 
 ha*.rb  @aspiers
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -16,8 +16,12 @@
...
 **/*manila*  @toabctl
 
-**/*keystone*  @rhafer @cmurphy
+**/*keystone*  @rhafer
 **/*monasca*  @jgrassler
 
 ha*.rb  @aspiers
 *_ha.rb  @aspiers
+
+/chef/openstack/keystone/ @cmurphy
```
dirkmueller: do you mean chef/cookbooks/keystone here? or am I misunderstanding how this works?

cmurphy: Just not enough caffeine yet

PR #1980, opened 2019-01-24
neutron: Reduce number of dhcp agents to tolerate node failures

dirkmueller:
If in a neutron cluster are less than configured number of
nodes online, neutron will try to start squeeze the dhcp agents
on the remaining nodes, which does not seem to actually work
and does not actually improve availability. In larger deployments
the duplicated rules seem to cause the dhcp agents to trip over
and we have rolling outages.
    
By only starting dhcp agents for the "smallest" online cluster size
we can avoid that. In a 3 node cluster we can tolerate a single
node failure, starting two agents. in a 5 node we can tolerate 2
nodes failure so 3 agents should be started and so on.

### DIFF
--- a/chef/cookbooks/neutron/templates/default/neutron.conf.erb
+++ b/chef/cookbooks/neutron/templates/default/neutron.conf.erb
@@ -13,7 +13,7 @@ global_physnet_mtu = <%= @mtu_value %>
 use_ssl = <%= @ssl_enabled ? "True" : "False" %>
 api_workers = <%= [node["cpu"]["total"] / 2, 1, 16].sort[1] %>
 rpc_workers = <%= @rpc_workers %>
-dhcp_agents_per_network = <%= @network_nodes_count %>
+dhcp_agents_per_network = <%= (@network_nodes_count + 1) / 2 %>
 <% if @dvr_enabled -%>
 router_distributed = True
 <% end -%>


AbelNavarro:
[Review of the DHCP - Agent Failover mechanism](https://trello.com/c/7BB2euN4/357-review-of-the-dhcp-agent-failover-mechanism)

chef/cookbooks/neutron/templates/default/neutron.conf.erb
```diff
@@ -13,7 +13,7 @@ global_physnet_mtu = <%= @mtu_value %>
 use_ssl = <%= @ssl_enabled ? "True" : "False" %>
 api_workers = <%= [node["cpu"]["total"] / 2, 1, 16].sort[1] %>
 rpc_workers = <%= @rpc_workers %>
-dhcp_agents_per_network = <%= @network_nodes_count %>
+dhcp_agents_per_network = <%= ((@network_nodes_count + 1) / 2).floor %>
```
rhafer: AFAIK ruby would be doing integer division here by default. So the `.floor` call is somewhat useless I think. Also if you're really aiming for a quorum of the cluster nodes to host the network I think you need something like:

`(@network_nodes_count / 2) + 1)` 

As quorum means more than half of the nodes.

But I guess for this patch a quorum is not really needed. How about just capping the maximum at some configurable value? (e.g. use 1 of non HA setups, 2 for 2 and 3 node clusters and 3 for everything else).  So e.g. something like:

```
max_agents_per_net=3 #chef attribute?
[(@network_nodes_count / 2) + 1, max_agents_per_network].min
```
I guess the cases where it makes sense to run a single network on more than 3 agents are rather limited.

dirkmueller: you're correct that calling that "quorum" is incorrect. and floor is sort of redundant. I can remove that. 

I think the current formular does exactly that except for introducing yet another configuration variable that needs to be documented and set by a customer.

dirkmueller:
@rhafer for nodecounts up to 6 your and my proposal are the same I think. It's just that it scales further if you have larger clusters, which seems okay to me..

PR #1981, opened 2019-01-24
monasca: Detect failure during monasca-installer run

toabctl:
Currently, the monasca-installer run can fail but the Crowbar
deployment just continues. This might result in follow-up errors.
The reason is, that the return code of the run-monasca-installer
script is not evaluated. Instead the return code of awk is
evaluated (which is usually zero).
So set "pipefail" to also detect run-monasca-installer failures and
abort the chef-client run in that case.


toabctl:
Untested

toabctl:
triggered a "with-monasca" run...

PR #1985, opened 2019-01-24
[5.0] database: Hide SSL options from database UI

skazi0:
This is a replacement for package patch to simplify future modifications
to the sources without causing conflicts.

(cherry picked from commit 8df86463d039a2cbb6a86dfd155dfbb286d393e1)

port of #1769


skazi0:
This was a misunderstanding. Rick confirmed that this should not be ported to C8.

PR #1987, opened 2019-01-29
mysql: saner sysctl values for galera nodes

Itxaka:
Adds sysctl values for swappiness obtained from the tuned application
profiles for virtual-hosts in order to have less aggresive swap
in galera nodes

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -364,3 +364,22 @@ haproxy_loadbalancer "galera" do
   servers ha_servers
   action :nothing
 end.run_action(:create)
+
+directory "create /etc/sysctl.d for galera values" do
+  path "/etc/sysctl.d"
+  mode "755"
+end
+
+template "galera_sysctl.conf" do
+  path "/etc/sysctl.d/90-galera_sysctl.conf"
+  mode "0644"
+  variables(
+    swappiness: node[:mysql][:sysctl][:swappiness]
+  )
+end
+
+bash "reload galera_sysctl.conf" do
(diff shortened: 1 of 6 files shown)


Itxaka:
[Saner sysctl defaults for compute nodes and database nodes](https://trello.com/c/hSbEqGyx/336-saner-sysctl-defaults-for-compute-nodes-and-database-nodes)

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -364,3 +364,22 @@
...
+template "galera_sysctl.conf" do
+  path "/etc/sysctl.d/90-galera_sysctl.conf"
+  mode "0644"
+  variables(
+    swappiness: node[:mysql][:sysctl][:swappiness]
+  )
+end
+
+bash "reload galera_sysctl.conf" do
+  code "/sbin/sysctl -e -q -p /etc/sysctl.d/90-galera_sysctl.conf"
```
nicolasbock: Why do we need the `-e` / `--ignore` option here? Isn't that just hiding a potential bug in the barclamp?

Itxaka: Its used on other parts of the cookbooks  when applying sysctl changes. Im guessing it had an historic reason for it, which I cannot find anywhere, so I opted to keep it the same for consistency reasons :thinking:

dirkmueller:
imho this should be moved to provisioner or some other barclamp in core to ensure that it is being rolled out everywhere. 

potentially we should just use tuned instead of reinvinting wheels (I know, our wheel is going to be rounder)

dirkmueller:
move this to provisioner-base ?

Itxaka:
wow sorry, I do not get notifications regarding this so I thougth it was just lingering around :o @dirkmueller

PR #1989, opened 2019-01-31
monasca: Fix grafana config for monasca-persister

toabctl:
The monasca-persister check is done via the process plugin and
it's *not* a http_status check.
So use the same grafana config as we already use for
monasca-notification.

This fixes the Grafana UI when accessing the dashboard node under
"grafana/d/Monasca_dashboard/suse-openstack-cloud-monitoring-monasca"
. The Persister bar there was always black due to the wrong config.

### DIFF
--- a/chef/cookbooks/horizon/files/default/grafana-monasca.json
+++ b/chef/cookbooks/horizon/files/default/grafana-monasca.json
@@ -237,7 +237,13 @@
             }
           ],
           "metricNameColor": "#000000",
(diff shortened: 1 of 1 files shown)


toabctl:
This is only needed when https://github.com/crowbar/crowbar-openstack/pull/1953 got merged.

toabctl:
I just added the commit to https://github.com/crowbar/crowbar-openstack/pull/1953 so this can be closed...

PR #1990, opened 2019-01-31
[WIP] rabbitmq: Switch back to client-local master-locator (bsc#112387)

rhafer:
It seems "min-master" is creating issue in RabbitMQ when the target
node for the queue is not fully up yet. This is basically a workaround
for something that appears to be a bug in RabbitMQ itself
(https://bugzilla.suse.com/show_bug.cgi?id=1123872).

WIP: Do not merge yet. I just submitted it for launch a few CI runs to check whether it actually makes a difference

### DIFF
--- a/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
+++ b/chef/cookbooks/rabbitmq/templates/default/rabbitmq.config.erb
@@ -35,7 +35,6 @@
 <% end -%>
 <% if @cluster_enabled -%>
    {cluster_partition_handling, <%= @cluster_partition_handling %>},
-   {queue_master_locator, <<"min-masters">>},
 <% end -%>
 <% if @hipe_compile -%>
    {hipe_compile, true},


Itxaka:
This makes sense as due to the difference of timeframes between a normal install vs an upgrade, as during the normal install there is quite some time for rabbit to be brought up (remember that the pacemaker agent for rabbitmq restarts the nodes several times!)

Although, couldn't we make it so we wait until the rabbitmq cluster is ready before continuing the upgrade part? As far as I understand this only affect the upgrade path right? Sorry, cant remember if during the upgrade we have control over the services enough as to do that.

rhafer:
> Although, couldn't we make it so we wait until the rabbitmq cluster is ready before continuing the upgrade part?

@Itxaka Isn't there still a possible issue when restarting rabbitmq on any node (e.g. for maintenance reasons)?

Gsantomaggio:
The problem can be reproduced by using 2 RabbitMQ nodes in the cluster with the `master-locator` policy enabled.


pump the cluster using this script:
```
from __future__ import absolute_import, unicode_literals, print_function

from kombu import Connection  # noqa


with Connection('amqp://guest:guest@localhost:5673//') as conn:
    for i in range(1,3000):
        simple_queue = conn.SimpleQueue('@aaaaaaaaa@@@@@'+str(i))
        simple_queue.close()

```

Stop the second node ( the one without client connected) and the restart it:

And you get the error:
````
** Reason for termination ==
** {{noproc,
        {gen_server,call,
            [{rabbit_amqqueue_sup_sup,rabbit@GaS},
             {start_child,
                 [{amqqueue,
                      {resource,<<"/">>,queue,
                          <<"@@@a@@aaaaaaaaa@simple_queue_292">>},
                      true,false,none,[],none,[],[],[],undefined,[],[],live,0},
                  declare]},
             infinity]}},
    [{gen_server,call,3,[{file,"gen_server.erl"},{line,214}]},
     {rabbit_amqqueue_sup_sup,start_queue_process,3,
         [{file,"src/rabbit_amqqueue_sup_sup.erl"},{line,42}]},
     {rabbit_amqqueue,declare,6,[{file,"src/rabbit_amqqueue.erl"},{line,341}]},
     {rabbit_channel,handle_method,3,
         [{file,"src/rabbit_channel.erl"},{line,1346}]},
````                                                                

what you see client side as:

```
 File "/home/gabriele/tests/stress_queues/.venv/local/lib/python2.7/site-packages/amqp/connection.py", line 510, in on_inbound_method
    method_sig, payload, content,
  File "/home/gabriele/tests/stress_queues/.venv/local/lib/python2.7/site-packages/amqp/abstract_channel.py", line 126, in dispatch_method
    listener(*args)
  File "/home/gabriele/tests/stress_queues/.venv/local/lib/python2.7/site-packages/amqp/connection.py", line 639, in _on_close
    (class_id, method_id), ConnectionError)
amqp.exceptions.InternalError: (0, 0): (541) INTERNAL_ERROR
(
````

Itxaka:
> @Itxaka Isn't there still a possible issue when restarting rabbitmq on any node (e.g. for maintenance reasons)?

Mmmh, it could be, yes, although for a simple restart shouldn't, if the master is dead and the cluster has to be redone, the same issue could happen, even if we have not seen it anywhere yet.

Gsantomaggio:
[RabbitMQ user group thread](https://groups.google.com/d/msg/rabbitmq-users/xEWZCmPXI-Q/yDYlPBC5EwAJ)

GitHub issue:  https://github.com/rabbitmq/rabbitmq-server/issues/1869

rhafer:
This can be close. I think we have a different fix (in oslo.messaging and rabbitmq now)

PR #1991, opened 2019-02-01
tempest: Install monasca tempest plugin unconditionally

toabctl:
Just install the plugin (as we do with the other tempest plugins) and
rely on the [service_available]monitoring parameter. If that parameter
is set to false, no monasca tempest tests will be execute.

### DIFF
--- a/chef/cookbooks/tempest/recipes/install.rb
+++ b/chef/cookbooks/tempest/recipes/install.rb
@@ -32,19 +32,11 @@ package "euca2ools"
 package "openstack-tempest-test"
 
 ["barbican", "cinder", "designate", "heat", "ironic", "keystone", "magnum",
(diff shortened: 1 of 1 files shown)


toabctl:
This needs https://review.openstack.org/#/c/634401/ (monasca-common is needed by the monasca tempest plugin)

toabctl:
> can we somehow keep the conditional on x86?

@dirkmueller Interessting. I thought we don't allow to assign the `monasca-server` and `monasca-master` role when `arch=!x86_64`. And if that would be true, then we don't need to preserve this here (because the service will never be installed and never ends up in the keystone catalog. 
But looks like you can currently assign the roles to a non-X86_64 node?!

toabctl:
Let's continue on #2018

PR #1992, opened 2019-02-01
Monasca: Add Python monasca-persister

witekest:
Optionally Python implementation of monasca-persister can be installed
by setting `mon_persister_impl = "python"`.


jgrassler:
I concur with @toabctl here. Let's shelve this until https://github.com/crowbar/crowbar-openstack/pull/1953 has landed and implement dual persister support in pure Crowbar then. Sorry about the duplication of effort, but I wanted this implementation to be available as well just in case https://github.com/crowbar/crowbar-openstack/pull/1953  didn't get done in time - which is no longer a concern in my book.

On a related note: please keep the branch around, just in case (we may be able to use it for persister comparison performance testing in the time window between https://github.com/crowbar/crowbar-openstack/pull/1953 landing and dual persister support getting implemented in the pure Crowbar version.

toabctl:
I think this can be closed. `monasca-installer` is history. We might want to add support for the java version in crowbar but that can/should be done in another PR. So closing for now.

PR #1995, opened 2019-02-03
galera_bootstrap: Wait for initial bootstrap start prior creating users

dirkmueller:
The first start of galera can be a bit slow, so we're hitting a
race with the initial sst user creation which in some cases
might otherwise immediately die with "wsrep state not in Sync".


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -90,6 +100,29 @@
       action :start
     end
 
+    ruby_block "wait mysql initialisation" do
```
rhafer: AFAICS this is basically a copy of the code from line 280. Apart from the different user name, static timeout and the `not_if` condition (which wouldn't hurt here). Could you please turn this into a chef definition with username and timeout as paramters?

PR #1996, opened 2019-02-04
Revert "Revert "keystone: Make register requests object-oriented""

cmurphy:
This adds back the new object-oriented keystone session handling
introduced in ff7a6ec5. It fixes the issue that caused the original
revert by explicitly resetting the session object at the end of the
:update_one_endpoint action so that the old protocol is not still active
for the next keystone_register command.

This reverts commit 158755c15afd4baf97f34e7b52922c37ef311bcd.

### DIFF
--- a/chef/cookbooks/keystone/libraries/helpers.rb
+++ b/chef/cookbooks/keystone/libraries/helpers.rb
@@ -142,4 +142,116 @@ module KeystoneHelper
   def self.cache_reset
     @cache = {}
   end
+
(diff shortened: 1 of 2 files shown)


cmurphy:
This should be ready now, the CI built this with https://github.com/SUSE-Cloud/automation/pull/3194 included

cmurphy:
I've rebased this in order to re-run CI and ensure this still works, please reapprove.

PR #2000, opened 2019-02-05
hoirzon: search for role while installing heat-plugin (SCRD-7430)

sjamgade:
heat proposal is still not saved while it is being applied as a
result corresponding horizon plugin will not get installed

And thus requires another chef-client run.

This additional check will get the package installed even while heat
proposal is being applied

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -205,11 +205,10 @@ end
 
 # install horizon heat plugin if needed
 heat_ui_pkgname = "openstack-horizon-plugin-heat-ui"
-unless Barclamp::Config.load("openstack", "heat").empty?
-  package heat_ui_pkgname do
-    action :install
-    notifies :reload, "service[horizon]"
-  end
+package heat_ui_pkgname do
+  action :install
+  notifies :reload, "service[horizon]"
+  only_if { RoleHelper.config_for_role_exists?("heat") }
 end
 
 if node[:platform_family] == "suse"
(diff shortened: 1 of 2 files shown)


chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -205,11 +205,10 @@
...
 heat_ui_pkgname = "openstack-horizon-plugin-heat-ui"
-unless Barclamp::Config.load("openstack", "heat").empty?
-  package heat_ui_pkgname do
-    action :install
-    notifies :reload, "service[horizon]"
-  end
+package heat_ui_pkgname do
+  action :install
+  notifies :reload, "service[horizon]"
+  only_if { RoleHelper.config_for_role_exists?("heat") }
```
jgrassler: Shouldn't that be "heat-server"? (that's the only role in the Heat barclamp).

sjamgade: heat-server is a "sub-role", what I check for is heat-config-, which indicates heat proposal exists

toabctl:
@sjamgade why does this work for the other plugins but not for Heat? Or is it broken for the other plugins (sahara, ironic, magnum, ...), too?

sjamgade:
@toabctl its broken for all, I did not touch other plugins as I want to keep the changeset small.

chef/cookbooks/horizon/recipes/server.rb
jgrassler: Ah, ok.

toabctl:
@sjamgade could you fix it for all? shouldn't be to difficult to review given that it's the same pattern everywhere....

sjamgade:
the other reason I do not want to touch the PR, is I needed 3 CI rebuild (~8 hrs) to get it correctly tested. I know the pattern is same, but the turn around time is excruciatingly painful

toabctl:
@sjamgade please do a follow up commit which fixes the rest. thanks!

PR #2003, opened 2019-02-06
ironic: Switch from classic drivers to hardware types

skazi0:
Classic drivers are now deprecated. Updated barclamp to use hardware
types and interfaces instead.


gosipyan:
After retrying apply PR i have new issue please look http://crowbar.vp9.cloud.suse.de/crowbar/ironic/1.0/proposals/default?attr_raw=true

skazi0:
@gosipyan good catch! glance cookbook should be fixed now.

cmurphy:
Has the upgrade case been considered here? From the doc it looks like enabled_drivers and the new options should be simultaneously enabled for a period:
https://docs.openstack.org/ironic/pike/admin/upgrade-to-hardware-types.html#configuration
and nodes need to be migrated:
https://docs.openstack.org/ironic/pike/admin/upgrade-to-hardware-types.html#migrating-nodes

chef/cookbooks/glance/recipes/api.rb
```diff
@@ -92,9 +92,9 @@
   notifies :restart, "service[#{node[:glance][:api][:service_name]}]"
 end
 
-# ensure swift tempurl key only if some agent_* drivers are enabled in ironic
+# ensure swift tempurl key only if "direct" deploy interface is enabled in ironic
```
guangyee: Should we be checking on boot interface 'ilo-virtual-media' instead? Only virtual media uses Swift TempURL right?

chef/data_bags/crowbar/migrate/ironic/301_drivers_to_hardware_types.rb
```diff
@@ -0,0 +1,200 @@
+def upgrade(ta, td, a, d)
+  # migrate classic drivers to new form
+  unless a["enabled_drivers"].nil? || a["enabled_drivers"].empty?
+    # based on https://docs.openstack.org/ironic/pike/admin/upgrade-to-hardware-types.html
+    mapping = {
```
guangyee: Wow. We support all of them in SOC?

skazi0: no, but the mapping was provided by the upstream so why not include it in the migration?

chef/cookbooks/glance/recipes/api.rb
skazi0: First sentence here: https://docs.openstack.org/ironic/pike/install/configure-glance-swift.html says that all "direct" deployments use it.

skazi0:
@cmurphy we can't have `enabled_drivers` filled in Rocky as it fails to start with that. Automatic modification of ironic nodes during upgrade could be tricky but if we decided to do it, it should be addressed during 8->9 upgrade work which has not started officially yet.

gosipyan:
@skazi0 thanks , i was able successfully deploy ironic after applying fix.

chef/data_bags/crowbar/migrate/ironic/301_drivers_to_hardware_types.rb
guangyee: That's fine. So as long as we indicate in our doc what's been tested and supported. Usually, anything that we include and visible to the customers, they'll expect us to support it. :-)

chef/cookbooks/glance/recipes/api.rb
guangyee: Interesting. I hope the doc's right. So far, from our testing, only virtual media needs TempURL. Not sure if we support other drivers. At least we haven't been able to test them as we are lacking hardware.

skazi0: But testing this doesn't require any special hardware. You can use `ipmitool` hardware type with `direct` deploy interface. This is similar to classic `agent_ipmitool` driver.

PR #2004, opened 2019-02-07
Fix MariaDB log rotation

sandonovsuse:
Add /etc/logrotate.d/mariadb-soc, containing correct file
references. Depends on 9295bcc - mysql root credentials
needed for mysqladmin to flush rotated log files (bsc#1112767)


chef/cookbooks/mysql/recipes/server.rb
```diff
@@ -286,3 +286,10 @@
     password: node[:database][:mysql][:server_root_password]
   )
 end
+
+template "/etc/logrotate.d/mariadb-soc" do
```
jsuchome: Why the suffix? I guess you can go directly with "/etc/logrotate.d/mariadb" (and it is even referenced from the logrotate script below)

dirkmueller:
@sandonovsuse so how does that help? wouldn't it be better to use the password in ~/.my.cnf so that it works without password? isn't there a logrotate in the package that we should leverage?

there is a PR that does that: https://github.com/crowbar/crowbar-openstack/pull/1701 we could backport it.

chef/cookbooks/mysql/recipes/server.rb
sandonovsuse: Yes, no reason - it should have been "/etc/logrotate.d/mariadb"

sandonovsuse:
> isn't there a logrotate in the package that we should leverage?
> 
logrotate script included the package only rotates mysqld.log, not the error log - which grew heavily on a production system (bsc#1112767). The only change is to include error and slow logs. 
> @sandonovsuse so how does that help? wouldn't it be better to use the password in ~/.my.cnf so that it works without password? 
> there is a PR that does that: #1701 we could backport it.

Agreed, 9295bcc should be the commit from PR #1701. Without the PR, it wouldn't work.

toabctl:
> Agreed, [9295bcc](https://github.com/crowbar/crowbar-openstack/commit/9295bcc4ecf57ddb900065784095458336f04ce7) should be the commit from PR #1701. Without the PR, it wouldn't work.

@sandonovsuse here is the backport for the `/root/.my.cnf`file which includes the credentials: https://github.com/crowbar/crowbar-openstack/pull/2020 .

rsalevsky:
Can we please fix this in the MariaDB package https://build.opensuse.org/package/show/server:database/mariadb?

toabctl:
I also created
- https://build.opensuse.org/request/show/676156 to rotate all logfiles in the package itself (so we don't need to override the logrotate script from the package
- https://github.com/crowbar/crowbar-openstack/pull/2021 to switch to the default logfile so at least rotation for `mysqld.log` works as expected

toabctl:
> Can we please fix this in the MariaDB package https://build.opensuse.org/package/show/server:database/mariadb?

like https://build.opensuse.org/request/show/676156 ?

rsalevsky:
> > Can we please fix this in the MariaDB package https://build.opensuse.org/package/show/server:database/mariadb?
> 
> like https://build.opensuse.org/request/show/676156 ?

Yes

toabctl:
@sandonovsuse I think with:
- https://build.opensuse.org/request/show/676156 and
- https://github.com/crowbar/crowbar-openstack/pull/2020

both accepted/merged, this PR is no longer needed. Do you agree and can we close this?

sandonovsuse:
> @sandonovsuse I think with:
> 
>     * https://build.opensuse.org/request/show/676156 and
> 
>     * #2020
> 
> 
> both accepted/merged, this PR is no longer needed. Do you agree and can we close this?

yes, looks great - thx

PR #2006, opened 2019-02-08
novnc: use tcp mode of haproxy only during ssl (SCRD-7127)

sjamgade:
this is a followup of bc439f8c686bb4e4533687d604508c74e538212d as mentioned in the #1993.

dropping http check when not using ssl is loss of feature which can be
avoided.

This PR add that feature back and still fixes the issue mentioned in
SCRD-7127


JanZerebecki:
Superseded by https://github.com/crowbar/crowbar-openstack/pull/2101

PR #2008, opened 2019-02-08
monasca: Update notification config in UI

witekest:
Update Monasca barclamp UI to new configuration option names [e4f9ccbf1b236c1f42a86e04ce4bb395f4aec237] for
notification engine.


crowbar_framework/config/locales/monasca/en.yml
```diff
@@ -46,13 +46,14 @@ en:
           backlog_send_rate: 'Maximum number of messages to send at one time when communication with the monasca-api is restored (backlog_send_rate)'
           amplifier: 'Publish extra metrics to the API by adding this number of "amplifier" dimensions. For load testing purposes only; set to 0 for production use (amplifier)'
         master_notification_header: 'Notification configuration'
```
jgrassler: nit: There's one last remnant of `master` here :-)

witekest: There are actually more _master_ variables in the schema. I have moved `tsdb` to its own section in the barclamp view.

PR #2009, opened 2019-02-08
keystone: Add basic osprofiler support

toabctl:
Adding osprofiler[1] support is useful for profiling OpenStack accross
different services. This can be used to find bottlenecks or slow code
inside the different OpenStack services.
osprofiler needs to be enabled in the different services and this
commit starts with keystone.
Once osprofiler is enabled (via the Crowbar attributes), it can be
used like this:

1) Call a command with the openstack client (which supports
osprofiler). Note: You have to know (and use) the HMAC key that is in the
Crowbar attribute stored. If you don't use that HMAC key, osprofiler
won't do anything.

$ openstack --os-profile 9tcwL56PlpAx endpoint list
...
Trace ID: e860c2b0-6945-4767-bbd9-c42c08e6417c

Beside the usual output (in this case the output of "endpoint list"),
there will be a Trace ID written to the command line.
With this Trace ID, the whole trace can be shown with:

$ osprofiler trace show --html --connection-string messaging:// \
    --transport-url \
    "rabbit://openstack:DNUdYtnlMPPY@192.168.192.81:5672/openstack" \
    e860c2b0-6945-4767-bbd9-c42c08e6417c

Note: Currently we use the messaging driver so osprofiler sends it's
trace points to the rabbitmq topic "profiler" which can then be
consumed with the --transport-url.
There are other drivers which might be better (but need to setup other
storage backends) so for now, the messaging driver should be good.

[1] https://docs.openstack.org/osprofiler/latest/

### DIFF
--- a/chef/cookbooks/keystone/libraries/helpers.rb
+++ b/chef/cookbooks/keystone/libraries/helpers.rb
@@ -104,6 +104,12 @@ module KeystoneHelper
       "service_password" => current_node[cookbook_name][:service_password])
   end
 
+  def self.profiler_settings(current_node, cookbook_name)
+    instance = current_node[cookbook_name][:keystone_instance] || "default"
+    node = search_for_keystone(current_node, instance)
+    node["keystone"]["osprofiler"]
+  end
+
   private_class_method def self.search_for_keystone(node, instance)
     if @keystone_node && @keystone_node.include?(instance)
       Chef::Log.info("Keystone server found at #{@keystone_node[instance].name} [cached]")
--- a/chef/data_bags/crowbar/migrate/keystone/302_add_osprofiler.rb
+++ b/chef/data_bags/crowbar/migrate/keystone/302_add_osprofiler.rb
@@ -1,0 +1,16 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs.key? "osprofiler"
+    attrs["osprofiler"] = template_attrs["osprofiler"]
+    unless defined?(@@osprofiler_hmac_keys)
+      service = ServiceObject.new "fake-logger"
+      @@osprofiler_hmac_keys = service.random_password
+    end
+    attrs["osprofiler"]["hmac_keys"] << @@osprofiler_hmac_keys
+  end
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs.delete("osprofiler")
+  return attrs, deployment
+end
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -247,6 +247,8 @@ register_auth_hash = { user: node[:keystone][:admin][:username],
                        password: node[:keystone][:admin][:password],
                        project: node[:keystone][:admin][:project] }
 
+profiler_settings = KeystoneHelper.profiler_settings(node, @cookbook_name)
+
 template node[:keystone][:config_file] do
     source "keystone.conf.erb"
     owner "root"
@@ -266,7 +268,8 @@ template node[:keystone][:config_file] do
       max_active_keys: max_active_keys,
       protocol: node[:keystone][:api][:protocol],
       frontend: node[:keystone][:frontend],
-      rabbit_settings: fetch_rabbitmq_settings
+      rabbit_settings: fetch_rabbitmq_settings,
+      profiler_settings: profiler_settings
     )
     if node[:keystone][:frontend] == "apache"
       notifies :create, resources(ruby_block: "set origin for apache2 restart"), :immediately
--- a/chef/cookbooks/keystone/templates/default/keystone.conf.erb
+++ b/chef/cookbooks/keystone/templates/default/keystone.conf.erb
@@ -107,3 +107,11 @@ driver = sql
 [token]
 expiration = <%= @token_expiration %>
 provider = <%= node[:keystone][:token_format] %>
+
+<% if @profiler_settings[:enabled] -%>
+[profiler]
+enabled = true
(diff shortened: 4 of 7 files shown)


toabctl:
An example how the html output looks like can be found here: https://w3.suse.de/~tbechtold/osprofiler/example.html

chef/data_bags/crowbar/migrate/keystone/302_add_osprofiler.rb
```diff
@@ -0,0 +1,13 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs.key? "osprofiler"
+    attrs["osprofiler"] = template_attrs["osprofiler"]
+    service = ServiceObject.new "fake-logger"
```
dirkmueller: you need to set the @@ instance variable as we discussed yesterday ;)

toabctl: @dirkmueller done (hopefully correct now)

chef/data_bags/crowbar/migrate/keystone/302_add_osprofiler.rb
```diff
@@ -0,0 +1,16 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs.key? "osprofiler"
+    attrs["osprofiler"] = template_attrs["osprofiler"]
+    unless defined?(@@osprofiler_hmac_key)
```
dirkmueller: Name should be osprofiler_hmac_keys (needs to match underscored stitched together path of the attribute)

toabctl: Fixed

chef/data_bags/crowbar/migrate/keystone/302_add_osprofiler.rb
```diff
@@ -0,0 +1,16 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs.key? "osprofiler"
+    attrs["osprofiler"] = template_attrs["osprofiler"]
+    unless defined?(@@osprofiler_hmac_keys)
```
cmurphy: Can you elaborate on this part? I see the pattern in other migrations but it looks like a relic of some old cookbook syntax? For example https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/migrate/cinder/006_generate_db_password.rb#L2-L4

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -104,6 +104,12 @@ def self.keystone_settings(current_node, cookbook_name)
       "service_password" => current_node[cookbook_name][:service_password])
   end
 
+  def self.profiler_settings(current_node, cookbook_name)
+    instance = current_node[cookbook_name][:keystone_instance] || "default"
+    node = search_for_keystone(current_node, instance)
+    node["keystone"]["osprofiler"]
+  end
```
cmurphy: This is hardcoded for keystone, is the idea that other cookbooks will have their own helper method to get their own profiler settings?

chef/data_bags/crowbar/migrate/keystone/302_add_osprofiler.rb
toabctl: I learned about that from @dirkmueller some days ago. The migration is run 2x (one time for the proposal, and one time for the role). To have the same password in both, we need to store the password in a class variable.
See https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/schema_migration.rb#L49 and https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/lib/schema_migration.rb#L59

chef/cookbooks/keystone/libraries/helpers.rb
toabctl: No. All cookbooks need to use the same secret. Otherwise the profiler does not work. The reason why I added theses attributes to keystone is:
- we don't have a common openstack barclamp where to add these kind of things
- keystone is the first cookbook for openstack that you usually deploy
- we have a similar thing with `keystone_settings` already there

So after this is merged, I would create follow-up commits for the other services that all use the `profiler_settings()` method to get the options needed for the service config files

cmurphy: I see, does that mean it's all-or-nothing? If osprofiler is enabled for keystone then it's enabled everywhere?

chef/data_bags/crowbar/migrate/keystone/302_add_osprofiler.rb
cmurphy: Thanks for clarifying

chef/cookbooks/keystone/libraries/helpers.rb
toabctl: @cmurphy No. You need to enable it per service. The tracepoints are only generated for theses services, where osprofiler is enabled.
But from a crowbar POV, it's all-or-nothing. If you enable osprofiler in keystone, it will be enabled for all services (where we added support for osprofiler via the config chef templates)

cmurphy: > But from a crowbar POV, it's all-or-nothing

That's what I meant. I'm not sure if there's any advantage or disadvantage of not being able to configure it per-service, I guess the point of having it at all is to be able to trace through all services so this probably makes sense.

toabctl: What we definitly need globally for all services is the `hmac_keys`. We could split the enable/disable later per service but currently I don't see any usecase for that.

nicolasbock:
I was 3 minutes too slow with my approval :cry:

toabctl:
> I was 3 minutes too slow with my approval cry

You'll get another chance! :)

PR #2011, opened 2019-02-11
Ceilometer fixes

toabctl:


toabctl:
Note: there is still the problem that we don't have a storage backend available for Ceilometer.See SCRD-7576

sjamgade:
I assumed ceilometer polling agent is somehow passing data to monasca. is that not so ?

toabctl:
> I assumed ceilometer polling agent is somehow passing data to monasca. is that not so ?

I asked @witekest and there is monaca-ceilometer but we don't use that atm. Given that we have no storage backend, ceilometer is currently useless with Crowbar...

PR #2012, opened 2019-02-11
monasca: Set Kafka log message format version

witekest:
Monasca components still work with old message format version.


toabctl:
@witekest out of interesst - is this fixing a current problem? I wonder how that worked for me without this setting....

witekest:
> @witekest out of interesst - is this fixing a current problem? I wonder how that worked for me without this setting....

I have updated the commit message adding motivation.

PR #2015, opened 2019-02-11
monasca: Use `monasca_db` for installing DB schema (SCRD-7422)

witekest:
Use `monasca_db` schema migration command to install the configuration
database schema for Monasca.

### DIFF
--- a/chef/cookbooks/monasca/recipes/database.rb
+++ b/chef/cookbooks/monasca/recipes/database.rb
@@ -64,29 +64,6 @@ crowbar_pacemaker_sync_mark "wait-monasca_database"
   end
 end
 
-# FIXME(toabctl):the package contains the mon_mysql.sql file
-# would be good if the file is in a extra package
-package "openstack-monasca-api"
-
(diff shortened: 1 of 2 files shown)


chef/cookbooks/monasca/recipes/common.rb
```diff
@@ -13,4 +14,39 @@
 # See the License for the specific language governing permissions and
 # limitation.
 
-### Common steps, if needed
+# Provide monasca-api configuration; needed for database recipe
+
+package "openstack-monasca-api"
```
toabctl: With that, we would install the package also on nodes with the `monasca-agent` and `monasca-log-agent` role because the `common` recipe is included there.
Maybe create a new recipe called `monasca_api_config` and create the config there. that can then be included in the `database` recipe.

witekest: Good catch. I have moved now schema initialization to _monasca-api_ recipe. At that point all necessary pieces are already available: the package, configuration and DB instance.

PR #2018, opened 2019-02-13
tempest: Conditionally install nonrequired plugins

cmurphy:
If a tempest plugin is installed then by default its tests will be run
unless explicitly turned off in tempest.conf. This breaks our
tempestfull CI because we do not install magnum by default. This change
ensures only plugins for barclamps we have applied will be installed.

### DIFF
--- a/chef/cookbooks/tempest/recipes/install.rb
+++ b/chef/cookbooks/tempest/recipes/install.rb
@@ -31,20 +31,36 @@ package "euca2ools"
 
 package "openstack-tempest-test"
 
-["barbican", "cinder", "designate", "heat", "ironic", "keystone", "magnum",
- "manila", "neutron"].each do |service|
-  package "python-#{service}-tempest-plugin"
-end
-
-["keystone", "swift", "glance", "cinder", "neutron", "nova",
- "heat", "ceilometer", "sahara"].each do |component|
+[
+  "barbican",
+  "ceilometer",
+  "cinder",
+  "designate",
+  "glance",
+  "heat",
+  "ironic",
+  "keystone",
+  "magnum",
+  "manila",
+  "neutron",
+  "nova",
+  "sahara",
+  "swift"
+].each do |component|
   package "python-#{component}client"
 end
 
-if node[:kernel][:machine] == "x86_64" &&
-    !node_search_with_cache("roles:monasca-server").empty?
-  ["api", "log-api"].each do |component|
-    package "python-monasca-#{component}"
-  end
-  package "python-monasca-tempest-plugin"
+[
+  "barbican",
+  "cinder",
+  "designate",
+  "heat",
+  "ironic",
+  "keystone",
+  "magnum",
+  "manila",
+  "monasca",
+  "neutron"
+].each do |component|
+  package "python-#{component}-tempest-plugin" if RoleHelper.config_for_role_exists?(component)
 end


chef/cookbooks/tempest/recipes/install.rb
```diff
@@ -31,18 +31,52 @@
...
+ "ceilometer", "sahara"].each do |component|
   package "python-#{component}client"
 end
 
+if RoleHelper.config_for_role_exists?("barbican")
+  package "python-barbicanclient"
+  package "python-barbican-tempest-plugin"
+end
+
+# TODO(colleen) Enable the designate plugin when we have a designate barclamp
```
sjamgade: any specific reason for keeping it commented ?
if there is no barclamp there will be no config

cmurphy: That's true, I can uncomment it.

chef/cookbooks/tempest/recipes/install.rb
```diff
@@ -31,18 +31,51 @@
...
 
-["keystone", "swift", "glance", "cinder", "neutron", "nova",
- "heat", "ceilometer", "sahara"].each do |component|
+["keystone", "glance", "cinder", "neutron", "nova",
+ "ceilometer", "sahara"].each do |component|
   package "python-#{component}client"
 end
 
+if RoleHelper.config_for_role_exists?("barbican")
+  package "python-barbicanclient"
```
dirkmueller: I think we could still install this always, right? its only about the tempest-plugin ...

chef/cookbooks/tempest/recipes/install.rb
```diff
@@ -31,18 +31,51 @@
...
+  package "python-barbican-tempest-plugin"
+end
+
+if RoleHelper.config_for_role_exists?("designate")
```
dirkmueller: two thoughts: a) why is the "enable/disable" in [services_available] in tempest.conf not enough? I don't understand this part? b) DRY.. we can do a forloop like

```
["designate", "ironic",].each |service| do
  if RoleHelper.config_for_role_exists? service
    package "python-#{service}client"
    package "python-#{service}-tempest-plugin"
```

chef/cookbooks/tempest/recipes/install.rb
cmurphy: We don't install it at all without this change. I can remove this line to make the barbican support identical to what we had before, but I didn't think it made sense to install it when barbican isn't deployed.

chef/cookbooks/tempest/recipes/install.rb
cmurphy: a) I went for this approach first because the way tempest works upstream it to run all the tests in all the plugins, so it doesn't really make sense to me to install the plugin and then disable it, plus why install a package you don't need? But also, the issue I was trying to solve was that the magnum tests are always being run, and now that I look, the magnum service *is* disabled in [service_available] and yet that setting is being ignored.
b) I can do that, this file pretty inconsistent about what plugins and clients get installed (there's a sahara client but no tempest plugin for example) so I was trying to make this readable without installing things we don't need or want.

chef/cookbooks/tempest/recipes/install.rb
dirkmueller: ah ok. I wonder if python-#{component}-templest-plugin is missing a requires on python-barbicanclient?

chef/cookbooks/tempest/recipes/install.rb
dirkmueller: a) right, I understand that its supposed to not run - why is it run? is there a typo in the service_available line?

Please note I don't object to the approach of not installing the package (I commented not "needs change" in the review), I think its a good approach. however so far we have two knobs for disabling/enabling testing coverage: installing plugins and the tempest config file that should enable/disable testing for a particular service. doing both seems violation from a cleanliness perspective, hence I'm asking why one is not enough. 

The check conditions are also different "active proposal" vs "has an entry in keystone service catalog" which can cause various funny discrepancies. all I'm looking after (and thats a generic advice) to do this condition only once in one place in one way :)

cmurphy: Okay, you're right to be suspicious. The setting wasn't being respected because the magnum plugin needs this patch: https://review.openstack.org/638619

Other plugins such as the heat plugin have the same problem. We can try to fix them, but I still think that not installing them in the first place is the better solution. That way time and console log space are not wasted on skipping tests.

cmurphy:
Reorganized to DRY things up.

cmurphy:
@dirkmueller can you re-review?

chef/cookbooks/tempest/recipes/install.rb
```diff
@@ -31,20 +31,43 @@
...
+  "keystone",
+  "magnum",
+  "manila",
+  "monasca",
+  "neutron"
+].each do |component|
+  package "python-#{component}-tempest-plugin" if RoleHelper.config_for_role_exists?(component)
 end
 
 if node[:kernel][:machine] == "x86_64" &&
```
witekest: Monasca tempest tests have been moved to a separate repo/package, so we can remove this if-block completely.

cmurphy: Not sure I understand, the monasca tempest tests are not in python-monasca-tempest-plugin?

witekest: They had been in _monasca-api_ and _monasca-log-api_ repos before. That's the reason for the if-block starting at that line, I guess.
Then, as for all other projects, tests have been moved to a dedicated repo.
It should be enough to install _python-monasca-tempest-plugin_ and the client.

toabctl: Also `python-monasca-tempest-plugin` is now in the openstack.group (no longer x86_64 only) so I agree that the whole block can be removed.

cmurphy: done

PR #2021, opened 2019-02-14
mysql: Do not set a custom logfile for mysqld (bsc#1112767)

toabctl:
the mariadb package comes with a logrotate
script (/etc/logrotate.d/mariadb) that currently only rotates the
default /var/log/mysql/mysqld.log file.
So let's use that file instead of a custom path to fix log rotation.

Note: If "slow_query_logging" is enabled, this file is still not
rotated. This will hopefully be fixed in the package itself[1]

[1] https://build.opensuse.org/request/show/676156

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -201,7 +201,7 @@ pacemaker_primitive service_name do
     "check_user" => "monitoring",
     "socket" => "/var/run/mysql/mysql.sock",
     "datadir" => node[:database][:mysql][:datadir],
-    "log" => "/var/log/mysql/mysql_error.log"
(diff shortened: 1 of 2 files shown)


rhafer:
@toabctl Did you by chance check if the supportconfig tools collects the the whole directory? Or do we need a change there as well?

Looks good otherwise.

toabctl:
> @toabctl Did you by chance check if the supportconfig tools collects the the whole directory? Or do we need a change there as well?

@rhafer . Yes. supportconfig collects all files in var/log/mysql . See https://github.com/SUSE-Cloud/supportutils-plugin-suse-openstack-cloud/blob/suse-openstack-cloud-7/suse_openstack_cloud#L293

toabctl:
Backport for stable/5.0 is https://github.com/crowbar/crowbar-openstack/pull/2022

PR #2023, opened 2019-02-15
mysql: improve galera HA setup (bsc#1122875)

aojea:
This patch applies the option "on-marked-down shutdown-sessions"
to our haproxy galera configuration to deal with the situation that
happens keeping connections open when a backend is declared down by
the health check but the mariadb server is still alive

Depends-On: https://github.com/crowbar/crowbar-ha/pull/347

Signed-off-by: aojeagarcia <aojeagarcia@suse.com>

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -350,6 +350,8 @@ ha_servers = ha_servers.each do |n|
   n["fall"] = 2
   # lower the interval checking after first failure is found
   n["fastinter"] = 1000
+  # shutdown connection when backend is marked down
+  n["on_marked_down_shutdown"] = true
 end
 
 haproxy_loadbalancer "galera" do


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -350,6 +350,8 @@
   n["fall"] = 2
   # lower the interval checking after first failure is found
   n["fastinter"] = 1000
+  # shutdown connection when backend is marked down
+  n["onmarkdown"] = n["name"]
```
aojea: I need help here, I'm totally guessing here :sweat_smile:

toabctl: then let's add the WIP label :)

jsuchome: Looks like it should be just `true`

rhafer: Yeah, with the current state of https://github.com/crowbar/crowbar-ha/pull/347 `true` would be the right value here.

toabctl:
commit msg first line should start with `mysql: `

aojea:
I have to test it locally

chef/cookbooks/mysql/recipes/ha_galera.rb
dirkmueller: has been adjusted

PR #2024, opened 2019-02-15
[4.0] mysql: Do not set a custom logfile for mysqld (bsc#1112767)

toabctl:
the mariadb package comes with a logrotate
script (/etc/logrotate.d/mariadb) that currently only rotates the
default /var/log/mysql/mysqld.log file.
So let's use that file instead of a custom path to fix log rotation.

Note: If "slow_query_logging" is enabled, this file is still not
rotated. This will hopefully be fixed in the package itself[1]

[1] https://build.opensuse.org/request/show/676156

(cherry picked from commit aabd06b7a291a2e5a3465238957f32390c8fffdd)
(cherry picked from commit a0bb13aeddef74775ab15610d4b2def6cec8ba0c)


cmurphy:
You have double cherry picks here

> (cherry picked from commit aabd06b)
> (cherry picked from commit a0bb13a)

I think it's okay to just cherry pick the one from master instead of cherry picking the cherry pick

PR #2025, opened 2019-02-15
[4.0] ceilometer: Use pacemaker to handle expirer cron link (bsc#1113107)

toabctl:
With recent openstack-ceilometer packaging changes (Newton[1],
Pike[2]), the ceilometer-expirer cronjob is no longer installed on all
nodes where the openstack-ceilometer-collector package is
installed. Instead the cronjob is installed in
/usr/share/ceilometer//openstack-ceilometer-expirer.cron .
This is needed to avoid parallel runs of the cronjob from different
nodes which lead to database deadlocks.
Now use pacemaker to handle a symlink in /etc/cron.daily so the
cronjob is executed on a single node in the cluster.

[1] https://build.opensuse.org/request/show/676177
[2] https://build.opensuse.org/request/show/676174

### DIFF
--- a/chef/cookbooks/ceilometer/recipes/server_ha.rb
+++ b/chef/cookbooks/ceilometer/recipes/server_ha.rb
@@ -23,6 +23,39 @@ haproxy_loadbalancer "ceilometer-api" do
   action :nothing
 end.run_action(:create)
 
+# setup the expirer cronjob only on a single node to not
+# run into DB deadlocks (bsc#1113107)
+crowbar_pacemaker_sync_mark "wait-ceilometer_expirer_cron"
+
+expirer_transaction_objects = []
+
+ceilometer_expirer_cron_primitive = "ceilometer-expirer-cron"
+pacemaker_primitive ceilometer_expirer_cron_primitive do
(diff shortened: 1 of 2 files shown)


toabctl:
Note:
- not yet tested
- not needed in Master because there is no longer a ceilometer backend which needs an expirer (and the expirer is of course also gone)

chef/cookbooks/ceilometer/recipes/server_ha.rb
```diff
@@ -23,6 +23,39 @@
   action :nothing
 end.run_action(:create)
 
+# setup the expirer cronjob only on a single node to not
+# run into DB deadlocks (bsc#1113107)
+crowbar_pacemaker_sync_mark "wait-ceilometer_expirer_cron"
+
+service_transaction_objects = []
```
vuntz: I'd probably name this `expirer_transaction_objects`, or something like this, to avoid the confusion with the `transaction_objects` variable that comes later and that is used for services. But not a huge deal.

toabctl: Done

toabctl:
Forward-port for SOC8 is https://github.com/crowbar/crowbar-openstack/pull/2028

PR #2027, opened 2019-02-18
ceilometer: Install Monasca publisher (SCRD-7531)

sjamgade:
Add support of ceilosca (monasca publisher for ceilometer)


jodavis-suse:
Looks good to me.  Wish we could easily backport this to C8. :)

jodavis-suse:
> Looks good to me. Wish we could easily backport this to C8. :)

Of course, the mkcloud build doesn't agree with me.  I have no clue what this means "FATAL: NoMethodError: undefined method `[]' for nil:NilClass"

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -103,6 +105,14 @@
     notifies :reload, resources(service: "apache2")
 end
 
+monasca_server = node_search_with_cache("roles:monasca-server").first
+if monasca_server.nil?
+  Chef::Log.warn("No monasca-server found. Skip Ceilometer setup.")
+  return
```
toabctl: Is that correct? We would skip setting up the other ceilometer templates in case monasca is not installed.

jodavis-suse: It is IFF we aren't supplying another option for a storage backend for Ceilometer.  We don't want to go the route of having to package and support Gnocchi, and the database backend is very problematic (bug 1118111, bug 1113107, etc).  If we don't provide another option for storage, then we should not set up Ceilometer. (note the ceilometer-ceilosca-monasca combination is the only option in CLM as well)

witekest: Yes, that's correct. Monasca is the only supported backend.
Please compare the [user story description](https://jira.suse.de/browse/SCRD-7531).

toabctl: The we should skip the whole ceilometer baclamp if monasca is not deployed. So ceilometer depends on monasca.

jgrassler: Unfortunately it does as of Rocky, yes.

witekest: Does this line address that? 
https://github.com/crowbar/crowbar-openstack/pull/2027/files#diff-98abc126c10c8146eb49c90cdaa95956R15

toabctl: - I don't think that is enough. There is eg also https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/models/ceilometer_service.rb#L77 . But I'm not sure how the deps in crowbar work.

chef/cookbooks/ceilometer/templates/default/monasca-field-definitions.yaml.erb
```diff
@@ -0,0 +1,228 @@
+dimensions:
```
toabctl: Is there anything dynamic in this file? Maybe we should add it to the package?

chef/cookbooks/ceilometer/recipes/common.rb
witekest: I've added `monasca` to barclamp requirements.

chef/cookbooks/ceilometer/templates/default/monasca-field-definitions.yaml.erb
witekest: It is a configuration file. I have kept it the same way as deployed in [Ardana](https://github.com/ArdanaCLM/ceilometer-ansible/blob/master/roles/_CEI-CMN/templates/monasca-field-definitions.yaml.j2).

chef/cookbooks/ceilometer/recipes/server.rb
```diff
@@ -115,11 +124,23 @@
   port keystone_settings["admin_port"]
   auth register_auth_hash
   user_name keystone_settings["service_user"]
-  project_name keystone_settings["service_tenant"]
+  project_name monasca_project
   role_name "admin"
   action :add_access
 end
 
+keystone_register "give ceilometer user access" do
```
cmurphy: This resource has the same name as the one above it, which means it's actually merging the the two resources instead of creating a unique one. Please give them unqiue names like "grant ceilometer user admin role on monasca project" and "grant ceilometer user monasca-user role on monasca project"

chef/cookbooks/ceilometer/templates/default/monasca-field-definitions.yaml.erb
cmurphy: > Maybe we should add it to the package?

Even if it's static, doesn't it make more sense to keep it in git? We could make it a file (in the files/ directory) instead of a template. Or is this copied from the source code? In that case it would make more sense to have the package install it.

sjamgade: I guess we can drop this file as upstream dropped it in https://review.openstack.org/#/c/637027/

@jodavis-suse may have more info on that

chef/cookbooks/ceilometer/recipes/server.rb
sjamgade: done

chef/cookbooks/ceilometer/templates/default/monasca-field-definitions.yaml.erb
witekest: > I guess we can drop this file as upstream dropped it in https://review.openstack.org/#/c/637027/
> 
> @jodavis-suse may have more info on that

We still need this. I've cleaned the config files though to reflect meters removed from Ceilometer.

witekest: > > Maybe we should add it to the package?
> 
> Even if it's static, doesn't it make more sense to keep it in git? We could make it a file (in the files/ directory) instead of a template. Or is this copied from the source code? In that case it would make more sense to have the package install it.

We don't have this file in that shape upstream. I moved it to `files` directory.

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -1,3 +1,11 @@
+monasca_server = node_search_with_cache("roles:monasca-server").first
+if monasca_server.nil?
+  Chef::Log.warn("No monasca-server found. Skip Ceilometer setup.")
```
toabctl: Given that ceilometer depends on monasca, this should never happen so I think it can be removed.

toabctl: but then the following code does not work. so forget my comment.

witekest: In the last comment you report about the case when although the Monasca dependency is satisfied the `monasca_server` variable still cannot be set. If we remove that check here Ceilometer won't be configured correctly.

I must admit I still don't understand handling the dependencies between barclamps but I'd rather make the code more robust here (and in the remaining recipes) instead of remove this check.

dirkmueller: I think the chef/cookbooks/ceilometer/metadata.rb needs to be updated

chef/cookbooks/ceilometer/recipes/common.rb
```diff
@@ -10,6 +18,7 @@
 memcached_instance("ceilometer-server") if is_controller
 
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+monasca_project = node[:monasca][:service_tenant]
```
dirkmueller: this should be using "monasca_server" instead of "node"

chef/cookbooks/ceilometer/recipes/server.rb
```diff
@@ -79,7 +79,16 @@
...
+cookbook_file node[:ceilometer][:monasca][:field_definitions] do
+  source "monasca-field-definitions.yaml"
+  owner "root"
+  group node[:ceilometer][:group]
+  mode "0640"
+  notifies :restart, resources(service: "apache2")
+end
+
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+monasca_project = node[:monasca][:service_tenant]
```
dirkmueller: same nere, :monasca only exists on monasca_server node

chef/cookbooks/ceilometer/recipes/common.rb
witekest: > I think the chef/cookbooks/ceilometer/metadata.rb needs to be updated

Already done [here](https://github.com/crowbar/crowbar-openstack/pull/2027/files#diff-98abc126c10c8146eb49c90cdaa95956R15).

chef/cookbooks/ceilometer/recipes/common.rb
witekest: Thanks, fixed.

chef/cookbooks/ceilometer/recipes/server.rb
witekest: Fixed

witekest:
> This happens because the config file resource is not defined. That's because it is skipped (because of this PR) in `ceilometer/recipe/common.rb` but needed in `ceilometer/recipes/agent.rb`

I moved the checks to from common to all other recipes.

cmurphy:
The CI is failing: `Failed to talk to service proposal create: 412: {"error":"Cannot create a proposal for ceilometer, please create a monasca proposal first."}` - if ceilometer hard-depends on monasca now, then mkcloud either needs to be updated to deploy monasca by default or to not deploy ceilometer by default.

toabctl:
Ceilometer is now only deployed when monasca is deployed (on SOC9). See https://github.com/SUSE-Cloud/automation/commit/6e43def153d75d9b1358c994c8968752b940be40

witekest:
> Ceilometer is now only deployed when monasca is deployed (on SOC9). See [SUSE-Cloud/automation@6e43def](https://github.com/SUSE-Cloud/automation/commit/6e43def153d75d9b1358c994c8968752b940be40)

I have also rebuilt Jenkins jobs with `want_monasca_proposal=1` which runs nicely.

toabctl:
@dirkmueller @cmurphy could you re-review please?

PR #2028, opened 2019-02-19
[5.0] ceilometer: Use pacemaker to handle expirer cron link (bsc#1113107)

toabctl:
With recent openstack-ceilometer packaging changes (Newton[1],
Pike[2]), the ceilometer-expirer cronjob is no longer installed on all
nodes where the openstack-ceilometer-collector package is
installed. Instead the cronjob is installed in
/usr/share/ceilometer//openstack-ceilometer-expirer.cron .
This is needed to avoid parallel runs of the cronjob from different
nodes which lead to database deadlocks.
Now use pacemaker to handle a symlink in /etc/cron.daily so the
cronjob is executed on a single node in the cluster.

[1] https://build.opensuse.org/request/show/676177
[2] https://build.opensuse.org/request/show/676174

(cherry picked from commit fc5971cbc019d1bf93723f401cd083647feae737)


jodavis-suse:
nice

PR #1707, opened 2018-07-18
Designate Barclamp

dirkmueller:

PR #2033, opened 2019-02-21
Designate barclamp

sjamgade:
PR #1707 was fairly long, so lets use this to get this merged


dirkmueller:
thanks.

toabctl:
Do we still care about commit messages? No message and things like `XXX neutron net-update` are not very useful imo...

PR #2034, opened 2019-02-21
glance: run db sync and load metadefs

cmurphy:
In bdf0cce we removed the glance-registry recipe which also removed the
glance-manage db_sync and glance-manage db_load_metadefs executions. We
didn't notice this was a problem at first because it seems that the
glance API service is capable of running the db sync itself, but the
metadef loading was still not happening. This causes tempest failures
because the contents of GET /v2/metadefs/resource_types will be empty.
This patch fixes the issue by adding back this setup steps for the API
service.

### DIFF
--- a/chef/cookbooks/glance/recipes/api.rb
+++ b/chef/cookbooks/glance/recipes/api.rb
@@ -175,4 +175,48 @@ end
 
 crowbar_pacemaker_sync_mark "create-glance_register_service" if ha_enabled
 
+template node[:glance][:manage][:config_file] do
+  source "glance-manage.conf.erb"
+  owner "root"
+  group node[:glance][:group]
+  mode 0o640
+end
+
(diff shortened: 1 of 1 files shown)


chef/cookbooks/glance/recipes/api.rb
```diff
@@ -67,7 +67,7 @@
   source "glance-api.conf.erb"
   owner "root"
   group node[:glance][:group]
-  mode 0640
+  mode 0o640
```
toabctl: why this change?

cmurphy: My mistake, hound was angry about the other template and I fixed the wrong one.

cmurphy: This is fixed, can you re-review?

PR #2035, opened 2019-02-21
SES integration and (a bit of) Ceph cleanup

skazi0:
Together with https://github.com/crowbar/crowbar-core/pull/1799 this is an alternative to: https://github.com/crowbar/crowbar-openstack/pull/1977

This PR enable barclamps to use SES settings imported with https://github.com/crowbar/crowbar-core/pull/1799

After this PR the supported Ceph deployments would be:
- SES with configuration imported from yaml file
- Manual Ceph deployment with complete pre-configuration (users and pools)

Not supported (but should keep working):
- Manual Ceph (empty) with automatic configuration done by Crowbar (using admin_keyring)
- Internal Ceph cluster deployed by Crowbar/Ceph barclamp

Open points/ideas:
- [x] using the same user/key for multiple services (access rights to keyring file)
- [ ] nova ephemeral ceph options (?)
- [x] empty keyring for nova (no user and key in ses yaml)
- [x] re-enable internal ceph support
- [ ] add ceph sanity checks similar to those which were done in external-ceph-with-autosetup scenario before
- [ ] set use_ses=true in newly added cinder backends if ses config is available (client side/JavaScript -> tricky)


chef/cookbooks/glance/recipes/ceph.rb
```diff
@@ -17,66 +17,45 @@
...
-  admin_keyring admin_keyring
+if admin_keyring.empty? || !File.exist?(admin_keyring)
+  Chef::Log.error("Ceph keyring file for Glance not set or missing.")
+  return
 end
+
+# If ceph.conf and admin keyring are available
+# we have to check ceph cluster status
+cmd = ["ceph", "-k", admin_keyring, "-c", ceph_conf, "-s"]
+check_ceph = Mixlib::ShellOut.new(cmd)
```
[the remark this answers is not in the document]
hemna: If we are removing the ability to deploy ceph itself, and are using the SES integration, then we don't have to 
worry about admin keyring, because we should have a client keyring for each of the services as described in the ses_config.yml that comes back from SES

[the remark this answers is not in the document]
skazi0: true. I removed `admin_keyring` altogether.

chef/cookbooks/cinder/recipes/ceph.rb
```diff
@@ -17,124 +17,18 @@
...
-      Chef::Log.info("Using external ceph cluster for cinder #{volume[:backend_name]} backend, with automatic setup.")
-    else
-      Chef::Log.info("Using external ceph cluster for cinder #{volume[:backend_name]} backend, with no automatic setup.")
-      next
-    end
-
-    cmd = ["ceph", "-k", admin_keyring, "-c", ceph_conf, "-s"]
-    check_ceph = Mixlib::ShellOut.new(cmd)
-
-    unless check_ceph.run_command.stdout.match("(HEALTH_OK|HEALTH_WARN)")
```
stefannica: Same thing here: shouldn't we keep this check ?

skazi0: Previously we had following scenarios:
1. internal (fully crowbar managed) ceph cluster
2. external (pre-deployed but not pre-configured) ceph cluster with automatic configuration done by crowbar
3. external (fully deployed and configured) ceph cluster

This change removes #1 and #2 and adds #4 = external (SES managed) ceph cluster.
The health checks in previous implementation were done only in case #2 which we remove completely.

chef/cookbooks/glance/recipes/ceph.rb
```diff
@@ -17,66 +17,34 @@
...
+    Chef::Log.fatal(message)
+    raise message
   end
+  ceph_conf = SesHelper.ceph_conf_path
+  # Copy settings from SES config
+  node.default[:glance][:rbd][:store_ceph_conf] = ceph_conf
+  node.default[:glance][:rbd][:store_pool] = ses_config["glance"]["rbd_store_pool"]
+  node.default[:glance][:rbd][:store_user] = ses_config["glance"]["rbd_store_user"]
 
-  # If ceph.conf and admin keyring will be available
```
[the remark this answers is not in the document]
skazi0: see comment above.

chef/cookbooks/glance/recipes/ceph.rb
```diff
@@ -17,66 +17,34 @@
...
-  check_ceph = Mixlib::ShellOut.new(cmd)
-
-  unless check_ceph.run_command.stdout.match("(HEALTH_OK|HEALTH_WARN)")
-    Chef::Log.info("Ceph cluster is not healthy, skipping the ceph setup for glance")
-    return
+  ses_config "glance" do
+    action :create
   end
+else # Manually deployed and configured Ceph
+  ceph_conf = node[:glance][:rbd][:store_ceph_conf]
```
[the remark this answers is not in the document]
skazi0: My understanding was that the user/operator needs to fully deploy and configure ceph cluster and distribute all needed configs to respective nodes. The only thing done via crowbar in this case is selection of users/pools to be used for services.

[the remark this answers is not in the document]
skazi0: Note that this scenario was there before this change and it's left intact.

chef/cookbooks/cinder/recipes/ceph.rb
stefannica: Thanks, I think I understand better the previously supported use-cases now, but I still don't see that as a good reason for not reusing this kind of check. I mean, let's think about this for a minute: the Crowbar user uploads a ceph configuration via the UI, which you can even validate syntactically and semantically, but still that gives you no guarantee that: a. the external SES cluster is accessible from the glance/cinder/nova nodes and b. the user and keyring configured for a particular service are valid. If you ignore both a. and b. while applying the barclamp, you end up with a broken deployment, which will be far more difficult to troubleshoot after the fact.

On the other hand, if you could run a simple check that validates both a. and b. (which is what looks like `["ceph", "-k", admin_keyring, "-c", ceph_conf, "-s"]` is doing), and you fail early with a nice error message, you'll spare the admin a lot of pain and time. I just quickly did that on a virtual Ardana CI setup and it works wonderfully:

```
# ceph -n client.cloud-ardana-ci-slot20-glance -c /etc/ceph/ceph.conf -s
  cluster:
    id:     7e8be23a-87f0-38e7-acea-f2666ba8d422
    health: HEALTH_OK
 
  services:
    mon:     1 daemons, quorum openstack-ses-cloud-ardana-ci-slot20
    mgr:     openstack-ses-cloud-ardana-ci-slot20(active)
    mds:     cephfs-1/1/1 up  {0=openstack-ses-cloud-ardana-ci-slot20=up:active}
    osd:     1 osds: 1 up, 1 in
    rgw:     1 daemon active
    rgw-nfs: 1 daemon active
 
  data:
    pools:   11 pools, 928 pgs
    objects: 350 objects, 706MiB
    usage:   1.73GiB used, 38.2GiB / 39.9GiB avail
    pgs:     928 active+clean
```

stefannica: Better even, you could encode that directly in the SES resource provider and reduce code duplication.

chef/cookbooks/glance/recipes/ceph.rb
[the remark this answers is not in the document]
stefannica: Thanks, yes, that makes sense now.

chef/cookbooks/cinder/templates/default/cinder.conf.erb
```diff
@@ -194,14 +194,9 @@ rbd_pool = <%= volume['rbd']['pool'] %>
     <% unless volume['rbd']['user'].empty? -%>
 rbd_user = <%= volume['rbd']['user'] %>
     <% end -%>
-    <% if volume['rbd']['use_crowbar'] -%>
-rbd_exclusive_cinder_pool = True
```
dirkmueller: This should be set imho in the external ses case..

skazi0: What if someone wants to use the same pool for glance and cinder?

chef/cookbooks/nova/recipes/ceph.rb
```diff
@@ -17,78 +17,46 @@
...
+  ceph_conf = volume[:rbd][:config_file]
 
-  if volume[:rbd][:use_crowbar]
-    ceph_conf = "/etc/ceph/ceph.conf"
-    admin_keyring = "/etc/ceph/ceph.client.admin.keyring"
-  else
-    ceph_conf = volume[:rbd][:config_file]
-    admin_keyring = volume[:rbd][:admin_keyring]
-
+  unless volume[:rbd][:uses_ses]
```
stefannica: I think you have a typo here: it's `use_ses`

chef/cookbooks/cinder/recipes/ceph.rb
hemna: We don't really have an admin_keyring anymore, so you'd have to pick one of the keyrings that are gleaned from the ses_config.yml.  The check is still a good thing to do, to make sure that 1) the keyring is good and that 2) we can talk to the external ceph cluster.

chef/cookbooks/cinder/templates/default/cinder.conf.erb
hemna: The default in cinder is False currently.

chef/data_bags/crowbar/migrate/cinder/303_remove_ceph_admin_keyring.rb
```diff
@@ -0,0 +1,17 @@
...
+    backend["rbd"].delete("admin_keyring")
+  end
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["volume_defaults"]["rbd"]["admin_keyring"] = template_attrs["volume_defaults"]["rbd"]["admin_keyring"]
+  attrs["volumes"].each do |backend|
+    next unless backend.key? "rbd"
+    backend["rbd"]["admin_keyring"] = template_attrs["volume_defaults"]["rbd"]["admin_keyring"]
```
hemna: When do we ever have an admin keyring?   For ceph clusters via SES, we won't have an admin keyring, and I don't think admins would want to provide an admin keyring for security reasons.

skazi0: this is the "downgrade" part of migration.

skazi0:
@dirkmueller @stefannica I reworked the PR to keep old functionality in (as much as possible).
**NOTE** This was not tested yet at all!

chef/cookbooks/cinder/templates/default/cinder.conf.erb
dirkmueller: I understand the upstream default, I also understand that its entirely inadequate for deployments at scale. even in a mid sized deployment the disk space checking in the hot path will cause performance to be unacceptably low. we should set it imho.

PR #2040, opened 2019-02-22
monasca: Set heap size for Elasticsearch

witekest:
The default installation of Elasticsearch is configured with a 1 GB
heap. For just about every deployment, this number is usually too small
[1]. This change allows setting `heap_size`.

elasticearch.tunables.max_open_files_soft_limit isn't used anymore. It
gets removed here, instead `limit_nofile` should be used.

elasticearch.tunables.memory_lock has is actually decoupled from
elasticsearch.bootstrap_memory_lock which it was supposed to control.
The latter has anyway a recommended default value. The setting is very
technical, so it makes sense to hide it from end user.

[1] https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html

### DIFF
--- a/chef/cookbooks/monasca/recipes/elasticsearch.rb
+++ b/chef/cookbooks/monasca/recipes/elasticsearch.rb
@@ -54,7 +54,8 @@ template "/etc/systemd/system/elasticsearch.service.d/override.conf" do
   variables(
     elasticsearch_limit_nproc: node[:monasca][:elasticsearch][:tunables][:limit_nproc],
     elasticsearch_limit_nofile: node[:monasca][:elasticsearch][:tunables][:limit_nofile],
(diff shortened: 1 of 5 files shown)


witekest:
> You write in the commit msg: `` The default installation of Elasticsearch is configured with a 1 GB heap. For just about every deployment, this number is usually too small [1]. This change allows setting `heap_size`. ``
> 
> But then you set the default heap size to 1GB? While we know that it is too small?

I've seen problems starting ES in mkcloud env with too small sized nodes and not being able to reserve the required heap space. But you're probably right that it's better to provide a more reasonable default of 4GB and configure a lower value in mkcloud and/or CI. In ansible installer we originally had 10GB and reduced it to 4GB. The official documentation recommends to give 50% of the available memory, which I have added to [Crowbar documentation issue](https://jira.suse.de/browse/SCRD-7786).

I'll update the default value to 4GB here.

PR #2042, opened 2019-02-25
designate: start/run mdns service and install rndc/pools files

sjamgade:
designate uses this internal dns server to perform zone sync between
master.

Without this mdns, desingate will not be able to perform zone xfers
(AXFR), neither primary or secondary.


dirkmueller:
In the current usage that would be the worker role, not the api role. Why do you put it on api?

sjamgade:
what do you mean by current usage ?

IIUC the worker/central talk to mdns over rabbit based rpc and the (m)dns server listens on 0.0.0.0 so I assumed we could put it anywhere. 
Could you point the hole in my understanding ?

chef/cookbooks/designate/recipes/common.rb
```diff
@@ -56,3 +58,45 @@
...
+      "host" => "127.0.0.1",
+      "port" => 53,
+      "rndc_host" => dnsmaster,
+      "rndc_port" => 953,
+      "rndc_key_file" => "/etc/designate/rndc.key"
+    }
+  }]
+}]
+
+file "/etc/designate/pools.crowbar.yaml" do
```
dirkmueller: this file should probably be a template not a file resource and it is missing from this PR, did you forget to git add it ?

sjamgade: That would be an empty template, so included it as `file` resource, which create a file with contents of `content` variable

sjamgade:
@dirkmueller I am not rendering a template, rather just dumping `pools` as yaml in the file

chef/cookbooks/designate/recipes/common.rb
```diff
@@ -56,3 +58,45 @@
...
+  "description" => "Default BIND9 Pool",
+  "id" => "794ccc2c-d751-44fe-b57f-8894c9f5c842",
+  "attributes" => {},
+  "ns_records" => [{ "hostname" => "#{dnsserv[:fqdn]}.", "priority" => 1 }],
+  "nameservers" => [{ "host" => "127.0.0.1", "port" => 53 }],
+  "also_notifies" => dnsslaves.map { |ip| { "host" => ip, "port" => 53 } },
+  "targets" => [{
+    "type" => "bind9",
+    "description" => "BIND9 Server 1",
+    "masters" => [{ "host" => mdnsaddr, "port" => 5354 }],
```
dirkmueller: how's that supposed to work? this is the address of node's admin interfce, but you're starting mdns in the api chef role (which would be on another node). Unless I'm missing something the  ip address should be referring to the api VIP or the mdns should be started here..

chef/cookbooks/designate/recipes/common.rb
dirkmueller: ah, right, thanks, got it now.

chef/cookbooks/designate/recipes/common.rb
sjamgade: thats correct switched to data from `network_settings`

chef/cookbooks/designate/recipes/common.rb
```diff
@@ -56,3 +58,49 @@
     memcached_servers: memcached_servers
   )
 end
+
+dnsserv = node_search_with_cache("roles:dns-server").first
+dnsmaster = dnsserv[:dns][:master_ip]
+dnsslaves = dnsserv[:dns][:slave_ips]
+mdnsaddr = if node[:designate][:ha][:enabled]
```
dirkmueller: you're setting this to a port number but call it "addr". thats confusing.

chef/cookbooks/designate/recipes/common.rb
dirkmueller: but you're using the port as host... thats not gonna work ?

dirkmueller:
@sjamgade sure it can run anywhere, but it should only run once imho. and hence it should be imho on the backend (because its the hidden DNS master that the actual dns-server should sync against). but I'm not debating about that too much - please note that the pools.yaml is created as part of a different role so I think it would make most sense to have mdns run on the same node where that pools.yaml is?

chef/cookbooks/designate/recipes/common.rb
```diff
@@ -56,3 +58,49 @@
...
+    "masters" => [{ "host" => mdnsaddr, "port" => 5354 }],
+    "options" => {
+      "host" => "127.0.0.1",
+      "port" => 53,
+      "rndc_host" => dnsmaster,
+      "rndc_port" => 953,
+      "rndc_key_file" => "/etc/designate/rndc.key"
+    }
+  }]
+}]
```
dirkmueller: where is the designate-manage code that loads this pools.yaml?

sjamgade: I dont intend to do that as the customer might want to have some custom config, 
the automation repo already does that

https://github.com/SUSE-Cloud/automation/blob/2fce65dff3f44f8f1047109b94c0d288e561922f/scripts/qa_crowbarsetup.sh#L4043

sjamgade:
@dirkmueller I have moved the code to a separate recipe for mdns and added it to the role_designate_api. so that should put mdns and the file on the same node. 

Also the host address should be fixed now

chef/cookbooks/designate/recipes/common.rb
dirkmueller: Hmm, interesting, I would not have done it that way. I think you can run it again after initial deployment, right? so we should run it ocne and the customer wanting to modify can still recreate the pool (as long as it isn't used)

chef/cookbooks/designate/libraries/helpers.rb
```diff
@@ -27,6 +27,7 @@ def network_settings(node)
 
       @network_settings ||= {
         ip: @ip,
+        cluster_admin_ip: @cluster_admin_ip,
```
dirkmueller: you could call this mdns_bind_host and depending on ha_enabled set it to cluster_admin_ip or ip

chef/cookbooks/designate/recipes/mdns.rb
```diff
@@ -0,0 +1,69 @@
...
+mdnsaddr = if node[:designate][:ha][:enabled]
+  network_settings[:cluster_admin_ip]
+else
+  network_settings[:ip]
+end
+
+pools = [{
+  "name" => "default-bind",
+  "description" => "Default BIND9 Pool",
+  "id" => "794ccc2c-d751-44fe-b57f-8894c9f5c842",
```
dirkmueller: hardcoded id?

chef/cookbooks/designate/recipes/mdns.rb
```diff
@@ -0,0 +1,69 @@
...
+  "attributes" => {},
+  "ns_records" => [{ "hostname" => "#{dnsserv[:fqdn]}.", "priority" => 1 }],
+  "nameservers" => [{ "host" => "127.0.0.1", "port" => 53 }],
```
dirkmueller: this should be the list of dns_server ips, not 127.0.0.1?

chef/cookbooks/designate/recipes/mdns.rb
```diff
@@ -0,0 +1,69 @@
...
+  }]
+}]
+
+file "/etc/designate/pools.crowbar.yaml" do
+  owner "root"
+  group node[:designate][:group]
+  mode "0640"
+  content pools.to_yaml
+end
+
```
dirkmueller: call designate-manage here when the file is created for the first time?

sjamgade: should we update the pool if the file gets is modified because the dns proposal changed ?

chef/cookbooks/designate/recipes/mdns.rb
sjamgade: One could have multiple pools in desginate. And
Desginate needs to have a default pool, this pools
id is hardcoded in the designate conf. By reusing that
id we let designate know about crowbar's deployement of
dns servers looks like.
This pool id can be generated by in proposal, but this will change
with every delete/create cycle of proposal. This might mess 
up the designate configuration.
 
if we intend to call `designate-manage pool update --file` with the changed config

sjamgade:
@dirkmueller I have updated the PR based on our discussion

chef/cookbooks/designate/recipes/mdns.rb
```diff
@@ -0,0 +1,76 @@
...
+  "name" => "default-bind",
+  "description" => "Default BIND9 Pool",
+  "id" => "794ccc2c-d751-44fe-b57f-8894c9f5c842",
+  "attributes" => {},
+  "ns_records" => [{ "hostname" => "#{dns[:fqdn]}.", "priority" => 1 }],
+  "nameservers" => dnsservs.map { |ip| { "host" => ip, "port" => 53 } },
+  "also_notifies" => dnsslaves[1, dnsslaves.length].map { |ip| { "host" => ip, "port" => 53 } },
+  "targets" => [{
+    "type" => "bind9",
+    "description" => "BIND9 Server 1",
```
dirkmueller: This could be maybe a better description but I don't have a good suggestion.

PR #2005, opened 2019-02-07
neutron: Added option to use L3 HA with Keepalived

dirkmueller:
This is a fixed version of https://github.com/crowbar/crowbar-openstack/pull/1550 which seems to rot without reviews for a while already.

neutron-ha-tool helps moving routers from one node to other active
nodes. This is triggered when the L3 agent on that particular network
node is down. This behavior is not compatible with L3 HA.

L3 HA implementation uses one master router and several others in
standby. If a router/agent is down, it will be discovered through VRRP
protocol [1] and another one will take the master role.

Both L3 HA and neuton-ha-tool are incompatible because they both has
their own way of balancing routers on fallen nodes.

This implementation lets the user configure L3 HA, which will disable
neutron-ha-tool. If L3 HA is deactivated, neutron-ha-tool comes back.

Notice that switching between neutron-ha-tool and L3 HA is possible, but
the routers created are not updated to reflect this change. If they were
created as HA routers they will remain that way. The contrary is also
true. It is up to the user to manually change those routers or delete
them and create new ones.

Co-Authored-By: Itxaka <igarcia@suse.com>
Co-Authored-By: Abel Navarro <abel.navarro@gmail.com>

[1] https://wiki.openstack.org/wiki/Neutron/L3_High_Availability_VRRP

PR #2043, opened 2019-02-26
[4.0] neutron: Added option to use L3 HA with Keepalived

dirkmueller:
neutron-ha-tool helps moving routers from one node to other active
nodes. This is triggered when the L3 agent on that particular network
node is down. This behavior is not compatible with L3 HA.

L3 HA implementation uses one master router and several others in
standby. If a router/agent is down, it will be discovered through VRRP
protocol [1] and another one will take the master role.

Both L3 HA and neuton-ha-tool are incompatible because they both has
their own way of balancing routers on fallen nodes.

This implementation lets the user configure L3 HA, which will disable
neutron-ha-tool. If L3 HA is deactivated, neutron-ha-tool comes back.

Notice that switching between neutron-ha-tool and L3 HA is possible, but
the routers created are not updated to reflect this change. If they were
created as HA routers they will remain that way. The contrary is also
true. It is up to the user to manually change those routers or delete
them and create new ones.

Co-Authored-By: Itxaka <igarcia@suse.com>
Co-Authored-By: Abel Navarro <abel.navarro@gmail.com>

[1] https://wiki.openstack.org/wiki/Neutron/L3_High_Availability_VRRP

(cherry picked from commit 6c2034309569d66099f49a3de8697a812c9bf176)

Backport-Of: #2005


AbelNavarro:
[L3 HA](https://trello.com/c/OJyN9gjj/155-l3-ha)

PR #2044, opened 2019-02-27
horizon: Fix configuration loading for Monasca (SCRD-8179)

witekest:
Monasca plugin should only be installed when all configuration options
are already available. Otherwise if Horizon barclamp is installed first
and Monasca afterwards the installation fails with the following error
message:

  FATAL: NoMethodError: undefined method `[]' for nil:NilClass

  Relevant File Content:

  22>> grafana_password = monasca_cfg["db_grafana"]["password"]

### DIFF
--- a/chef/cookbooks/horizon/recipes/monasca_ui.rb
+++ b/chef/cookbooks/horizon/recipes/monasca_ui.rb
@@ -18,8 +18,7 @@ if monasca_server.nil?
   Chef::Log.warn("No monasca-server found.")
   return
 end
-monasca_cfg = Barclamp::Config.load("openstack", "monasca")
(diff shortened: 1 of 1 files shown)


toabctl:
Hm. I still don't understand in which case this happens

witekest:
> Hm. I still don't understand in which case this happens

Steps to reproduce:
#. Commit Horizon barclamp.
#. Commit Monasca barclamp.

As the error message suggests configuration options are still not initialized at that stage. In particular `monasca_cfg["db_grafana"]`.

dirkmueller:
that's surprising.. sounds like a bug elsewhere.

witekest:
Thanks, the problem was in loading Monasca configuration. I've updated the commit message.

PR #2046, opened 2019-02-28
openstack: Fix depedency ordering name for galera

dirkmueller:
The correct resource to base ordering on is ms-galera not galera.


sjamgade:
ICBW but would this break if there is no database ha ?

jgrassler:
> ICBW but would this break if there is no database ha ?

@dirkmueller can you answer this? Right now, I'd still +1 if it did break because having a HA cloud without database HA is a bit of a fringe case and without this change HA clouds are broken across the board. If it broke for a cloud without database HA we should create a followup though.

dirkmueller:
@sjamgade the pacemaker resource is automatically removing references to resources that don't exist in the cluster (which is why that bug occurred)

PR #2047, opened 2019-02-28
rabbit: fix mirroring regex

Itxaka:
When the regex for mirroring queues was introduced on
341f6b059b33690e607e9a3da83cf250c97323d0 a typo was introduced
which skipped some of the default queues.

Restore it to the proper regex


toabctl:
retriggered HA CI

Itxaka:
retriggered!

dirkmueller:
retriggering doesn't help if it relies on a bug that is in this tree, the patch would need to be rebased..

PR #2051, opened 2019-03-04
designate: add rndc_secret_key feild to proposal

sjamgade:
Designate needs an rndc key to CRUD zones on the real bind9 master
server, and updates to bind9 can only be sent via a key that is
mentioned in its(bind9) config.

So the key is supposed to be shared across the nodes, (admin and
designate-worker) therefore adding it to the proposal.  one could
rsync the key between the nodes but this opens up differnt
types of race conditions/syncronization problems.

This key will be used while creating the key-file on admin and
designate-worker node.

the openssl module is part of ruby standard library. And using md5 hexdigest
as that is what rndc expects, a normal 6/8 charachter password was
deemed invalid during trial runs

### DIFF
--- a/chef/data_bags/crowbar/template-designate.json
+++ b/chef/data_bags/crowbar/template-designate.json
@@ -11,6 +11,7 @@
       "service_user": "designate",
       "service_password": "",
       "memcache_secret_key": "",
+      "rndc_secret_key": "",
       "api": {
         "protocol": "http",
         "bind_open_address": true,
--- a/chef/data_bags/crowbar/template-designate.schema
+++ b/chef/data_bags/crowbar/template-designate.schema
@@ -20,6 +20,7 @@
             "service_user": { "type": "str", "required": true },
             "service_password": { "type": "str", "required": true },
             "memcache_secret_key": { "type": "str", "required": true },
+            "rndc_secret_key": { "type": "str", "required": true },
             "api": {
               "type": "map", "required": true, "mapping": {
                 "protocol": { "type" : "str", "required" : true },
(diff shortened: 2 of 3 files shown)


toabctl:
@sjamgade typo in the commit msg `feild` instead of `field`

chef/data_bags/crowbar/template-designate.json
```diff
@@ -11,6 +11,7 @@
       "service_user": "designate",
       "service_password": "",
       "memcache_secret_key": "",
+      "rndn_secret_key": "",
```
toabctl: Don't we need a migration ?

sjamgade: ICBW but given this is a first time designate will be deployed by customers, so the proposal should not exist

toabctl: ok. works for me.

toabctl: Where do you use this new attribute?

sjamgade: this attribute will be used in different PR. I am trying to limit the size of PR.

sjamgade: that was a typo, now fixed.

toabctl: > this attribute will be used in different PR. I am trying to limit the size of PR.

This PR is so small. I don't see why adding the usecase for the attribute would hurt here.
You can add another commit and if the PR is then very large, we can still split it into multiple PRs.

sjamgade:
closing this in favor of https://github.com/crowbar/crowbar-core/pull/1804

PR #2052, opened 2019-03-04
tempest: Improve error message when getting the floating ID

toabctl:
Print stdout and stderr in case the openstack CLI call fails to get
the floating network ID.
Currently it's difficult to know why getting the floating network ID
fails. The error message is just:

RuntimeError: Cannot fetch ID of floating network

This currently happens in the monasca CI job.


toabctl:
https://ci.suse.de/job/openstack-mkcloud/159371/console is an example where this PR might help for debugging.

PR #2057, opened 2019-03-13
neutron: enable lbaasv2 again

dirkmueller:


dirkmueller:
(just for seeing if it would be feasible to get lbaasv2 for GM)

toabctl:
Closing. Use https://github.com/crowbar/crowbar-openstack/pull/2060 instead which has a proper revert commit.

PR #2065, opened 2019-03-19
rabbit: raise failure-timeout

dirkmueller:
The failure timeout of 30s is far too low. Essentially it means that a
failed node is considered ready after 30s. Given that any start or
stop operation takes considerably more than 30s. We should only expire
failures after around 30 minutes to prevent flapping services.


JanZerebecki:
Changed commit message to match regex checked by travis.
Sounds good to me, but I don't know the finer details of these timings, really needs someone with rabitmq clustering knowledge.

PR #2066, opened 2019-03-21
[5.0] db: Raise default connection limit to 2048

vuntz:
if you deploy the minimal of 10 services (each with the minimum
of 15 database threads) with the default 4 workers on a 3 node cluster,
you're at 1800 connections max already, so going with 1400 seems
to a reasonable default.

It seems the previous setting of 1400 is still not enough, with
#1952 we need
higher limits. 2048 should be safer


dirkmueller:
Did we file a ticket for documentation?

PR #2068, opened 2019-03-22
database: apply password update (SCRD-8438)

sjamgade:
Currently the create action for mysql database does not check of changed
password.

Without this patch if the admin changes database password, this change
will be applied to the config files however the database is not informed
about this change. As a result that service has looses access to the
database.

This patch relies on the mysql.user table of the database to check of
user existence as well as password hash matching.  Every "input" to
any query is quoted before usage, to avoid any security related error.

An alternate approach to this would be to check if the user connect to
the database with the given credentials, but that requires determining
the correct host string which is not readily available in this context
and will require collecting user specific connection params (including
ssl) and will also require handling of connection close in error as well
as non error cases

### DIFF
--- a/chef/cookbooks/database/libraries/provider_database_mysql_user.rb
+++ b/chef/cookbooks/database/libraries/provider_database_mysql_user.rb
@@ -32,15 +32,43 @@ class Chef
         end
 
         def action_create
-          Chef::Log.info("Creating user '#{new_resource.username}@#{new_resource.host}'")
-          username = client.escape(new_resource.username)
-          host = client.escape(new_resource.host)
-          create_sql = "CREATE USER IF NOT EXISTS '#{username}'@'#{host}'"
-          if new_resource.password
-            password = client.escape(new_resource.password)
-            create_sql += " IDENTIFIED BY '#{password}'"
+          # create
+          # new_resource can have password defined or not;
+          # search a user
+          # if found,
+          #   if password hashes match
+          #     do nothing
+          #   else
+          #     update password
+          # else
+          #   create user (with or without password)
+          nr = new_resource
+          username = "'#{client.escape(nr.username)}'"
+          host = "'#{client.escape(nr.host)}'"
+          userspec = "#{username}@#{host}"
+          createuser = "CREATE USER IF NOT EXISTS #{userspec}"
+          if nr.password
+            password = "'#{client.escape(nr.password)}'"
+            alteruser = "ALTER USER #{userspec} IDENTIFIED BY #{password}"
+            createuser = "#{createuser} IDENTIFIED BY #{password}"
           end
-          client.query(create_sql)
+
+          checkqry = "SELECT user,host FROM mysql.user WHERE user=#{username} AND host=#{host}"
+          if client.query(checkqry).count > 0
+            # PASSWORD is the mysql's internal password hashing function
+            checkqry = "#{checkqry} AND password=PASSWORD(#{password})"
+            if nr.password && client.query(checkqry).count.zero?
+              Chef::Log.info("Updating password : #{userspec}")
+              query = alteruser
+            else
+              # user found but either passowrd not defied or no change
+              query = nil
+            end
+          else
+            Chef::Log.info("Creating user #{userspec}")
+            query = createuser
+          end
+          client.query(query) if query
         ensure
           close_client
         end


chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -32,15 +32,43 @@ def load_current_resource
...
+          #     do nothing
+          #   else
+          #     update password
+          # else
+          #   create user (with or without password)
+          nr = new_resource
+          username = "'#{client.escape(nr.username)}'"
+          host = "'#{client.escape(nr.host)}'"
+          userspec = "#{username}@#{host}"
+          createuser = "CREATE USER IF NOT EXISTS #{userspec}"
```
jsuchome: `IF NOT EXISTS` looks like not needed when you do the query for existence anyway...

sjamgade: I left it there since it was there in the original code, it doesnt hurt to have it. I can remove it in the next round of changes if at all any because getting the gating green after 2 round of rebuild is a long and frustrating

jsuchome:
It looks correct, I have not tested it though

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
jsuchome: Right, it does not hurt (I'm just trying to understand the change)

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -32,15 +32,43 @@ def load_current_resource
...
+
+          checkqry = "SELECT user,host FROM mysql.user WHERE user=#{username} AND host=#{host}"
+          if client.query(checkqry).count > 0
+            # PASSWORD is the mysql's internal password hashing function
+            checkqry = "#{checkqry} AND password=PASSWORD(#{password})"
+            if nr.password && client.query(checkqry).count.zero?
+              Chef::Log.info("Updating password : #{userspec}")
+              query = alteruser
+            else
+              # user found but either passowrd not defied or no change
```
jsuchome: Typo in comment ;-) (_passowrd_)

chef/cookbooks/database/libraries/provider_database_mysql_user.rb
```diff
@@ -32,15 +32,43 @@ def load_current_resource
...
+          if nr.password
+            password = "'#{client.escape(nr.password)}'"
+            alteruser = "ALTER USER #{userspec} IDENTIFIED BY #{password}"
+            createuser = "#{createuser} IDENTIFIED BY #{password}"
+          end
+
+          checkqry = "SELECT user,host FROM mysql.user WHERE user=#{username} AND host=#{host}"
+          if client.query(checkqry).count > 0
+            # PASSWORD is the mysql's internal password hashing function
+            checkqry = "#{checkqry} AND password=PASSWORD(#{password})"
```
jsuchome: Is the `password` defined at this stage?

jsuchome: Oh, you have extra check for `nr.password` before executing the query ... it's a bit confusing :-)

sjamgade: password is not defined when `nr.password` is undefined 
so the next `if` statement will short circuit because `nr.password`  is not defined and thus the query will never be used.

sjamgade: ah you this will throw 'Undefined local variable error' correct

sjamgade:
```
irb(main):021:0> def a                                                                                                             
irb(main):022:1>   if nil
irb(main):023:2>    c = 123
irb(main):024:2>   end
irb(main):025:1>   puts "#{c}"                                                                                                     
irb(main):026:1> end
=> :a
irb(main):027:0> a()

=> nil
irb(main):028:0>
```

jsuchome: If it works, than let it be, I think it would be little more obvious if you'd create the query only after you verified you need it

sjamgade: yes it works, as variable get defined, but not initialized, so its value remains `nil`

PR #2069, opened 2019-03-25
neutron: properly define region (bsc#1128753)

dirkmueller:
Unlikey everywhere else, the key to set is region not region_name

### DIFF
--- a/chef/cookbooks/neutron/templates/default/neutron_lbaas.conf.erb
+++ b/chef/cookbooks/neutron/templates/default/neutron_lbaas.conf.erb
@@ -2,11 +2,8 @@
 interface_driver = <%= @interface_driver %>
 [quotas]
 [service_auth]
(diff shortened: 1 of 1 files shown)


chef/cookbooks/neutron/templates/default/neutron_lbaas.conf.erb
```diff
@@ -9,7 +9,7 @@ auth_url = <%= KeystoneHelper.versioned_service_URL(@keystone_settings["protocol
 admin_project_name = <%= @keystone_settings['service_tenant'] %>
```
KeithMnemonic: I think this is also wrong and was the reason i needed to add neutron to admin tenant as admin role. According to this 
https://github.com/openstack/neutron-lbaas/blob/stable/pike/neutron_lbaas/common/keystone.py#L39 it needs to be admin_tenant_name.  I tested this and it works without adding neutron

dirkmueller: done

PR #2071, opened 2019-03-25
Increase sync_mark timeouts (SCRD-8179)

witekest:
It has been observed that DB migrations for Grafana and Manila take
more time than the configured default 60 s timeout for sync_marks. This
change doubles the timeout in Horizon and Manila barclamps to 120 s.


witekest:
> Thanks. This is a million times better than changing the global default.

All the credit goes to @jgrassler as he investigated it and provided all details.

vuntz:
No idea why this is not merged while there are three approvals, so merging...

PR #2072, opened 2019-03-25
tempest: update neutron extension list

dirkmueller:
The tag extension is deprecated as of Queens, and has been
replaced by tagging.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -23,19 +23,12 @@ require "open3"
 nova = get_instance("roles:nova-controller")
 keystone_settings = KeystoneHelper.keystone_settings(nova, "nova")
 
-alt_comp_user = keystone_settings["default_user"]
-alt_comp_pass = keystone_settings["default_password"]
-alt_comp_tenant = keystone_settings["default_tenant"]
-
 # Will only be set if this cloud is actually running heat
 heat_trusts_delegated_roles = nil
 
 tempest_comp_user = node[:tempest][:tempest_user_username]
 tempest_comp_pass = node[:tempest][:tempest_user_password]
 tempest_comp_tenant = node[:tempest][:tempest_user_tenant]
-
-tempest_adm_user = node[:tempest][:tempest_adm_username]
-tempest_adm_pass = node[:tempest][:tempest_adm_password]
 
 # manila (share)
 tempest_manila_settings = node[:tempest][:manila]
@@ -81,9 +74,9 @@ comp_environment << "OS_IDENTITY_API_VERSION='#{keystone_settings["api_version"]
 openstackcli = "#{comp_environment} openstack --insecure"
 
 # for admin usage (listing the available services)
-adm_environment = "OS_USERNAME='#{tempest_adm_user}' "
-adm_environment << "OS_PASSWORD='#{tempest_adm_pass}' "
-adm_environment << "OS_PROJECT_NAME='#{tempest_comp_tenant}' "
+adm_environment = "OS_USERNAME='#{keystone_settings["admin_user"]}' "
+adm_environment << "OS_PASSWORD='#{keystone_settings["admin_password"]}' "
+adm_environment << "OS_PROJECT_NAME='#{keystone_settings["admin_project"]}' "
 adm_environment << "OS_AUTH_URL='#{auth_url}' "
 adm_environment << "OS_IDENTITY_API_VERSION='#{keystone_settings["api_version"]}'"
 openstackcli_adm = "#{adm_environment} openstack --insecure"
@@ -91,9 +84,8 @@ openstackcli_adm = "#{adm_environment} openstack --insecure"
 enabled_services = `#{openstackcli_adm} service list -f value -c Type`.split
 
 users = [
-          {"name" => tempest_comp_user, "pass" => tempest_comp_pass, "role" => "member"},
-          {"name" => tempest_adm_user, "pass" => tempest_adm_pass, "role" => "admin" }
-        ]
+  { "name" => tempest_comp_user, "pass" => tempest_comp_pass, "role" => "member" }
+]
 
 roles = [ 'anotherrole' ]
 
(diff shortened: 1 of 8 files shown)


dirkmueller:
@toabctl yes, main difference is for tempest full runs (which should be the actual target for testing the product but isn't..)

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -357,17 +365,29 @@
...
+  end
+  if neutrons[0][:neutron][:use_dvr] then
+    neutron_api_extensions += ", dvr"
+  end
+  if neutrons[0][:neutron][:l3_ha][:use_l3_ha] then
+    neutron_api_extensions += ", l3-ha"
   end
 end
 
+## Run Designate tests when available
```
toabctl: Can you enable that with an if condition when a designate node is available?

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -443,33 +463,20 @@
...
-  # Default: 300
-  validation_ssh_timeout = 450
-  use_interface_attach = false
-  use_rescue = false
-  use_suspend = false
-  use_vnc = node[:kernel][:machine] != "aarch64"
-  use_run_validation = false
-  use_config_drive = false
-end
+use_interface_attach = false
```
toabctl: Are theses the default? Previously we only set theses options to `false` in the Xen case. So this might actually decrease the test coverage.

chef/cookbooks/tempest/templates/default/tempest.conf.erb
```diff
@@ -46,9 +46,12 @@ block_migration_for_live_migration = true
 vnc_console = <%= @use_vnc %>
 rescue = <%= @use_rescue %>
 interface_attach = <%= @use_interface_attach %>
-personality = false
+personality = True
+volume_multiattach = True
```
toabctl: Not all drivers support this (see https://docs.openstack.org/cinder/latest/reference/support-matrix.html#driver-support-matrix ). So I guess we need a condition depending on the enabled backends?

cmurphy:
Do these new tests pass tempestfull? I'm happy about more tests but we're already struggling to get the tempestfull job green and it basically gets ignored, it doesn't do a lot of good to add more broken tests that will still be ignored.

chef/cookbooks/tempest/recipes/config.rb
dirkmueller: default is True. I can remove the setting (it was set to true in the "if we have at least one kvm compute node") case _ so I think in practice for our CI it made little difference.

chef/cookbooks/tempest/recipes/config.rb
dirkmueller: done

chef/cookbooks/tempest/templates/default/tempest.conf.erb
dirkmueller: yeah, onl LVM seems to support it. so lets turn it off for now to address @cmurphy 's concern of not enabling stuff that we might not be able to get passing.

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -357,17 +365,26 @@
...
+  "subnet_allocation",
+  "trunk",
+  "trunk-details"
+].join(", ")
 
 unless neutrons[0].nil?
-  if neutrons[0][:neutron][:use_lbaas] then
-    neutron_api_extensions += ",lbaasv2,lbaas_agent_schedulerv2,lb-graph,lb_network_vip"
+  neutron_attr = neutrons[0][:neutron]
+  if neutron_attr[:use_lbaas] then
```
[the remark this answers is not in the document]
dirkmueller: fixed

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -357,17 +365,29 @@
...
+  "trunk",
+  "trunk-details"
+].join(", ")
 
 unless neutrons[0].nil?
   if neutrons[0][:neutron][:use_lbaas] then
-    neutron_api_extensions += ",lbaasv2,lbaas_agent_schedulerv2,lb-graph,lb_network_vip"
+    neutron_api_extensions += ", lbaasv2, lbaas_agent_schedulerv2, lb-graph, lb_network_vip"
+  end
+  if neutrons[0][:neutron][:use_dvr] then
```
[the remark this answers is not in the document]
dirkmueller: fixed

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -357,17 +365,29 @@
...
+    neutron_api_extensions += ", dvr"
+  end
+  if neutrons[0][:neutron][:l3_ha][:use_l3_ha] then
```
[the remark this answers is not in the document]
dirkmueller: fixed

dirkmueller:
@cmurphy so I started looking at the errors in tempestful, thats why you see me touching the config file.. ;-)

I don't have a full run with results yet.

dirkmueller:
without the patch:

```
- Passed: 1973
- Skipped: 386
- Failed: 11
```

with the patch:

```
 - Passed: 2030
 - Skipped: 291
 - Failed: 10
```

dirkmueller:
@cmurphy with https://github.com/SUSE-Cloud/automation/pull/3283 the job should be green.

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -81,19 +74,18 @@
...
 adm_environment << "OS_IDENTITY_API_VERSION='#{keystone_settings["api_version"]}'"
 openstackcli_adm = "#{adm_environment} openstack --insecure"
 
 enabled_services = `#{openstackcli_adm} service list -f value -c Type`.split
 
 users = [
-          {"name" => tempest_comp_user, "pass" => tempest_comp_pass, "role" => "member"},
-          {"name" => tempest_adm_user, "pass" => tempest_adm_pass, "role" => "admin" }
-        ]
+  {"name" => tempest_comp_user, "pass" => tempest_comp_pass, "role" => "member"}
```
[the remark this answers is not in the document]
dirkmueller: fixed

chef/cookbooks/tempest/templates/default/tempest.conf.erb
dirkmueller: I have turned it off.

cmurphy:
Commit b83a923036a2b73a08f76df317b4a4845b91f6f6 has additional changes that aren't related to the title, "tempest: Avoid syntax error in the ironic case ", would be nicer for git blame/bisect if that was extracted or squashed.

PR #2073, opened 2019-03-26
memcache: Use first array element as fallback (SCRD-8255)

s-t-e-v-e-n-k:
When this code was refactored, a fallback of node[:memcache][:port] was
used, since the original code in the nova and horizon cookbooks used it.
However, node is not defined in this context, so the fallback causes a
NameError instead. Use the first array element as the fallback instead.


s-t-e-v-e-n-k:
(I have a backport locally for SOC8, SOC7 is not affected)

PR #2074, opened 2019-03-26
[5.0] ceilometer: Install package which contains cron file (bsc#1130414)

toabctl:
Commit 2f441560b uses pacemaker to handle the cron symlink. But to
work correctly, the link
destination (/usr/share/ceilometer/openstack-ceilometer-expirer.cron)
needs to be there and this file is in the
openstack-ceilometer-collector package.
So install the package to solve this.


toabctl:
Note: This is not needed for master because there is no expierer anymore

toabctl:
Backport for stable/4.0 is https://github.com/crowbar/crowbar-openstack/pull/2090

PR #2075, opened 2019-03-26
glance: use the correct db_synced flag

sjamgade:
the currently used flag is not set anywhere as a result this migration is run
everytime the recipe is run.


dirkmueller:
interestingly this isn't an issue in 5.0 or older.. so seems like a new regression only

PR #2078, opened 2019-03-28
tempest: Improve ironic test results (was: Use fake-hardware driver)

skazi0:
Following changes are included to make basic ironic testing possible:
1. Use fake-hardware driver
Current default in ironic_tempest_plugin config is still "fake"
with comment: "change to fake-hardware when Ocata is no longer supported".
2. Change endpoint type to public
Two tests in ironic suite check endpoint urls and they expect public ones.
3. Disable ironic inspector tests
Ironic barclamp doesn't support Inspector yet.

With above plus https://github.com/SUSE-Cloud/automation/pull/3295 I got following results:
```
======
Totals
======
Ran: 201 tests in 1949.0000 sec.
 - Passed: 178
 - Skipped: 21
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 2
Sum of execute time for each test: 578.9197 sec.
```

Remaining two failed tests are for:
- BFV (Boot From Volume) - not implemented in Ironic barclamp yet
- BaremetalBasicOps - some "real" testing of Ironic node operations. This seems to require more complete environment setup (e.g. pre-created node with specific driver and features).

### DIFF
--- a/chef/cookbooks/tempest/templates/default/tempest.conf.erb
+++ b/chef/cookbooks/tempest/templates/default/tempest.conf.erb
@@ -24,7 +24,7 @@ aws_secret = <%= @ec2_secret %>
 aws_access = <%= @ec2_access %>
 
 [baremetal]
-endpoint_type = internalURL
+endpoint_type = public
 max_microversion = latest
 driver = fake-hardware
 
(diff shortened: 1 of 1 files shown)


skazi0:
@dirkmueller at some point we could remove this and leave the default (which is "fake" right now and will be "fake-hardware" somewhere in the future). Also see this: https://github.com/openstack/ironic-tempest-plugin/blob/master/ironic_tempest_plugin/tests/api/admin/base.py#L28

PR #2079, opened 2019-03-29
[5.0] Neutron LBaaS fixes

dirkmueller:
Backport of #2069


dirkmueller:
potentially needs backport to SOC7, to be checked.

PR #2081, opened 2019-04-02
monasca: Add openvswitch plugin (SCRD-7571)

witekest:
Add Monasca OVS plugin configuration on all neutron-network nodes.
Plugin configuration is automatically enabled and can be disabled in
Crowbar UI.

Monasca OVS plugin autodetection is disabled on Monasca server node.

Add `monasca-agent` role to `delegate_authorized_roles` to enable
collecting tenant metrics.

### DIFF
--- a/chef/cookbooks/monasca/providers/agent_plugin_ovs.rb
+++ b/chef/cookbooks/monasca/providers/agent_plugin_ovs.rb
@@ -1,0 +1,49 @@
+#
+# Copyright 2019 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require "json"
+require "yaml"
+
+action :create do
+  config = {
+    "username" => new_resource.username,
+    "user_domain_name" => new_resource.user_domain_name,
+    "password" => new_resource.password,
+    "project_name" => new_resource.project_name,
+    "project_domain_name" => new_resource.project_domain_name,
+    "auth_url" => new_resource.auth_url,
+    "region_name" => new_resource.region_name,
+    "check_router_ha" => new_resource.check_router_ha
+  }.merge(node[:monasca][:agent][:plugins][:ovs])
+
+  # be sure the package is installed. that way "/etc/monasca/agent/conf.d/" is available
+  # and also the user and group are there
+  package "openstack-monasca-agent"
+
+  # NOTE(toabctl): convert/parse first to/from json. Otherwise we have unwanted markers
+  # like "- !ruby/hash:Mash" in the yaml output
+  ovs_conf = JSON.parse({ "init_config" => config,
+                          "instances" => [] }.to_json).to_yaml
+
+  # write OVS plugin config
+  file "/etc/monasca/agent/conf.d/ovs.yaml" do
+    content ovs_conf
+    owner node[:monasca][:agent][:user]
+    group node[:monasca][:agent][:group]
+    mode "0640"
+    notifies :restart, resources(service: node[:monasca][:agent][:agent_service_name]), :delayed
+  end
+end
--- a/chef/cookbooks/monasca/templates/default/monasca-api.conf.erb
+++ b/chef/cookbooks/monasca/templates/default/monasca-api.conf.erb
@@ -623,7 +623,7 @@ agent_authorized_roles = monasca-agent
 # Roles that are allowed to POST metrics on
 # behalf of another tenant
 #  (list value)
-#delegate_authorized_roles = admin
+delegate_authorized_roles = monasca-agent
 
 [keystone_authtoken]
 www_authenticate_uri = <%= @keystone_settings['public_auth_url'] %>
--- a/chef/cookbooks/neutron/recipes/monitor_monasca_ovs.rb
+++ b/chef/cookbooks/neutron/recipes/monitor_monasca_ovs.rb
@@ -1,0 +1,37 @@
+#
+# Copyright 2019 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+return unless node["roles"].include?("monasca-agent")
+
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+
+if node[:monasca][:agent][:monitor_ovs]
+  node["roles"].each do |role|
+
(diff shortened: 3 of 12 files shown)


chef/cookbooks/monasca/templates/default/monasca-api.conf.erb
```diff
@@ -624,6 +624,7 @@ agent_authorized_roles = monasca-agent
 # behalf of another tenant
 #  (list value)
 #delegate_authorized_roles = admin
```
cmurphy: delete this?

chef/data_bags/crowbar/migrate/monasca/315_add_ovs_plugin_settings.rb
```diff
@@ -0,0 +1,19 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["agent"]["monitor_ovs"] = template_attrs["agent"]["monitor_ovs"] unless
+    attrs["agent"].key?("monitor_ovs")
+
+  attrs["agent"]["plugins"] = template_attrs["agent"]["plugins"] unless
+    attrs["agent"].key?("plugins")
```
cmurphy: I could be wrong but I don't think this will do what you want, the template already has a pre-existing "plugins" dictionary, so either this will not add in the new "ovs" key or it will overwrite the old version with the defaults in the template instead of the user-set values. I think this should be `attrs["agent"]["plugins"]["ovs"] = template_attrs... etc`

chef/cookbooks/monasca/templates/default/monasca-api.conf.erb
witekest: Done

chef/data_bags/crowbar/migrate/monasca/315_add_ovs_plugin_settings.rb
witekest: Good catch. Thank you.

chef/cookbooks/monasca/providers/agent_plugin_ovs.rb
```diff
@@ -0,0 +1,48 @@
...
+action :create do
+  config = {
+    "username" => new_resource.username,
+    "user_domain_name" => new_resource.user_domain_name,
+    "password" => new_resource.password,
+    "project_name" => new_resource.project_name,
+    "project_domain_name" => new_resource.project_domain_name,
+    "auth_url" => new_resource.auth_url,
+    "region_name" => new_resource.region_name
+  }.merge(node[:monasca][:agent][:plugins][:ovs])
```
toabctl: in the other providers, we use `node[:monasca][:agent_plugin_config][:ovs_instances]` to store the different configs. IIRC the idea was, that a use could also store custom configs in the node object if needed.
Is there a reason to do this differently here?

chef/cookbooks/monasca/providers/agent_plugin_ovs.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  # be sure the package is installed. that way "/etc/monasca/agent/conf.d/" is available
+  # and also the user and group are there
+  package "openstack-monasca-agent"
+
+  # NOTE(toabctl): convert/parse first to/from json. Otherwise we have unwanted markers
+  # like "- !ruby/hash:Mash" in the yaml output
+  process_conf = JSON.parse({ "init_config" => config,
```
toabctl: nit: ovs_conf would be a better name

chef/cookbooks/monasca/templates/default/monasca-api.conf.erb
```diff
@@ -623,7 +623,7 @@ agent_authorized_roles = monasca-agent
 # Roles that are allowed to POST metrics on
 # behalf of another tenant
 #  (list value)
-#delegate_authorized_roles = admin
+delegate_authorized_roles = monasca-agent
```
toabctl: This is unrelated to the ovs agent enablement? if so, can you move it to another commit please?

chef/cookbooks/neutron/recipes/monitor_monasca_ovs.rb
```diff
@@ -0,0 +1,36 @@
+#
```
toabctl: Shouldn't that be in the file `monitor_monasca.rb` which already contains the http_check configuration?

chef/cookbooks/neutron/recipes/monitor_monasca_ovs.rb
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+return unless node["roles"].include?("monasca-agent")
+
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+
+if node[:monasca][:agent][:monitor_ovs]
```
toabctl: Is there a reason to **not** monitor OVS if monasca is deployed? So why do we need to make it configurable?

chef/cookbooks/monasca/providers/agent_plugin_ovs.rb
witekest: Instances section of OVS plugin configuration [should be empty](https://github.com/openstack/monasca-agent/blob/master/docs/Ovs.md).

chef/cookbooks/monasca/providers/agent_plugin_ovs.rb
witekest: Done.

chef/cookbooks/monasca/templates/default/monasca-api.conf.erb
witekest: Done.

chef/cookbooks/neutron/recipes/monitor_monasca_ovs.rb
witekest: http_check from `monasca_monasca.rb` should be configured on Neutron server.
OVS plugin should be configured on all Neutron agents.
The recipes are included in `role_neutorn_server.rb` and `role_neutron_network_network.rb` respectively.

chef/cookbooks/neutron/recipes/monitor_monasca_ovs.rb
witekest: I have just followed the same pattern as for libvirt and Ceph plugins. OVS plugin was not supported till now, so adding the configuration option increases visibility for end users, who not necessarily read release notes. I don't have a strong opinion though.

chef/data_bags/crowbar/template-monasca.json
```diff
@@ -32,6 +32,18 @@
             "vm_ping_check_enable": true,
             "vm_probation": 300,
             "vnic_collection_period": 0
+          },
+          "ovs": {
+            "cache_dir": "/dev/shm",
+            "neutron_refresh": 14400,
+            "network_use_bits": false,
+            "check_router_ha": true,
```
toabctl: `check_router_ha` should be `false` if no router HA is used. Documentation says:

_check_router_ha will check router HA status if set to true. This should be set to false if not configuring routers for HA, as setting this to true will cause the plugin to make additional neutron calls._

I think you need to check for `neutron[:neutron][:l3_ha][:use_l3_ha]`

Given that this depends on `neutron[:neutron][:l3_ha][:use_l3_ha]`, I would remove the config option and just set it correctly depending on `use_l3_ha`

chef/data_bags/crowbar/template-monasca.json
```diff
@@ -32,6 +32,18 @@
             "vm_ping_check_enable": true,
             "vm_probation": 300,
             "vnic_collection_period": 0
+          },
+          "ovs": {
+            "cache_dir": "/dev/shm",
```
toabctl: do you know how much cached data will be produced? on a mkcloud deployment, I have 6 GB free on `/dev/shm`

witekest: This includes only values from the last Neutron API call. In my test env I measured ~1KB per router/port. I don't expect any issues here.

chef/data_bags/crowbar/template-monasca.json
witekest: Good point. Done.

PR #2083, opened 2019-04-02
nova: verify glance signatures if able (SCRD-8578)

cmurphy:
Without this patch, the `test_signed_image_upload_boot_failure` barbican
tempest tests fails because nova fails to prevent booting from an
unverified signed image. This test requires the
verify_glance_signatures option be set in the nova config on the compute
node. However, it must only be set if barbican is actually available,
otherwise nova can't boot anything.


cmurphy:
This won't work, see comments in jira issue

cmurphy:
Jira issue was closed.

PR #2086, opened 2019-04-03
ironic: Use IP instead of DNS name for Swift

skazi0:
Swift endpoint in Glance section is used to build TempURLs. These
are then used by Ironic agent running from a ramdisk image. It is
better to use IP here as proper DNS configuration might not be
available in the image and Swift host would not be resolved.

### DIFF
--- a/chef/cookbooks/ironic/libraries/helpers.rb
+++ b/chef/cookbooks/ironic/libraries/helpers.rb
@@ -30,6 +30,18 @@ module IronicHelper
       "#{env} openstack #{insecure}"
     end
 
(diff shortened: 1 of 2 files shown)


cmurphy:
Is this associated with a bug?

skazi0:
@cmurphy nope. it's part of SCRD-7852 -> an attempt to make ironic barclamp testable with tempest.

skazi0:
I'm not sure how this will behave in HA case. Previous version handled this properly but this might not work as expected. Back to testing...

skazi0:
I've tested this new version with HA and non-HA and it seems to work as expected.

PR #2088, opened 2019-04-04
designate: Fix variables initialization in mdns

witekest:
Fit typo in variable name and `dnsservs` array initialisation.


sjamgade:
@witekest I have pushed a new commit with the way I had expected it to be. Feel free to squash or drop it if you feel this is incorrect.

witekest:
> @witekest I have pushed a new commit with the way I had expected it to be. Feel free to squash or drop it if you feel this is incorrect.

`dnsmaster` is a string and I was getting a TypeError.

PR #2089, opened 2019-04-05
ceilometer: add a polling section in ceilometer conf (SCRD-8720)

aagate:
* Add a [polling] section to ceilometer.conf with cfg_file property which
  points to polling_pipeline.yaml

* removed default_domain_id and default_domain_name from  [service_credentials]
  and added project_domain_name and user_domain name to [service_credentials]
  in polling.conf to prevent ceilometer polling agent from failing with
  "need user domain" and "need project domain" error.

* added a new pipeline yaml file for polling agent, called polling_pipeline.yml

* added swift meters which are polling meters to polling_pipeline.yml

### DIFF
--- a/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
+++ b/chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
@@ -54,10 +54,10 @@ interface = internalURL
 auth_type = password
 auth_url = <%= @keystone_settings['internal_auth_url'] %>
 project_name = <%= @monasca_project %>
-default_domain_id = <%= @keystone_settings["admin_domain_id"] %>
-default_domain_name = <%= @keystone_settings["admin_domain"] %>
 username = <%= @keystone_settings['service_user'] %>
 password = <%= @keystone_settings['service_password'] %>
+user_domain_name = <%= @keystone_settings["admin_domain"] %>
+project_domain_name = <%= @keystone_settings["admin_domain"]%>
 
 [service_types]
 neutron_lbaas_version = v2
@@ -105,3 +105,6 @@ service_region_name= <%= @keystone_settings['endpoint_region'] %>
 
 [publisher_notifier]
 metering_topic=metering
+
+[polling]
+cfg_file=/etc/ceilometer/polling_pipeline.yaml
--- a/chef/cookbooks/ceilometer/templates/default/polling_pipeline.yaml.erb
+++ b/chef/cookbooks/ceilometer/templates/default/polling_pipeline.yaml.erb
@@ -1,0 +1,15 @@
+---
+sources:
+    - name: swift_source
+      interval: <%= @swift_interval %>
+      meters:
+        - "storage.objects"
+        - "storage.objects.size"
+        - "storage.objects.containers"
+      sinks:
+        - meter_sink
+sinks:
+    - name: meter_sink
+      transformers:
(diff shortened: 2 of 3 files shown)


chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
```diff
@@ -58,6 +58,8 @@ default_domain_id = <%= @keystone_settings["admin_domain_id"] %>
 default_domain_name = <%= @keystone_settings["admin_domain"] %>
```
witekest: Please remove `default_domain_id` and `default_domain_name`.

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
```diff
@@ -105,3 +107,6 @@ service_region_name= <%= @keystone_settings['endpoint_region'] %>
 
 [publisher_notifier]
 metering_topic=metering
+
+[polling]
+cfg_file=/etc/ceilometer/pipeline.yaml
```
witekest: `pipeline.yaml` contains also notification originated measurements. Is it OK like that or should we provide a separate pipeline configuration for pollsters only as configured in Ardana?

aagate: It is OK to have a same pipeline.yaml for both notification agent and polling agent, ceilometer in crowbar based deployment was always set up so that there was a single pipeline.yaml. I was trying to keep the changes to minimum at this stage. Having said that having a different pipeline.yaml is better, since customer has better control over what services are being polled. If you think its better I can make a new pipeline-polling template, and copy it to /etc/ceilometer/pipeline-polling.yaml and have it poll only swift meters. (similar to ardana based install)

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
aagate: Thanks! will remove default_domain_id and default_domain_name

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
vuntz: @aagate I'd say give it a go if you can get it done today :-)

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
aagate: - done

chef/cookbooks/ceilometer/templates/default/ceilometer.conf.erb
aagate: done

chef/cookbooks/ceilometer/templates/default/polling_pipeline.yaml.erb
```diff
@@ -0,0 +1,15 @@
+---
+sources:
+    - name: swift_source
+      interval: <%= @swift_interval %>
+      meters:
+        - "storage.objects"
+        - "storage.objects.size"
+        - "storage.objects.containers"
+      sinks:
+        - meter_sink
```
vuntz: Do we need to drop this from the other pipeline.yaml?

witekest:
Installed in my environment and could see Swift measurements.

PR #2092, opened 2019-04-09
neutron: Switch to openvswitch firewall driver

dirkmueller:


dirkmueller:
(just testing this, yes it needs to be configurable)

PR #2095, opened 2019-04-10
neutron: enable designate integration

sjamgade:
the config enabled by this flag lets neutron create dns records when
floating ip are created and assigned.
The config was previously added but was not enabled as the barclamp was
still under development 

This config is not really needed for designate or neutron to work, but they help enable certain features
this is where the flag is used https://github.com/crowbar/crowbar-openstack/blob/a671c048ec071a13170d055b4dfc5a6906596cd0/chef/cookbooks/neutron/templates/default/neutron.conf.erb#L36-L48

### DIFF
--- a/chef/cookbooks/neutron/recipes/common_config.rb
+++ b/chef/cookbooks/neutron/recipes/common_config.rb
@@ -119,6 +119,17 @@ if neutron[:neutron][:use_infoblox]
   infoblox_settings = neutron[:neutron][:infoblox]
 end
 
+designate_public_uri = nil
+designate_server = node_search_with_cache("roles:designate-server").first
+unless designate_server.nil?
+  node_designate = designate_server[:designate]
+  public_host = CrowbarHelper.get_host_for_public_url(designate_server,
+                                                    node_designate[:api][:protocol] == "https",
+                                                    node_designate[:ha][:enabled])
+  api_protocol = node_designate[:api][:protocol]
+  designate_public_uri = "#{api_protocol}://#{public_host}:#{node_designate[:api][:bind_port]}/v2"
+end
+
 template neutron[:neutron][:config_file] do
     cookbook "neutron"
     source "neutron.conf.erb"
@@ -132,8 +143,8 @@ template neutron[:neutron][:config_file] do
       sql_max_pool_overflow: neutron[:neutron][:sql][:max_pool_overflow],
       sql_pool_timeout: neutron[:neutron][:sql][:pool_timeout],
       debug: neutron[:neutron][:debug],
-      # TODO
-      designate_enabled: false,
+      designate_enabled: (not designate_server.nil?),
+      designate_public_uri: designate_public_uri,
       bind_host: bind_host,
       bind_port: bind_port,
       use_syslog: neutron[:neutron][:use_syslog],
--- a/chef/cookbooks/neutron/recipes/post_install_conf.rb
+++ b/chef/cookbooks/neutron/recipes/post_install_conf.rb
@@ -211,6 +211,25 @@ execute "add_ironic_network_to_router" do
   action :nothing
 end
 
+domain_float = node[:neutron][:floating_dns_domain]
+domain_fixed = node[:neutron][:dns_domain]
+
+execute "update_dns_domain_for_fixed_network" do
+  command "#{env} neutron net-update fixed --dns-domain #{domain_fixed}"
+  not_if "#{openstack_cmd} network show fixed -f value -c dns_domain | grep -q #{domain_fixed}"
+  retries 5
+  retry_delay 10
+  action :nothing
+end
+
+execute "update_dns_domain_for_floating_network" do
+  command "#{env} neutron net-update floating --dns-domain #{domain_float}"
+  not_if "#{openstack_cmd} network show floating -f value -c dns_domain | grep -q #{domain_float}"
+  retries 5
+  retry_delay 10
+  action :nothing
+end
+
 execute "Neutron network configuration" do
   command "#{openstack_cmd} network list &>/dev/null"
   retries 5
@@ -226,6 +245,8 @@ execute "Neutron network configuration" do
   notifies :run, "execute[set_router_gateway]", :delayed
   notifies :run, "execute[add_fixed_network_to_router]", :delayed
   notifies :run, "execute[add_ironic_network_to_router]", :delayed
+  notifies :run, "execute[update_dns_domain_for_fixed_network]", :delayed
+  notifies :run, "execute[update_dns_domain_for_floating_network]", :delayed
 end
 
 # This is to trigger all the above "execute" resources to run :delayed, so that
@@ -235,4 +256,3 @@ execute "Trigger Neutron network configuration" do
   command "true"
   notifies :run, "execute[Neutron network configuration]", :delayed
 end
-
--- a/chef/data_bags/crowbar/template-neutron.json
+++ b/chef/data_bags/crowbar/template-neutron.json
@@ -10,8 +10,8 @@
       "max_header_line": 16384,
       "debug": false,
       "create_default_networks": true,
-      "dns_domain": "openstack.local",
-      "floating_dns_domain": "floating.openstack.local",
+      "dns_domain": "openstack.local.",
+      "floating_dns_domain": "floating.openstack.local.",
       "rpc_workers": 1,
       "use_lbaas": true,
       "lbaasv2_driver": "haproxy",
@@ -192,7 +192,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 305,
+      "schema-revision": 306,
       "element_states": {
         "neutron-server": [ "readying", "ready", "applying" ],
         "neutron-network": [ "readying", "ready", "applying" ],
--- a/chef/cookbooks/neutron/templates/default/neutron.conf.erb
+++ b/chef/cookbooks/neutron/templates/default/neutron.conf.erb
@@ -37,16 +37,17 @@ default_log_levels = <%= @default_log_levels.join(", ") %>
 external_dns_driver = designate
 
 [designate]
+auth_type = password
 url = <%= @designate_public_uri %>
-admin_auth_url = <%= @keystone_settings['internal_auth_url'] %>
-admin_tenant_name = <%= @keystone_settings['service_tenant'] %>
-admin_username = <%= @keystone_settings['service_user'] %>
-admin_password = <%= @keystone_settings['service_password'] %>
+username = <%= @keystone_settings['service_user'] %>
+password = <%= @keystone_settings['service_password'] %>
+auth_url = <%= @keystone_settings['internal_auth_url'] %>
 region_name = <%= @keystone_settings['endpoint_region'] %>
-admin_user_domain = <%= @keystone_settings['admin_domain'] %>
-admin_project_domain = <%= @keystone_settings['admin_domain'] %>
+project_domain_name = <%= @keystone_settings['default_user_domain'] %>
+user_domain_name = <%= @keystone_settings['default_user_domain'] %>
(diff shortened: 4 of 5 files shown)


KeithMnemonic:
But is the neutron barclamp creating the ext-net ? If so you need to see the dns_name on that network. Is that also being done? And are you using the ml2 driver? that is another step needed.

sjamgade:
Thanks for input, I was not aware of these steps. let me wip this

dirkmueller:
there is a pending XXX for this for floating (this is the name of ext-net)

jgrassler:
What is a "pending XXX"?

chef/cookbooks/neutron/recipes/server.rb
```diff
@@ -445,3 +445,27 @@
     only_if { ha_enabled }
   end
 end
+
+floating_dns_domain = node[:neutron][:floating_dns_domain]
```
dirkmueller: put all of this in post_install_conf after the floating etc is being created? otherwise execution order might be wrong.

dirkmueller:
@jgrassler https://github.com/crowbar/crowbar-openstack/pull/2033/commits/7ca9b937c9a49f0b7de956e72fd6583305e10ecb

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -236,3 +236,25 @@ def mask_to_bits(mask)
   notifies :run, "execute[Neutron network configuration]", :delayed
 end
 
+floating_dns_domain = node[:neutron][:floating_dns_domain]
+env = {}
+env["OS_USERNAME"] = keystone_settings["service_user"]
```
dirkmueller: this is already on line 73ff, can we reuse that? I don't see the difference.

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -236,3 +236,25 @@ def mask_to_bits(mask)
...
+env["OS_PASSWORD"] = keystone_settings["service_password"]
+env["OS_PROJECT_NAME"] = keystone_settings["service_tenant"]
+env["OS_AUTH_URL"] = keystone_settings["internal_auth_url"]
+env["OS_REGION_NAME"] = keystone_settings["endpoint_region"]
+env["OS_IDENTITY_API_VERSION"] = keystone_settings["api_version"]
+env["OS_INTERFACE"] = "internal"
+env["OS_USER_DOMAIN_NAME"] = "Default"
+env["OS_PROJECT_DOMAIN_NAME"] = "Default"
+
+if CrowbarOpenStackHelper.config_for_role_exists?("designate")
```
dirkmueller: why do we do this only when designate exists?

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -236,3 +236,25 @@ def mask_to_bits(mask)
...
+  bash "Update dns for external network" do
```
dirkmueller: please add this as a phase 2 tasks in the "execute block" on line 215ff after floating etc has been created.

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -236,3 +236,25 @@ def mask_to_bits(mask)
...
+    user "root"
+    env env
+    code "
+      openstack network show floating -f value -c dns_domain |\
```
dirkmueller: this could be a only_if statement.

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -119,6 +119,18 @@
   infoblox_settings = neutron[:neutron][:infoblox]
 end
 
+designate_public_uri = nil
+designate_enabled = false
+if CrowbarOpenStackHelper.config_for_role_exists?("designate")
+  designate_enabled = true
+  node_designate = node_search_with_cache("roles:designate-server").first[:designate]
+  public_host = CrowbarHelper.get_host_for_public_url(node,
```
dirkmueller: this should be node_designate.

jgrassler:
> @jgrassler [7ca9b93](https://github.com/crowbar/crowbar-openstack/commit/7ca9b937c9a49f0b7de956e72fd6583305e10ecb)

Interesting. i just looked at what Sumit's got and that config section lacked domain information. Upon further investigation we stumbled upon this little gem:

https://github.com/openstack/neutron/blob/stable/rocky/neutron/services/externaldns/drivers/designate/driver.py#L54

tl;dr: the Designate integration in Neutron does not support Keystone V2. We'll try to come up with a patch.

toabctl:
> tl;dr: the Designate integration in Neutron does not support Keystone V2. We'll try to come up with a patch.

I think you can just set the domain info and set `auth_type = password` (then you land here: https://github.com/openstack/neutron/blob/stable/rocky/neutron/services/externaldns/drivers/designate/driver.py#L50). iirc using `password` and having the needed info will automatically select the correct version. But you could also set `auth_type = v3password` to be sure to use v3. @cmurphy might know more about this...

cmurphy:
Yes, just use auth_type=password and set user_domain_name and project_domain_name

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -211,6 +211,16 @@ def mask_to_bits(mask)
   action :nothing
 end
 
+doamin_floating = node[:neutron][:floating_dns_domain]
+
+execute "update_dns_domain_for_floating_network" do
+  command "#{env} neutron net-update floating --dns-domain #{doamin_floating}"
+  not_if "#{openstack_cmd} network show floating -f value -c dns_domain | grep -q #{doamin_floating}"
```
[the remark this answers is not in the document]
sjamgade: 1 char more should be fine

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -119,6 +119,19 @@
   infoblox_settings = neutron[:neutron][:infoblox]
 end
 
+designate_public_uri = nil
+designate_enabled = false
+if CrowbarOpenStackHelper.config_for_role_exists?("designate")
+  designate_enabled = true
+  designate_server = node_search_with_cache("roles:designate-server").first
```
dirkmueller: as you're doing a node_search_with_cache anyway, you can skip the "config_for_role" part completely - you're using anyway the node attributes. then you can do this:

```
  designate_public_uri = nil
  designate_server = node_search_with_cache("roles:designate-server").first
  unless designate_server.nil?
    ...

   designate_enabled: not designate_server.nil?

```

sjamgade: Done
@dirkmueller could you please re-approve

sjamgade:
This should be ready for reviews !

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -211,6 +211,25 @@ def mask_to_bits(mask)
   action :nothing
 end
 
+doamin_floating = node[:neutron][:floating_dns_domain]
```
jgrassler: Typo: "doamin" and it's pretty consistent throughout the section...any reason for that?

chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -226,6 +245,8 @@ def mask_to_bits(mask)
   notifies :run, "execute[set_router_gateway]", :delayed
   notifies :run, "execute[add_fixed_network_to_router]", :delayed
   notifies :run, "execute[add_ironic_network_to_router]", :delayed
+  notifies :run, "execute[update_dns_domain_for_fixed_network]", :delayed
```
dirkmueller: I believe this should only be done when designate integration is enabled?

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -119,6 +119,16 @@
...
 end
 
+designate_server = node_search_with_cache("roles:designate-server").first
+unless designate_server.nil?
+  node_designate = designate_server[:designate]
+  public_host = CrowbarHelper.get_host_for_public_url(designate_server,
+                                                    node_designate[:api][:protocol] == "https",
+                                                    node_designate[:ha][:enabled])
+  api_protocol = node_designate[:api][:protocol]
+  designate_public_uri = "#{api_protocol}://#{public_host}:#{node_designate[:api][:bind_port]}/v2"
```
vuntz: `designate_public_uri` is not defined outside of this block; I'd like to have an explicit `designate_public_uri = nil` for the other case

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -10,8 +10,8 @@
       "max_header_line": 16384,
       "debug": false,
       "create_default_networks": true,
-      "dns_domain": "openstack.local",
-      "floating_dns_domain": "floating.openstack.local",
+      "dns_domain": "openstack.local.",
```
vuntz: Should we have a migration to add this `.` at the end of the attribute, for upgrades?

chef/cookbooks/neutron/recipes/post_install_conf.rb
sjamgade: not necessary, it can be done otherwise too, putting it in there just makes it explicit

chef/cookbooks/neutron/recipes/common_config.rb
sjamgade: last time I checked that was implicit,

chef/data_bags/crowbar/template-neutron.json
KeithMnemonic: You also need to make sure the "." is in the neutron.conf override or wherever it is configured. https://docs.openstack.org/mitaka/networking-guide/config-dns-int.html

dirkmueller: or just fix it up as part of the chef recipe.. I am not sure we want to dns propagate a domain called "openstack.local" anyway (or anything else that is *local)

chef/cookbooks/neutron/recipes/common_config.rb
sjamgade: done

chef/data_bags/crowbar/template-neutron.json
sjamgade: added one for both the domains

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -119,6 +119,17 @@
   infoblox_settings = neutron[:neutron][:infoblox]
 end
 
+designate_public_uri = nil
+designate_server = node_search_with_cache("roles:designate-server").first
```
cmurphy: What if there are multiple nodes, as in an HA case?

jgrassler: That doesn't pose a problem: we just need a node object for any node with the designate role. A little further down we invoke `CrowbarHelper.get_host_for_public_url()` on that node object and that function will Do The Right Thing on an HA cloud and yield the HA cluster's public host (I've checked in QA's HA enabled test environment).

cmurphy: Okay, thanks!

chef/cookbooks/neutron/recipes/common_config.rb
```diff
@@ -132,8 +142,8 @@
       sql_max_pool_overflow: neutron[:neutron][:sql][:max_pool_overflow],
       sql_pool_timeout: neutron[:neutron][:sql][:pool_timeout],
       debug: neutron[:neutron][:debug],
-      # TODO
-      designate_enabled: false,
+      designate_enabled: (not designate_server.nil?),
```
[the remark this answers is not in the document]
cmurphy: If you end up needing to alter this PR again you may as well fix this hound complaint

PR #2096, opened 2019-04-10
nova, neutron: Hide the ZVM bits

vuntz:
This is not supported, and not working. No need to expose this.

Note that this is specifically not changing the cookbooks for this, as
the changes there may be more intrusive.


dirkmueller:
it is actually working - there is work for this undergoing and it was demoed at SUSECon this year..

vuntz:
We don't want to go that way.

PR #2097, opened 2019-04-10
neutron: Hide the Cisco API bits

vuntz:
This is not functional right now, hide it until it is supported.

Note that this is specifically not changing the cookbooks for this, as
the changes there may be more intrusive.


vuntz:
We don't want to go that way.

PR #2098, opened 2019-04-10
Octavia: Barclamp

dirkmueller:
This is the barclamp to deploy Octavia in crowbar.

Octavia: ha fixed and other fixes

Important changes
- Octavia-api is now in ha using haproxy
- all octavia services will be restarted by systemd if fails
- syncronization to create the octavia network route in each worker

Octavia: small fixes

octavia: various cosmetic fixes

* Fix barclamp name capitalization.
* Add ilausuch as Octavia codeowner
* Polish Octavia barclamp's README.rdoc.
* Removed API section in Octavia barclamp UI
* Added validation for certificate country code.

octavia: bind host/port fixes

* Remove bind host/port for API and health manager from data bag
* Add  bind_open_address setting for API and health manager to data bag
* Determine listen address/port based on bind_open_address and cloud type
  (single controller/HA)
* Move logic for determining bind host/port from template to octavia_conf
  definition.

octavia: network definiton in Octavia barclamp

This commit moves the definition of the Octavia management
network to the Octavia barclamp. Previously the Octavia
barclamp worked on the assumption that this network would be
defined in the Network barclamp.

octavia: use IP address for binding.

Some octavia services can only bind to raw IP addresses (i.e.
they won't perform reverse lookups on host names). They wil
fail when using the result of CrowbarHelper.get_host_for_admin_url().
Hence his commit switches over to raw IP addresses.

Octavia: Fix undeclared cluster_vhostname

Octavia: Clean up openstack calls

octavia: Fix name of octavia barclamp

It needs to be lowercase, as the template file for proposal is using
lowercase.

octavia: Only show the keyname attribute if ssh access is enbled

octavia: Fix text in the UI

There were some capitalization issues, and more.

Octavia: Fix router creation

Octavia: Prepare the octavia database form the cookbook

octavia: create and use OctaviaHelper.

This also shortens do block in octavia_conf define to a Hound
friendly length by using OctaviaHelper.

Make service specific config file sections conditional.

octavia: address Hound comments in helper.

Revert "Make service specific config file sections conditional."

This reverts commit 136bfe09555fc057b82d14c67b435debf6d285cd.

octavia: switch back to static health manager bind port

Octavia: fixed helpers

octavio: add timeout settings for amphora build.

Octavia: removed subdoers

Octavia: Fix database creation content

Octavia: Simplification of openstack command

+ some hound problems solved

Octavia: simplify network_settings method

Octavia: Deleted certificates generation & more

- Removed migrations
- Fix problem when octavia router is created

Octavia: removed random passphrase initialization (certificates)

octavia: fix non-HA port for API.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -18,6 +18,7 @@
 
 **/*keystone*  @rhafer
 **/*monasca*  @jgrassler
+**/*octavia*  @ilausuch
 
 ha*.rb  @aspiers
 *_ha.rb  @aspiers
--- a/bin/crowbar_octavia
+++ b/bin/crowbar_octavia
@@ -1,0 +1,22 @@
+#!/usr/bin/env ruby
+#
+# Copyright 2019, SUSE LINUX Products GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+require File.join(File.expand_path(File.dirname(__FILE__)), "barclamp_lib")
+@barclamp = "octavia"
+@timeout = 3600
+
+main
--- a/chef/cookbooks/octavia/README.rdoc
+++ b/chef/cookbooks/octavia/README.rdoc
@@ -1,0 +1,33 @@
+DESCRIPTION
+===========
+
+Chef Cookbook to install and configure OpenStack Octavia.
+
+REQUIREMENTS
+============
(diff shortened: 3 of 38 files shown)


dirkmueller:
This is not supposed to be a hostile fork of #2098 , but this one is only adding new code and hides the barclamp with user_managed: false.

the idea is, if we can agree to that, to merge that part and then work on test automation and finishing up the implementation.

jgrassler:
> This is not supposed to be a hostile fork of #2098 , but this one is only adding new code and hides the barclamp with user_managed: false.

What new code does it add over https://github.com/crowbar/crowbar-openstack/pull/2082 ? It's hard to tell since the commit is already squashed...

dirkmueller:
@jgrassler none, its the same code. the only change is user_manged: false

vuntz:
I think I had mentioned it, but apparently no: I'd feel more comfortable having only a dummy barclamp (and hidden, as you did) for now, which really means removing the cookbook content. That way we can apply the barclamp, and the work that is more challenging (getting the cookbook to the state that is desired) is actually left to a separate state.

I'd keep the barclamp attributes as I guess that will help prepare the CI side...

dirkmueller:
original pr continues

PR #2099, opened 2019-04-10
tempest: Fix Heat flavor id

skazi0:
ID 8 was already used in nova cookbook.
Taking 20 to leave some space for the future (and skip 13).

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -278,7 +278,7 @@ end
 
 flavor_ref = "6"
 alt_flavor_ref = "7"
-heat_flavor_ref = "8"
+heat_flavor_ref = "20"
 
 bash "create_yet_another_tiny_flavor" do
   code <<-EOH


chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -278,7 +278,7 @@
 
 flavor_ref = "6"
 alt_flavor_ref = "7"
-heat_flavor_ref = "8"
+heat_flavor_ref = "13"
```
dirkmueller: as we don't have to be overly picky on numbering, how about leaving some more space and call it "20" ?  I think 20 is an awesome number.

skazi0: Well... I took 20 already for ironic flavor (another PR)... but OK, 21 is even better :o)

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -278,7 +278,7 @@
 
 flavor_ref = "6"
 alt_flavor_ref = "7"
-heat_flavor_ref = "8"
+heat_flavor_ref = "20"
```
vuntz: Can't we use `openstack flavor create` which doesn't seem to require this thing?

vuntz: Something like `openstack flavor create -c id -f value --ram 128 --disk 0 --vcpus 1 myflavor` (and this way we get the id automatically)

skazi0: `nova flavor-create auto` does the same. I think the point here is that we have fixed IDs which need to be placed in config files and we don't need to look them up after creation.

vuntz: Right, but the problem seems to be that these fixed IDs may conflict, so I was wondering why not avoid the conflict altogether.

I guess it's fine to do the minimal change for now, but if we don't need fixed IDs, it's probably best to stop using them in the future.

skazi0: Agree. I just followed the "broken windows" pattern :o)

PR #2100, opened 2019-04-10
ironic: Install deploy image with ironic

skazi0:
The official SUSE deploy image package will be installed by default
together with ironic-server role. Kernel and ramdisk images will be
uploaded to glance.

### DIFF
--- a/chef/cookbooks/ironic/recipes/post_install.rb
+++ b/chef/cookbooks/ironic/recipes/post_install.rb
@@ -1,0 +1,69 @@
+#
+# Copyright 2019 SUSE
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Cookbook Name:: ironic
+# Recipe:: post_install
+#
+
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+
+# the image is uploaded via the glance API
+glance_config = Barclamp::Config.load("openstack", "glance", node[:ironic][:glance_instance])
+glance_insecure = CrowbarOpenStackHelper.insecure(glance_config)
+openstack_args_glance = glance_insecure || keystone_settings["insecure"] ? "--insecure" : ""
+
+env = "OS_USERNAME='#{keystone_settings["service_user"]}' "
+env << "OS_PASSWORD='#{keystone_settings["service_password"]}' "
+env << "OS_PROJECT_NAME='#{keystone_settings["service_tenant"]}' "
+env << "OS_AUTH_URL='#{keystone_settings["internal_auth_url"]}' "
+env << "OS_INTERFACE=internal "
+env << "OS_IDENTITY_API_VERSION=3 "
+
+kernel_image_prefix = "ir-deploy-kernel"
+ramdisk_image_prefix = "ir-deploy-ramdisk"
+image_path = "/srv/tftpboot/openstack-ironic-image"
+
+# vmlinux symlink from openstack-ironic-image points to something like:
+#   openstack-ironic-image.x86_64-9.0.0.kernel.4.12.14-95.13-default
+# find all x.y.z version substrings and pick the first one to be used
+# as version suffix
+image_version_cmd = "$(readlink #{image_path}/vmlinux | egrep -o '[0-9]+\\.[0-9]+\\.[0-9]+' | head -n1)"
+
+openstack_cmd = "#{env} openstack"
+
+bash "upload_ironic_deploy_kernel_image" do
+  code "#{openstack_cmd} #{openstack_args_glance} image create \
+  --disk-format aki --container-format aki --public \
+  --file #{image_path}/vmlinux #{kernel_image_prefix}-#{image_version_cmd}"
+  not_if "#{openstack_cmd} #{openstack_args_glance} image list -f value -c Name | grep -q #{kernel_image_prefix}-#{image_version_cmd}"
+  action :nothing
+end
+
+bash "upload_ironic_deploy_ramdisk_image" do
+  code "#{openstack_cmd} #{openstack_args_glance} image create \
+  --disk-format ari --container-format ari --public \
+  --file #{image_path}/initrd #{ramdisk_image_prefix}-#{image_version_cmd}"
+  not_if "#{openstack_cmd} #{openstack_args_glance} image list -f value -c Name | grep -q #{ramdisk_image_prefix}-#{image_version_cmd}"
+  action :nothing
+end
+
+# This is to trigger the above resource to run :delayed, so that they run at
+# the end of the chef-client run, after the ironic services have been restarted
+# (in case of a config change)
+execute "trigger-ironic-post-commands" do
+  command "true"
+  notifies :run, "bash[upload_ironic_deploy_kernel_image]", :delayed
+  notifies :run, "bash[upload_ironic_deploy_ramdisk_image]", :delayed
+end
--- a/chef/cookbooks/ironic/attributes/default.rb
+++ b/chef/cookbooks/ironic/attributes/default.rb
@@ -20,6 +20,7 @@ when "rhel", "suse"
       "openstack-ironic",
       "openstack-ironic-api",
       "openstack-ironic-conductor",
+      "openstack-ironic-image-x86_64",
       "python-ironicclient",
       "python-openstackclient",
       "qemu-tools"
--- a/chef/cookbooks/ironic/recipes/role_ironic_server.rb
+++ b/chef/cookbooks/ironic/recipes/role_ironic_server.rb
@@ -17,4 +17,5 @@
 if CrowbarRoleRecipe.node_state_valid_for_role?(node, "ironic", "ironic-server")
   include_recipe "ironic::server"
   include_recipe "ironic::tftp"
+  include_recipe "ironic::post_install"
 end


chef/cookbooks/ironic/recipes/post_install.rb
```diff
@@ -0,0 +1,63 @@
...
+env = "OS_USERNAME='#{keystone_settings["service_user"]}' "
+env << "OS_PASSWORD='#{keystone_settings["service_password"]}' "
+env << "OS_PROJECT_NAME='#{keystone_settings["service_tenant"]}' "
+env << "OS_AUTH_URL='#{keystone_settings["internal_auth_url"]}' "
+env << "OS_INTERFACE=internal "
+env << "OS_IDENTITY_API_VERSION=3"
+
+openstack_cmd = "#{env} openstack"
+
+kernel_image_name = "suse-ironic-vmlinux"
```
dirkmueller: it would be better if this is a more specific name that also includes a version number. this way we have a way to update/upload a new image when it becomes available?

chef/cookbooks/ironic/recipes/post_install.rb
```diff
@@ -0,0 +1,63 @@
...
+ramdisk_image_name = "suse-ironic-initrd"
+image_path = "/srv/tftpboot/openstack-ironic-image"
+
+execute "upload_ironic_deploy_kernel_image" do
+  command "#{openstack_cmd} #{openstack_args_glance} image create \
+  --disk-format aki --container-format aki --public \
+  --file #{image_path}/vmlinux #{kernel_image_name}"
+  not_if "#{openstack_cmd} #{openstack_args_glance} image list -f value -c Name | grep -q #{kernel_image_name}"
```
dirkmueller: needs to check for !ha_enabled || is_founder_node?

chef/cookbooks/ironic/recipes/post_install.rb
```diff
@@ -0,0 +1,63 @@
...
+  action :nothing
+end
+
+execute "upload_ironic_deploy_ramdisk_image" do
+  command "#{openstack_cmd} #{openstack_args_glance} image create \
+  --disk-format ari --container-format ari --public \
+  --file #{image_path}/initrd #{ramdisk_image_name}"
+  not_if "#{openstack_cmd} #{openstack_args_glance} image list -f value -c Name | grep -q #{ramdisk_image_name}"
```
dirkmueller: needs to check for !ha_enabled || is_founder_node?

skazi0: ironic barclamp doesn't currently support HA so it can be placed only on one node.

chef/cookbooks/ironic/recipes/post_install.rb
skazi0: ironic barclamp doesn't currently support HA so it can be placed only on one node.

chef/cookbooks/ironic/recipes/post_install.rb
skazi0: hmm... this would cause some problems when we need to find the IDs of those images in glance (I planned to grep for "suse-ironic-vmlinux").

guangyee: FWIW, Ardana have the version number in the image names for the reasons Dirk mentioned.

skazi0: do you have some link to the ardana code which does that? which version is added to the name?

skazi0:
Versioning of images in glance added.

chef/cookbooks/ironic/recipes/post_install.rb
```diff
@@ -0,0 +1,67 @@
...
+env << "OS_PROJECT_NAME='#{keystone_settings["service_tenant"]}' "
+env << "OS_AUTH_URL='#{keystone_settings["internal_auth_url"]}' "
+env << "OS_INTERFACE=internal "
+env << "OS_IDENTITY_API_VERSION=3 "
+
+kernel_image_prefix = "ir-deploy-kernel"
+ramdisk_image_prefix = "ir-deploy-ramdisk"
+image_path = "/srv/tftpboot/openstack-ironic-image"
+
+image_version_env = "IMAGE_VERSION=$(readlink #{image_path}/vmlinux | egrep -o '[0-9]+\\.[0-9]+\\.[0-9]+' | head -n1)"
```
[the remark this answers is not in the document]
guangyee: This would work so as long as the file naming convention stayed the same.

chef/cookbooks/ironic/recipes/post_install.rb
```diff
@@ -0,0 +1,71 @@
...
+glance_config = Barclamp::Config.load("openstack", "glance", node[:ironic][:glance_instance])
+glance_insecure = CrowbarOpenStackHelper.insecure(glance_config)
+openstack_args_glance = glance_insecure || keystone_settings["insecure"] ? "--insecure" : ""
+
+env = "OS_USERNAME='#{keystone_settings["service_user"]}' "
+env << "OS_PASSWORD='#{keystone_settings["service_password"]}' "
+env << "OS_PROJECT_NAME='#{keystone_settings["service_tenant"]}' "
+env << "OS_AUTH_URL='#{keystone_settings["internal_auth_url"]}' "
+env << "OS_INTERFACE=internal "
+env << "OS_IDENTITY_API_VERSION=3 "
```
toabctl: Don't we need the domain as well? Like `OS_DEFAULT_DOMAIN=#{keystone_settings["admin_domain"]}`

chef/cookbooks/ironic/recipes/post_install.rb
```diff
@@ -0,0 +1,71 @@
...
+ramdisk_image_prefix = "ir-deploy-ramdisk"
+image_path = "/srv/tftpboot/openstack-ironic-image"
+
+# vmlinux symlink from openstack-ironic-image points to something like:
+#   openstack-ironic-image.x86_64-9.0.0.kernel.4.12.14-95.13-default
+# find all x.y.z version substrings and pick the first one to be used
+# as version suffix
+image_version_env = "IMAGE_VERSION=$(readlink #{image_path}/vmlinux | egrep -o '[0-9]+\\.[0-9]+\\.[0-9]+' | head -n1)"
+
+# note: semicolon after #{image_version_env} makes this variable available to current shell,
```
toabctl: Isn't that like using `environment` in the `bash` chef resource (which I would prefer if that is true)?

chef/cookbooks/ironic/recipes/post_install.rb
skazi0: I'm not sure. There are at least a couple of places which didn't set this (e.g. glance/recipes/setup.rb).

chef/cookbooks/ironic/recipes/post_install.rb
skazi0: Will try that. Thanks for the hint.

chef/cookbooks/ironic/recipes/post_install.rb
```diff
@@ -0,0 +1,71 @@
...
+openstack_args_glance = glance_insecure || keystone_settings["insecure"] ? "--insecure" : ""
+
+env = "OS_USERNAME='#{keystone_settings["service_user"]}' "
+env << "OS_PASSWORD='#{keystone_settings["service_password"]}' "
+env << "OS_PROJECT_NAME='#{keystone_settings["service_tenant"]}' "
+env << "OS_AUTH_URL='#{keystone_settings["internal_auth_url"]}' "
+env << "OS_INTERFACE=internal "
+env << "OS_IDENTITY_API_VERSION=3 "
+
+kernel_image_prefix = "ir-deploy-kernel"
```
toabctl: Can we name this `ironic-deploy-kernel` ? `ir` might be a bit cryptic. Or is there some length limitation?

chef/cookbooks/ironic/recipes/post_install.rb
```diff
@@ -0,0 +1,71 @@
...
+ramdisk_image_prefix = "ir-deploy-ramdisk"
```
toabctl: same here

chef/cookbooks/ironic/recipes/post_install.rb
skazi0: This was changed based on Ardana naming.

chef/cookbooks/ironic/recipes/post_install.rb
skazi0: This is tricky. Note that value of `IMAGE_VERSION` is and output of a command that it executed together with the command which uses it. With `environment` property you need to have the value already in chef compile phase.

Maybe a better option is to not set any variable but use the `$(readlink ...)` part directly in the commands.

skazi0: @toabctl I changed the PR to inline the version command in the openstack commands. The amount of work is the same but the structure is more straightforward imo.

chef/cookbooks/ironic/recipes/post_install.rb
skazi0: see: https://github.com/ArdanaCLM/ironic-ansible/blob/master/roles/ironic-common/tasks/deploy-image-upload.yml#L54

guangyee: Yes, I think we should keep the names consistent with Ardana to avoid further confusion.

chef/cookbooks/ironic/recipes/post_install.rb
```diff
@@ -0,0 +1,69 @@
...
+  --disk-format ari --container-format ari --public \
+  --file #{image_path}/initrd #{ramdisk_image_prefix}-#{image_version_cmd}"
+  not_if "#{openstack_cmd} #{openstack_args_glance} image list -f value -c Name | grep -q #{ramdisk_image_prefix}-#{image_version_cmd}"
+  action :nothing
+end
+
+# This is to trigger the above resource to run :delayed, so that they run at
+# the end of the chef-client run, after the ironic services have been restarted
+# (in case of a config change)
+execute "trigger-ironic-post-commands" do
```
toabctl: Is this tested with HA? What if 3 cluster nodes want to upload the image at the same time?

skazi0: Current ironic barclamp doesn't support HA (i.e. you can't assign role to cluster, only to single node).

PR #2101, opened 2019-04-11
Revert "novnc: disabled ssl based checks via haproxy"

cmurphy:
The bug in nova that was causing the CPU spikes when haproxy does health
checks is now fixed[1], so we can reenable TLS health checks for
nova-novncproxy.

[1] https://review.openstack.org/649375

This reverts commit bc439f8c686bb4e4533687d604508c74e538212d.

### DIFF
--- a/chef/cookbooks/nova/recipes/controller_ha.rb
+++ b/chef/cookbooks/nova/recipes/controller_ha.rb
@@ -55,12 +55,7 @@ if node[:nova][:use_novnc]
   haproxy_loadbalancer "nova-novncproxy" do
     address "0.0.0.0"
     port node[:nova][:ports][:novncproxy]
-    # novnc proxy does not like empty ssl packet followed by an RST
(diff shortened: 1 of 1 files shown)


cmurphy:
Needs https://review.openstack.org/649375 to merge and make it into the package.

aspiers:
Sounds reasonable to me.  CI is failing:

    rabbitmq-port-blocker: migration-threshold=3 fail-count=1 last-failure='Thu Apr 11 06:20:27 2019'

but that's probably unrelated.

cmurphy:
At the moment, it probably is related. The issue is that haproxy hammers the nova-novncproxy service and the nova bug causes massive CPU load on the controller, which could take out rabbitmq or anything else running on the controller. Once the upstream patch lands and makes it into the package, the issue should be gone and we can recheck the HA job.

cmurphy:
@jgrassler totally agreed, I just had it on my to-do list and didn't want to forget about it

cmurphy:
This should probably be looked at again.

JanZerebecki (approved):

PR #2102, opened 2019-04-11
nova, neutron: Hide the ZVM attributes

vuntz:
This is not supported, so keep the attributes for the raw view.


toabctl:
@dirkmueller didn't you say somewhere that this is working and somebody is also improving it?

vuntz:
> @dirkmueller didn't you say somewhere that this is working and somebody is also improving it?

Note that it's not removing the feature at all, it's only requiring users to edit the attributes in the raw view. That makes the zvm support less visible in the UI. Which is actually fine, since it's not supported.

PR #2104, opened 2019-04-15
designate: allow worker on cluster.

sjamgade:
[1] suggests that workers are there for designate-central to get work
done. So having more of them is a good idea.
A worker does not really have an idea of api so no vip or haproxy
config is required.

1 - https://docs.openstack.org/designate/rocky/contributor/architecture.html

### DIFF
--- a/chef/cookbooks/designate/definitions/designate_service.rb
+++ b/chef/cookbooks/designate/definitions/designate_service.rb
@@ -16,7 +16,6 @@
 define :designate_service do
   designate_service_name = "designate-#{params[:name]}"
   ha_enabled = node[:designate][:ha][:enabled]
-  use_crowbar_pacemaker_service = ha_enabled && node[:pacemaker][:clone_stateless_services]
 
   package "openstack-designate-#{params[:name]}"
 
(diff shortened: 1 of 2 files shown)


sjamgade:
either way admins could have them deployed on multiple nodes without necessarily calling it HA

jgrassler:
Feel free to review but please do not merge, yet. We'll give this a spin in a QA environment first.

cmurphy:
If it can already be deployed individually to multiple nodes why does it need to be part of a cluster?

dirkmueller:
@cmurphy count is set to 1, so only one node can be assigned atm

jgrassler:
It makes sense to have it running on all cluster nodes if an HA cluster is available (it is just a RabbitMQ driven backend server after all). This pull request is just a shortcut for convenience.

vuntz:
Am I reading correctly that the service will use `provider Chef::Provider::CrowbarPacemakerService if use_crowbar_pacemaker_service` ?

If yes, this needs to be fixed and systemd needs to be used instead. You can compare with how it's done for other services (like glance), which has a similar code structure for the cookbook.

guangyee:
It should be on all cluster nodes right, unless Designate can tolerate single-point-of-failure for the workers node? fwiw, looking at the Ardana input models, it appears Designate workers are running on all the clustered nodes.

dirkmueller:
@guangyee thats what the code does now.

cmurphy:
Is it expected that the ha_want_designate job is failing?

sjamgade:
https://github.com/SUSE-Cloud/automation/pull/3443 and https://jira.suse.com/browse/SOC-9342  is needed for tempest (smoke) test to pass.
thats why the ha_want_designate job is failing.

PR #2107, opened 2019-04-16
ironic: Use IP also in [swift] section

skazi0:
Swift settings in [glance] already use IP for endpoint URL. This
is used for deployment images.

If config drive is enabled and stored in Swift, this endpoint URL
is not used but instead standard keystone adapter is created based
on [swift] section of Ironic configuration. Added override makes
it possible to use deploy image which can't resolve Crowbar's
internal domain names.


dirkmueller:
so why can it reach crowbar internal ip addresses but not hostnames? I am confused about that.

skazi0:
@dirkmueller I think the deploy image doesn't have proper resolver setup. I tried the same with upstream coreos deploy image and it also didn't work with hostnames.

skazi0:
testing gating with https://github.com/crowbar/crowbar-openstack/pull/2116

skazi0:
testing gating again with #2116 (new version)

PR #2108, opened 2019-04-16
neutron: disable metering agent if no ceilometer

dirkmueller:
Also cleanup code a bit to use the proper entry points rather
than dotted path notations (which cause a warning)


rhafer (dismissed):

rhafer (dismissed):

toabctl (dismissed):

JanZerebecki:
Changed commit message to match regex checked by travis.

JanZerebecki (dismissed):

PR #2109, opened 2019-04-16
designate: do not install the keystone_authtoken on worker nodes

sjamgade:
the desigante-worker does not need keystone auth token, it commnunicates
only with desginate-central over rpc.

the diagram gives a hint of this [1] or else one could look the code [2]

1 -
https://docs.openstack.org/designate/rocky/contributor/architecture.html
2-
https://opendev.org/openstack/designate/src/branch/master/designate/worker/service.py


dirkmueller:
rebased on top of the HA fix

PR #2110, opened 2019-04-17
neutron: raise validation error if domain names dont end with a dot(.)

sjamgade:
This is needed for designate. Designate will try to create zones
based on these names which get updated into the fixed and floating
network.

### DIFF
--- a/crowbar_framework/app/models/neutron_service.rb
+++ b/crowbar_framework/app/models/neutron_service.rb
@@ -411,6 +411,11 @@ class NeutronService < OpenstackServiceObject
       validate_external_networks proposal["attributes"]["neutron"]["additional_external_networks"]
     end
 
+    ["dns_domain", "floating_dns_domain"].each do |domain|
+      validation_error "#{domain} should end with a dot"\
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/neutron_service.rb
```diff
@@ -96,6 +96,12 @@ def save_proposal!(prop, options = {})
         end
       end
     end
+    neutron = prop.raw_data[:attributes][:neutron]
```
dirkmueller: I think silently changing the data we save is not a nice design. given that this is just json with a schema, can you add it as a schema validation rule? this can be solved by a simple regexp.

sjamgade: then I would have to figure out why exactly the json failed to parse because user could edit it in RAW mode and there could be a json error anywhere. So if its bad design to change data silently, I will just add it as a validation error.

sjamgade: done.

sjamgade:
Rebased on top of HA fix

jgrassler:
@dirkmueller can you take another look at this, too? Thanks!

PR #2111, opened 2019-04-17
tempest: enable multiattach for NetApp + LVM (SCPM-97)

dirkmueller:
just disable it for the backends that are known to not support it


dirkmueller:
@flaviodsr unintended change, removed this part.

PR #2112, opened 2019-04-17
neutron: restart neutron-ha-tool when the config file changes

dirkmueller:
When we toggle ssl/non-ssl in keystone, neutron-ha-tool
gets a new config file but doesn't get restarted, so it just
repeatedly crashes all the way until it causes a pacemaker
failcount exceeded and then its dead. We should try better.

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -108,6 +108,15 @@ if use_l3_agent
       )
     end
 
+    service "neutron-l3-ha-service" do
+      supports status: true, restart: true
+      subscribes :restart, resources(file: "/etc/neutron/neutron-l3-ha-service.yaml")
+      subscribes :restart, resources(template: "/root/.openrc")
+      subscribes :restart, resources(file: "/etc/neutron/os_password")
+
+      provider Chef::Provider::CrowbarPacemakerService
+    end
+
     # Reload systemd when unit file changed
     bash "reload systemd after neutron-l3-ha-service update" do
       code "systemctl daemon-reload"


dirkmueller:
(needs testing, and probably does not work as is)

dirkmueller:
yeah, not sure about that problem, it happens quite frequently recently.

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -154,6 +154,7 @@
       agent "systemd:neutron-l3-ha-service"
       op node[:neutron][:ha][:neutron_l3_ha_resource][:op]
       action :update
+      subscribes :restart, resources("template[/etc/neutron/neutron-l3-ha-service.yaml]")
```
cmurphy: This is a file resource, not a template resource, so this won't match anything

https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/neutron/recipes/network_agents_ha.rb#L79-L85

dirkmueller: fixed

chef/cookbooks/neutron/recipes/network_agents_ha.rb
```diff
@@ -154,6 +154,9 @@
       agent "systemd:neutron-l3-ha-service"
       op node[:neutron][:ha][:neutron_l3_ha_resource][:op]
       action :update
+      subscribes :restart, resources(file: "/etc/neutron/neutron-l3-ha-service.yaml")
+      subscribes :restart, resources(template: "/root/.openrc")
+      subscribes :restart, resources(file: "/etc/neutron/os_password")
```
vuntz: Just curious: any reason to not use the `"template[/root/.openrc]"` syntax? Is it to crash if the type is not correct (file vs template)?

sjamgade:
```Error executing action `restart` on resource 'pacemaker_primitive[neutron-l3-ha-service]'``` from CI logs

chef/cookbooks/neutron/recipes/network_agents_ha.rb
dirkmueller: @vuntz this semed to be the more prevalent syntax. I tried both, both don't work.

dirkmueller:
yes, thats why there is "do not merge" and "wip" and "untested, probably does not work" on the PR

vuntz:
Ah, I hadn't realized. In case it's not already solved on your hard disk: the `restart` action needs to be on a `service` resource, not a pacemaker resource. So we need to define a service for neutron-l3-ha-service, make sure the service uses the pacemaker provider, and that should do it.

dirkmueller:
@vuntz thanks, that did it

PR #2115, opened 2019-04-25
[4.0] keystone: fix SSL HA scenario (SCRD-9026)

stefannica:
The SSL+HA combo is currently broken (only) in Crowbar SOC7. The
keystone barclamp is failing with the following failure:

```
FATAL: Errno::ENOENT: ruby_block[synchronize signing keys for founder and remember them for non-HA case] (keystone::server line 347) had an error: Errno::ENOENT: No such file or directory @ rb_sysopen - /etc/keystone/ssl/certs/ca.pem
```

This happens because there's a code block in the SOC7 keystone
server recipe that is synchronizing the keystone certificates
between the founder and the non-founder nodes. This code block
is a legacy [from the days when PKI tokens were still supported](https://github.com/crowbar/crowbar-openstack/blob/51bbfcf0a613c5aca975d32b8f83b6102b0d4623/chef/cookbooks/keystone/recipes/server.rb#L311-L380),
and [was re-added](https://github.com/crowbar/crowbar-openstack/pull/743) after [support for PKI tokens was removed](https://github.com/crowbar/crowbar-openstack/pull/702/commits/da320389b2591b44e4088dbcdcb164998cabaf47)
for the only reason that the [ceph rados gateway role still requires
these certificates to be present in SOC7](https://github.com/crowbar/crowbar-ceph/pull/83).

This commit does the following to remedy the situation:
 - uses the keystone SSL certificate path attributes instead of
 the hard-coded path values. This fixes the CI jobs, because [the
 CI now explicitly configures the CA certificate attribute](https://github.com/SUSE-Cloud/automation/commit/4194b0ee6fbe05305d17bca2166ca87fd5cefc02) in the
 keystone barclamp to point to the CA certificate installed with
 the SUSE `ca-certificates' package
 - remove the code that synchronizes the certificate file contents
 between the founder and non-founder nodes, because this is no
 longer required. The only thing that is required by the ceph
 barclamp is that [the certificate file contents be saved as node
 attributes](https://github.com/crowbar/crowbar-ceph/blob/stable/4.0/chef/cookbooks/ceph/recipes/radosgw_keystone.rb#L112-L139), to be [converted to nss db format](http://docs.ceph.com/docs/jewel/radosgw/keystone/) . With the
 synchronization code out of the way, the sync marks also become
 useless and are removed.

UPDATE: The `openstack-keystone` SOC7 package should install the certificates in the correct paths (`/etc/keystone/ssl/certs/ca.pem`). This fix only helps to hide the real root cause of the issue. For some reason that I couldn't identify, the post-install section of the `openstack-keystone` package that is responsible for generating these certificates doesn't work in the SSL+HA scenario when invoked from the chef recipes. When reinstalled manually after the fact, the package works fine and the certificates are put in place. The recipe doesn't fail either, because failures in the the post-install package script aren't reported as package installation failures.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -335,21 +335,13 @@ ruby_block "mark node for keystone db_sync" do
   subscribes :create, "execute[keystone-manage db_sync]", :immediately
 end
 
-if ha_enabled
-  crowbar_pacemaker_sync_mark "create-keystone_db_sync"
+crowbar_pacemaker_sync_mark "create-keystone_db_sync" if ha_enabled
 
-  # Make sure the certificates code is run on the founder first
-  crowbar_pacemaker_sync_mark "wait-keystone_certificates" do
-    fatal true
-  end
-end
-
-ruby_block "synchronize signing keys for founder and remember them for non-HA case" do
-  only_if { (!ha_enabled || (ha_enabled && CrowbarPacemakerHelper.is_cluster_founder?(node))) }
+ruby_block "store certificate file contents for ceph radosgw" do
+  only_if { (!ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)) }
   block do
-    ca = File.open("/etc/keystone/ssl/certs/ca.pem", "rb", &:read)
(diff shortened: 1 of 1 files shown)


stefannica:
Not needed anymore, the SOC7 SSL+HA jobs are running fine now.

PR #2116, opened 2019-04-27
neutron: use crm_resource restart for restarting neutron-l3-ha-service

dirkmueller:
When we restart via systemd, it can happen that the service already
crashed and then 'service neutron-l3-ha-service status" will (correctly)
report that the service isn't running. And then the restart is not being
executed due to:

  INFO: Ignoring restart action for neutron-l3-ha-service service since not running on this node (d52-54-77-77-01-01)

which will then later cause a pacemaker failcount. the pacemaker
resource restart hopefully does not have this problem.

### DIFF
--- a/chef/cookbooks/neutron/recipes/network_agents_ha.rb
+++ b/chef/cookbooks/neutron/recipes/network_agents_ha.rb
@@ -109,10 +109,10 @@ if use_l3_agent
     end
 
     service "neutron-l3-ha-service" do
-      supports status: true, restart: true
-      subscribes :restart, resources(file: "/etc/neutron/neutron-l3-ha-service.yaml")
-      subscribes :restart, resources(template: "/root/.openrc")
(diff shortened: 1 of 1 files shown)


skazi0:
@dirkmueller the gating job still fails with the same symptomps :o(

sjamgade:
```
neutron-l3-ha-service.log
 [2019-05-07T16:35:25.390173 #23874]  INFO -- : error counter: bumped to 11   
chef logs
[2019-05-07T16:38:49+00:00] INFO: Ignoring restart action for neutron-l3-ha-service service since not running on this node (d52-54-77-77-01-01)  
service was running on d52-54-77-77-01-01
```

By default neutron-ha-tool restart is subscribed as  **delayed** so all restarts are queued, So the actual restart happens after chef is done. 

So may be changing it to not-delayed should help

PR #2122, opened 2019-05-09
swift: Reset syncmarks when applying proposal

skazi0:
Each barclamp which uses HA should reset syncmarks when applying proposal.

### DIFF
--- a/chef/data_bags/crowbar/template-swift.json
+++ b/chef/data_bags/crowbar/template-swift.json
@@ -105,10 +105,10 @@
       },
       "elements": {},
       "element_order": [
+        [ "swift-proxy" ],
         [ "swift-storage" ],  
         [ "swift-ring-compute" ], 
         [ "swift-storage" ],
-        [ "swift-proxy" ],
         [ "swift-dispersion" ]
       ],
       "element_run_list_order": {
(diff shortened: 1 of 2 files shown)


skazi0:
@aspiers @dirkmueller @Itxaka This is very tricky piece. Currently swift doesn't reset syncmarks when applying proposal and adding this is not trivial because of mix of controller/non-controller roles in swift barclamp. With no syncmark reset, all previous (usually database, rabbit, haproxy and keystone) syncmarks are already set when we start chef-client run when applying swift proposal. This way, when we reach swift recipes, nodes could be very much out of sync (which happens for me quite often). I see following options:
1. add 'sync' type syncmark somewhere in the beginning of swift cookbook to force the nodes to line up (no idea if this will be enough)
2. rework the controller/non-controller roles somehow (maybe add new "dummy" ones?) to model this scenario better (no idea how to do it)
3. something else?

Itxaka:
Adding the sync should be straightforward and the easiest way to achieve sync between the nodes no? If then during the swift cookbooks they still desync we can add more sync marks so they keep on sync.

skazi0:
I'm testing this now: https://github.com/crowbar/crowbar-openstack/pull/2160

skazi0:
closing (again) in favor of https://github.com/crowbar/crowbar-openstack/pull/2160

PR #2114, opened 2019-04-24
database: Raise and align promote/demote timeouts (bsc#1131791)

dirkmueller:
Looks like 5 minutes is too short in some cases
when a new node is joining. I have observed the 5 min timeout
in a small cloud of 100 nodes already that has been operated
for some time, so we might be better off being conservative.

PR #2124, opened 2019-05-13
database: Increase the default "promote" timeout (bsc#1131791)

toabctl:
When a galera cluster is out of sync and needs a resync, the cluster
may decide to wipe the available data on the out-of-sync nodes and do
a full resync to *all* (except the bootstrap node of course) cluster
nodes. In a 3-node cluster, it means that the data needs to be
transfered 2x.
To not run into timeouts in that case, increase the timeout to 10
minutes.

### DIFF
--- a/chef/data_bags/crowbar/template-database.json
+++ b/chef/data_bags/crowbar/template-database.json
@@ -43,7 +43,7 @@
               "timeout": "60s"
             },
             "promote": {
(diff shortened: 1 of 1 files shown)


toabctl:
Already a PR available: https://github.com/crowbar/crowbar-openstack/pull/2114

PR #2127, opened 2019-05-14
database: Make wsrep_provider_options configurable (fate#327745)

toabctl:
It is sometimes useful to be able to add extra options (like
gcache.size, gcs.fc_debug, ...) to the galera wsrep_provider_options
configuration variable.
This can now be done via the Crowbar RAW view.

### DIFF
--- a/chef/cookbooks/mysql/templates/default/galera.cnf.erb
+++ b/chef/cookbooks/mysql/templates/default/galera.cnf.erb
@@ -3,7 +3,7 @@ wsrep_on = ON
 wsrep_provider = /usr/lib64/galera-3/libgalera_smm.so
 wsrep_cluster_address = "<%= @cluster_addresses %>"
 # values recommended by mysqltuner.pl
-wsrep_provider_options = "gmcast.listen_addr=tcp://<%= @node_address %>:4567;gcs.fc_limit = <%= @wsrep_slave_threads * 5 %>;gcs.fc_factor = 0.8"
+wsrep_provider_options = "gmcast.listen_addr=tcp://<%= @node_address %>:4567;gcs.fc_limit = <%= @wsrep_slave_threads * @gcs_fc_limit_multiplier %>;gcs.fc_factor = <%= @gcs_fc_factor %>;<%= @wsrep_provider_options_custom %>"
 wsrep_slave_threads = <%= @wsrep_slave_threads %>
 
 # Maximum number of rows in write set
--- a/chef/data_bags/crowbar/migrate/database/301_make_wsrep_provider_options_configurable.rb
+++ b/chef/data_bags/crowbar/migrate/database/301_make_wsrep_provider_options_configurable.rb
@@ -1,0 +1,13 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["mysql"]["wsrep_provider_options_custom"] = template_attrs["mysql"]["wsrep_provider_options_custom"] unless attrs["mysql"]["wsrep_provider_options_custom"]
+  attrs["mysql"]["gcs_fc_limit_multiplier"] = template_attrs["mysql"]["gcs_fc_limit_multiplier"] unless attrs["mysql"]["gcs_fc_limit_multiplier"]
+  attrs["mysql"]["gcs_fc_factor"] = template_attrs["mysql"]["gcs_fc_factor"] unless attrs["mysql"]["gcs_fc_factor"]
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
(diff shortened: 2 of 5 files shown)


chef/data_bags/crowbar/template-database.json
```diff
@@ -15,6 +15,8 @@
         "expire_logs_days": 10,
         "bootstrap_timeout": 600,
         "wsrep_slave_threads" : 1,
+        "gcs_fc_limit_multiplier" : 5,
+        "wsrep_provider_options_custom" : [],
```
vuntz: Wondering out loud: is it easier/less confusing to use a sequence of string, or a single string (where we would require the user to use `;`)?

toabctl: I don't care. I can also switch to a single string instead of a list of strings. What do you prefer?

alexdepalex: So how would that work in batch? Like `wsrep_provider_options_custom: [ "gcache.size=2G" ]`

vuntz: > What do you prefer?

No strong feeling on my side. Just raising this in case someone has strong feelings ;-)

chef/data_bags/crowbar/migrate/database/301_make_wsrep_provider_options_configurable.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(ta, td, a, d)
```
rhafer: Would you mind switching these to the longer parameter names we started using recently (not so much for making hound happpy, just for making that better understandable? 

`def upgrade(template_attrs, template_deployment, attrs, deployment)`

See e.g. https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/migrate/database/300_noop.rb

chef/cookbooks/mysql/templates/default/galera.cnf.erb
```diff
@@ -3,7 +3,7 @@ wsrep_on = ON
 wsrep_provider = /usr/lib64/galera-3/libgalera_smm.so
 wsrep_cluster_address = "<%= @cluster_addresses %>"
 # values recommended by mysqltuner.pl
-wsrep_provider_options = "gmcast.listen_addr=tcp://<%= @node_address %>:4567;gcs.fc_limit = <%= @wsrep_slave_threads * 5 %>;gcs.fc_factor = 0.8"
+wsrep_provider_options = "gmcast.listen_addr=tcp://<%= @node_address %>:4567;gcs.fc_limit = <%= @wsrep_slave_threads * @gcs_fc_limit_multiplier %>;gcs.fc_factor = 0.8";<%= @wsrep_provider_options_custom %>
```
rhafer: I think we need to make `gcs.fc_factor` an attribute as well if we really want to offer configurabilty here. Otherwise we're moving ourselves into a weird corner.

toabctl: done

chef/data_bags/crowbar/migrate/database/301_make_wsrep_provider_options_configurable.rb
toabctl: done

chef/data_bags/crowbar/migrate/database/301_make_wsrep_provider_options_configurable.rb
```diff
@@ -0,0 +1,14 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+def upgrade(ta, td, a, d)
```
rhafer: Forgot to delete this line?

toabctl: fixed

chef/data_bags/crowbar/template-database.json
toabctl: @alexdepalex, yes. Or with multiple options like:
```
    "wsrep_provider_options_custom": [
      "gcache.size=2G",
      "debug=yes"
    ],
```
which results in:
```
root@d52-54-77-77-01-01:~ # grep wsrep_provider_options /etc/my.cnf.d/75-galera.cnf 
wsrep_provider_options = "gmcast.listen_addr=tcp://192.168.206.82:4567;gcs.fc_limit = 5;gcs.fc_factor = 0.8";gcache.size=2G;debug=yes
```

toabctl:
Backports are:
- 4.0 : https://github.com/crowbar/crowbar-openstack/pull/2130
- 5.0 : https://github.com/crowbar/crowbar-openstack/pull/2129

PR #2131, opened 2019-05-16
nova: Show stdout/stderr when "openstack flavor list" fails

toabctl:
When running into the problem that the "openstack flavor list" command
fails, we want to know the error and not only see a comment that the
nova api might be down.
This is useful for debugging.


toabctl:
Backport 5.0: https://github.com/crowbar/crowbar-openstack/pull/2134

PR #2138, opened 2019-05-21
neutron: Stop creating / watching .openrc

dirkmueller:
On a toggle of the keystone services configuration, the .openrc
is written too early and neutron-l3-ha-service is restarted
prior the configuration yaml being updated. So given that
the l3-ha-service only fetches information from the yaml, we don't
need to trigger restart and we don't need to write the .openrc.


dirkmueller:
duplicate of #2137

PR #2141, opened 2019-05-23
[4.0] neutron: Don't restart l3-ha on .openrc change + timeout extension

skazi0:
l3-ha service doesn't use .openrc so it doesn't need to be restarted
when that file is modified.

Second commit removes redundant generation of .openrc in neutron cookbook.
Third one increases grace period for l3-ha service from 10x10s to 10x30s. This is needed to improve reliability of CI running in ECP.

port of #2137


skazi0:
rebased to get the gating fixes included

PR #2142, opened 2019-05-29
nova: Don't retry creating existing flavors

skazi0:
In some cases the flavor create call succeeds but client still returns
non-zero status. Retries of the create call fail with "Flavor already
exists" and the retry loop never succeeds. Added check is executed in
every loop turn and will stop reytring if the flavor already exists.

Example scenario where flavor might be correctly created but client
doesn't return zero is when one of HA nodes executes flavor create
commands while others perform delayed restart of nova API after config
files are modified. If the "create" request hits the API just before
restart it could be accepted but the client might not get the correct
response back.

PR #2134, opened 2019-05-20
nova: Show stdout/stderr when "openstack flavor list" fails

toabctl:
When running into the problem that the "openstack flavor list" command
fails, we want to know the error and not only see a comment that the
nova api might be down.
This is useful for debugging.

(cherry picked from commit b8776b650f511dc9b42e0e579dc6f49de590a690)

PR #2143, opened 2019-05-30
[5.0] nova: Don't retry creating existing flavors

skazi0:
In some cases the flavor create call succeeds but client still returns
non-zero status. Retries of the create call fail with "Flavor already
exists" and the retry loop never succeeds. Added check is executed in
every loop turn and will stop reytring if the flavor already exists.

Example scenario where flavor might be correctly created but client
doesn't return zero is when one of HA nodes executes flavor create
commands while others perform delayed restart of nova API after config
files are modified. If the "create" request hits the API just before
restart it could be accepted but the client might not get the correct
response back.

port of #2142

### DIFF
--- a/chef/cookbooks/nova/recipes/flavors.rb
+++ b/chef/cookbooks/nova/recipes/flavors.rb
@@ -127,6 +127,8 @@ ruby_block "Flavor creation" do
         )
         flavor_create.command command
         flavor_create.retries 5
+        # don't retry after "Flavor with ID ... already exists"
+        flavor_create.not_if "#{openstack} flavor show #{id}"
 
         # delay the run of this resource until the end of the run
         run_context.notifies_delayed(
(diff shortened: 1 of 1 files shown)


nicolasbock:
Could you look into the CI failure? Seems a bit strange that it's failing on what looks like an SSL problem?

skazi0:
@nicolasbock this looks like the log quoted by @toabctl here: https://github.com/crowbar/crowbar-openstack/pull/2134#issuecomment-493910660
@toabctl is there some fix for this?

toabctl:
> @toabctl is there some fix for this?

Nope. https://github.com/crowbar/crowbar-openstack/pull/2134 just made the real failure visible.

nicolasbock:
At this point I can't see how the CI failure is related to your change here @skazi0 , but the failure is in Nova and I would feel better if we addressed it before merging your change.

PR #2144, opened 2019-05-30
[4.0] nova: Don't retry creating existing flavors

skazi0:
In some cases the flavor create call succeeds but client still returns
non-zero status. Retries of the create call fail with "Flavor already
exists" and the retry loop never succeeds. Added check is executed in
every loop turn and will stop reytring if the flavor already exists.

Example scenario where flavor might be correctly created but client
doesn't return zero is when one of HA nodes executes flavor create
commands while others perform delayed restart of nova API after config
files are modified. If the "create" request hits the API just before
restart it could be accepted but the client might not get the correct
response back.

port of #2142


nicolasbock:
The HA failures are already in Keystone and don't seem to be related to the PR.

skazi0:
rebased to get the gating fixes included

PR #2145, opened 2019-05-31
[4.0] keystone: Use correct paths when syncing certs

skazi0:
The sync failed when certs and/or keys were located in non-default paths.

PR #2146, opened 2019-06-03
[4.0] rabbitmq: Fix ACL of SSL key after uid/gid change + keystone: Use correct paths when syncing certs

skazi0:
In shared storage based HA setup, rabbitmq uses fixed uid/gid=91.
This user/group modification was done after (optional) SSL certificate
generation. The ACLs on the SSL key were incorrect making rabbitmq
unable to start because with EACCESS errors.

Second commit is taken from https://github.com/crowbar/crowbar-openstack/pull/2145 to avoid cross-PR dependencies in gating.

The sync failed when certs and/or keys were located in non-default paths.

**Note:** forward ports need to include only the rabbitmq part as the keystone change is not relevant for newer versions.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -347,9 +347,9 @@ end
 ruby_block "synchronize signing keys for founder and remember them for non-HA case" do
   only_if { (!ha_enabled || (ha_enabled && CrowbarPacemakerHelper.is_cluster_founder?(node))) }
   block do
-    ca = File.open("/etc/keystone/ssl/certs/ca.pem", "rb", &:read)
-    signing_cert = File.open("/etc/keystone/ssl/certs/signing_cert.pem", "rb", &:read)
-    signing_key = File.open("/etc/keystone/ssl/private/signing_key.pem", "rb", &:read)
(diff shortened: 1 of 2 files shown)


skazi0:
This will hopefully fix the gating failures we have in stable/4.0 branch.

skazi0:
Temporarily added this fix https://github.com/crowbar/crowbar-openstack/pull/2145 as both are probably needed to have the gating pass.

cmurphy:
Can you clarify more what the issue was that the commit "keystone: Use correct paths when syncing certs" is fixing? That part of the code isn't doing HTTPS but actually configuration needed for PKI token signing.

I think we're also supposed to fix in master first and then backport?

skazi0:
@cmurphy besides Jan's [comment](https://github.com/crowbar/crowbar-openstack/pull/2145#issuecomment-498426909), the problem is that gating jobs have (recently) enabled `want_all_ssl` which triggered all kinds of SSL failures. In SOC7 it covered some code paths which seem not to be tested very well. In later branches those parts are either removed or not used by default so these problems didn't pop up earlier.

skazi0:
Replaced cherry-picked commit with updated one.

skazi0:
@skazi0 related issue: SOC-9026

PR #2150, opened 2019-06-06
Configurable timeout for Galera pre-sync

AbelNavarro:
We have a pacemaker sync-mark to wait for all nodes to have all galera
packages and configuration files installed. The default timeout is 60
seconds for these sync-marks.

While installing all cluster members at once, like in an initial
installation, all nodes take about the same time to install the packages
and reach the sync-mark about the same time as well. But when adding a
new node to the cluster, the already installed nodes reach the sync-mark
much faster and the default timeout proves insuficient.

Based on tests in developer environments, this takes roughly 70 seconds.
We will be setting this to 180 seconds to give enough margin.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -183,6 +183,7 @@ end
 # all the required packages and configurations installed before we create the
 # pacemaker resources
 crowbar_pacemaker_sync_mark "sync-database_before_ha" do
+  timeout node[:database][:mysql][:presync_timeout]
   revision node[:database]["crowbar-revision"]
 end
 
--- a/chef/data_bags/crowbar/migrate/database/302_add_presync_timeout.rb
+++ b/chef/data_bags/crowbar/migrate/database/302_add_presync_timeout.rb
@@ -1,0 +1,11 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs["mysql"]["presync_timeout"]
+    attrs["mysql"]["presync_timeout"] = template_attrs["mysql"]["presync_timeout"]
+  end
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["mysql"].delete("presync_timeout") unless template_attrs["mysql"].key?("presync_timeout")
+  return attrs, deployment
+end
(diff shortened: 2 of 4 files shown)


AbelNavarro:

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -183,6 +183,7 @@
 # all the required packages and configurations installed before we create the
 # pacemaker resources
 crowbar_pacemaker_sync_mark "sync-database_before_ha" do
+  timeout node[:database][:mysql][:presync_timeout]
```
toabctl: Do we need this to be configurable? From your description, it would be enough to just increase the timeout

AbelNavarro: You have a good point here, but we haven't tried this in customers' environment and being able to configure it in case it takes longer is a bonus.

rhafer: Do you have any idea where that time is actually spent? AFAICS the most time consuing thing that the new node would be doing is installing the packages.  (Most of the other time consuming steps should actually be skipped when adding a new node)

I really doubt that that can take more than 3 minutes.

AbelNavarro: The logs didn't show anything specific, just when started and finished, which happened in about 70 seconds in my developer environment.

AbelNavarro:
> However I think you're using the wrong schema revision. The database barclamp in master is already at 301. Also you forgot to increase the revision in https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/template-database.json#L89

Fixed that. Thanks for noticing the error @rhafer . I'm getting rusty.

chef/data_bags/crowbar/migrate/database/302_add_presync_timeout.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(ta, td, a, d)
```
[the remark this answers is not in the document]
rhafer: In case you want to make hound happy you might wanna change this to what we use in the newer migrations scripts:
https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/migrate/database/300_noop.rb#L1

AbelNavarro:
made Hound (thus rhafer too) happy.

PR #2152, opened 2019-06-17
monasca: use string keys for attrs in migrations

jgrassler:
Symbolic keys cause the migrations to fail during upgrade due
to these keys not being found. This commit fixes the problem.

### DIFF
--- a/chef/data_bags/crowbar/migrate/monasca/302_add_mondb.rb
+++ b/chef/data_bags/crowbar/migrate/monasca/302_add_mondb.rb
@@ -1,20 +1,24 @@
 def upgrade(template_attrs, template_deployment, attrs, deployment)
   unless attrs.key? "db_monapi"
-    attrs[:db_monapi] = template_attrs[:db_monapi]
+    attrs["db_monapi"] = template_attrs[:db_monapi]
     service = ServiceObject.new "fake-logger"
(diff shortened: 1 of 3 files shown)


chef/data_bags/crowbar/migrate/monasca/302_add_mondb.rb
```diff
@@ -4,17 +4,21 @@ def upgrade(template_attrs, template_deployment, attrs, deployment)
...
     attrs[:db_monapi][:password] = service.random_password
   end
+
   # use the database password that is already available
-  if attrs[:master].key? "database_monapi_password"
-    attrs[:db_monapi][:password] = attrs[:master][:database_monapi_password]
-    attrs[:master].delete(:database_monapi_password)
+  if attrs.key? "master"
+    if attrs["master"].key? "database_monapi_password"
+      attrs[:db_monapi][:password] = attrs["master"]["database_monapi_password"]
```
cmurphy: Should these be strings too?

jgrassler: Technically it shouldn't matter: I believe the problem happens due to JSON serialization flattening symbols into strings (and `attrs` is the result of deserializing these strings back into Ruby strings). I still homogenized everything to use string keys though - it does make things more consistent.

PR #2153, opened 2019-06-17
case-insensitive role name lookup (SOC-9339)

guangyee:
We should be using case-insensitive role name comparison
when doing a role lookup since we only support mysql backend
now. See

https://docs.openstack.org/keystone/rocky/admin/identity-case-insensitive.html#roles

### DIFF
--- a/chef/cookbooks/keystone/providers/register.rb
+++ b/chef/cookbooks/keystone/providers/register.rb
@@ -412,7 +412,12 @@ def find_id(item_name, path, dir, key = "name", ret = "id")
       data2hash = {}
 
       data.each do |item|
-        my_item_id = item[ret] if item[key] == item_name
+        # NOTE: for Keystone with MySQL backend, which is the default since
+        # Cloud 8, we should be doing case-insensitive comparison when lookup
+        # ID by name. For more information, see
+        # https://docs.openstack.org/keystone/rocky/admin
+        # /identity-case-insensitive.html
+        my_item_id = item[ret] if item[key].casecmp(item_name).zero?
         data2hash[[path, item[key]]] = item[ret]
       end
       KeystoneHelper.cache_update(data2hash) if my_item_id


chef/cookbooks/keystone/providers/register.rb
```diff
@@ -412,7 +412,10 @@ def find_id(item_name, path, dir, key = "name", ret = "id")
       data2hash = {}
 
       data.each do |item|
-        my_item_id = item[ret] if item[key] == item_name
+        if case_insensitive
+          my_item_id = item[ret] if item[key] == item_name
+        else
+          my_item_id = item[ret] if item[key].casecmp(item_name) == 0
```
cmurphy: Need an `end` here

cmurphy: Shouldn't these be swapped? If we want a case-sensitive comparison, use `==`, if we want case-insensitive, use `casecmp`

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -412,7 +412,10 @@ def find_id(item_name, path, dir, key = "name", ret = "id")
       data2hash = {}
 
       data.each do |item|
-        my_item_id = item[ret] if item[key] == item_name
+        if case_insensitive
```
cmurphy: Technically it's not just the role that would have this problem, everything that uses the mysql backend would have the same issue (i.e. we would get a conflict if we tried to recreate user `gyee` as `Gyee`). So could we just always use the case-insensitive check instead of `==`?

guangyee: Good point! Let me default it to true instead.

chef/cookbooks/keystone/providers/register.rb
guangyee: ah, you're right!

chef/cookbooks/keystone/providers/register.rb
cmurphy: We could just also remove the flag entirely, is there any reason to keep it?

guangyee: I was thinking to keep it for backward compatibility, just in case we need a case-sensitive comparison. But now that you asked, I can't think of any! Let me just remove it entirely to simplify things.

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -412,7 +412,12 @@ def find_id(item_name, path, dir, key = "name", ret = "id")
       data2hash = {}
 
       data.each do |item|
-        my_item_id = item[ret] if item[key] == item_name
+        # NOTE: for Keystone with MySQL backend, which is the default since
+        # Cloud 8, we should be doing case-insensitive comparison when lookup
+        # ID by name. For more information, see
+        # https://docs.openstack.org/keystone/rocky/admin
+        # /identity-case-insensitive.html
+        my_item_id = item[ret] if item[key].casecmp(item_name) == 0
```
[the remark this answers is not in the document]
jgrassler: I took the liberty of addressing this one.

chef/cookbooks/keystone/providers/register.rb
```diff
@@ -412,7 +412,12 @@ def find_id(item_name, path, dir, key = "name", ret = "id")
       data2hash = {}
 
       data.each do |item|
-        my_item_id = item[ret] if item[key] == item_name
+        # NOTE: for Keystone with MySQL backend, which is the default since
+        # Cloud 8, we should be doing case-insensitive comparison when lookup
+        # ID by name. For more information, see
+        # https://docs.openstack.org/keystone/rocky/admin
```
skazi0: IMO it's better to keep the URL on one line even if hound doesn't like it.

PR #2155, opened 2019-06-21
[5.0] Configurable timeout for Galera pre-sync

AbelNavarro:
Backport of #2150 

We have a pacemaker sync-mark to wait for all nodes to have all galera
packages and configuration files installed. The default timeout is 60
seconds for these sync-marks.

While installing all cluster members at once, like in an initial
installation, all nodes take about the same time to install the packages
and reach the sync-mark about the same time as well. But when adding a
new node to the cluster, the already installed nodes reach the sync-mark
much faster and the default timeout proves insuficient.

Based on tests in developer environments, this takes roughly 70 seconds.
We will be setting this to 180 seconds to give enough margin.

(cherry picked from commit 4f1d42ed8beb5e35b4428904adccde5cbd6ab570)


AbelNavarro:

PR #2159, opened 2019-07-01
monasca: move Grafana DB creation

jgrassler:
This commit moves Grafana DB creation to the Horizon barclamp
since this is where the Grafana service is being deployed.

### DIFF
--- a/chef/cookbooks/horizon/recipes/monasca_ui.rb
+++ b/chef/cookbooks/horizon/recipes/monasca_ui.rb
@@ -47,6 +47,39 @@ package "grafana" do
   action :install
 end
 
+grafana_db = node[:monasca][:db_grafana]
+
+database "create #{grafana_db[:database]} database" do
+  connection db_settings[:connection]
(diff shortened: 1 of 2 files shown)


jgrassler:
Note: this one is crucial for the upgrade since a controller only upgrade does not necessarily apply the Monasca barclamp (unless the monasca-server role is assigned to a controller). Consequently, the `grafana` database (which used to be in the MariaDB instance on the Monasca node) won't be created in the regular MariaDB instance, causing the chef run to fail.

toabctl:
> Note: this one is crucial for the upgrade since a controller only upgrade does not necessarily apply the Monasca barclamp (unless the monasca-server role is assigned to a controller). Consequently, the `grafana` database (which used to be in the MariaDB instance on the Monasca node) won't be created in the regular MariaDB instance, causing the chef run to fail.

@jgrassler can you add this note to the commit msg please?

jgrassler:
Ok, note added.

PR #2161, opened 2019-07-04
swift: Sync HA nodes (SOC-9683)

skazi0:
In most barclamps using syncmarks existing syncmarks are reset when applying
proposal. In Swift barclamp, swift-proxy is the only "HA" role and it is not
executed as the first one (swift-storage and swift-ring-compute need to be
run first). Because of that, first chef-client run during proposal commit
might not include the founder node and syncmark mechanism would fail in all
cookbooks.

Because syncmarks are not reset when applying Swift proposal, nodes can be
out of sync beyond normal syncmark timeout. This 'sync' syncmark should
line them up and correct any time offsets.


skazi0:
I got three green runs in a row with this patch (cloud8 version) in an env where I had these timeouts constantly.

skazi0:
backport to SOC7 is probably not needed and we don't want to risk breaking an old-and-stable release.

PR #2162, opened 2019-07-05
[4.0] Make ovs of_inactivity_probe configurable from neutron barclamp

zzaimeche:
This patch allows the user to change
the ovs inactivity_probe timeout from the neutron barclamp, in the 'raw'
view. Previously, this value was always set to the OVS default, 5.

It provides crowbar support for this upstream patch:

https://review.opendev.org/#/c/663024/

### DIFF
--- a/chef/data_bags/crowbar/template-neutron.json
+++ b/chef/data_bags/crowbar/template-neutron.json
@@ -43,7 +43,8 @@
       "ovs": {
         "tunnel_csum": false,
         "of_interface": "native",
-        "ovsdb_interface": "native"
+        "ovsdb_interface": "native",
+        "of_inactivity_probe": 10
       },
       "apic": {
         "hosts": "",
@@ -194,7 +195,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 123,
+      "schema-revision": 124,
       "element_states": {
         "neutron-server": [ "readying", "ready", "applying" ],
         "neutron-network": [ "readying", "ready", "applying" ],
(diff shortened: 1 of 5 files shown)


chef/data_bags/crowbar/migrate/neutron/307_add_ovs_of_inactivity_probe.rb
```diff
@@ -0,0 +1,15 @@
+def upgrade(tattr, tdep, attr, dep)
```
nicolasbock: The migration should be `124`.

chef/data_bags/crowbar/template-neutron.json
```diff
@@ -194,7 +195,7 @@
     "neutron": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 123,
+      "schema-revision": 307,
```
nicolasbock: Revision is `124`.

chef/data_bags/crowbar/migrate/neutron/307_add_ovs_of_inactivity_probe.rb
zzaimeche: Yeah, it's a backport to Newton. Newton is EoL upstream so there's no corresponding upstream patch for that branch; we have it in for Pike and up. I can remove the note about the conflicts.

chef/data_bags/crowbar/template-neutron.json
zzaimeche: Ok, thanks, I wasn't sure if 124 would conflict with a later patch in soc8 so I left it the same as the version on master, but I can change that.

nicolasbock: No, it shouldn't. We jump to the next hundred with each release, i.e. SOC 8 -> 200, SOC 9 -> 300.

zzaimeche: Cool, thanks for explaining. :)

nicolasbock: No problem :smile:

chef/data_bags/crowbar/migrate/neutron/307_add_ovs_of_inactivity_probe.rb
nicolasbock: You also need to rename the migration file :smile:

PR #2164, opened 2019-07-08
horizon: Install designate plugin when configured (SOC-9695)

dirkmueller:


dirkmueller:
@toabctl yes, since cloud8

PR #2166, opened 2019-07-10
[5.0] designate: allow worker on cluster (SOC-9632)

dirkmueller:
[1] suggests that workers are there for designate-central to get work
done. So having more of them is a good idea.
A worker does not really have an idea of api so no vip or haproxy
config is required.

1 - https://docs.openstack.org/designate/rocky/contributor/architecture.html

(cherry picked from commit a0a087728dbd15559852bc58cb6400ed23a43a95)


nicolasbock:
I restarted the CI job. It's pending right now.

PR #2170, opened 2019-07-16
designate: Update ns_records with all nameservers (SOC-9636)

ritesh216:


ritesh216:
Need tp update commit message.

nicolasbock:
You can force push so that you don't have to close a PR to update the commit message.

PR #2173, opened 2019-07-19
Tempest fixes for nova and magnum (SOC-9298)

flaviodsr:
Nova:
- disable `swap_volume` tests when cinder is using ceph backend as it is
not supported.
- replaced cirros uec used by tempest to qcow2 image

Fix for: `tempest.scenario.test_minimum_basic.TestMinimumBasicScenario.test_minimum_basic_scenario`

Magnum:
- magnum barclamp: set `verify_ca = false`. The magnum barclamp does not
support setting `openstack_ca_file` which leaves magnum unusable when
deploying crowbar using self-signed certificates. Setting `verify_ca =
false` ensures that it will work for any case.
- tempest barclamp: for magnum set `dns_nameserver` with the crowbar DNS
server; and add `member` role to the list of roles given to users created by tempest.

Fix for: `magnum_tempest_plugin.tests.api.v1.test_cluster.ClusterTest.test_create_list_sign_delete_clusters`

### DIFF
--- a/chef/cookbooks/tempest/templates/default/tempest.conf.erb
+++ b/chef/cookbooks/tempest/templates/default/tempest.conf.erb
@@ -13,9 +13,7 @@ admin_username = <%= @keystone_settings['admin_user'] %>
 admin_project_name = <%= @keystone_settings['default_tenant'] %>
 admin_password = <%= @keystone_settings['admin_password'] %>
 admin_domain_name = Default
-<%- unless @tempest_roles.empty? %>
 tempest_roles = <%= @tempest_roles %>
-<%- end %>
 <%- if @enabled_services.include? "baremetal" %>
 create_isolated_networks = False
 <%- end %>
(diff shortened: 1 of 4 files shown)


chef/cookbooks/tempest/templates/default/tempest.conf.erb
```diff
@@ -134,6 +132,7 @@ image_id = <%= @magnum_settings['image_id'] %>
 nic_id = floating
 flavor_id = <%= @magnum_settings['flavor_id'] %>
 master_flavor_id = <%= @magnum_settings['master_flavor_id'] %>
+dns_nameserver = <%= @crowbar_node_ip %>
```
rsalevsky: The Admin Node isn't always the DNS server. There is a DNS Barclamp which contains the DNS Servers.

`node_search_with_cache("roles:dns-server")` gives you all DNS Servers

flaviodsr: done! thanks!

PR #2174, opened 2019-07-22
Cleanup clone_stateless_services leftovers (SOC-9842)

skazi0:
Since clone_stateless_services attribute was dropped in
1cf82d3ff3142a700ac0e2c55c20944bf364e868 and implicitly set to false,
we don't need some parts of code which were active when it was set to
true.

### DIFF
--- a/crowbar_framework/lib/openstack/upgrade.rb
+++ b/crowbar_framework/lib/openstack/upgrade.rb
@@ -29,7 +29,11 @@ module Openstack
     # so the next time that Chef runs, will synchronize and update the
     # OpenStack service database
     def self.unset_db_synced
-      # we can't search by roles (like 'keystone-server') because at
+      # keystone is not included here because we keep it running for some longer
+      # during the upgrade and don't what the chef-recipes to trigger the database
+      # schema migrations.
+
+      # we can't search by roles (like 'neutron-server') because at
       # this point the nodes maybe don't have roles assigned anymore
       components = [
         :ceilometer, :cinder, :glance, :heat,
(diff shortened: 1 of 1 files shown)


crowbar_framework/lib/openstack/upgrade.rb
```diff
@@ -29,27 +29,22 @@ def self.post_upgrade
...
-          node["pacemaker"]["clone_stateless_services_orig"].nil? ||
-            node["pacemaker"]["clone_stateless_services_orig"]
-        else
-          true
-        end
 
         complete_components = components.clone
-        complete_components << "keystone" if clone_stateless
+        # run keystone db_sync only in non-ha scenarios
+        complete_components << "keystone" unless node["pacemaker"]
```
jsuchome: Ah, I think you'll need some better check for "non-ha" scenario, I think `node["pacemaker"]` could exist with some default values (?)

skazi0: This is how it was done before... and I think it would not hurt much if we never include keystone here.

rhafer: Yeah, I think it's best to always skip keystone here. We just have to make sure that the corresponding code (`keystone_migrate_and_restart`) in crowbar-core is called for non-ha setups. (AFAICS this is the case, but I might be missing something)

skazi0: AFAICS `keystone_migrate_and_restart` is currently only called from `upgrade_first_cluster_node`. I wanted to keep these changes as small as possible to avoid any unwanted side effects.

skazi0: @jsuchome you were right about this one, see: https://github.com/crowbar/crowbar-openstack/pull/2178

PR #2175, opened 2019-07-23
designate: Deploy producer on server nodes (SOC-9766)

s-t-e-v-e-n-k:
Designate has another service 'producer', which is used to run
long-running periodic tasks. Install the package and start the service
on nodes with the server role.


s-t-e-v-e-n-k:
This will require two packages to be added to Cloud (or at least I can't find them there), which are python2-pymemcache and python2-future.

chef/cookbooks/designate/templates/default/designate.conf.erb
```diff
@@ -48,6 +48,11 @@ memcache_secret_key = <%= node[:designate][:memcache_secret_key] %>
 memcache_pool_socket_timeout = 1
 <% end -%>
 
+<% if @with_authtoken and @ha_enabled -%>
+[coordination]
+backend_url = memcached://<%= @memcached_servers[0] %>/
```
dirkmueller: so why we're listing here only one out of all the memcached servers? what if that one is down?

I think a singleton service would be less painful..

chef/cookbooks/designate/recipes/api.rb
```diff
@@ -16,6 +16,8 @@
 # Recipe:: api
 #
 
+package "openstack-designate-producer"
```
dirkmueller: isn't that done by the designate_service resource already?

dirkmueller: Ah, I see, this is to install the package so that the singleton service (which is done after the sync mark?) will always be able to start. a comment would be helpful (of a thing like "if ha_enabled", needs to be moved a few lines down then)

PR #2177, opened 2019-07-25
Upgrade: restore Monasca DB [SOC-9772]

jgrassler:
This pull request will - once it's done - restore the Monasca and Grafana databases from the dump created by https://github.com/crowbar/crowbar-core/pull/1883 . This is necessary to move these databases from the MariaDB instance on the `monasca-server` node to the one on the controller node(s).

### DIFF
--- a/chef/cookbooks/horizon/recipes/monasca_ui.rb
+++ b/chef/cookbooks/horizon/recipes/monasca_ui.rb
@@ -20,6 +20,7 @@ if monasca_server.nil?
 end
 grafana_password = monasca_server[:monasca][:db_grafana][:password]
 db_settings = fetch_database_settings
+db_host = db_settings[:address]
 
 # Used for creating data source
 grafana_base_url = ::File.join(MonascaUiHelper.dashboard_local_url(node), "/grafana")
@@ -78,6 +79,33 @@ database_user "grant privileges to the #{grafana_db[:user]} database user" do
   require_ssl db_settings[:connection][:ssl][:enabled]
   action :grant
   only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
+# When upgrading from Cloud 8, restore the Monasca database dumped from the
+# Monasca node and stamp it with Alembic revision information. We stop
+# grafana-server as part of this to make sure there's no concurrent write
+# access to the DB.
+if node["crowbar_upgrade_step"] == "done_os_upgrade"
+  service "stop grafana-server for DB restore" do
+    service_name "grafana-server"
+    supports status: true, restart: true, start: true, stop: true
+    action [:disable, :stop]
+    not_if { File.exist?("/var/lib/crowbar/upgrade/grafana_db_restored") }
+  end
+
+  execute "restore Grafana DB" do
+    command  " /usr/bin/zcat /var/lib/crowbar/upgrade/monasca-grafana-database.dump.gz"\
+             " | /usr/bin/mysql"\
+             "     -h #{db_host}"\
+             "     -u #{grafana_db[:user]}"\
+             "   \"-p#{grafana_db[:password]}\""\
+             " #{grafana_db[:database]}"\
+             " && touch /var/lib/crowbar/upgrade/grafana_db_restored"
+    subscribes :run, resources(service: "stop grafana-server for DB restore"), :immediately
+    action :run
+    not_if { File.exist?("/var/lib/crowbar/upgrade/grafana_db_restored") }
+    only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
+  end
 end
 
 template "/etc/grafana/grafana.ini" do
--- a/chef/cookbooks/monasca/recipes/monasca_api.rb
+++ b/chef/cookbooks/monasca/recipes/monasca_api.rb
@@ -22,6 +22,13 @@ package "openstack-monasca-api"
 monasca_servers = search(:node, "roles:monasca-server")
 monasca_net_ip = MonascaHelper.get_host_for_monitoring_url(monasca_servers[0])
 
+db_settings = fetch_database_settings
+db_host = db_settings[:address]
+
+mondb_user = node[:monasca][:db_monapi][:user]
+mondb_password = node[:monasca][:db_monapi][:password]
+mondb_database = node[:monasca][:db_monapi][:database]
+
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
 memcached_instance("monasca") if node["roles"].include?("monasca-server")
@@ -45,6 +52,28 @@ template "/etc/monasca/api.conf" do
     sql_connection: sql_connection,
     tsdb: tsdb
   )
+end
+
+# When upgrading from Cloud 8, restore the Monasca database dumped from the
+# Monasca node and stamp it with Alembic revision information.
+if node["crowbar_upgrade_step"] == "done_os_upgrade"
+  db_backup_path = "/var/lib/crowbar/upgrade/monasca-mon-database.dump.gz"
+  execute "restore Monasca DB" do
+    command "/usr/bin/zcat #{db_backup_path}"\
+             " | /usr/bin/mysql -h #{db_host} -u #{mondb_user} \"-p#{mondb_password}\""\
+             " #{mondb_database}"
+    not_if "monasca_db version" # If DB is stamped with an Alembic version, this already happened.
+    action :run
+    only_if File.exist?(db_backup_path)
+    notifies :run, "execute[stamp Monasca DB]", :immediately
+  end
+
+  execute "stamp Monasca DB" do
+    command "/usr/bin/monasca_db stamp --from-fingerprint"
+    not_if "monasca_db version" # Not needed if the DB is stamped already.
+    action :run
+    notifies :run, "execute[apply mon database schema migration]", :immediately
+  end
 end
 
 execute "apply mon database schema migration" do
--- a/crowbar_framework/lib/openstack/upgrade.rb
+++ b/crowbar_framework/lib/openstack/upgrade.rb
@@ -37,7 +37,7 @@ module Openstack
       # this point the nodes maybe don't have roles assigned anymore
       components = [
         :ceilometer, :cinder, :glance, :heat,
-        :manila, :neutron, :nova
+        :manila, :neutron, :nova, :monasca
       ]
       NodeObject.all.each do |node|
         save_it = false
@@ -46,7 +46,7 @@ module Openstack
         # run keystone db_sync only in non-ha scenarios
         complete_components << "keystone" if node["keystone"] && !node["keystone"]["ha"]["enabled"]
         complete_components.each do |component|
-          [:db_synced, :api_db_synced].each do |flag|
+          [:db_synced, :api_db_synced, :db_monapi_synced].each do |flag|
             if node[component] && node[component][flag]
               node[component][flag] = false
               save_it = true


chef/cookbooks/monasca/recipes/monasca_api.rb
```diff
@@ -54,6 +57,25 @@
   )
 end
 
+# When upgrading from Cloud 8, restore the Monasca database dumped from the
+# Monasca node and stamp it with Alembic revision information.
+if node["crowbar_upgrade_step"] == "done_os_upgrade"
+  execute "restore Monasca DB" do
+    command "/usr/bin/zcat /var/lib/crowbar/upgrade/monasca-metrics-database.dump.gz"\
+             " | /usr/bin/mysql -u #{mondb_user} \"-p#{mondb_password}\""
```
skazi0: what about host?

chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -80,6 +80,23 @@
...
 
+# When upgrading from Cloud 8, restore the Monasca database dumped from the
+# Monasca node and stamp it with Alembic revision information. We stop
+# grafana-server as part of this to make sure there's no concurrent write
+# access to the DB.
+if node["crowbar_upgrade_step"] == "done_os_upgrade"
+  execute "restore Grafana DB" do
+    command  "/usr/bin/systemctl stop grafana-server; "\
+             " /usr/bin/zcat /var/lib/crowbar/upgrade/monasca-grafana-database.dump.gz"\
+             " | /usr/bin/mysql -u #{grafana_db[:user]} \"-p#{grafana_db[:password]}\""\
```
skazi0: If I get it right. We're trying to restore a database dump created on `database-server` to a node running horizon. These don't need to be the same in HA setup with multiple clusters (e.g. one for database, one for services).

skazi0: i.e. to follow the pattern from metrics db, you should dump to the node with `horizon-server` role.

chef/cookbooks/monasca/recipes/monasca_api.rb
jgrassler: Done now.

chef/cookbooks/horizon/recipes/monasca_ui.rb
jgrassler: Yeah, the node running Horizon is the target, so that role is indeed more appropriate. Fixed now (I adjusted https://github.com/crowbar/crowbar-core/pull/1883 accordingly as well - it dumps the Grafana database to the `horizon-server` node now).

chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -80,6 +81,24 @@
   only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
+# When upgrading from Cloud 8, restore the Monasca database dumped from the
+# Monasca node and stamp it with Alembic revision information. We stop
+# grafana-server as part of this to make sure there's no concurrent write
+# access to the DB.
+if node["crowbar_upgrade_step"] == "done_os_upgrade"
+  execute "restore Grafana DB" do
+    command  "/usr/bin/systemctl stop grafana-server; "\
```
skazi0: can we do this stoping using standard chef resource?

chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -80,6 +81,24 @@
...
+             " /usr/bin/zcat /var/lib/crowbar/upgrade/monasca-grafana-database.dump.gz"\
+             " | /usr/bin/mysql"\
+             "     -h #{db_host}"\
+             "     -u #{grafana_db[:user]}"\
+             "   \"-p#{grafana_db[:password]}\""\
```
skazi0: The "restore" `mysql` calls miss the database parameter. The dumps don't contain the database create/use lines. The database name could be added on restore or `mysqldump` could generate `DROP/CREATE DATABASE`  statements but these would probably require more privs when restoring.

chef/cookbooks/monasca/recipes/monasca_api.rb
```diff
@@ -54,6 +60,25 @@
...
 
+# When upgrading from Cloud 8, restore the Monasca database dumped from the
+# Monasca node and stamp it with Alembic revision information.
+if node["crowbar_upgrade_step"] == "done_os_upgrade"
+  execute "restore Monasca DB" do
+    command "/usr/bin/zcat /var/lib/crowbar/upgrade/monasca-metrics-database.dump.gz"\
+             " | /usr/bin/mysql -h #{db_host} -u #{mondb_user} \"-p#{mondb_password}\""
+    not_if "monasca_db version" # If DB is stamped with an Alembic version, this already happened.
+    action :run
+    notifies :run, "execute[stamp monasca DB]", :immediately
```
skazi0: resource name typo `s/monasca/Monasca/`

chef/cookbooks/monasca/recipes/monasca_api.rb
```diff
@@ -54,6 +60,25 @@
...
+  end
+
+  execute "stamp Monasca DB" do
+    command "/usr/bin/monasca_db stamp"
```
skazi0: When I ran it like this it said that I need to pass some revision or use `--from-fingerprint`. When I tried with the latter I got:
`Schema fingerprint da39a3ee5e6b4b0d3255bfef95601890afd80709 does not match any known legacy revision.`
The docs say that this must be some non-standard schema https://docs.openstack.org/monasca-api/latest/admin/index.html

chef/cookbooks/monasca/recipes/monasca_api.rb
```diff
@@ -54,6 +60,25 @@
   )
 end
 
+# When upgrading from Cloud 8, restore the Monasca database dumped from the
+# Monasca node and stamp it with Alembic revision information.
+if node["crowbar_upgrade_step"] == "done_os_upgrade"
+  execute "restore Monasca DB" do
+    command "/usr/bin/zcat /var/lib/crowbar/upgrade/monasca-metrics-database.dump.gz"\
```
skazi0: the dump name is `monasca-mon-database.dump.gz`

chef/cookbooks/horizon/recipes/monasca_ui.rb
jgrassler: Depends on whether chef defers service resources or not (I didn't check and erred on the side of caution right away). I'll give it a try. I need to modify that part of the code anyway - I just thought of a scenario where it might cause problems for HA clouds.

chef/cookbooks/horizon/recipes/monasca_ui.rb
jgrassler: Fixed now.

chef/cookbooks/monasca/recipes/monasca_api.rb
jgrassler: Fixed now.

chef/cookbooks/monasca/recipes/monasca_api.rb
jgrassler: I have it on rather good authority that your edited comment is accurate: I wrote the monasca_db command and still forgot about that bit myself... :-)

Fixed this now. Fingers crossed that we're not producing a schema fingerprint that does not occur in https://github.com/openstack/monasca-api/blob/master/monasca_api/db/fingerprint.py#L27 anyway. For that fingerprinting hack is rather brittle and minor MariaDB updates with different defaults for column data types will break them.

chef/cookbooks/monasca/recipes/monasca_api.rb
jgrassler: Fixed now.

chef/cookbooks/monasca/recipes/monasca_api.rb
```diff
@@ -54,6 +60,26 @@
   )
 end
 
+# When upgrading from Cloud 8, restore the Monasca database dumped from the
+# Monasca node and stamp it with Alembic revision information.
+if node["crowbar_upgrade_step"] == "done_os_upgrade"
+  execute "restore Monasca DB" do
+    command "/usr/bin/zcat /var/lib/crowbar/upgrade/monasca-mon-database.dump.gz"\
+             " | /usr/bin/mysql -h #{db_host} -u #{mondb_user} \"-p#{mondb_password}\""\
+             " #{node["monasca"]["db_monapi"]["database"]}"
```
skazi0: I'd shorten it the way you did with `mondb_user` and `mondb_password` to keep things consistent.

jgrassler: Done.

chef/cookbooks/monasca/recipes/monasca_api.rb
```diff
@@ -22,6 +22,13 @@
 monasca_servers = search(:node, "roles:monasca-server")
 monasca_net_ip = MonascaHelper.get_host_for_monitoring_url(monasca_servers[0])
 
+db_settings = fetch_database_settings
+db_host = db_settings[:address]
+
+mondb_user = node[:monasca][:db_monapi][:user]
+mondb_password = node[:monasca][:db_monapi][:password]
+mondb_database = node["monasca"]["db_monapi"]["database"]
```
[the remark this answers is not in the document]
skazi0: 1. why strings not symbols?
2. what is the deal with this hound complaint?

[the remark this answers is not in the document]
jgrassler: The useless assignment came about due to me forgetting the string interpolation curly braces further down (fixed by force push 30s later).

String rather than symbol...copied from an existing line. Since Crowbar is occasionally funny about string-vs-symbol (i.e. sometimes only one of these two works) I left it that way.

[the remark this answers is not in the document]
skazi0: Yeah, there is some ruby magic in the `node` which makes both (mostly) work. My question came from the two lines above which use the same object but use symbols.

chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -80,6 +81,33 @@
...
+             " #{grafana_db[:database]}"\
+             " && touch /var/lib/crowbar/upgrade/grafana_db_restored"
+    subscribes :run, resources(service: "stop grafana-server for DB restore"), :immediately
+    action :run
+    not_if { File.exist?("/var/lib/crowbar/upgrade/grafana_db_restored") }
+    only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
dirkmueller: also testing for the dump actually being there already?

jgrassler: test for dump being there added

chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -80,6 +81,33 @@
...
+    service_name "grafana-server"
+    supports status: true, restart: true, start: true, stop: true
+    action [:disable, :stop]
+    not_if { File.exist?("/var/lib/crowbar/upgrade/grafana_db_restored") }
+  end
+
+  execute "restore Grafana DB" do
+    command  " /usr/bin/zcat /var/lib/crowbar/upgrade/monasca-grafana-database.dump.gz"\
+             " | /usr/bin/mysql"\
+             "     -h #{db_host}"\
```
dirkmueller: a little bit of quotes here wouldn't hurt..

chef/cookbooks/horizon/recipes/monasca_ui.rb
dirkmueller: is the quoting really correct here? it looks weird to have "-pFOO" as a cmd line parameter

chef/cookbooks/horizon/recipes/monasca_ui.rb
jgrassler: Quotes? For a host name? Do you insist on that? This will

1) Invalidate all existing reviews.
2) Invalidate all existing CI runs.
3) Require another 3 hours of manually test upgrading a cloud. And yes, that is needed. While the chances of a typo may be slim, it can occur, thus ruining tested, working code for the sake of neatness.

If you do insist please state so. If you do not insist, please state so as well. Nobody knows whether this is a hard veto or an offhand comment from you. Until that is known, other reviewers will hesitate to deliver any sort of verdict at all, causing this pull request to remain open even longer.

dirkmueller: I am not insisting otherwise I would have set a "request changes".. :-)

PR #2178, opened 2019-07-26
upgrade: Fix HA detection for keystone db_sync (SOC-9981)

skazi0:
Detection of HA scenario for keystone db_sync during upgrade used
incorrect logic relying on pacemaker attribute which is present also
in non-HA deployment.

### DIFF
--- a/crowbar_framework/lib/openstack/upgrade.rb
+++ b/crowbar_framework/lib/openstack/upgrade.rb
@@ -44,7 +44,7 @@ module Openstack
 
         complete_components = components.clone
         # run keystone db_sync only in non-ha scenarios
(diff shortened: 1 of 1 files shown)


jsuchome:
Didn't @rhafer suggest to get rid of that check completely here? But i guess this needs backports...

skazi0:
@jsuchome to remove this I would need to find a proper place to call the migrations in non-ha upgrade case as they don't seem to be started by the upgrade scripts now.

skazi0:
@jsuchome the condition was incorrect (triggering in ha instead of non-ha). please re-check.

skazi0:
I'm testing it in my env and I'm already on second node without a failure.

PR #2179, opened 2019-07-26
[4.0] magnum: retry flavor creation (SOC-9991)

stefannica:
Creating magnum flavors is done as a delayed action, alongside
all other delayed actions, such as restarting services.
If one of those restarted services is apache (e.g. because it
was triggered by another barclamp configuration change), then
keystone and other API services might not be available right
away, in which case the magnum flavor creation will fail.

Re-attempting magnum flavor creation fixes this issue.


stefannica:
Also testing with: https://ci.suse.de/blue/organizations/jenkins/cloud-maintenance-gating/detail/cloud-maintenance-gating/55/pipeline/

PR #2183, opened 2019-08-02
[5.0] horizon: fix Grafana in HA clouds (bsc#1116846)

skazi0:
In HA clouds, Grafana is started on all controllers. Upon first startup,
Grafana will run database migrations. If this happens concurrently on more than
one node, Grafana's database schema may end up in an inconsistent state. This
pull request ensures Grafana starts on the cluster founder first so the
migrations only happen there.

(cherry picked from commit 2062cc8d4626f20476fd727e019bca3ce2574e34)

port of #1988

### DIFF
--- a/chef/cookbooks/horizon/recipes/monasca_ui.rb
+++ b/chef/cookbooks/horizon/recipes/monasca_ui.rb
@@ -60,10 +60,27 @@ template "/etc/grafana/grafana.ini" do
   mode "0640"
 end
 
+crowbar_pacemaker_sync_mark "wait-grafana_migrations" if ha_enabled
+
+# Start Grafana server on cluster founder first to ensure database migrations
+# happen there...
+
 service "grafana-server" do
   supports status: true, restart: true, start: true, stop: true
   action [:enable, :start]
   subscribes :restart, resources(template: "/etc/grafana/grafana.ini")
+  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
+end
+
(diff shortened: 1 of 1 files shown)


chef/cookbooks/horizon/recipes/monasca_ui.rb
```diff
@@ -60,10 +60,27 @@
...
+
+crowbar_pacemaker_sync_mark "create-grafana_migrations" if ha_enabled
+
+# ...then start Grafana server on all other nodes.
+
 service "grafana-server" do
   supports status: true, restart: true, start: true, stop: true
   action [:enable, :start]
   subscribes :restart, resources(template: "/etc/grafana/grafana.ini")
+  only_if { ha_enabled && !CrowbarPacemakerHelper.is_cluster_founder?(node) }
```
rtamalin: Is this only_if correct? It would appear to only restart grafana if ha is enabled, and this is not the found node... But will that restart at all in a non-HA case?

Should it possibly be

    only_if !ha_enabled or !CrowbarPacemakerHelper.is_cluster_founder?(node)

skazi0: Yeah, this is one tricky case... not the usual "run anywhere for non-HA or on founder for HA" condition. That one is on line 72 :o)
The logic is such that in non-HA we use only one of these `service` calls but on HA case we use both and they are ordered by sync-marks so that the founder's service is started first (and runs the migrations).

PR #2188, opened 2019-08-05
support OpenID Connect WebSSO (SOC-9616)

guangyee:
Add support for OpenID Connect WebSSO. Configure Keystone and Horizon
to support OpenID Connect WebSSO per

https://docs.openstack.org/keystone/latest/admin/federation/configure_federation.html#setting-up-openid-connect

The creation of identity provider, protocol, and mapping will be done
out-of-band using OpenStack CLI as customers are likely to experiment
with different mappings. It is easier for them to use OpenStack CLI to make
those changes as oppose to having to re-run the Keystone proposal.

Initial testing will be done against Google OpenID Connection provider
only.


chef/cookbooks/horizon/templates/default/local_settings.py.erb
```diff
@@ -250,3 +250,20 @@ EXTERNAL_MONITORING = [
...
+WEBSSO_CHOICES = (
+    ("credentials", _("Keystone Credentials")),
+{% if (enable_oidc | bool) %}
+    ("openid", _("OpenID Connect")),
+{% endif %}
+)
+<% end %>
+
+<% if !node[:keystone][:ssl][:insecure] %>
+OPENSTACK_SSL_CACERT = <%=  node[:keystone][:ssl][:ca_certs]%>
```
cmurphy: I'm pretty sure both `OPENSTACK_SSL_CACERT` and the `ca_certs` in the keystone barclamp are actually about verifying incoming client connections, whereas the `insecure` option in the keystone barclamp is about whether the client should bother verifying the server (i.e. whether or not it's self-signed), so these really have nothing to do with each other.

chef/cookbooks/crowbar-openstack/templates/default/vhost-wsgi.conf.erb
```diff
@@ -1,5 +1,36 @@
...
+OIDCClaimPrefix <%= node[:keystone][:federation][:openidc][:claim_prefix] %>
+OIDCResponseType <%= node[:keystone][:federation][:openidc][:response_type] %>
+OIDCScope <%= node[:keystone][:federation][:openidc][:scope] %>
+OIDCProviderMetadataURL <%= node[:keystone][:federation][:openidc][:metadata_url] %>
+OIDCClientID  <%= node[:keystone][:federation][:openidc][:client_id] %>
+OIDCClientSecret <%= node[:keystone][:federation][:openidc][:client_secret] %>
+OIDCCryptoPassphrase <%= node[:keystone][:federation][:openidc][:passphrase] %>
+OIDCRedirectURI <%= node[:keystone][:federation][:openidc][:redirect_url] %>
+
+<Location /v3/OS-FEDERATION/identity_providers/<%= node[:keystone][:federation][:openidc][:identity_provider %>/protocols/openid/auth>
```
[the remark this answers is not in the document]
cmurphy: Not sure whether or how we should handle this but `mapped` is also a valid name for the protocol in this case

chef/cookbooks/crowbar-openstack/templates/default/vhost-wsgi.conf.erb
```diff
@@ -1,5 +1,36 @@
 Listen <%= @bind_host %>:<%= @bind_port %>
 
+<% if node[:keystone][:federation][:openidc][:enabled] %>
+# See
+# https://github.com/zmartzone/mod_auth_openidc/blob/master/auth_openidc.conf
+# FIXME(gyee): do we need to make the prefix configurable or just doc it as is?
```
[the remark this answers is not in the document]
guangyee: Sounds good. Let's make it immutable for now.

chef/cookbooks/crowbar-openstack/templates/default/vhost-wsgi.conf.erb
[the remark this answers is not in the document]
guangyee: The 'openid' is just an alias to 'mapped' right? https://github.com/openstack/keystone/blob/master/setup.cfg#L93
Do we anticipate supporting a custom auth plugin?

chef/cookbooks/crowbar-openstack/templates/default/vhost-wsgi.conf.erb
```diff
@@ -1,5 +1,36 @@
...
+    LogLevel debug
+</Location>
+<Location /v3/auth/OS-FEDERATION/identity_providers/<%= node[:keystone][:federation][:openidc][:identity_provider] %>/protocols/openid/websso>
+    Require valid-user
+    AuthType openid-connect
+    LogLevel debug
+</Location>
+<% end %>
+
 <VirtualHost <%= @bind_host %>:<%= @bind_port %>>
```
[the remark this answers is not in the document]
guangyee: Done

chef/cookbooks/crowbar-openstack/templates/default/vhost-wsgi.conf.erb
```diff
@@ -1,5 +1,36 @@
...
+OIDCProviderMetadataURL <%= node[:keystone][:federation][:openidc][:metadata_url] %>
+OIDCClientID  <%= node[:keystone][:federation][:openidc][:client_id] %>
+OIDCClientSecret <%= node[:keystone][:federation][:openidc][:client_secret] %>
+OIDCCryptoPassphrase <%= node[:keystone][:federation][:openidc][:passphrase] %>
+OIDCRedirectURI <%= node[:keystone][:federation][:openidc][:redirect_url] %>
+
+<Location /v3/OS-FEDERATION/identity_providers/<%= node[:keystone][:federation][:openidc][:identity_provider %>/protocols/openid/auth>
+    Require valid-user
+    AuthType openid-connect
+    LogLevel debug
```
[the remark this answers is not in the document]
guangyee: Done

chef/cookbooks/horizon/templates/default/local_settings.py.erb
```diff
@@ -250,3 +250,20 @@ EXTERNAL_MONITORING = [
...
 SECRET_KEY = "<%= @secret_key %>"
+
+# FIXME(gyee): if we ever going to support multiple protocols or multiple
+# IDPs, we'll need to refactor this accordingly.
+<% if node[:keystone][:federation][:openidc][:enabled] %>
+WEBSSO_ENABLED = True
+
+WEBSSO_CHOICES = (
+    ("credentials", _("Keystone Credentials")),
+{% if (enable_oidc | bool) %}
```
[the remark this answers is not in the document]
guangyee: Copy-n-paste mistake. I was copying this straight from my devstack env.

chef/cookbooks/horizon/templates/default/local_settings.py.erb
guangyee: In this case, we should be basing on node[:keystone][:ssl][:generate_certs] then? We only need this to verify self-signed certs. For certs signed by commercial CA, the signer cert should've already be in the system trust list.

chef/cookbooks/crowbar-openstack/templates/default/vhost-wsgi.conf.erb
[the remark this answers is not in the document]
cmurphy: [**@guangyee**](https://github.com/guangyee) commented on [Aug 6, 2019, 12:07 PM PDT](https://github.com/crowbar/crowbar-openstack/pull/2188#discussion_r311228252 "2019-08-06T19:07:59Z - Replied by Github Reply Comments"):
> The 'openid' is just an alias to 'mapped' right? https://github.com/openstack/keystone/blob/master/setup.cfg#L93

Yes

> Do we anticipate supporting a custom auth plugin?

Oh god no

chef/cookbooks/horizon/templates/default/local_settings.py.erb
cmurphy: I'm not sure it's needed at all, I don't think client cert verification is relevant to this feature.

guangyee: Sounds good. I'll back out this part.

chef/data_bags/crowbar/migrate/keystone/304_add_federation_openidc.rb
```diff
@@ -0,0 +1,15 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs.key? "federation"
+    if attrs["federation"]["openidc"]["passphrase"].nil? ||
+       attrs["federation"]["openidc"]["passphrase"].empty?
+      service = ServiceObject.new "fake-logger"
+      attrs["federation"]["openidc"]["passphrase"] = service.random_password
+    end
+  end
```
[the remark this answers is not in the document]
guangyee: Done

chef/data_bags/crowbar/template-keystone.json
```diff
@@ -177,6 +177,20 @@
...
+          "enabled": false,
+          "identity_provider": "",
+          "claim_prefix": "OIDC-",
+          "response_type": "id_token",
+          "scope": "openid email profile",
+          "metadata_url": "",
+          "client_id": "",
+          "client_secret": "",
+          "passphrase": "",
+          "redirect_url": ""
```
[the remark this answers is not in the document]
guangyee: Done

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -31,6 +31,10 @@
 # useful with .openrc
 package "python-openstackclient"
 
+# FIXME(gyee): need a more elegant way to update the default auth methods
+# if these happened to change from one release to another.
+auth_methods = "external,password,token,oauth1,mapped,application_credential"
```
cmurphy: external should be removed due to the reason here https://docs.openstack.org/keystone/latest/admin/external-authentication.html#configuration (and also for the reasons we discussed in the keystone meetings)

guangyee: good point! Let me remove that.

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -32,6 +32,64 @@ def self.unversioned_internal_auth_url(node, admin_host)
...
+    if node.key?("horizon") && node["horizon"].key?("apache")
+      horizon_url = ::File.join(dashboard_public_url(node), "/auth/websso")
+    else
+      horizon_url = ""
+    end
+
+    horizon_url
+  end
+
+  def self.trusted_dashboards(node):
```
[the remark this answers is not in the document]
guangyee: doh! Good that we catch this one before the next round of testing.

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -32,6 +32,64 @@ def self.unversioned_internal_auth_url(node, admin_host)
...
+
+  def self.trusted_dashboard_url(node)
+    # NOTE(gyee): since Horizon is depended on Keystone and will be deployed
+    # after Keystone, the Horizon node may not be available when Keystone
+    # is first deployed. However, chef executes the recipes periodically.
+    # On the next chef run after, after Horizon had deployed, we should be
+    # able to figure out the trusted_dashboard from the node, assuming Horizon
+    # is always deployed on the same node as Keystone. Otherwise, we'll need
+    # to do node search.
+    if node.key?("horizon") && node["horizon"].key?("apache")
```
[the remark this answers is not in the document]
guangyee: WTF you mean?

chef/data_bags/crowbar/migrate/keystone/304_add_federation_openidc.rb
```diff
@@ -0,0 +1,14 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs.key? "federation" && attrs["federation"].key? "openidc"
+    if attrs["federation"]["openidc"]["passphrase"].nil? || attrs["federation"]["openidc"]["passphrase"].empty?
+      service = ServiceObject.new "fake-logger"
+      attrs["federation"]["openidc"]["passphrase"] = service.random_password
+    end
+  end
+  return attrs, deployment
+end
```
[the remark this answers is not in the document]
guangyee: ???

chef/cookbooks/keystone/libraries/helpers.rb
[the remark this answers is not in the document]
cmurphy: I think it wants you to do 
```
horizon_url =
  if node.key?("horizon") && node["horizon"].key?("apache")
    ::File.join(dashboard_public_url(node), "/auth/websso") 
  else
    ""
  end
```
because somehow that's better :roll_eyes:

chef/data_bags/crowbar/migrate/keystone/304_add_federation_openidc.rb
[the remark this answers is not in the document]
cmurphy: hound is drunk

chef/cookbooks/keystone/libraries/helpers.rb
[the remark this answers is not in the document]
guangyee: Yuck! What the hell, let me just change it to hound's liking.

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -249,6 +306,15 @@
 
 profiler_settings = KeystoneHelper.profiler_settings(node, @cookbook_name)
 
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+
+# make sure remote_id_attribute is specified
+if keystone_settings["websso_enabled"]
+    raise "remote_id_attribute is required and cannot be an empty value" if
```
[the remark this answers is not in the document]
guangyee: hound is jerking me around!

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -249,6 +306,13 @@
 
 profiler_settings = KeystoneHelper.profiler_settings(node, @cookbook_name)
 
+keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+
+# make sure remote_id_attribute is specified
```
cmurphy: I would just hardcode this, since we know we're using openidc and we know the prefix then this should always be `HTTP_OIDC_ISS`

chef/data_bags/crowbar/template-keystone.json
```diff
@@ -177,14 +177,30 @@
         "hmac_keys": [],
         "trace_sqlalchemy": false,
         "connection_string": "messaging://"
+      },
+      "federation": {
+        "trusted_dashboards": [],
+        "remote_id_attribute": "",
```
cmurphy: I would just hardcode this in the config template

chef/data_bags/crowbar/template-keystone.json
```diff
@@ -177,14 +177,30 @@
...
+        "openidc": {
+          "enabled": false,
+          "identity_provider": "",
+          "response_type": "id_token",
+          "scope": "openid email profile",
+          "metadata_url": "",
+          "client_id": "",
+          "client_secret": "",
+          "passphrase": "",
+          "redirect_url": "",
```
cmurphy: The apache config setting is `OIDCRedirectURI` so I would use `redirect_uri` not `redirect_url`

chef/data_bags/crowbar/template-keystone.json
```diff
@@ -177,14 +177,30 @@
...
+          "remote_id_attribute": ""
```
cmurphy: Delete this too

chef/data_bags/crowbar/template-keystone.json
cmurphy: Should also be able make a reasonable guess at a default value (in the chef recipe, not here) same as with the trusted_dashboards

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -392,6 +392,27 @@
...
+# to be accessible by the browsers. Also, in some cases, the URL needs to be
+# fully qualified. For example, Google OpenID Connection requires the URL to
+# be FQDN instead of IP as it must match the authorized domain. Therefore,
+# the WEBSSO_KEYSTONE_URL in Horizon should contain the FQDN, depending
+# on the identity provider. For production deployments, we offers the user
+# the ability to override it because we don't know the user's network
+# topology beforehand. For example, the external endpoint could be handled
+# by an HTTP proxy which may not be known to crowbar.
+websso_keystone_url = keystone_settings["public_fqdn_auth_url"]
+unless node[:horizon][:websso_keystone_url].empty?
```
cmurphy: Getting `undefined method 'empty?' for nil:NilClass` here.

chef/cookbooks/keystone/templates/default/keystone.conf.erb
```diff
@@ -114,4 +114,22 @@ enabled = true
...
\ No newline at end of file
+<% end -%>
+
+<% if @websso_enabled %>
+[federation]
+<% @trusted_dashboards.each do |trusted_dashboard_url| -%>
+trusted_dashboard = <%= trusted_dashboard_url %>
+<% end -%>
+<% if not @federation_remote_id_attribute.empty? -%>
+remote_id_attribute = <%= @federation_remote_id_attribute %>
```
cmurphy: This isn't needed, the one in `[openid]` will override it

chef/data_bags/crowbar/template-horizon.json
```diff
@@ -45,7 +45,8 @@
         "apache2": {
           "LimitNOFILE": null
         }
-      }
+      },
+      "websso_keystone_url": ""
```
cmurphy: The schema revision number needs to be bumped here too.

But - maybe it would make more sense to just put this in the keystone barclamp? Then the user doesn't have to reapply two different barclamps when they want to set up federation for an existing cloud.

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -32,6 +32,79 @@ def self.unversioned_internal_auth_url(node, admin_host)
...
+      cluster_vhostname = CrowbarPacemakerHelper.cluster_vhostname(node)
+      host_fqdn = "#{cluster_vhostname}.#{node[:domain]}"
+    else
+      host_fqdn = node[:fqdn]
+    end
+
+    protocol = "http"
+    protocol = "https" if ssl_enabled
+
+    "#{protocol}://#{host_fqdn}"
```
cmurphy: This isn't right, there needs to be a `/auth/websso/` appended to it. Even if the user sets it in the proposal, we should be helpful and add the path because there's no way they will know to add it until they run into the auth failure.

chef/data_bags/crowbar/template-keystone.json
```diff
@@ -177,14 +177,30 @@
         "hmac_keys": [],
         "trace_sqlalchemy": false,
         "connection_string": "messaging://"
+      },
+      "federation": {
+        "trusted_dashboards": [],
+        "remote_id_attribute": "",
+        "openidc": {
+          "enabled": false,
+          "identity_provider": "",
```
cmurphy: I still think the recipe should create this and the protocol and the mapping. If the user already has to think about the name of the identity provider here, then it doesn't make sense to have them apply the proposal and then have to remember it again when they go and manually create the idp in keystone. I expect things to just work after I've enabled federation in crowbar.

chef/cookbooks/keystone/libraries/helpers.rb
guangyee: This is just the partial url, the websso one is specified in line 89 below.

chef/cookbooks/horizon/recipes/server.rb
guangyee: This is weird. websso_keystone_url is defined in the schema.

chef/cookbooks/keystone/recipes/server.rb
guangyee: Sounds good. I think ISS is a required attribute so we should be OK hardcoding it.

chef/cookbooks/keystone/templates/default/keystone.conf.erb
guangyee: Done.

chef/data_bags/crowbar/template-horizon.json
guangyee: Sounds good. Let me move it over to Keystone barclamp.

chef/data_bags/crowbar/template-keystone.json
guangyee: Done

chef/data_bags/crowbar/template-keystone.json
guangyee: The problem is we don't know the mapping in advance. For example, does user want local or ephemeral users? Any group to map to? If so, do we need to create those groups and assignments? Are there any other attributes in the id_token that they want to utilize? We can make some assumptions for a know provider like Google. But even that is likely to change in a production environment.

chef/data_bags/crowbar/template-keystone.json
guangyee: Done. I'll automagically populate redirect_uri if its not specified.

chef/data_bags/crowbar/template-keystone.json
guangyee: Done

chef/data_bags/crowbar/template-keystone.json
cmurphy: I still think this could be done programmatically. For group membership mapping, we could have another proposal attribute that is a dict that maps groups and roles to projects, and have chef check if they exist and create them if not. Same with local users, we could either parse the mapping to look for keyword `"local"` or have a separate attribute with a list of local users to create. But I'll concede that that is a bit of work and for a first pass this is probably fine.

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -32,6 +32,105 @@ def self.unversioned_internal_auth_url(node, admin_host)
     service_URL(node[:keystone][:api][:protocol], admin_host, node[:keystone][:api][:service_port])
   end
 
+  def self.keystone_public_fqdn(node)
+    ha_enabled = node[:keystone][:ha][:enabled]
+    ssl_enabled = node["keystone"]["api"]["protocol"] == "https"
+
+    if ha_enabled
```
rhafer: I think you'd need to first check if the pacemaker configuration is using the "Public name for public virtual IP" setting here. And if that is present use that as the public fqdn. You should be able to get that by calling  `CrowbarPacemakerHelper.cluster_haproxy_vpublic_name`

Only if that is empty you can derive the public fqdn from the internal vhostname. But AFAIK there "real" host_fqdn looks something like `public-<#{cluster_vhostname}.#{node[:domain]}>`. (See: http://docserv.suse.de/documents/SUSE_OpenStack_Cloud_9/suse-openstack-cloud-crowbar-deployment/single-html/#ann-depl-ostack-pacemaker-prop-name)

Ideally you'd just use `CrowbarHelper.get_host_for_public_url` (https://github.com/crowbar/crowbar-core/blob/master/chef/cookbooks/utils/libraries/helpers.rb#L36). But under certain cirumstances (non-SSL setup and no "Public name for public virtual IP" set) it that one will just return IP addresses. It might be worth to refactor that code a bit to be able to request the return of the fqdn in those cases as well.

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -32,6 +32,105 @@ def self.unversioned_internal_auth_url(node, admin_host)
...
+  # from the one that is outside of the firewall. And in some cases, the
+  # endpoint could be the corporate HTTP proxy.
+  # For now, we are prepopulating one that is only understood by Crowbar for
+  # testing/demo purpopses. In a production environment, it can be amended with
+  # other external endpoints.
+  def self.dashboard_public_url(node)
+    ha_enabled = node[:horizon][:ha][:enabled]
+    ssl_enabled = node[:horizon][:apache][:ssl]
+
+    if ha_enabled
```
rhafer: I think here my comment for `keystone_public_fqdn` applies as well.

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -32,6 +32,105 @@ def self.unversioned_internal_auth_url(node, admin_host)
...
+    node[:keystone][:federation][:openidc][:enabled]
+  end
+
+  def self.trusted_dashboard_url(node)
+    # NOTE(gyee): since Horizon is depended on Keystone and will be deployed
+    # after Keystone, the Horizon node may not be available when Keystone
+    # is first deployed. However, chef executes the recipes periodically.
+    # On the next chef run after, after Horizon had deployed, we should be
+    # able to figure out the trusted_dashboard from the node, assuming Horizon
+    # is always deployed on the same node as Keystone. Otherwise, we'll need
```
rhafer: Crowbar makes no guarantees that horizon and keystone are always deployed to the same set of nodes. I think you need to do that node search to be really correct. (You should be able to use `node_search_with_cache` for that.)

And of course you need to be prepared for that search returning nothing until horizon is deployed.

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -32,6 +32,105 @@ def self.unversioned_internal_auth_url(node, admin_host)
...
+  # to Keystone upon successfully authentication with the identity provider.
+  # Therefore, the Keystone redirect URL must be external or public as it needs
+  # to be accessible by the browsers. Also, in some cases, the URL needs to be
+  # fully qualified. For example, Google OpenID Connection requires the URL to
+  # be FQDN instead of IP as it must match the authorized domain. Therefore,
+  # the WEBSSO_KEYSTONE_URL in Horizon should contain the FQDN, depending
+  # on the identity provider. For production deployments, we offers the user
+  # the ability to override it because we don't know the user's network
+  # topology beforehand. For example, the external endpoint could be handled
+  # by an HTTP proxy which may not be known to crowbar.
```
rhafer: I wonder if `websso_keystone_url` is really needed since we already offer the ability to set custom external public names via the cluster configuration in the pacemaker barclamp, and for each individual node via the `public_name` attribute. But I might be missing something.

chef/data_bags/crowbar/migrate/keystone/304_add_federation_openidc.rb
```diff
@@ -0,0 +1,16 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs.key? "federation"
+    attrs["federation"] = template_attrs["federation"]
+    if attrs["federation"]["openidc"]["passphrase"].nil? ||
+        attrs["federation"]["openidc"]["passphrase"].empty?
+      service = ServiceObject.new "fake-logger"
+      attrs["federation"]["openidc"]["passphrase"] = service.random_password
```
cmurphy: Saw a strange issue where this passphrase wasn't populated if I installed directly from the update repo instead of running the barclamp reinstall after crowbar was already installed. Not sure if it was just a fluke or if there's an issue, will try to look more tomorrow.

chef/cookbooks/keystone/libraries/helpers.rb
guangyee: I'll pull in https://github.com/crowbar/crowbar-core/pull/1901 once it is landed.

chef/cookbooks/keystone/libraries/helpers.rb
guangyee: Done

chef/cookbooks/keystone/libraries/helpers.rb
guangyee: websso_keystone_url is for Horizon to redirect the browsers back to Keystone's public endpoint. And unlike the keystone_auth_url, it must be versioned.

chef/data_bags/crowbar/migrate/keystone/304_add_federation_openidc.rb
guangyee: I ran into this issue if do "mkcloud instonly", then do the proposal one by one.

chef/data_bags/crowbar/template-keystone.json
guangyee: My gut feeling is that mapping is something that most people will get it wrong the first time. And it will take some iterations to sort out what they really want. In that case, it's probably easier to ask them to just use the openstack CLI to calibrate it.

cmurphy:
CI is failing 
```
undefined method `node_search_with_cache' for KeystoneHelper:Module
```

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -32,6 +32,110 @@ def self.unversioned_internal_auth_url(node, admin_host)
...
+  def self.trusted_dashboard_url(node)
+    # NOTE(gyee): since Horizon is depended on Keystone and will be deployed
+    # after Keystone, the Horizon node may not be available when Keystone
+    # is first deployed. However, chef executes the recipes periodically.
+    # On the next chef run after, after Horizon had deployed, we should be
+    # able to figure out the trusted_dashboard from the node, assuming Horizon
+    # is always deployed on the same node as Keystone. Otherwise, we'll need
+    # to do node search.
+
+    horizon_server = CrowbarUtilsSearch.node_search_with_cache("roles:horizon-server").first
```
cmurphy: Need to pass the node object in, e.g. https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/octavia/libraries/helpers.rb#L19

guangyee: @cmurphy thanks! I was looking at the wrong example earlier. Will test the changes again today.

chef/data_bags/crowbar/migrate/keystone/304_add_federation_openidc.rb
cmurphy: Okay I think this needs to be a class variable like https://github.com/crowbar/crowbar-openstack/blob/41279a88ee163e3acd2dd034698c4eec65247448/chef/data_bags/crowbar/migrate/neutron/304_add_use_l3_ha.rb#L5-L10 I think that may fix the issue I was seeing

guangyee: thanks! let me give it a try

chef/cookbooks/crowbar-openstack/providers/wsgi.rb
```diff
@@ -67,6 +67,15 @@ def whyrun_supported?
...
         ssl_cacert: current_resource.ssl_cacert,
         timeout: current_resource.timeout,
+        openidc_enabled: current_resource.openidc_enabled,
+        openidc_provider: current_resource.openidc_provider,
+        openidc_response_type: current_resource.openidc_response_type,
+        openidc_scope: current_resource.openidc_scope,
+        openidc_metadata_url: current_resource.openidc_metadata_url,
+        openidc_client_id: current_resource.openidc_client_id,
+        openidc_client_secret: current_resource.openidc_client_secret,
+        openidc_passphrase: current_resource.openidc_client_secret,
```
cmurphy: This is pulling in the wrong attribute, causing the passphrase and the secret to be identical!

guangyee: ahhhhhhh! damn it! I thought I fixed that last week.

chef/data_bags/crowbar/migrate/keystone/304_add_federation_openidc.rb
```diff
@@ -0,0 +1,18 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs.key? "federation"
+    attrs["federation"] = template_attrs["federation"]
+    unless defined(@@federation_openidc_passphrase)
+      service = ServiceObject.new "fake-logger"
+      @@federation_openidc_passphrase = service.random_password
+    end
+
+    attrs["federation"]["openidc"]["passphrase"] = @@federation_openidc_passphrase
```
cmurphy: This still isn't working for me when I install using this patch as an update repo, I can't figure out why. Everything else looks good.

guangyee:
> One last thing is the commit message and PR description need to point to https://docs.openstack.org/keystone/latest/admin/federation/configure_federation.html#setting-up-openid-connect

Done

cmurphy:
[**@guangyee**](https://github.com/guangyee) commented on [Aug 22, 2019, 9:48 AM PDT](https://github.com/crowbar/crowbar-openstack/pull/2188#issuecomment-523986276 "2019-08-22T16:48:56Z - Replied by Github Reply Comments"):
> > One last thing is the commit message and PR description need to point to https://docs.openstack.org/keystone/latest/admin/federation/configure_federation.html#setting-up-openid-connect
> 
> Done

And also the commit message?

chef/data_bags/crowbar/migrate/keystone/304_add_federation_openidc.rb
cmurphy: I figured it out! It needs this https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/models/keystone_service.rb#L67-L69

chef/cookbooks/keystone/libraries/helpers.rb
rhafer: Sorry. I might have been not really clear. What I was trying to says is:

I wonder if that new additional proposal attribute `[:keystone][:federation][:websso_keystone_url]` is really needed. Assuming that the user configure the external name of the cluster correctly it might just be enough to always  use `public_auth_url(node, keystone_public_fqdn(node))` for the websso_keystone_url setting (currently it's only used if the new attribute is not set at all).
Or is it possible/useful that websso_keystone_url points to yet another name?

chef/data_bags/crowbar/template-keystone.json
rhafer: Yeah, I am leaning towards @guangyee's approach here. I was just testing this PR and also failed to get the mappings right on the first try. Having to fill in the mappings in a proposal attribute and going through the whole "update proposal" -> "reapply" cycle everytime would have been really time consuming.

Of course I can also understand @cmurphy's point here. Having to edit more things than just the proposal isn't the providing the best user experience either.

chef/cookbooks/keystone/libraries/helpers.rb
rhafer: Forget about my last comment. After going through the excersize of actually setting this up myself I can see a point in having a separate setting for `[:keystone][:federation][:websso_keystone_url]`which is not based on what ever is setup for the public name.

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -32,6 +32,96 @@ def self.unversioned_internal_auth_url(node, admin_host)
...
+  # NOTE(gyee): trusted_dashboard in Keystone can be multiple URLs.
+  # For example, in a typical production deployment, there can be multiple
+  # Horizon endpoints, depending on where Horizon is being accessed. For
+  # example, the endpoint for access inside the firewall could be different
+  # from the one that is outside of the firewall. And in some cases, the
+  # endpoint could be the corporate HTTP proxy.
+  # For now, we are prepopulating one that is only understood by Crowbar for
+  # testing/demo purpopses. In a production environment, it can be amended with
+  # other external endpoints.
+  def self.dashboard_public_url(node)
```
rhafer: Would you mind renaming the `node` variable here to something like `horizon_node` or `dashboard_node`? Otherwise it is too easy to confuse with the global chef  variable `node` that's present everywhere.

PR #2193, opened 2019-08-14
neutron-lbaas: remove loadbalancer/pool limit

dirkmueller:
This is not really configurable via CLI, and 10 is not a production
default. better have no limit.


chef/cookbooks/neutron/templates/default/neutron_lbaas.conf.erb
```diff
@@ -21,4 +28,3 @@ admin_project_domain = <%= @keystone_settings['admin_domain'] %>
 %>
 service_provider=<%= service_provider %>
 <% end -%>
-[certificates]
```
toabctl: unrelated to the commit

dirkmueller: splitted into a new commit.

swami-dvr:
Looks good to me.

PR #2195, opened 2019-08-16
Add additional debug output to memcached (SOC-10173)

rsalevsky:
In order to understand the isses which cause SOC-10173 better we need
addional debug output.


rsalevsky:
I will add the patch to the package fist to better understand what information is useful.

PR #2199, opened 2019-08-19
horizon: trigger django load as part of deployment (SOC-10011)

flaviodsr:
Due to the "laziness" of django the first request to horizon takes a long
time to complete. This change do a request to horizon as soon as it is
deployed so it triggers the django loading and make further requests to
be replied quickly.

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -519,7 +519,7 @@ template "#{node[:apache][:dir]}/sites-available/openstack-dashboard.conf" do
     grafana_url: grafana_url
   )
   if ::File.symlink?("#{node[:apache][:dir]}/sites-enabled/openstack-dashboard.conf") || node[:platform_family] == "suse"
-    notifies :reload, resources(service: "apache2"), :immediately
+    notifies :restart, resources(service: "apache2"), :immediately
   end
 end
 
@@ -586,4 +586,14 @@ unless monasca_ui_pkgname.nil?
       notifies :reload, "service[horizon]"
     end
   end
+end
+
+# Make a simple request to load cache right away.
+bash "build horizon cache" do
+  url = "http://localhost:#{bind_port}"
+  url = "http://#{bind_host}:#{bind_port}" if ha_enabled
+  code <<-EOH
+    curl --location --max-time 5 --silent --insecure --show-error #{url}
+    EOH
+  ignore_failure true
 end


Itxaka:
FYI, I dont this the message commit is correct. The compression of assets is one thing, and IIRC is done as part of the RPM install as a post command.

Usually the reason the first request to horizon is slow its because django its "lazy" and will defer loading stuff until is requested, so on that first query it will load everything.

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -587,3 +587,16 @@
     end
   end
 end
+
+# Make a simple request to load cache right away.
+bash "build horizon cache" do
+  if ha_enabled
```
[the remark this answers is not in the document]
rsalevsky:
```
url = "http://#{bind_host}:#{bind_port}"
url = "http://localhost:#{bind_port}" if ha_enabled
```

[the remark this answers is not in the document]
flaviodsr: done, thanks!

flaviodsr:
> FYI, I dont this the message commit is correct. The compression of assets is one thing, and IIRC is done as part of the RPM install as a post command.
> 
> Usually the reason the first request to horizon is slow its because django its "lazy" and will defer loading stuff until is requested, so on that first query it will load everything.

Thanks, I have updated the message commit accordingly, hope it is better now :)

Itxaka:
> Thanks, I have updated the message commit accordingly, hope it is better now :)

Much better reflects reality now, thanks!

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -587,3 +587,13 @@
     end
   end
 end
+
+# Make a simple request to load cache right away.
+bash "build horizon cache" do
+  url = "http://localhost:#{bind_port}"
+  url = "http://#{bind_host}:#{bind_port}" if ha_enabled
+  code <<-EOH
+    curl --location --max-time 30 --silent --insecure --show-error #{url}
```
dirkmueller: so that'll hang the deployment then for 30s?

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -519,7 +519,7 @@
     grafana_url: grafana_url
   )
   if ::File.symlink?("#{node[:apache][:dir]}/sites-enabled/openstack-dashboard.conf") || node[:platform_family] == "suse"
-    notifies :reload, resources(service: "apache2"), :immediately
+    notifies :restart, resources(service: "apache2"), :immediately
```
dirkmueller: why this change to restart?

dirkmueller:
@flaviodsr why does it take so long? the first one is doing the asset compilation, sure, but that shouldn't take more than like a minute or so. and we wait for 2 minutes..

chef/cookbooks/horizon/recipes/server.rb
flaviodsr: yes. how long do you suggest? probably 1 or 2 seconds would be enough.

chef/cookbooks/horizon/recipes/server.rb
flaviodsr: The reload for HA deployments is actually not doing anything. The curl fails with connection refused as apache is not listening on the expected `#{bind_port}` which for HA is 5580.

For no-HA it probably works because apache is already listening on the default port (80)

flaviodsr:
@dirkmueller I did measure how long the curl operation takes on each controller of a ha deployment and that's the result:
```
For the first request:
  controller1: real	2m06.887s
  controller2: real	1m16.289s
  controller3: real	1m44.089s

For the second request:
  controller1: real	0m0.755s
  controller2: real	0m3.262s
  controller3: real	0m0.976s
```

So it actually takes more than a minute (maybe due to the performance/load on ECP??). To confirm that those times have something more than the laziness of django requires a better understanding of how to debug horizon/django which I don't have.

chef/cookbooks/horizon/recipes/server.rb
dirkmueller: ah, so you're saying we need restart becasue the listen statements are changing? but is it then only happening once? or on every chef run? we should avoid doing this more than once.

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -587,3 +587,13 @@
     end
   end
 end
+
+# Make a simple request to load cache right away.
+bash "build horizon cache" do
+  url = "http://localhost:#{bind_port}"
+  url = "http://#{bind_host}:#{bind_port}" if ha_enabled
+  code <<-EOH
+    curl --location --max-time 5 --silent --insecure --show-error #{url}
```
dirkmueller: are you sure that we have curl installed at all? isn't haproxy doing the probing for us? why do we need to do this from the chef recipe? haproxy's backend health checks should be doing that arleady.

flaviodsr:
@dirkmueller as you suspected the root cause is `openstack-horizon-plugin-heat-ui`. The heat dashboard increases the cache build time by ~50 seconds.

With `openstack-horizon-plugin-heat-ui` installed:
```
    time_namelookup:  0.000084
       time_connect:  0.000242
    time_appconnect:  0.006281
   time_pretransfer:  0.006609
      time_redirect:  0.010017
 time_starttransfer:  78.891062
                    ----------
         time_total:  78.891213
```

Without `openstack-horizon-plugin-heat-ui`:
```    time_namelookup:  0.000103
       time_connect:  0.000307
    time_appconnect:  0.005039
   time_pretransfer:  0.005421
      time_redirect:  0.008593
 time_starttransfer:  25.933790
                    ----------
         time_total:  25.933951
```

On both cases the cache was cleaned before doing the request.

PR #2200, opened 2019-08-20
memcache: lookup memcached servers port only on local node (SOC-10173)

dirkmueller:
With #2073 merged
we were looking up the port on the node with the lowest mac address,
which however in compilation phase might not yet have that attribute.

By passing in the local node and moving to execution phase we should
always have the default attributes for memcached port visible.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -59,10 +59,6 @@ cluster_nodes = ha_enabled ? node[:pacemaker][:elements]["pacemaker-cluster-memb
 cluster_nodes = cluster_nodes.map { |n| Chef::Node.load(n) }
 cluster_nodes.sort_by! { |n| n[:hostname] }
 
-memcached_servers = MemcachedHelper.get_memcached_servers(
-  ha_enabled ? cluster_nodes : [node]
-)
-
 memcached_instance "keystone"
 
 # resource to set a flag when apache2 is restarted so we now which cookbook was the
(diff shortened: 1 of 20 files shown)


dirkmueller:
@toabctl the underlying problem is that the get_memcached_servers() code is stupid in the sense that it just picks memcached_nodes.first and looks up an attribute there. however it is not guaranteed to have read after write consistency for nodes other than the $currentnode, so selecting first() might fundamentally fail, always, by design of how chef works.

dirkmueller:
looks like this is actually a recent regression introduced by https://github.com/crowbar/crowbar-openstack/pull/2073

we might want to just revert that patch instead.

stefannica:
Was jut looking at #2073 ...

chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -262,7 +258,8 @@
         node[:keystone][:api][:protocol],
         my_admin_host, node[:keystone][:api][:admin_port]
       ),
-      memcached_servers: memcached_servers,
+      memcached_servers: MemcachedHelper.get_memcached_servers(node,
```
stefannica: you missed one inline condition here

chef/cookbooks/nova/recipes/config.rb
```diff
@@ -388,7 +383,8 @@
     vncproxy_cert_file: api_novnc_ssl_certfile,
     vncproxy_key_file: api_novnc_ssl_keyfile,
     serialproxy_public_host: public_api_host,
-    memcached_servers: memcached_servers,
+    memcached_servers:  MemcachedHelper.get_memcached_servers(node,
```
stefannica: another one here

chef/cookbooks/keystone/recipes/server.rb
dirkmueller: there is no inline condition here?

chef/cookbooks/nova/recipes/config.rb
dirkmueller: done

PR #2203, opened 2019-08-22
rabbitmq: Turn off hipe compile

dirkmueller:
It interacts badly with the pacemaker health monitoring and
based on rumours does not work with IPv6.


Gsantomaggio:
>rumours does not work with IPv6.

That's not correct,.
HIPE [does not work with erlang 22](http://erlang.org/pipermail/erlang-questions/2019-March/097512.html)

PR #2204, opened 2019-08-26
Octavia: Follow up patch addressing comments from last PR (SOC-9894)

matthewoliver:
This patch addresses the comments from the last PR (2082) that needed to be
followed up.

Most notable change is config management, which now utilises the conf.d
structure. Each element will touch the common
`octavia.conf.d/100-octavia.conf` but also touch it's individual conf as
`octavia.conf.d/110-#{name}.conf`.
Which makes more sense as octavia is run from the same config file. And
stops resetting bind_ips and ports.

The "enable ssh" and "keyname"  has also been removed from the 'custom' UI.
To enable and add an SSH keyname is now down only in 'raw' mode. And
done by [:ssh_access][:keyname] field. If it's nil or empty, it's
disabled otherwise enabled.

We also go one step further then the last PR and create the octavia
management network if it isn't there. The network name is defined by
[:amphora][:manage_net] and [:amphora][:manage_cidr] and defaults to:

  "manage_net": "lb-mgmt-net",
  "manage_cidr": "172.31.0.0/16",

These can also be changed in the 'raw' mode.


matthewoliver:
Still a WIP. Working through testing on my mkcloud host. Want to get something up to 1. get it off my computer and 2. so people can see it.

chef/cookbooks/octavia/templates/default/100-octavia.conf.erb
```diff
@@ -57,18 +36,6 @@ base_cert_dir = <%= @node[:octavia][:certs][:cert_path]%>
...
-event_streamer_driver = queue_event_streamer
-heartbeat_interval = <%= @node[:octavia][:health_manager][:heartbeat_interval] %>
-heartbeat_timeout = <%= @node[:octavia][:health_manager][:heartbeat_timeout] %>
-health_check_interval = <%= @node[:octavia][:health_manager][:health_check_interval] %>
-
-[house_keeping]
-
 [keystone_authtoken]
 auth_version = <%= @octavia_keystone_settings['api_version'] %>
 www_authenticate_uri = <%= @octavia_keystone_settings["public_auth_url"] %>
```
matthewoliver: In general there might be more sections to slit out into the other daemon configs. But assume what's in this patch is the minimum.

crowbar_framework/app/views/barclamp/octavia/_edit_attributes.html.haml
```diff
@@ -20,8 +20,8 @@
 
       = boolean_field %w(amphora ssh_access enabled),
         "data-showit" => "true",
-        "data-showit-target" => "#ssh_access_keyname_containe",
+        "data-showit-target" => "#ssh_access_keyname_container",
         "data-showit-direct" => "true"
 
-      #ssh_access_keyname_containe
+      #ssh_access_keyname_container
```
matthewoliver: Probably should just remove this whole ssh_access part, so it should probably only be accessed in "raw" expert mode.

matthewoliver:
Currently getting `openstack image create ` errors. So patch still needs some work.Maybe its access to the image.

matthewoliver:
OK I'm comfused hound.. all hound vilations are now "outdated" so what's the one I'm missing?

chef/cookbooks/octavia/attributes/default.rb
```diff
@@ -50,7 +50,6 @@
 default[:octavia][:amphora][:image_tag] = "amphora"
 default[:octavia][:amphora][:project] = "service"
 
-default[:octavia][:amphora][:ssh_access][:enabled] = false
 default[:octavia][:amphora][:ssh_access][:keypair] = ""
```
ilausuch: The name of the variable should be keyname instead of keypair. This is a previous error, sorry

chef/cookbooks/octavia/definitions/octavia_conf.rb
```diff
@@ -28,29 +28,36 @@
                                               "-f value -c id").stdout.delete "\n"
       node.save
     end
+    only_if do
```
ilausuch: I agree with this. But I am thinking that if someone wants to update the flavor for instance it will never update the node variable

chef/cookbooks/octavia/recipes/nova.rb
```diff
@@ -15,9 +15,10 @@
 
 require "chef/mixin/shell_out"
 
-package "openstack-octavia-amphora-image-x86_64"
 
+image = "openstack-octavia-amphora-image-x86_64"
 ha_enabled = node[:octavia][:ha][:enabled]
+package image if !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node)
```
ilausuch: I am sure there is a good reason to do this but I cannot figure it out now, Why in ha we don't install the package?

matthewoliver: Ha does install the package, but only on the cluster founder node. So the one that will push the image. 
Just so we don't install it on every node when we don't need to.

chef/cookbooks/octavia/attributes/default.rb
matthewoliver: Cool, thanks I'll fix that and push a new version up!

chef/cookbooks/octavia/definitions/octavia_conf.rb
matthewoliver: hmm, yeah, might need to think on this.

chef/cookbooks/octavia/recipes/nova.rb
ilausuch: Ok, good point thanks!

chef/cookbooks/octavia/definitions/octavia_conf.rb
matthewoliver: OK, so got a new version. We know grab the old amphora settings from the an old proposal when we make changes and save it to the nodes in question (this has to happen on the crowbar rails side). On the chef side, we can check to see if the old amphora setting have changed, if they have then we'll go ask for the update the information.

Testing this on my mkcloud host and seems to work. I can change the flavor and the config updates with the correct flavor id. And the openstack cli isn't run if nothing has changed.. so I think that's what we want. 

Anyway, please review the new code.

ilausuch: Good point, I could work during the best of cases. I am not sure if we should consider that chef breaks before update the new variables (flavor,..). The old_amphora variable would be updated but the variables not. Is it possible? I am not quite sure about that

chef/cookbooks/octavia/attributes/default.rb
ilausuch: Perfect!

nicolasbock:
Rebased

chef/cookbooks/octavia/recipes/nova.rb
```diff
@@ -76,15 +77,35 @@
...
   retries 5
   retry_delay 10
   action :run
 end
+
+manage_net = node[:octavia][:amphora][:manage_net]
+manage_cidr = node[:octavia][:amphora][:manage_cidr]
+
+execute "create_octavia_management_network" do
+  command "#{cmd} network create --project #{project_name} #{manage_net}"
```
stefannica: This will use the default provider network type, which could be vlan, vxlan or gre. The external network configuration (i.e. outside the umbrella of the LCM) then also needs to be changed to be able to route between the Openstack management network and this vlan/vxlan/gre tenant network. Shouldn't we add another parameter to control the segmentation type ? Or restrict it to VLAN, with the proper validation in place ? I think we need to make this explicitly clear rather than rely on neutron defaults.

chef/cookbooks/octavia/libraries/helpers.rb
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 SUSE Linux GmbH.
+# Copyright 2020 SUSE Linux GmbH.
```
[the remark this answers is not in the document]
nicolasbock: On our intra webpage it's still just saying `SUSE LINUX GmbH`. I agree with the year though :smile:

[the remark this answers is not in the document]
dirkmueller: @nicolasbock we definitely renamed. where did you find the old name?

chef/cookbooks/octavia/recipes/nova.rb
stefannica: To follow up, we're also missing the security group settings for this network, allowing amphora API and health-check traffic between the amphorae and Octavia controller.

More on this here: https://docs.openstack.org/octavia/latest/contributor/guides/dev-quick-start.html#load-balancer-network-configuration

chef/cookbooks/octavia/libraries/helpers.rb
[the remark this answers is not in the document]
dirkmueller: @nicolasbock I'm also fine with SUSE LLC. do it that way then.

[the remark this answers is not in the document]
nicolasbock: Thanks. I'll update the copyright notices.

stefannica:
Testing [the crowbar-openstack PR](https://github.com/crowbar/crowbar-openstack/pull/2204) alongside [the automation PR](https://github.com/SUSE-Cloud/automation/pull/3748) and [a submit request against the python-octavia package](https://build.opensuse.org/request/show/742140) with:
 * no-ha: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/508/pipeline
 * ha: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/507/pipeline

chef/cookbooks/octavia/recipes/nova.rb
nicolasbock: @stefannica Do you want to revoke your "request changes"?

nicolasbock:
@ilausuch Could you re-review?

nicolasbock:
Changed the commit messages slightly to pass the linter

chef/cookbooks/octavia/recipes/nova.rb
stefannica: As agreed, we'll solve the management network problem in a subsequent PR. For now, the workaround is to create the management network & subnet manually before applying the octavia barclamp (or to create a management network & subnet with a different name after having applied the octavia barclamp, followed by changing the `lb_mgmt_net` barclamp attribute value).

PR #2210, opened 2019-08-28
nova: raise neutron client timeout to 5 minutes

dirkmueller:
The 30s is definitely too low for environments with many
thousand ports in active state. 300 is also the set value
in Ardana.


dirkmueller:
@toabctl done

nicolasbock:
The commit message title still needs a Jira reference.

dirkmueller:
@nicolasbock why does it *need* a jira reference?

nicolasbock:
@dirkmueller Because that's how we configured our linter. I am just the messenger :wink:

dirkmueller:
@nicolasbock there is no agreement to enforce a jira ticket. its just meant as a reminder by Rick that people should think whether there is one.

nicolasbock:
Ah ok, thanks.

PR #2198, opened 2019-08-19
tempest: Install available clients and tempest plugins (SOC-10191)

rsalevsky:
We should install all available clients and tempest plugins.

PR #2213, opened 2019-08-29
Install barbican tempest plugin for SOC8 (SOC-10191)

gosipyan:
Barbican tests curently not running on SOC8 because
python-barbican-tempest-plugin not installed.


gosipyan:
Please ignore ,PR already exist with change https://github.com/crowbar/crowbar-openstack/pull/2198

PR #2214, opened 2019-08-29
tempest: add kibana_version to the identity senction of tempest.conf

sjamgade:
kibana_version conf is needed by monasca_log_api_tempest testcases as
this key-value pair is added headers of requests to kibana
https://opendev.org/openstack/monasca-log-api/src/tag/newton-eol/monasca_log_api_tempest/config.py#L40

sample failure: https://ci.suse.de/job/openstack-mkcloud/8437/consoleText

### DIFF
--- a/chef/cookbooks/tempest/templates/default/tempest.conf.erb
+++ b/chef/cookbooks/tempest/templates/default/tempest.conf.erb
@@ -92,6 +92,7 @@ alt_username = <%= @alt_comp_user %>
 alt_tenant_name = <%= @alt_comp_tenant %>
 alt_password = <%= @alt_comp_pass %>
 alt_domain_name = Default
+kibana_version = '4.6.3'
 
 [identity-feature-enabled]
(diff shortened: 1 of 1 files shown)


chef/cookbooks/tempest/templates/default/tempest.conf.erb
```diff
@@ -92,6 +92,7 @@ alt_username = <%= @alt_comp_user %>
 alt_tenant_name = <%= @alt_comp_tenant %>
 alt_password = <%= @alt_comp_pass %>
 alt_domain_name = Default
+kibana_version = '4.6.3'
```
toabctl: would it be possible to get the installed kibana version instead of hardcoding the version?

sjamgade: I am not sure how to carry that version from monasca-server node to controller where the tempest is installed without putting is much effort for SOC7 which I was told is EOL in couple of months from now.
Also Kibana is not something we update frequently so this version wont change as often.

What do you suggest ?

JanZerebecki:
Looks good, but needs rebase.

PR #2215, opened 2019-08-30
Designate default Bind9 pool config (SOC-10339)

ritesh216:
We have default Bind9 server pool for Designate.
We also apply DNS barclamp that configures Bind9 cluster
with one master and others as slaves.

With designate enabled, all Bind9 servers act as slaves for
the zones created by designate. We need to make this happen
via pool configuration.

Updating the default Bind9 pool configuration:
1. nameservers: Add all Bind9 servers as nameservers.
2. targets: One block for each of Bind9 servers.
2. targets/masters: Add designate-mdns(s) as masters.
3. also_notify: Leave empty.


chef/cookbooks/designate/recipes/mdns.rb
```diff
@@ -44,19 +50,21 @@
...
+    {
+      "type" => "bind9",
+      "description" => "BIND9 Server",
+      "masters" => hiddenmasters.map { |hm| { "host" => hm, "port" => 5354 } },
+      "options" => {
+        "host" => ip,
+        "port" => 53,
+        "rndc_host" => ip,
+        "rndc_port" => 953,
+        "rndc_key_file" => "/etc/designate/rndc.key"
```
matthewoliver: I might have to go check, I have a memory that we are only dropping this rndc.key on the crowbar DNS master
So, should we be dropping this on every dns node?

I always get confused with this, so nameservers is the nameservers designate will query to make sure a change has propagated. And targets are what it updates. So how does this work? We give a list of masters that the crowbar dns nameservers will use as masters they will request zone transfers from.
So, in our case, each target is a crowbar dns node, and designate will use the rndc.key to configure it to treat the  mdns as masters?

If we only have crowbar DNS master and the rest slaves, then do we only actually need one target (the master with the rndc.key set) as the slaves of this master will already just accept what the master says? Or as you have it here we want to configure the crowbar master and slaves to treat mdns as masters?

There is alot of dns in the above comment, so hopefully it makes sense :P

ritesh216: rndc.key is generated in Bind9 cookbook and is used by all Bind9 servers to allow Designate rndc access to them (chef/cookbooks/bind9/templates/default/designate-rndc-access.conf.erb). The same key is passed as node attribute and used in mdns (generally present on the controller) recipe; this is used by Designate mdns to do rndc operations on all Bind9s.

Yes pretty much. When we do openstack zone create; designate-mdns does a rndc addzone; which in effect in NOTIFY to all slaves (Bind9s); so that they can request Zone transfer (AXFR).

Actually, I had been trying to make that work for some days; finally got reply from Graham that Designate does not support that (one Bind9 as master and other slaves...). So mdns is master and rest slaves is the recommended way to go.

It all made sense. These are the questions I had been struggling with past few days :)

matthewoliver: Cool, nice. Great research and getting it straight from Graham means we can move forward!
I wonder if we should have a comment or probably even better document (in the documentation card) that this is the situation and the only way it's supported. Cause no doubt this is something others will ask when they ever need to set up designate ;)

And I see PR 1912 in crowbar-core now. So yeah. Great work!

chef/cookbooks/designate/recipes/mdns.rb
```diff
@@ -24,18 +24,24 @@
 dnsslaves = dns[:dns][:slave_ips].to_a
 dnsservers = [dnsmaster] + dnsslaves
```
matthewoliver: NIT: We don't use dnsmaster or dnsslaves anywhere else now. So we may not need to pull them out like this. Although it does make it easier to read.
We could optionally do something like:
`dnsservers = [dns[:dns][:master_ip]] + dns[:dns][:slave_ips].to_a`
And remove lines 23 and 24... but this is really just a very very minor NIT. I'm happy if it stays as it is. It definitely reads better this way :)

ritesh216: This master_ip business will be cleaned up via: https://jira.suse.com/browse/SOC-10456

s-t-e-v-e-n-k: I've pushed up https://github.com/crowbar/crowbar-openstack/pull/2224, which still requires them split out. :-(

nicolasbock: Can we resolve this conversation here @s-t-e-v-e-n-k ?

chef/cookbooks/designate/recipes/mdns.rb
nicolasbock: Can we resolve this discussion @matthewoliver ?

chef/cookbooks/designate/recipes/mdns.rb
```diff
@@ -24,18 +24,24 @@
...
-network_settings = DesignateHelper.network_settings(node)
+designate_servers = node_search_with_cache("roles:designate-server")
+
+# hidden masters are designate-mdns services, in ha this service will be running on multiple
+# hosts and any host can be asked to update a zone on the pool target(s).
+# We use the vip for the cluster in case of HA.
+hiddenmasters = if node[:designate][:ha][:enabled]
+  [CrowbarPacemakerHelper.cluster_vip(node, "admin")]
+else
+  designate_servers.map do |n|
```
matthewoliver: This is really just more of a confirmation rather then a problem. So hidden masters are really the mdns servers. So just confirming that mdns servers are on every "roles:desinate-server".

Looking at the chef roles, the "designate-server" role does include the mdns recipe. So cool. I guess this is fine. If we ever decide to split up the roles more (across servers) we can look at a more specific mdns role or something.

PR #2216, opened 2019-08-30
enable LDAP chase_referrals configuration (SOC-7364)

guangyee:
For Microsoft Active Directory deployments where users are not
visible/accessible via the global catalog, list users operation, as a
whole, will fail when one or more referrals are returned on an LDAP
lookup and the current (admin) user have no access to a given referral.

This patch exposes the chase_referrals option in Keystone to enable
customers who got stuck in the above situation to optionally turn
off referrals chasing.

By default, chase_referrals is set to an empty value to preserve
backward compatibility. This is a three state configuration.

Empty value means use the system default. This is all depended on the
underlying LDAP SDK and the system configuration (i.e. /etc/ldap.conf).
"0" means disable referrals chasing.
"1" means enable referrals chasing.

This patch is also exposes the insecure_debug option to help customers and
field engineers to quicky pinpoint the problem when needed.

### DIFF
--- a/chef/cookbooks/keystone/attributes/default.rb
+++ b/chef/cookbooks/keystone/attributes/default.rb
@@ -25,6 +25,7 @@ default[:keystone][:user] = "keystone"
 default[:keystone][:group] = "keystone"
 
 default[:keystone][:debug] = false
+default[:keystone][:insecure_debug] = false
 default[:keystone][:frontend] = "apache"
 default[:keystone][:domain_specific_drivers] = false
 default[:keystone][:domain_config_dir] = "/etc/keystone/domains"
@@ -63,6 +64,7 @@ default[:keystone][:ldap][:password] = ""
 default[:keystone][:ldap][:suffix] = "cn=example,cn=com"
 default[:keystone][:ldap][:page_size] = 0
 default[:keystone][:ldap][:alias_dereferencing] = "default"
+default[:keystone][:ldap][:chase_referrals] = ""
 default[:keystone][:ldap][:query_scope] = "one"
 
 default[:keystone][:ldap][:user_tree_dn] = ""
--- a/chef/data_bags/crowbar/migrate/keystone/305_add_ldap_configs_and_insecure_debug.rb
+++ b/chef/data_bags/crowbar/migrate/keystone/305_add_ldap_configs_and_insecure_debug.rb
@@ -1,0 +1,26 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs["ldap"].key?("chase_referrals")
+    attrs["ldap"]["chase_referrals"] = template_attrs["ldap"]["chase_referrals"]
+  end
+  attrs["domain_specific_config"].keys.each do |domain|
+    unless attrs["domain_specific_config"][domain]["ldap"].key?("chase_referrals")
+      attrs["domain_specific_config"][domain]["ldap"]["chase_referrals"] =
+        template_attrs["domain_specific_config"]["ldap_users"]["ldap"]["chase_referrals"]
+    end
+  end
+  attrs["insecure_debug"] = template_attrs["insecure_debug"] unless attrs.key?("insecure_debug")
+
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["ldap"].delete("chase_referrals") unless template_attrs["ldap"].key?("chase_referrals")
+  unless template_attrs["domain_specific_config"]["ldap_users"]["ldap"].key?("chase_referrals")
+    attrs["domain_specific_config"].keys.each do |domain|
+      attrs["domain_specific_config"][domain]["ldap"].delete("chase_referrals")
+    end
+  end
+  attrs.delete("insecure_debug") unless template_attrs.key?("insecure_debug")
+
+  return attrs, deployment
+end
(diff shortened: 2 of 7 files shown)


guangyee:
> Question about the commit message -
> 
> ```
> For this type of deployment, customer will be using
> domain-specific configuration where chasing referrals is not needed.
> ```
> 
> Why is this specific to a domain-specific configuration?

If there's no LDAP (global) admin user that have access to the global catalog, customers must use domain-specific configuration where each "domain" have its owner (domain) admin user what can fetch users for the given domain. In that case, there's no need to chase referrals.

> 
> This will need a data bag migration and a schema revision bump in template-keystone.json.

Why's migration needed? We are not setting it to a default value. By default it is unset.

cmurphy:
[**@guangyee**](https://github.com/guangyee) commented on [Sep 9, 2019, 1:50 PM PDT](https://github.com/crowbar/crowbar-openstack/pull/2216#issuecomment-529661296 "2019-09-09T20:50:28Z - Replied by Github Reply Comments"):
> > Question about the commit message -
> > 
> > ```
> > For this type of deployment, customer will be using
> > domain-specific configuration where chasing referrals is not needed.
> > ```
> > 
> > Why is this specific to a domain-specific configuration?
> 
> If there's no LDAP (global) admin user that have access to the global catalog, customers must use domain-specific configuration where each "domain" have its owner (domain) admin user what can fetch users for the given domain. In that case, there's no need to chase referrals.

I'm still not sure what a keystone domain (as opposed to an AD domain) has to do with chasing referrals, as the LDAP backend for a keystone domain may not map 1:1 with an AD domain and chasing referrals could still be needed.

> 
> > 
> > This will need a data bag migration and a schema revision bump in template-keystone.json.
> 
> Why's migration needed? We are not setting it to a default value. By default it is unset.

A migration is needed whenever any attribute is added. If you test this yourself by applying the code to an already-deployed crowbar, you can see that the new attributes are not visible in the proposal raw view.

guangyee:
> [**@guangyee**](https://github.com/guangyee) commented on [Sep 9, 2019, 1:50 PM PDT](https://github.com/crowbar/crowbar-openstack/pull/2216#issuecomment-529661296):
> 
> > > Question about the commit message -
> > > ```
> > > For this type of deployment, customer will be using
> > > domain-specific configuration where chasing referrals is not needed.
> > > ```
> > > 
> > > 
> > > Why is this specific to a domain-specific configuration?
> > 
> > 
> > If there's no LDAP (global) admin user that have access to the global catalog, customers must use domain-specific configuration where each "domain" have its owner (domain) admin user what can fetch users for the given domain. In that case, there's no need to chase referrals.
> 
> I'm still not sure what a keystone domain (as opposed to an AD domain) has to do with chasing referrals, as the LDAP backend for a keystone domain may not map 1:1 with an AD domain and chasing referrals could still be needed.
> 

You're right. An AD domain doesn't not necessary map to a Keystone domain. And Keystone domain-specific configuration is orthogonal to referral chasing. I'll take out the sentence.

> > > This will need a data bag migration and a schema revision bump in template-keystone.json.
> > 
> > 
> > Why's migration needed? We are not setting it to a default value. By default it is unset.
> 
> A migration is needed whenever any attribute is added. If you test this yourself by applying the code to an already-deployed crowbar, you can see that the new attributes are not visible in the proposal raw view.

Thanks for the info! No wonder I didn't ran into this issue. I was testing it with just 'mkcloud instonly' then manually apply the barclamps.

chef/cookbooks/keystone/attributes/default.rb
```diff
@@ -25,6 +25,7 @@
 default[:keystone][:group] = "keystone"
 
 default[:keystone][:debug] = false
+default[:keystone][:insecure_debug] = false
```
cmurphy: I thought I commented on this before but I can't find it. I'd prefer we didn't expose insecure_debug. The user can always add it in an override file if they need it, as it's not domain-specific, and in the situations where things are bad enough to need it then they probably don't want to go through the whole proposal process anyway, easier to just add the override file and restart apache. Exposing it as a barclamp option legitimizes it and runs the risk they might leave it on in production.

guangyee: Well, if customer is doing config override manually. There's also a chance of them accidentally leaving it in production. Upgrade probably won't detect it either. Chris requested this one because it helps the L3 folks actually see what get sent to the LDAP server.

I am a believer in "Security is a process, software is a tool." mindset. :-) Unless Chris come up with a compelling argument, I'll leave it out for now.

chef/data_bags/crowbar/migrate/keystone/305_add_ldap_configs_and_insecure_debug.rb
```diff
@@ -0,0 +1,30 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs.key? "ldap"
+    unless attrs["ldap"].key? "chase_referrals"
+      attrs["ldap"]["chase_referrals"] = template_attrs["ldap"]["chase_referrals"]
+    end
+    unless attrs["ldap"]["domain_specific_config"].key? "chase_referrals"
```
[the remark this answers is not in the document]
guangyee: Good catch! Let me fix.

chef/cookbooks/keystone/attributes/default.rb
cdevitasuse: Customers don't get the LDAP configuration correct the first time and rarely apply the barclamp once. From a troubleshooting perspective it should be easy to get it working by providing a way to get useful diagnostic output. Not including it only makes it harder for customers to find it in the documentation and put it in the override.  If it was in the barclamp it would be  "hey let's try that" to see if we get better debug before  pronouncing keystone has a problem.

cdevitasuse:
Tested and got the ldap parameters that keystone is using. Constructed ldapsearch from that output

PR #2219, opened 2019-09-03
[5.0] tempest: Skip failing tests (SOC-9799)

gosipyan:
Manila see bug SOC-10395
Ceilometer see bug lp 1673738

### DIFF
--- a/chef/cookbooks/tempest/files/default/run_filters/ceilometer.txt
+++ b/chef/cookbooks/tempest/files/default/run_filters/ceilometer.txt
@@ -1,1 +1,4 @@
 +ceilometer.*
+
+# BUG lp 1673738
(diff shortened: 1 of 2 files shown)


dirkmueller:
@gosipyan is this a backport from cloud9? then please add cherry-pick identifiers. if it isn't can we backpor tthe fixes from cloud9 to 8 rather than blacklisting testing?

gosipyan:
@dirkmueller Currently in SOC9 dont have ceilometer run_filter enabled , regarding manila those tests are passing on SOC9 but failing in SOC8(also mkcloud fulltempest) I have created bug ,not aware of any fix that can be backported from SOC9

dirkmueller:
@gosipyan right, ceilometer changed in cloud9, so thats explaining things. the manila part is actually being fixed by core services team..

gosipyan:
@dirkmueller should i remove test skip now or after manila fix merged?do you ave any reference to fix?

PR #2221, opened 2019-09-03
designate: cleanup producer HA deployment (SOC-9766)

s-t-e-v-e-n-k:
The previous PR merged before I managed to push up some fixes to silence
Hound, and correct the sync marks.

### DIFF
--- a/chef/cookbooks/designate/recipes/api.rb
+++ b/chef/cookbooks/designate/recipes/api.rb
@@ -16,8 +16,6 @@
 # Recipe:: api
 #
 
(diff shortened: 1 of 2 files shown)


matthewoliver:
ahh you have hound complaining about something but everything is 'outdated' too. Yay, thanks hound. Bad dog! If you don't like something tell us what it is. 
Because otherwise the patch looks good to go, ie. a nice tidy up, installing the package when it's actually required etc. And a wait point for the non-leaders to wait on while the leader does it's thing.

PR #2224, opened 2019-09-05
designate: Handle one dns-server node for mdns (SOC-10456)

s-t-e-v-e-n-k:
The mdns recipe was written with the assumption that the dns-server role
would be applied to multiple servers, which means one of them would be
elected master with a master_ip property set. This does not exist if
dns-server is only applied to one role, so handle that case as well.

Also drive-by correcting the recipe name in the header comment.

### DIFF
--- a/chef/cookbooks/designate/recipes/mdns.rb
+++ b/chef/cookbooks/designate/recipes/mdns.rb
@@ -13,16 +13,15 @@
 # limitations under the License.
 #
 # Cookbook Name:: designate
-# Recipe:: api
+# Recipe:: mdns
 #
 
 require "yaml"
 
 dns_all = node_search_with_cache("roles:dns-server")
-dns = dns_all.first
-dnsmaster = dns[:dns][:master_ip]
-dnsslaves = dns[:dns][:slave_ips].to_a
-dnsservers = [dnsmaster] + dnsslaves
+dnsservers = dns_all.map do |n|
+  Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address
+end
 
 designate_servers = node_search_with_cache("roles:designate-server")
 


chef/cookbooks/designate/recipes/mdns.rb
```diff
@@ -13,15 +13,20 @@
...
-dnsmaster = dns[:dns][:master_ip]
-dnsslaves = dns[:dns][:slave_ips].to_a
+if dns[:dns].key? "master_ip"
+  dnsmaster = dns[:dns][:master_ip]
+  dnsslaves = dns[:dns][:slave_ips].to_a
+else
+  dnsmaster = Chef::Recipe::Barclamp::Inventory.get_network_by_type(dns, "admin").address
+  dnsslaves = []
+end
 dnsservers = [dnsmaster] + dnsslaves
```
ritesh216: Can't we do:
dnsservers = dns_all.map do |n|
    Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address
end

and call it a day?

nicolasbock: Hmm, I don't quite see (yet) how that is the same as what @s-t-e-v-e-n-k proposes. Could you elaborate @ritesh216 ?

ritesh216: @nicolasbock We do not use dnsmaster and dnsslaves in the mdns recipe. So my proposal is to get rid of those two and directly compose list of IPs that dns-servers listen on.

nicolasbock: Thanks @ritesh216 . I was wondering how

```
[dnsmaster] + dnsslaves
```

is identical to

```
dnsservers = dns_all.map do |n|
  Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address
end
```

That's not clear to me right now by looking at the logic defining `dnsservers` and `dnsslaves`.

ritesh216: Here is the master_ip and slave_ips assignments:
https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/dns_service.rb#L155

So, as I understand it, among all the nodes running dns-server (Bind9), first one is chosen as master and its admin IP becomes the master IP; and the admin IPs of other nodes becomes the slave IPs.

Since, in Designate, we treat all dns-servers (Bind9) equally; we do not need to keep track of dnsmaster and dnsslaves. Hence, we compile just the admin network IPs for all nodes running dns-server and use it via dnsservers.

s-t-e-v-e-n-k: @ritesh216 We do still use dnsmaster actually, but I think I see a way forward ...

s-t-e-v-e-n-k: And dnsslaves, so my way forward isn't so clear.

s-t-e-v-e-n-k: Okay, I see now. This can be cleaned up when https://github.com/crowbar/crowbar-openstack/pull/2215 lands.

matthewoliver:
of course let's see what CI thinks. But it pretty straight forward IMHO.

ritesh216:
Looks good to me.

PR #2225, opened 2019-09-05
tempest: remove manila test from blacklist

flaviodsr:
bsc#1137262 was fixed so all manila tempest tests should be passing now.


flaviodsr:
Tested in https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar-tests/detail/openstack-crowbar-tests/45/tests

PR #2231, opened 2019-09-10
Install barbican tempest plugin for SOC8 (SOC-10191)

gosipyan:
Barbican tests curently not running on SOC8 because
python-barbican-tempest-plugin not installed.

### DIFF
--- a/chef/cookbooks/tempest/recipes/install.rb
+++ b/chef/cookbooks/tempest/recipes/install.rb
@@ -42,3 +42,9 @@ if node[:kernel][:machine] == "x86_64" &&
     package "python-monasca-#{component}"
   end
 end
+
+[
+  "barbican"
+].each do |component|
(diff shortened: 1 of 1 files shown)


chef/cookbooks/tempest/recipes/install.rb
```diff
@@ -31,6 +31,8 @@
 
 package "openstack-tempest-test"
 
+package "python-barbican-tempest-plugin"
```
dirkmueller: can we backport the logic from master? there are more tempest plugins missing:

```
openstack.group:      <package name="python-barbican-tempest-plugin" supportstatus="unsupported"/>
openstack.group:      <package name="python-designate-tempest-plugin" supportstatus="unsupported"/>
openstack.group:      <package name="python-monasca-tempest-plugin" supportstatus="unsupported"/>
openstack.group:      <package name="python-heat-tempest-plugin" supportstatus="unsupported"/>
```

PR #2232, opened 2019-09-10
[5.0] Reinstall monasca node

sjamgade:
the chef clinet run after os install during the reinstallation of the
node expects the node is ready with monasca services(eg: moansca-api)
This is false assumption as the monasca-installer (installed on
monasca-master) has not yet run, thus causing disruption as monasca-setup
tries to configure services for monasca-agent.

The monasca-master node has keys which indicate if monasca-installer has
finished successfully. This patch remove those keys when a node with
monasca-server role is being "reinstalled". Thus informing monasca-agent
that monasca services are no longer installed.

This patch does not solve the problem- "re-running moansca-installer"
for the services to be installed. That is a bigger problem as
monasca-master needs the nodes with monasca-server role to be in ready
state with no parallel execution of chef-client.


chef/cookbooks/crowbar-openstack/libraries/helpers.rb
```diff
@@ -36,7 +36,8 @@ def fetch_rabbitmq_settings(barclamp=@cookbook_name)
     def no_monasca_server_or_master
       monasca_master = node_search_with_cache("roles:monasca-master").first
       monasca_server = node_search_with_cache("roles:monasca-server").first
-      monasca_master.nil? || monasca_server.nil?
+      inst = monasca_master[:monasca] && monasca_master[:monasca][:installed]
```
witekest: We should check if `monasca_master.nil?` first.

sjamgade:
This is one of the ways the [second problem from here](https://jira.suse.com/browse/SOC-10475?focusedCommentId=955951&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-955951) show up and I cannot think of way to fix that. We should look for more help on this one

chef/data_bags/crowbar/migrate/monasca/205_allow_node_transition_notification.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  deployment["config"]["transitions"] = template_deployment["config"]["transitions"]
```
jgrassler: Can you please add a check for the presence of these two settings here (in case we ever backport this feature to Cloud 7)?

jgrassler: Sorry. I didn't pay attention to the crucial detail yesterday: this is not an attribute but one of the keys under deployment. That doesn't require a such a check, of course. Please disregard my comment above :-)

sjamgade:
needed a rebase and added another commit to fix the issue discovered in Witek's comment

PR #2237, opened 2019-09-11
Octavia: Hide UI until complete (SOC-10550)

matthewoliver:
This patch sets `user_managed: false` so that the UI doesn't appear
in crowbar. This will be changed back to `true` once we've finished the
barclamp and it's been released.


matthewoliver:
Just corrected the commit message.

PR #2241, opened 2019-09-13
Allow setting ElasticSearch path.repo from Crowbar (SOC-10440)

witekest:
To create database backup the directory for storing snapshots must be
configured. Administrator should be able to specify target directories
for these snapshots and so the variable should be exposed in Crowbar
proposal.

Although [monasca][elasticsearch][repo_dir] variable has been present in
the data_bag, it hasn't been used in the template. This change allows
setting path.repo configuration setting in Crowbar proposal as
documented.


sjamgade:
LGTM, waiting for the CI

PR #2243, opened 2019-09-16
Add tempest filter for designate (SOC-10288)

gosipyan:
Add tempest filter to run designate tests

### DIFF
--- a/chef/cookbooks/tempest/files/default/run_filters/designate.txt
+++ b/chef/cookbooks/tempest/files/default/run_filters/designate.txt
@@ -1,0 +1,1 @@
++designate_tempest_plugin.*


chef/cookbooks/tempest/files/default/run_filters/designate.txt
```diff
@@ -0,0 +1 @@
+*designate_tempest_plugin.*
```
flaviodsr: I think there is a typo, if you are including the regex the line should start with `+` instead of `*`.

gosipyan: Thanks! fixed

PR #2244, opened 2019-09-16
[5.0] designate: Mark as user managed (SOC-10233)

s-t-e-v-e-n-k:
The designate barclamp has been marked as not user manageable since it
had some issues. Since those have been corrected, mark it as user
manageable, so it appears in the web UI.

(cherry picked from commit 3b753ac93d36410c72ffc8a81e97e4320c603e97)


ritesh216:
Looks good; do we want to make it user visible in SOC 8, I am wondering if this needs PMs go ahead...

PR #2247, opened 2019-09-17
Add pci passthrough filter (SOC-10624)

kwu83tw:
Add pci passthrough filter to default all filters.


nicolasbock:
Still looks good.

PR #2248, opened 2019-09-18
rabbitmq: fix migrations (SOC-10623)

jgrassler:
Two of the Cloud 8 migrations for RabbitMQ have a bug where
they never add the fields they are supposed to add to the
proposal. This commit fixes the issue.

See https://jira.suse.com/browse/SOC-10623 for details. This will need a backport to SOC8 as well since the missing migration breaks the Cloud 7 to Cloud 8 upgrade.

Setting "do not merge yet" since we are currently testing this on a QA cloud.

### DIFF
--- a/chef/data_bags/crowbar/migrate/cinder/205_add_resource_limits.rb
+++ b/chef/data_bags/crowbar/migrate/cinder/205_add_resource_limits.rb
@@ -1,5 +1,5 @@
 def upgrade(ta, td, a, d)
-  a["resource_limits"] = ta["resource_limits"] unless a["resource_limits"]
+  a["resource_limits"] = ta["resource_limits"] unless a.key?("resource_limits")
   return a, d
 end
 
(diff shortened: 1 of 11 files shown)


dirkmueller:
even further: 

```ruby
migrate/nova/207_add_resource_limits.rb:  a["resource_limits"] = ta["resource_limits"] unless a["resource_limits"]

migrate/rabbitmq/200_add_resource_limits.rb:  a["resource_limits"] = ta["resource_limits"] unless a["resource_limits"]
```

jgrassler:
Yeah. `grep -r unless crowbar-openstack/chef/data_bags | grep -v 'key\?' | grep -v 'nil\?' | grep -v 'include\?' | grep -v 'defined\?' | grep -v 'empty\?' | grep -v 'is_a?'` says this is a very common mistake :-)

On it...

jgrassler:
Just tested this and together with https://github.com/crowbar/crowbar-openstack/pull/2260 the Cloud 7 to Cloud 8 upgrade passes now.

PR #2251, opened 2019-09-18
enable LDAP chase_referrals configuration (SOC-7364)

guangyee:
For Microsoft Active Directory deployments where users are not
visible/accessible via the global catalog, list users operation, as a
whole, will fail when one or more referrals are returned on an LDAP
lookup and the current (admin) user have no access to a given referral.

This patch exposes the chase_referrals option in Keystone to enable
customers who got stuck in the above situation to optionally turn
off referrals chasing.

By default, chase_referrals is set to an empty value to preserve
backward compatibility. This is a three state configuration.

Empty value means use the system default. This is all depended on the
underlying LDAP SDK and the system configuration (i.e. /etc/ldap.conf).
"0" means disable referrals chasing.
"1" means enable referrals chasing.

This patch is also exposes the insecure_debug option to help customers and
field engineers to quicky pinpoint the problem when needed.

(cherry picked from commit 0550b00be1fb56aa80faa66739eb6abd82cceaef)

### DIFF
--- a/chef/data_bags/crowbar/migrate/keystone/208_add_ldap_configs_and_insecure_debug.rb
+++ b/chef/data_bags/crowbar/migrate/keystone/208_add_ldap_configs_and_insecure_debug.rb
@@ -1,0 +1,26 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs["ldap"].key?("chase_referrals")
+    attrs["ldap"]["chase_referrals"] = template_attrs["ldap"]["chase_referrals"]
+  end
+  attrs["domain_specific_config"].keys.each do |domain|
+    unless attrs["domain_specific_config"][domain]["ldap"].key?("chase_referrals")
+      attrs["domain_specific_config"][domain]["ldap"]["chase_referrals"] =
+        template_attrs["domain_specific_config"]["ldap_users"]["ldap"]["chase_referrals"]
+    end
+  end
+  attrs["insecure_debug"] = template_attrs["insecure_debug"] unless attrs.key?("insecure_debug")
+
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["ldap"].delete("chase_referrals") unless template_attrs["ldap"].key?("chase_referrals")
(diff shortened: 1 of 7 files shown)


chef/data_bags/crowbar/migrate/keystone/208_add_ldap_configs_and_insecure_debug.rb
```diff
@@ -0,0 +1,26 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  unless attrs["ldap"].key?("chase_referrals")
+    attrs["ldap"]["chase_referrals"] = template_attrs["ldap"]["chase_referrals"]
+  end
+  attrs["domain_specific_config"].keys.each do |domain|
+    unless attrs["domain_specific_config"][domain]["ldap"].key?("chase_referrals")
+      attrs["domain_specific_config"][domain]["ldap"]["chase_referrals"] =
+        template_attrs["domain_specific_config"]["ldap"]["chase_referrals"]
```
cmurphy: `template_attrs["domain_specific_config"]["ldap_users"]["ldap"]["chase_referrals"]`

chef/data_bags/crowbar/migrate/keystone/208_add_ldap_configs_and_insecure_debug.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+  end
+  attrs["insecure_debug"] = template_attrs["insecure_debug"] unless attrs.key?("insecure_debug")
+
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["ldap"].delete("chase_referrals") unless template_attrs["ldap"].key?("chase_referrals")
+  unless template_attrs["domain_specific_config"]["ldap"].key?("chase_referrals")
```
cmurphy: `template_attrs["domain_specific_config"]["ldap_users"]["ldap"].key?("chase_referrals")`

chef/data_bags/crowbar/migrate/keystone/208_add_ldap_configs_and_insecure_debug.rb
guangyee: good catch!

PR #2252, opened 2019-09-19
nova: set default attribute for max_threads_per_process

dirkmueller:
Otherwise there will be a syntax error in the libvirt config
and the cloud is down afterwards.

### DIFF
--- a/chef/cookbooks/nova/attributes/default.rb
+++ b/chef/cookbooks/nova/attributes/default.rb
@@ -71,7 +71,7 @@ default[:nova][:libvirt_type] = "kvm"
 #
 
 default[:nova][:kvm][:ksm_enabled] = false
-
+default[:nova][:kvm][:max_threads_per_process] = 0
 #
 # VMware Settings
 #


jgrassler:
> Otherwise there will be a syntax error in the libvirt config
and the cloud is down afterwards.

Where/how does that happen? The data bag both enforces the presence of that attribute and sets a default after all:

https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/template-nova.schema#L133
https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/template-nova.json#L77

So there must be something subtle that allows this to happen anyway...what is that subtle mechanism?

dirkmueller:
@jgrassler it is using the node injected attributes, but node attribute injection is only done after applying the barclamp in the new schema. until then, the chef attributes matter.

jgrassler:
Ah, there's the subtle mechanism. Thanks :-)

stefannica:
@dirkmueller you need to do something about that commit message though

rsalevsky:
Jira issue should be https://jira.suse.com/browse/SOC-10660

dirkmueller:
@rsalevsky that jira issue has nothing to do with this.

PR #2254, opened 2019-09-19
Revert "designate: Mark as user managed (SOC-10233)"

rsalevsky:
Designate is not ready for customers.

This reverts commit f9a0441889a1e44dbbf19fc01ba82ffc449998c5.


rsalevsky:
@nicolasbock The PO should know. In this case we don't even have a little bit CI for it. So it's in no way ready for customers/production as we can not guarantee it will work with the next MU.

nicolasbock:
Thanks

PR #2255, opened 2019-09-19
tempest: don't rely on service catalogue (SOC-10633)

stefannica:
The way that the tempest chef recipe figures out which OpenStack
services can be tested is by inspecting the openstack service
catalogue to retrieve a list of enabled services, instead of
using the barclamp configuration which is more accurate. Worse
even, it does so in the chef compilation phase, before the
running openstack configuration is updated to reflect changes
in the barclamp configuration.

Sometimes, this results in a tempest configuration that doesn't
reflect the list of services that are actuallly deployed. In some of
the more extreme cases, such as updating the admin password, the
OpenStack CLI command used to retrieve the service catalogue
fails completely and returns an empty list, which results in tempest
disabling all services in its configuration.

Inspecting the barclamp role assignments is a more reliable way of
determining which services are actually enabled.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -73,15 +73,24 @@ comp_environment << "OS_AUTH_URL='#{auth_url}' "
 comp_environment << "OS_IDENTITY_API_VERSION='#{keystone_settings["api_version"]}'"
 openstackcli = "#{comp_environment} openstack --insecure"
 
-# for admin usage (listing the available services)
-adm_environment = "OS_USERNAME='#{keystone_settings["admin_user"]}' "
-adm_environment << "OS_PASSWORD='#{keystone_settings["admin_password"]}' "
-adm_environment << "OS_PROJECT_NAME='#{keystone_settings["admin_project"]}' "
-adm_environment << "OS_AUTH_URL='#{auth_url}' "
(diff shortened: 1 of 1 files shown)


stefannica:
> "metering" => "ceilometer-server" should be "metering": "ceilometer-server" and so on

that would normally work, but in this case I can't use the new syntax because the keys use `-` characters.

chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -73,15 +73,24 @@
...
+  "container-infra" => "magnum-server",
+  "baremetal" => "ironic-server",
+  "logs_v2" => "monasca-server",
+  "logs-search" => "monasca-server",
+  "monitoring" => "monasca-server"
+}
 
-enabled_services = `#{openstackcli_adm} service list -f value -c Type`.split
+enabled_services = service_role_map.reject do |service_name, role_name|
+  search(:node, "roles:#{role_name}").first.nil?
```
guangyee: Any noticeable performance impact here since we are doing node search?

stefannica: @guangyee yes, I expect it will be faster to lookup things in the chef configuration then to talk to OpenStack :). I don't have any hard numbers to show you, but intuitively, interacting with an external system should be slower.

PR #2260, opened 2019-09-20
[5.0] Make ovs of_inactivity_probe configurable from neutron barclamp

dirkmueller:
*We should not be merging changes to attributes in 4.0 without fixing 5.0. *

This patch allows the user to change
the ovs inactivity_probe timeout from the neutron barclamp, in the 'raw'
view. Previously, this value was always set to the OVS default, 5.

It provides crowbar support for this upstream patch:

https://review.opendev.org/#/c/663024/
(cherry picked from commit c0a4d9e49d042c80b118404a3472727c41403145)

### DIFF
--- a/chef/data_bags/crowbar/template-neutron.schema
+++ b/chef/data_bags/crowbar/template-neutron.schema
@@ -48,7 +48,8 @@
                     }},
                     "ovs": { "type": "map", "required": true, "mapping": {
                       "tunnel_csum": { "type": "bool", "required": true },
-                      "ovsdb_interface": { "type": "str", "required": true }
+                      "ovsdb_interface": { "type": "str", "required": true },
+                      "of_inactivity_probe": { "type": "int", "required": true }
                     }},
                     "apic": { "type": "map", "required": true, "mapping": {
                       "hosts": { "type" : "str", "required" : true },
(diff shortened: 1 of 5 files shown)


chef/data_bags/crowbar/template-neutron.schema
```diff
@@ -48,7 +48,8 @@
                     }},
                     "ovs": { "type": "map", "required": true, "mapping": {
                       "tunnel_csum": { "type": "bool", "required": true },
-                      "ovsdb_interface": { "type": "str", "required": true }
+                      "ovsdb_interface": { "type": "str", "required": true },
+                      "of_interface": { "type": "str", "required": true }
```
nicolasbock: That should be `of_inactivity_probe`

dirkmueller:
@nicolasbock fixed thanks

nicolasbock:
Did you add the fix to #2256 accidentally?

dirkmueller:
@nicolasbock yes, fixed.

chef/data_bags/crowbar/template-neutron.schema
dirkmueller: done

PR #2264, opened 2019-09-20
[5.0] tempest: don't rely on service catalogue (SOC-10633)

stefannica:
(backport of #2255)

The way that the tempest chef recipe figures out which OpenStack
services can be tested is by inspecting the openstack service
catalogue to retrieve a list of enabled services, instead of
using the barclamp configuration which is more accurate. Worse
even, it does so in the chef compilation phase, before the
running openstack configuration is updated to reflect changes
in the barclamp configuration.

Sometimes, this results in a tempest configuration that doesn't
reflect the list of services that are actuallly deployed. In some of
the more extreme cases, such as updating the admin password, the
OpenStack CLI command used to retrieve the service catalogue
fails completely and returns an empty list, which results in tempest
disabling all services in its configuration.

Inspecting the barclamp role assignments is a more reliable way of
determining which services are actually enabled.

(cherry picked from commit 82eac9d338bdb337a9fc154a1626ed9baa05befb)

NOTE: updated to include the "alarming" service (aodh), which was not present in cloud9.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -78,15 +78,25 @@ comp_environment << "OS_AUTH_URL='#{auth_url}' "
 comp_environment << "OS_IDENTITY_API_VERSION='#{keystone_settings["api_version"]}'"
 openstackcli = "#{comp_environment} openstack --insecure"
 
-# for admin usage (listing the available services)
-adm_environment = "OS_USERNAME='#{tempest_adm_user}' "
-adm_environment << "OS_PASSWORD='#{tempest_adm_pass}' "
-adm_environment << "OS_PROJECT_NAME='#{tempest_comp_tenant}' "
-adm_environment << "OS_AUTH_URL='#{auth_url}' "
-adm_environment << "OS_IDENTITY_API_VERSION='#{keystone_settings["api_version"]}'"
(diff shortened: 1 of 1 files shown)


chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -78,15 +78,25 @@
...
+  "orchestration" => "heat-server",
+  "dns" => "designate-server",
+  "data-processing" => "sahara-server",
+  "database" => "trove-server",
+  "sharev2" => "manila-server",
+  "container-infra" => "magnum-server",
+  "baremetal" => "ironic-server",
+  "logs_v2" => "monasca-server",
+  "logs-search" => "monasca-server",
+  "monitoring" => "monasca-server",
```
cmurphy: I guess it doesn't hurt to have it here but technically this branch doesn't need the monasca, designate, or ironic stuff.

stefannica: Thanks, that's a good point. The undesired side-effect of using barclamp config to detect enabled services is if you want to backport features from cloud 9, you will most likely miss the fact that you have to manually augment this list to include new services. This is certainly the case with designate, which is already available in cloud 8, but for which the tempest changes haven't been back-ported yet. I kept the other unused values as a "just in case" measure.

PR #2267, opened 2019-09-24
Designate: Filter out the admin node (SOC-10658)

matthewoliver:
The admin node shouldn't be listening on the public network. As such it
can't used as a designate server.

This patch filters it out of the pools list, as crowbar already has a
check to not use the admin node in barclamps that is all we need.

### DIFF
--- a/chef/cookbooks/designate/recipes/mdns.rb
+++ b/chef/cookbooks/designate/recipes/mdns.rb
@@ -19,6 +19,10 @@
 require "yaml"
 
 dns_all = node_search_with_cache("roles:dns-server")
+
+# filter out the crowbar node
+dns_all.select! { |node| node["crowbar"]["admin_node"].nil? }
+
 dnsservers = dns_all.map do |n|
   Chef::Recipe::Barclamp::Inventory.get_network_by_type(n, "admin").address
 end


chef/cookbooks/designate/recipes/mdns.rb
```diff
@@ -19,6 +19,10 @@
 require "yaml"
 
 dns_all = node_search_with_cache("roles:dns-server")
+
+# filter out the crowbar node
+dns_all = dns_all.map { |node| node if node["crowbar"]["admin_node"].nil? }.compact
```
s-t-e-v-e-n-k: dns_all.select!{ |node| node["crowbar"]["admin_node"].nil? }

matthewoliver: oh, that's nice!

matthewoliver:
So doing some tests, the rails check was unneeded as there is already a check that disallows the crowbar node from being used in a barclamp. So all we actually need to do is filter it out of the pools config.

ritesh216:
Looks good concise one liner; looks ready for prime time :)

matthewoliver:
Lol, it's true. After all that, Steve come up with the perfect line that made all mine redundant :)

PR #2271, opened 2019-09-26
Designate: Add dns_domain_ports config (SOC-10740)

ritesh216:
We need to replace 'dns' with 'dns_domain_ports' in ml2.conf extension_drivers.
This adds the capability to assign dns-domain to ports.


nicolasbock:
You could add a note that the `dns` extension is a subset of the  `dns_domain_ports` extension and making this change adds features (DNS on ports) without removing features we were already using. (https://docs.openstack.org/neutron/rocky/admin/config-dns-int.html)

PR #2273, opened 2019-09-26
Designate: Remove enable worker config (SOC-10740)

ritesh216:
Worker service is enabled by default in Rocky for Designate.
This also gets rid of the warning in the logs.


ritesh216:
Closing this pull request. Looks like the warning in the log is misleading; we still need to have worker enabled in the configuration.

PR #2276, opened 2019-09-30
[5.0] Designate: Add dns_domain_ports config (SOC-10740)

ritesh216:
We need replace 'dns' with 'dns_domain_ports' in ml2.conf extension_drivers.
This adds the capability to assign dns-domain to ports.


matthewoliver:
this look good. But is there a reason there are 2 commits? Can these be merged into 1?

ritesh216:
> this look good. But is there a reason there are 2 commits? Can these be merged into 1?

Done, had missed the cherry-pick comment in commit message. Now corrected.

PR #2277, opened 2019-10-01
Fix Cloud 8 no-op migrations (SOC-10623)

jgrassler:
Various Cloud 8 proposal migrations have a bug where they
never add the fields they are supposed to add to the
proposal. This is caused by evaluating the proposal hash key
in question as a Boolean truth value the way it happens in
this example:

  unless a["resource_limits"]

This will return an empty string which evaluates to True in a
boolean context. The correct approach is to use Hash.key?().
This commit fixes all occurences of this problem for Cloud 8.

(cherry picked from commit f479dbe1063817967dbf8938f8d07786f1a05096)

This is a partial (only the Cloud 8 bits) backport of https://github.com/crowbar/crowbar-openstack/pull/2248/

### DIFF
--- a/chef/data_bags/crowbar/migrate/rabbitmq/200_add_resource_limits.rb
+++ b/chef/data_bags/crowbar/migrate/rabbitmq/200_add_resource_limits.rb
@@ -1,5 +1,5 @@
 def upgrade(ta, td, a, d)
-  a["resource_limits"] = ta["resource_limits"] unless a["resource_limits"]
+  a["resource_limits"] = ta["resource_limits"] unless a.key?(["resource_limits"])
   return a, d
 end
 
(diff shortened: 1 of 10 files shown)


chef/data_bags/crowbar/migrate/rabbitmq/200_add_resource_limits.rb
```diff
@@ -1,5 +1,5 @@
 def upgrade(ta, td, a, d)
-  a["resource_limits"] = ta["resource_limits"] unless a["resource_limits"]
+  a["resource_limits"] = ta["resource_limits"] unless a.key?(["resource_limits"])
```
skazi0: too many brackets?

PR #2278, opened 2019-10-01
tempest:install designate tempest plugin for SOC8 (SOC-10288)

gosipyan:
python-designate-tempest-plugin need to be installed on SOC8


gosipyan:
@s-t-e-v-e-n-k  yes  https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/tempest/recipes/install.rb#L53

PR #2282, opened 2019-10-07
Handle no dns-server in tempest config (SOC-10856)

ritesh216:
The config recipe in tempest assumes there is always a node
running dns-server role. This may not be true.

The patch handles such case and uses search instead of
node_search_with_cache.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -539,11 +539,15 @@ tempest_roles = ["member"]
 barbicans = search(:node, "roles:barbican-controller") || []
 tempest_roles += ["creator"] unless barbicans.empty?
 
-dns_server_node = node_search_with_cache("roles:dns-server").first
-dns_server_node_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(
-  dns_server_node,
-  "admin"
-).address
+dns_server_nodes = search(:node, "roles:dns-server") || []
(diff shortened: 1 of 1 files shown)


chef/cookbooks/tempest/recipes/config.rb
```diff
@@ -539,11 +539,15 @@
 barbicans = search(:node, "roles:barbican-controller") || []
 tempest_roles += ["creator"] unless barbicans.empty?
 
-dns_server_node = node_search_with_cache("roles:dns-server").first
-dns_server_node_ip = Chef::Recipe::Barclamp::Inventory.get_network_by_type(
-  dns_server_node,
-  "admin"
-).address
+dns_server_nodes = search(:node, "roles:dns-server") || []
```
ritesh216: Need to understand why this search works and node_search_with_cache does not work. This is seen only when Designate is enabled. Please comment if you know, I am a bit puzzled.

matthewoliver: We may need to go delve into the code, or add some logging to peer into the depths of crowbar. Is it finding something in the cache so never goes bother searching.. maybe what it finds isn't something we expect. Like doesn't match a false value, but isn't what we want?? 

NOTE: I haven't looked yet, just speculating.

nicolasbock: I'll set up a test environment and see whether I can understand this.

nicolasbock:
Closing

PR #2280, opened 2019-10-07
[5.0] Fix nova tempest tests (SOC-9298, SOC-10844)

rsalevsky:
- disable `swap_volume` tests when cinder is using ceph backend as it is
not supported.
- replace the uec image used in tempest for a qcow2 image

(cherry picked from commit 07864bdff573b4ad150530168f879e2e215c896b)

PR #2283, opened 2019-10-08
[5.0] glance: Set barbican auth endpoint (bsc#1123191)

flaviodsr:
By default, the castellan key manager library uses
'http://localhost/identity/v3' as its keystone auth endpoint (works in
devstack!). When barbican uploads a signed image to glance, if glance
doesn't override the auth endpoint it will fail to use the key manager.
The symptom of this is a tempest error that looks like:

  TypeError: sendall() argument 1 must be string or buffer, not callable-iterator

which arises from a combination of barbican, glance, and tempest
handling the authentication error very very poorly.

(cherry picked from commit cf46a4461f08234daf06d7ff77e5ea7196e0078e)


rsalevsky:
Merged into https://github.com/crowbar/crowbar-openstack/pull/2280

PR #2285, opened 2019-10-14
mysql: stop service for mysql_install_db (SOC-10717)

jgrassler:
mysql_install_db will attempt to acquire various exclusive
locks on files in /var/lib/mysql before it bootstraps the
database. If the mysql service is already running, it will
hold these same locks, thus preventing mysql_install_db from
working.

This situation can occur if the critical section guarded by
node[:database][:galera_bootstrapped] fails between start and
stop of the temporary mysqld server. This commonly happens if
WSREP replication fails during user creation. In this case,
subsequent barclamp runs will fail due to mysql_install_db
being unable to acquire the locks the leftover mysqld holds.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -41,10 +41,26 @@ node_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admi
 
 unless node[:database][:galera_bootstrapped]
   if CrowbarPacemakerHelper.is_cluster_founder?(node)
+    case node[:platform_family]
+    when "rhel", "fedora"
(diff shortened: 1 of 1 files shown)


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -41,10 +41,26 @@
 
 unless node[:database][:galera_bootstrapped]
   if CrowbarPacemakerHelper.is_cluster_founder?(node)
+    case node[:platform_family]
+    when "rhel", "fedora"
+      mysql_service_name = "mysqld"
+    else
+      mysql_service_name = "mysql"
+    end
```
jsuchome: I know it's not related to this PR, but we could probably stop assuming rhel family will ever use this recipe...

PR #2288, opened 2019-10-17
mysql: fix WSREP sync race (SOC-10717)

jgrassler:
The single node Galera cluster the barclamp uses for
bootstrapping may take a few seconds to report WSREP sync. If
the barclamp creates users before that point in time, user
creation will fail with

  "WSREP replication failed. Check your wsrep connection state
  and retry the query."

This commit makes the barclamp wait for the single-node
bootstrapping cluster to "sync" to prevent this race condition
from happening.

### DIFF
--- a/chef/cookbooks/mysql/recipes/ha_galera.rb
+++ b/chef/cookbooks/mysql/recipes/ha_galera.rb
@@ -39,6 +39,12 @@ end
 
 node_address = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
 
+# This is needed because the single node bootstrap cluster will take a
+# while until it reports "Sync". If that takes too long, user creation will
+# fail.
+early_sync_retry_delay = 10
+early_sync_retries = 2
+
 unless node[:database][:galera_bootstrapped]
   if CrowbarPacemakerHelper.is_cluster_founder?(node)
 
@@ -91,6 +97,14 @@ unless node[:database][:galera_bootstrapped]
       service_name mysql_service_name
       supports status: true, restart: true, reload: true
       action :start
+    end
+
+    execute "Test WSREP state early" do
+      retries early_sync_retries
+      retry_delay early_sync_retry_delay
+      command "mysql -u 'monitoring' -N -B " \
+        "-e \"SHOW STATUS WHERE Variable_name='wsrep_local_state_comment';\" | grep Synced"
+      action :run
     end
 
     database_user "create state snapshot transfer user" do
@@ -183,7 +197,7 @@ end
 # all the required packages and configurations installed before we create the
 # pacemaker resources
 crowbar_pacemaker_sync_mark "sync-database_before_ha" do
-  timeout node[:database][:mysql][:presync_timeout]
+  timeout node[:database][:mysql][:presync_timeout] + (early_sync_retry_delay * early_sync_retries)
   revision node[:database]["crowbar-revision"]
 end
 


chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -93,13 +93,26 @@
...
 
+    # This is needed because the single node bootstrap cluster will take a
+    # while until it reports "Sync". If that takes too long, user creation will
+    # fail.
+    execute "Test WSREP state early" do
+      retries 2
+      retry_delay 60
+      command "mysql -u 'monitoring' -N -B " \
+        "-e \"SHOW STATUS WHERE Variable_name='wsrep_local_state_comment';\" | cut -f 2 |
+grep Synced"
```
jsuchome: As you are grepping, why do you need `cut` before?

jsuchome:
Won't the new block increase the whole bootstrapping part, thus making the whole part timeout somewhere?
In other words, don't you need to raise some timeout on non-founder nodes?

chef/cookbooks/mysql/recipes/ha_galera.rb
jgrassler: Oh...Copy and past artifact from where i got this: https://github.com/crowbar/crowbar-openstack/blob/277923b892dfa9134e2cc1941592547a68bdc8c7/chef/cookbooks/mysql/recipes/ha_galera.rb#L273 - absolutely not needed. Removed.

jgrassler:
> Won't the new block increase the whole bootstrapping part, thus making the whole part timeout somewhere?
> In other words, don't you need to raise some timeout on non-founder nodes?

[Sorry for editing your comment - didn't know that was even possible]

Good point. I added some code to account for that to the first sync mark's timeout.

chef/cookbooks/mysql/recipes/ha_galera.rb
```diff
@@ -93,13 +93,28 @@
...
+    end
+
     database_user "create state snapshot transfer user" do
       connection db_connection
       username "sstuser"
       password node[:database][:mysql][:sstuser_password]
       host "localhost"
       provider db_settings[:user_provider]
       action :create
+      subscribes :create, "execute[Test WSREP state early]"
```
jsuchome: Why `subscribes` here? You run "Test WSREP state early" always during the bootstrap. And `database_user` has also `action :create` already so it will be run anyway...

jgrassler: Resource ordering...sometimes chef defers exec resources and only runs them after all the other resources. I'm not entirely sure why it happens (it took me quite a while to get everything to execute in the desired sequence in https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/horizon/recipes/monasca_ui.rb - ever since I'm paranoid about this). This would be a problem here, so I added the `subscribes` to make sure the execute does not get deferred.

jsuchome: OK. But then maybe there is no need for `action :create` and you can use `:nothing` instead?
Although it probably does not matter.

I admit my chef skills got rusty

rhafer: > ..sometimes chef defers exec resources and only runs them after all the other resources. I'm not entirely sure why it happens 

Hm, this sounds a bit like an urban legend to me. Unless the resource itself is triggered by some (e.g. delayed) notification or resources being executed in the compile phrase (i.e. using `run_action`)  I don't think I have ever seen resources being executed in a different order than defined in the recipe.

> But then maybe there is no need for action :create and you can use :nothing instead?

Right. You either want `action: nothing` and `notify` or `subscribe` here. Or `action: create` without any notification. I am not entirely sure what chef's behavior is with `action: create` and `subscribe` it might be trying to prevent executing the same resource twice.

jgrassler:
The code as it looks now (at the time of the most recent force push) has been tested on a Cloud 8 hardware cloud and the database barclamp deployed fine. So as far as I'm concerned this is ready.

PR #2290, opened 2019-10-18
[4.0] Do not use obsoleted --endpoint-type option with CLI

jsuchome:
Correct option is --os-endpoint-type which by default uses
the value of OS_ENDPOINT_TYPE env var.

### DIFF
--- a/chef/cookbooks/neutron/recipes/post_install_conf.rb
+++ b/chef/cookbooks/neutron/recipes/post_install_conf.rb
@@ -77,7 +77,7 @@ env << "OS_INTERFACE=internal "
 env << "OS_USER_DOMAIN_NAME=Default "
 env << "OS_PROJECT_DOMAIN_NAME=Default "
 env << "OS_IDENTITY_API_VERSION=3"
-neutron_args = "--endpoint-type internalURL"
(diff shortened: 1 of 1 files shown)


jsuchome:
TBH this does not cause any errors unless you have misconfigured public endpoint (which is actually used here instead of internal one when the wrong option is used)

jsuchome:
This is no port, but direct fix for SOC7.

I'm not sure about SOC8: the same code is currently not using endpoint-type at all so it is not affected by obsoleted option. So I understand SOC8 is using public endpoint by default. I can add the option to SOC8 codebase if that's what you mean, @dirkmueller

skazi0:
@jsuchome for SOC8+ we switched to `openstack` client which uses `OS_INTERFACE` instead of `OS_ENDPOINT_TYPE` but later there were some new calls added which used `neutron` cli for operations which were not ported to `openstack`.
I guess we still need `OS_ENDPOINT_TYPE=internalURL` for those calls in SOC8+.

See: 
https://github.com/crowbar/crowbar-openstack/pull/2095
https://github.com/crowbar/crowbar-openstack/pull/1223

jsuchome:
OK, I'll prepare SOC8 version as well

PR #2291, opened 2019-10-18
[5.0] Set OS_ENDPOINT_TYPE to be used with CLI

jsuchome:
Correct option is --os-endpoint-type which by default uses
the value of OS_ENDPOINT_TYPE env var.

(cherry picked from commit 87ad941d55a348b1d83a50fde25e29db40beb559)

port of https://github.com/crowbar/crowbar-openstack/pull/2290


chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -77,6 +77,7 @@ def mask_to_bits(mask)
 env << "OS_USER_DOMAIN_NAME=Default "
 env << "OS_PROJECT_DOMAIN_NAME=Default "
 env << "OS_IDENTITY_API_VERSION=3"
```
toabctl: You need to add a space at the end here. CI is failing because of the missing space

jsuchome: Oh, thanks for noticing!

PR #2294, opened 2019-10-21
Use correct admin password for openstack commands

nicolasbock:
During a password update for the `admin` user the Designate barclamp
is compiled before the password has been updated. We therefore need to
use the old admin password.

Co-Authored-By: Gayane Osipyan <gosipyan@suse.com>

### DIFF
--- a/chef/cookbooks/designate/recipes/common.rb
+++ b/chef/cookbooks/designate/recipes/common.rb
@@ -36,9 +36,15 @@ api_protocol = node[:designate][:api][:protocol]
 
 resource_project_id = ""
 keystone_settings = KeystoneHelper.keystone_settings(node, :designate)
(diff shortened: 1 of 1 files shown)


jeremy-moffitt:
updated screenshot:
amphora_up

jsuchome:
This looks suspicious. Why aren't other barclamps affected?

nicolasbock:
Hmm, good point. I'll go through the other barclamps to check. Thanks!

jsuchome:
I mean ... it seems that no other barclamp has check like this - besides keystone, obviously. So is there some dependency problem with designate that it is run before keystone?

stefannica:
Should be closed in favor of #2297

nicolasbock:
Closing in favor of #2297

PR #2296, opened 2019-10-23
Fix the migration that tried to access Array as a Hash

jsuchome:
Commit that introduces the problem: cb5347d6a47565c2ef1aebf10ef80d980114e046

This tries to fix the problem introduced with https://github.com/crowbar/crowbar-openstack/pull/2218

~However it's  possibly not good enough as it only checks first entry in the list...~


jsuchome:
I did not test it yet, neither in SOC9 setup nor in upgrade. Force push is just when I found better solution than the preliminary one I had before.

nicolasbock:
Would be so nice to have unit tests in Chef :smile:

jsuchome:
OK, so basic test works: with the patch applied, migration does not produce any errors. And the new keys are present in `neutron-config-default` role. Is that enough?

PR #2297, opened 2019-10-25
designate: move keystone resource lookup to convergence (SOC-10887)

guangyee:
To avoid weird problems such as race conditions due to
changing Keystone endpoint protocols, admin password, etc.

### DIFF
--- a/chef/cookbooks/designate/recipes/common.rb
+++ b/chef/cookbooks/designate/recipes/common.rb
@@ -36,21 +36,23 @@ api_protocol = node[:designate][:api][:protocol]
 
 resource_project_id = ""
 keystone_settings = KeystoneHelper.keystone_settings(node, :designate)
-env = {
-  "OS_USERNAME" => keystone_settings["admin_user"],
-  "OS_PASSWORD" => keystone_settings["admin_password"],
-  "OS_PROJECT_NAME" => keystone_settings["admin_tenant"],
-  "OS_AUTH_URL" => keystone_settings["internal_auth_url"],
-  "OS_IDENTITY_API_VERSION" => "3"
-}
 
 if node["roles"].include?("designate-server")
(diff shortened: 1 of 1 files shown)


chef/cookbooks/designate/recipes/common.rb
```diff
@@ -36,21 +36,23 @@
 
 resource_project_id = ""
 keystone_settings = KeystoneHelper.keystone_settings(node, :designate)
-env = {
```
stefannica: This is just to confirm something I had some doubts about earlier: it wasn't also necessary to also move this variable into the ruby block, running the openstack command there is enough.

stefannica:
@guangyee by the way, you need a Jira ticket number to make Travis happy. Could you please create one to track this issue ?

chef/cookbooks/designate/recipes/common.rb
nicolasbock: I don't think I understand what you are trying to say :smile:

guangyee:
> @guangyee by the way, you need a Jira ticket number to make Travis happy. Could you please create one to track this issue ?

Can you reuse SOC-10887?

nicolasbock:
@gosipyan Could you check this PR?

guangyee:
> i have triggered job and apply patch manually it failed during test setup
> https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/526/pipeline/139/

Looks like the dns-server role need to be added to the controller as well.

https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/designate/recipes/mdns.rb#L21-#L28
Otherwise, we don't have any backend configured.

@gosipyan can you please reapply the DNS barclamp with dns-server added to the controller?

PR #2303, opened 2019-11-11
octavia: add SSL support to octavia-api (SOC-10906)

stefannica:
The Octavia API service doesn't have intrinsic support for
terminating SSL traffic. The alternatives are to terminate
SSL traffic at haproxy level (which isn't done anywhere else
in Crowbar), or to deploy the API service as an Apacke WSGI
application. This PR implements the latter and it also solves
the problem described in the following paragraphs.

The octavia system group and user are created when the
openstack-octavia package is installed on the controller nodes.
This happens automatically when the octavia barclamp is
applied.
However, setting up the paths for the certificates required for
the two-way SSL amphorae communication needs to be done in advance,
before applying the octavia barclamp, and those paths need to have
their ownership set to the octavia system user/group, which aren't
yet available at that time.

This change proposes a solution to this problem, where the admin
can set up those certificate paths on the controller nodes with
root (or other user) as owner, while the Octavia barclamp is
responsible for switching ownership over to the octavia user/group
during application.

This PR is complemented by a [corresponding automation PR](https://github.com/SUSE-Cloud/automation/pull/3767).

### DIFF
--- a/chef/cookbooks/octavia/recipes/api.rb
+++ b/chef/cookbooks/octavia/recipes/api.rb
@@ -13,6 +13,9 @@
 # limitations under the License.
 #
 
+include_recipe "apache2"
+include_recipe "apache2::mod_wsgi"
+
 octavia_config = Barclamp::Config.load("openstack", "octavia")
 cmd = OctaviaHelper.get_openstack_command(node, octavia_config)
 
@@ -24,4 +27,19 @@ package "openstack-octavia-api" if ["rhel", "suse"].include? node[:platform_fami
 
 include_recipe "#{@cookbook_name}::database"
 
-octavia_service "api"
+crowbar_openstack_wsgi "WSGI entry for octavia-api" do
+  bind_host OctaviaHelper.bind_host(node, "api")
+  bind_port OctaviaHelper.bind_port(node, "api")
+  daemon_process "octavia-api"
+  script_alias "/usr/bin/octavia-wsgi"
+  user node[:octavia][:user]
+  group node[:octavia][:group]
+  ssl_enable node[:octavia][:api][:protocol] == "https"
+  ssl_certfile node[:octavia][:ssl][:certfile]
+  ssl_keyfile node[:octavia][:ssl][:keyfile]
+  ssl_cacert node[:octavia][:ssl][:ca_certs] if node[:octavia][:ssl][:cert_required]
+end
+
+apache_site "octavia-api.conf" do
+  enable true
+end
--- a/chef/cookbooks/octavia/recipes/common.rb
+++ b/chef/cookbooks/octavia/recipes/common.rb
@@ -17,3 +17,9 @@ package "python-octaviaclient"
 package "octavia" do
   package_name "openstack-octavia" if %w[rhel suse].include?(node[:platform_family])
 end
+
+execute "chown certs path" do
+  command "chown -R #{node[:octavia][:user]}:#{node[:octavia][:group]} \
+    #{node[:octavia][:certs][:cert_path]}"
+  action :run
+end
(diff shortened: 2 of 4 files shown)


stefannica:
Tested in the unified CI with: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/561/pipeline

chef/cookbooks/octavia/recipes/api.rb
```diff
@@ -24,4 +27,19 @@
...
-octavia_service "api"
+crowbar_openstack_wsgi "WSGI entry for octavia-api" do
+  bind_host OctaviaHelper.bind_host(node, "api")
+  bind_port OctaviaHelper.bind_port(node, "api")
+  daemon_process "octavia-api"
+  script_alias "/usr/bin/octavia-wsgi"
+  user node[:octavia][:user]
+  group node[:octavia][:group]
+  ssl_enable node[:octavia][:api][:protocol] == "https"
+  ssl_certfile node[:octavia][:ssl][:certfile]
```
guangyee: We don't need to set ssl_certfile, ssl_keyfile, and ssl_cacert if ssl_enable is false.

chef/cookbooks/octavia/recipes/api.rb
```diff
@@ -24,4 +27,19 @@
 
 include_recipe "#{@cookbook_name}::database"
 
-octavia_service "api"
+crowbar_openstack_wsgi "WSGI entry for octavia-api" do
+  bind_host OctaviaHelper.bind_host(node, "api")
+  bind_port OctaviaHelper.bind_port(node, "api")
```
guangyee: Do we differentiate between SSL and non-SSL port?

stefannica:
> Tested in the unified CI with: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/561/pipeline

Re-testing after fixing some bugs with: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/565/pipeline

chef/cookbooks/octavia/recipes/api.rb
stefannica: True, we don't need to, but we don't have to exclude them either, which is what all the other barclamps are doing. I'm just keeping things consistent :)

chef/cookbooks/octavia/recipes/api.rb
stefannica: No, but we differentiate between HA port and non-HA port. That's what these functions are meant to do. These are the same functions that are currently used to populate the host/port in the octavia-api service configuration file.

chef/cookbooks/octavia/recipes/common.rb
```diff
@@ -17,3 +17,9 @@
 package "octavia" do
   package_name "openstack-octavia" if %w[rhel suse].include?(node[:platform_family])
 end
+
+execute "chown certs path" do
+  command "chown -R #{node[:octavia][:user]}:#{node[:octavia][:group]} \
```
nicolasbock: Will this not fail when the user has not created `cert_path`?

stefannica: Yes, it will fail, but I actually prefer it that way. This is better than having the Octavia barclamp deployment pass but not having any certificates installed for Octavia, which is far more difficult to debug.

nicolasbock: It would be even better if we had a validator in the barclamp though. Then the user would get direct feedback on the missing certs.

nicolasbock: Ok, strike that. I don't think what I was thinking of is going to work.

nicolasbock:
If @KeithMnemonic is happy we can merge this.

PR #2306, opened 2019-11-15
designate: declare all mdns servers as master on pool config (SOC-10952)

flaviodsr:
When creating zones the DNS server notify other servers about the new
zone, however the notification is only accepted when it comes from
masters.

Using the VIP as master causes the zone creation to fail as the
notification is refused because it does not come from the VIP, but from
the node IP address. This change fixes it by declaring all mdns server
as master on the default pool defined by crowbar.


flaviodsr:
@nicolasbock I did test powering off one of the controllers, however designate still show the service as UP and consequently failing on some tests. This is the case with/without this change and I also see the same behavior on Ardana.

PR #2307, opened 2019-11-19
horizon: install lbaas horizon dashboard (SOC-10883)

stefannica:
Add the LBaaS dashboard to the UI if lbaas is enabled in the
neutron barclamp.

### DIFF
--- a/chef/cookbooks/horizon/recipes/server.rb
+++ b/chef/cookbooks/horizon/recipes/server.rb
@@ -103,6 +103,16 @@ unless neutron_fwaas_ui_pkgname.nil?
   end
 end
 
+# install horizon neutron lbaas plugin if needed
+neutron_server = node_search_with_cache("roles:neutron-server").first
+unless neutron_server.nil?
+  package "openstack-horizon-plugin-neutron-lbaas-ui" do
+    action :install
+    notifies :reload, "service[horizon]"
+    only_if { neutron_server[:neutron][:use_lbaas] }
+  end
+end
+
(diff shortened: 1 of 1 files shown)


chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -103,6 +103,26 @@
   end
 end
 
+# install horizon neutron lbaas plugin if needed
+neutron_lbaas_ui_pkgname =
+  case node[:platform_family]
+  when "suse"
+    "openstack-horizon-plugin-neutron-lbaas-ui"
+  when "rhel"
```
jsuchome: Does the separate branch for RHEL package make any sense? I know we have it elsewhere, but ...

jsuchome: I don't think so. And AFAIK it never ran there. We were just trying to be good open source citizens. Now it matters even less, that's why I wouldn't even add such new code.... but it does not hurt, so no need to rebase it just because of this

stefannica: I removed the rhel section :-)

stefannica:
I fixed the hound complaints and I also removed any Octavia references from this PR, because this should be just about adding the lbaas dashboard to horizon. I will address Octavia related logic in another PR, where I will also address adding the Octavia dashboard.

stefannica:
Adding the Octavia horizon dashboard with all its related concerns is handled in #2311

PR #2308, opened 2019-11-19
designate: add support for SSL (SOC-10877)

flaviodsr:
This change adds support for configuring designate with SSL enabled.


flaviodsr:
@nicolasbock yes, I will do that once I have this tested.

flaviodsr:
Tested in: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/576/pipeline/

chef/cookbooks/horizon/recipes/server.rb
```diff
@@ -432,7 +433,7 @@
     || trove_insecure \
     || sahara_insecure \
     || manila_insecure \
-    || ceilometer_insecure,
```
nicolasbock: Why are we dropping ceilometer here?

flaviodsr: `ceilometer_insecure` is on that list two times. I am removing one of them.

nicolasbock: Ah thanks, I didn't notice.

PR #2313, opened 2019-11-20
octavia: Update configuration parameters (SOC-10904)

nicolasbock:
The current logic in the Octavia barclamp does not detect cases where
the IDs of Octavia related resources has changed, but the names of
those resources hasn't. This happens for instance, if the user
deletes the management network for some reason.

### DIFF
--- a/chef/cookbooks/octavia/definitions/octavia_conf.rb
+++ b/chef/cookbooks/octavia/definitions/octavia_conf.rb
@@ -16,28 +16,20 @@
 define :octavia_conf do
   amphora = node[:octavia][:amphora]
   net_name = amphora[:manage_net]
-  old_amp = if node[:octavia][:old_amphora].nil?
-    {}
-  else
-    node[:octavia][:old_amphora]
-  end
 
   ruby_block "Check for amphora changes" do
     block do
-      if node[:octavia][:sec_group_id].nil? || amphora[:sec_group] != old_amp[:sec_group]
-        node.set[:octavia][:sec_group_id] = shell_out("#{params[:cmd]} security group show "\
-                                                      "#{amphora[:sec_group]} "\
-                                                      "-f value -c id").stdout.delete "\n"
-      end
-      if node[:octavia][:flavor_id].nil? || amphora[:flavor] != old_amp[:flavor]
-        node.set[:octavia][:flavor_id] =  shell_out("#{params[:cmd]} flavor show "\
-                                                    "#{amphora[:flavor]}" \
-                                                    " -f value -c id").stdout.delete "\n"
-      end
-      if node[:octavia][:net_id].nil? || net_name != old_amp[:manage_net]
-        node.set[:octavia][:net_id] = shell_out("#{params[:cmd]} network show #{net_name} "\
-                                                "-f value -c id").stdout.delete "\n"
-      end
+      sec_group_id = shell_out("#{params[:cmd]} security group show " \
+                               "#{amphora[:sec_group]} -f value -c id").stdout.chomp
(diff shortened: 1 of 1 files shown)


chef/cookbooks/octavia/definitions/octavia_conf.rb
```diff
@@ -16,27 +16,25 @@
 define :octavia_conf do
   amphora = node[:octavia][:amphora]
   net_name = amphora[:manage_net]
-  old_amp = if node[:octavia][:old_amphora].nil?
-    {}
-  else
-    node[:octavia][:old_amphora]
-  end
+
+  sec_group_id = shell_out("#{params[:cmd]} security group show " \
```
stefannica: If you're running OpenStack commands outside of a ruby_block, you'll run into failures when the admin password is changed or the SSL/no-SSL switch is exercised by the CI. The reason for this is that these commands will run in the chef compilation phase, when the configuration updates are not yet applied to the affected services, instead of running in the chef execution phase. See [here](https://coderanger.net/two-pass/) for more details on that interesting subject :).

nicolasbock: Right, good point. I moved the commands into the ruby block.

nicolasbock:
Another review @stefannica ? Thanks!

chef/cookbooks/octavia/definitions/octavia_conf.rb
```diff
@@ -16,27 +16,25 @@
...
-  end
 
   ruby_block "Check for amphora changes" do
     block do
-      if node[:octavia][:sec_group_id].nil? || amphora[:sec_group] != old_amp[:sec_group]
-        node.set[:octavia][:sec_group_id] = shell_out("#{params[:cmd]} security group show "\
-                                                      "#{amphora[:sec_group]} "\
-                                                      "-f value -c id").stdout.delete "\n"
+      sec_group_id = shell_out("#{params[:cmd]} security group show " \
+                               "#{amphora[:sec_group]} -f value -c id").stdout.chomp
```
stefannica: This is a lot better, but there's still one thing I'm worried about: if something happens and this command fails for any reason, the option in the configuration file will be set to an empty value and that value may or may not cause the service to crash after restart. Wouldn't it make sense to fail the execution here, before that happens ?

nicolasbock: Good point. I fixed it now. Please have another look. Thanks!

nicolasbock:
Please review again @stefannica . Thanks!

chef/cookbooks/octavia/definitions/octavia_conf.rb
```diff
@@ -16,27 +16,28 @@
...
+      flavor_id = shell_out("#{params[:cmd]} flavor show "\
+                            "#{amphora[:flavor]} -f value -c id").stdout.chomp
+      mgmt_net_id = shell_out("#{params[:cmd]} network list " \
+                              "--format value --column ID --column Name " \
+                              "| grep #{net_name} | cut -d ' ' -f 1").stdout.chomp
+
+      if !sec_group_id.empty? && \
+          (node[:octavia][:sec_group_id].nil? || node[:octavia][:sec_group_id] != sec_group_id)
+        node.set[:octavia][:sec_group_id] = sec_group_id
```
stefannica: Maybe it's just me, but I think the second part of this condition is redundant. You could get away with just:
```
      if !sec_group_id.empty?
        node.set[:octavia][:sec_group_id] = sec_group_id
```

PR #2315, opened 2019-11-21
[5.0] designate backports

gosipyan:
Besides bringing the configuration closer to the ardana configuration,
those changes also gets all designate tempest tests to pass.

The configuration changes are:
  - Increase the default quota for zones from 10 to 100
  - Enable admin api
  - Enable `quotas` and `zones` admin extensions
  - Enable all extensions from v2 api

(cherry picked from commit 48dfc155dc4847979652e9e7ce869258e7416a42)

designate: add support for SSL (SOC-10877)

This change adds support for configuring designate with SSL.

(cherry picked from commit c1bf75c)

designate: declare all mdns servers as master on pool config (SOC-10952)

When creating zones the DNS server notify other servers about the new
zone, however the notification is only accepted when it comes from
masters

Using the VIP as master causes the zone creation to fail as the
notification is refused because it does not come from the VIP, but from
the node IP address. This change fixes it by declaring all mdns server
as master on the default pool defined by crowbar.

(cherry picked from commit 92848ad)

### DIFF
--- a/chef/cookbooks/designate/attributes/default.rb
+++ b/chef/cookbooks/designate/attributes/default.rb
@@ -26,6 +26,13 @@ default[:designate][:ha][:enabled] = false
 # Ports to bind to when haproxy is used for the real ports
 default[:designate][:ha][:ports][:api_port] = 5574
 
+default[:designate][:ssl][:certfile] = "/etc/designate/ssl/certs/signing_cert.pem"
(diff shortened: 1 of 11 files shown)


gosipyan:
job green all tests passed https://ci.suse.de/job/openstack-crowbar/601/

gosipyan:
need https://github.com/crowbar/crowbar-core/pull/1940  merged

PR #2318, opened 2019-11-25
Octavia: Create listener port for health manager (SOC-10904)

nicolasbock:
Create a port that will be bound so that the health manager can
communicate with the `lb-mgmt-net` and the amphorae.

### DIFF
--- a/chef/cookbooks/octavia/recipes/nova.rb
+++ b/chef/cookbooks/octavia/recipes/nova.rb
@@ -89,6 +89,13 @@ execute "create_amphora_image" do
   action :run
 end
 
+# execute "create octavia user key" do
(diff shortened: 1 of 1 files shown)


rhafer:
I really think messing around with the `neutron-(openvswitch|linuxbridge)-agent` managed bridges and ports behind the agent's back is a very bad idea. It will be a nightmare to maintain and debug. It might be ok for a devstack like use case but I don't think this should be done in production. Please don't go down this path.

nicolasbock:
@rhafer Thanks for the comment! This approach does feel very hackish, I completely agree. I'll think about this some more and see if I can find another approach that achieves the same through more "official" channels :smile:

nicolasbock:
Abandoning as per @rhafer 's comment. Not the right path forward...

PR #2320, opened 2019-12-03
octavia: configuration fixes (SOC-10989,SOC-10990)

stefannica:
This PR fixes some Octavia configuration options, either deprecated or missing.

Remaining Octavia deprecated configuration options:
 - disables deprecated V1 API [1]
 - use the service_token_roles_required and service_token_roles
 parameters in the keystone auth config. Adding this will ensure that
 token validation sent to keystone by Octavia service has the configured
 service token role to pass otherwise keystone will reject that request [2]

Even though Octavia uses the service catalogue to auto-configure
the Barbican service endpoint, it helps if we use an endpoint
hard-coded in the configuration files because there is no other
way to restart Octavia services in case the Barbican endpoint
also changes (e.g. switch from HTTP to HTTPs or vice-versa).

This commit also sets the insecure option not just for Barbican,
but also for other services that Octavia interacts with (Nova
and Neutron).

[1] https://wiki.openstack.org/wiki/Neutron/LBaaS/Deprecation#What_about_the_Octavia_v1_API.3F
[2] https://docs.openstack.org/releasenotes/keystonemiddleware/ocata.html

### DIFF
--- a/chef/cookbooks/octavia/definitions/octavia_conf.rb
+++ b/chef/cookbooks/octavia/definitions/octavia_conf.rb
@@ -56,6 +56,7 @@ define :octavia_conf do
             octavia_db_connection: fetch_database_connection_string(node[:octavia][:db]),
             neutron_endpoint: OctaviaHelper.get_neutron_endpoint(node),
             nova_endpoint: OctaviaHelper.get_nova_endpoint(node),
+            barbican_endpoint: OctaviaHelper.get_barbican_endpoint(node),
             octavia_keystone_settings: KeystoneHelper.keystone_settings(node, "octavia"),
             rabbit_settings: fetch_rabbitmq_settings,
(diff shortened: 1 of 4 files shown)


stefannica:
Tested with: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/637/pipeline

stefannica:
Testing with: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/637/pipeline

stefannica:
Also verified manually that TLS load-balancers are still functional after this change (procedure documented [here](https://jira.suse.com/browse/SOC-10989?focusedCommentId=973473&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-973473)).

PR #2323, opened 2019-12-10
keystone: Add OS_INTERFACE env var to .openrc (SOC-11006)

skazi0:
New openstack client(s) use OS_INTERFACE instead of OS_ENDPOINT_TYPE.
The default is 'public' so if we don't set it, clients will use mix of
internal and public calls depending on the implementation.


bbobrov:
i would just remove `keystone` from the commit message, because nothing about keystone is changed here. `cli` would be a better keyword, or nothing at all

PR #2327, opened 2019-12-12
tempest: Update default image on schema (SOC-11023)

flaviodsr:
This change adds a schema migrate function for updating the default
tempest image which has been changed in [1].

1. https://github.com/crowbar/crowbar-openstack/commit/07864bdff573b4ad150530168f879e2e215c896b

### DIFF
--- a/chef/data_bags/crowbar/migrate/tempest/302_update_tempest_image.rb
+++ b/chef/data_bags/crowbar/migrate/tempest/302_update_tempest_image.rb
@@ -1,0 +1,10 @@
+def upgrade(ta, td, a, d)
+  a["tempest_test_images"]["x86_64"] = ta["tempest_test_images"]["x86_64"]
+  return a, d
+end
+
+def downgrade(ta, td, a, d)
+  a["tempest_test_images"]["x86_64"] = ta["tempest_test_images"]["x86_64"]
+  return a, d
+end
+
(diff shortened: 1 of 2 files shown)


chef/data_bags/crowbar/migrate/tempest/302_update_tempest_image.rb
```diff
@@ -0,0 +1,4 @@
+def upgrade(ta, td, a, d)
+  a["tempest_test_images"]["x86_64"] = ta["tempest_test_images"]["x86_64"]
+  return a, d
+end
```
nicolasbock: The downgrade is missing. Is that intentional?

stefannica: @nicolasbock this is a very particular case for which I don't think there's any precedent: we're changing the default value of a parameter. In this case, the downgrade functionality doesn't give you enough parameters to figure out if the current attribute value is the default one (the "current" template is missing). For that reason, and also because we're only touching the tempest barclamp, I suggested to Flavio that we could just skip the downgrade piece. If you have any suggestion on how this can be done though, we could use some.

flaviodsr: ok. I did add a downgrade function and it does the same, just set the value to what is on the template.

nicolasbock: Thanks!

PR #2331, opened 2019-12-13
octavia: Add ssh key to health manager (SOC-11025)

nicolasbock:
The amphora ssh key needs to be known by the health manager so that it
can respawn dead amphora with the correct ssh key.

Co-Authored-By: Keith Berger <keith.berger@suse.com>


nicolasbock:
I looked through the Octavia code @KeithMnemonic and the reason why we need to set this in the health manager is that it doesn't call the worker to create the new amphora, but includes the the worker code directly. This means that the worker code inherits the configuration options read by the health manager and we need to set the ssh key separately in the health manager section of the Octavia configuration. I checked and found the same code path in the house keeping process also.

PR #2332, opened 2019-12-16
octavia: Add anti-affinity settings (SOC-11026)

nicolasbock:
Add the anti-affinity setting to the relevant configuration files. The
affected services are the worker, the health-manager, and the
house-keeping service.


nicolasbock:
Thanks!

PR #2333, opened 2019-12-17
Do not read data from barclamp that has not been saved yet.

jsuchome:
Recipe X should not access barclamp data looking for X in the same
run where this data should be saved for the first time.

This actually seems to have only one effect: not getting the "insecure"
value correctly if SSL was set up only for one component (neutron or nova)
but not for keystone.

### DIFF
--- a/chef/cookbooks/neutron/recipes/post_install_conf.rb
+++ b/chef/cookbooks/neutron/recipes/post_install_conf.rb
@@ -65,8 +65,7 @@ vni_start = [node[:neutron][:vxlan][:vni_start], 0].max
 
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
-neutron_config = Barclamp::Config.load("openstack", "neutron")
-ssl_insecure = CrowbarOpenStackHelper.insecure(neutron_config) || keystone_settings["insecure"]
+ssl_insecure = CrowbarOpenStackHelper.insecure(node[:neutron]) || keystone_settings["insecure"]
 
 env = "OS_USERNAME='#{keystone_settings["service_user"]}' "
 env << "OS_PASSWORD='#{keystone_settings["service_password"]}' "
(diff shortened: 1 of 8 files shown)


jsuchome:
There's no Jira story/bug report yet, just a trial baloon. Am I missing something obvious?

jsuchome:
@nicolasbock I think most of the cases you are mentioning are looking for _other_ barclamp data, e.g. cinder for glance etc. In such case it it's OK.

nicolasbock:
Sorry, I should have read your commit message more carefully :/

```
./chef/cookbooks/octavia/recipes/health_manager.rb\016:octavia_config = Barclamp::Config.load("openstack", "octavia")
```

```
./chef/cookbooks/octavia/recipes/housekeeping.rb\016:octavia_config = Barclamp::Config.load("openstack", "octavia")
```

```
./chef/cookbooks/octavia/recipes/worker.rb\015:octavia_config = Barclamp::Config.load("openstack", "octavia")
```

```
./chef/cookbooks/octavia/recipes/api.rb\019:octavia_config = Barclamp::Config.load("openstack", "octavia")
```

```
./chef/cookbooks/octavia/recipes/nova.rb\022:octavia_config = Barclamp::Config.load("openstack", "octavia")
```

```
./crowbar_framework/lib/openstack/data_bag_config.rb\048:        keystone_config = Crowbar::DataBagConfig.load(
```

The last one I am not sure about. It does load the `openstack` config but is not in a recipe. In any case, it seems that Octavia is a bit time offender :wink:

jsuchome:
Yeah, Octavia seems problematic, unfortunately I do not know it at all. I'll look at it tomorrow

skazi0:
@jsuchome is this the place where the data is saved: https://github.com/crowbar/crowbar-openstack/blob/master/crowbar_framework/app/models/openstack_service_object.rb#L29 ?
If so, then I think it's worth adding some ref to this to the commit message.

jsuchome:
`crowbar_framework/lib/openstack/data_bag_config.rb` is fine I believe, it's also used from other barclamps looking at keystone...

jsuchome:
@nicolasbock @skazi0 updated.

PR #2337, opened 2020-01-03
Install openstack client for neutron recipes

jsuchome:
This is only relevant when neutron nodes are in separate cluster
and openstack client is not installed by some other (most likely keystone)
dependency.

### DIFF
--- a/chef/cookbooks/neutron/recipes/post_install_conf.rb
+++ b/chef/cookbooks/neutron/recipes/post_install_conf.rb
@@ -117,6 +117,8 @@ else
   Chef::Log.error("networking plugin '#{networking_plugin}' invalid for creating provider networks")
 end
 
+package "python-openstackclient"
+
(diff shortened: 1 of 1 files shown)


jsuchome:
That is not a problem. If user needs it for manual actions, he/she can install it manually. But it is needed for the actions done in post install recipe

nicolasbock:
True :smile:

JanZerebecki:
Travis complains about Commit fe0df29a62:
````
Title does not match regex (^(?!SOC).*\((((bsc#|SOC-)[0-9]+(,\s)?)+|trivial|typo|noref)\)$)
````

jsuchome:
I know. First I want to find out if we really need it, and if so, I'll create SOC reference

jsuchome:
@nicolasbock @skazi0 I've added SOC reference, and I need a new review...

PR #2339, opened 2020-01-10
[4.0] Install openstack client for neutron recipes (SOC-11039)

sandonovsuse:
This is only relevant when neutron nodes are in separate cluster
and openstack client is not installed by some other (most likely keystone)
dependency.

(cherry picked from commit f5249b2)


JanZerebecki:
Backport of https://github.com/crowbar/crowbar-openstack/pull/2337 .
But SOC7 only gets important fixes. This seems not to be one. Please reopen if you think otherwise.

PR #2340, opened 2020-01-10
[4.0] Terminate ssl on haproxy (bsc#1149535)

bbobrov:

The change depends on https://github.com/crowbar/crowbar-ha/pull/372


JanZerebecki:
Would need to be refreshed with the changes from master.
That being said this branch is only open for really important changes, I get the impression this doesn't count. Please reopen if you feel differently.

PR #2341, opened 2020-01-13
tempest: remove port_admin_state_change workaround (SOC-10029)

stefannica:
(reverts #2189)

Changing the admin state of a port was broken due to a bug on
vhost net and tun/tap. This bug has been fixed on SLES12SP4, so
we can again run tempest test cases targeting this feature.

### DIFF
--- a/chef/cookbooks/tempest/recipes/config.rb
+++ b/chef/cookbooks/tempest/recipes/config.rb
@@ -409,7 +409,6 @@ neutron_api_extensions = [
   "trunk-details"
 ].join(", ")
 
-neutron_ml2_mechanism_drivers = []
 unless neutrons[0].nil?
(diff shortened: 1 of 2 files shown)


stefannica:
Testing with: https://ci.suse.de/blue/organizations/jenkins/cloud-crowbar9-job-gate-linuxbridge-deploy-x86_64/detail/cloud-crowbar9-job-gate-linuxbridge-deploy-x86_64/152/

stefannica:
The tempest test that verifies the port admin state disable/enable use case is still failing. I suspect this is because the SOC9 Crowbar ECP CI is still using an old SP4 SLES image, because the new one breaks rebooting the controller nodes (https://jira.suse.com/browse/SOC-10960). This PR is blocked until SOC-10960 is fixed.

PR #2343, opened 2020-01-14
tempest: tempest run filters as templates (SOC-11052)

stefannica:
The tempest run filters employed by Crowbar are modeled
as static files instead of being generated at runtime as
chef templates. This makes it difficult to select which test
cases are enabled/disabled depending on which features and
options are actually configured (as is the case with Octavia
and LBaaS test cases).

This change re-models tempest run filters as templates.


stefannica:
Testing with ECP job: https://ci.suse.de/blue/organizations/jenkins/cloud-crowbar9-job-ha-x86_64/detail/cloud-crowbar9-job-ha-x86_64/252/pipeline

PR #2346, opened 2020-01-24
Terminate ssl on haproxy (bsc#1149535)

bbobrov:

Depends on https://github.com/crowbar/crowbar-ha/pull/373

### DIFF
--- a/chef/data_bags/crowbar/template-nova.json
+++ b/chef/data_bags/crowbar/template-nova.json
@@ -116,7 +116,9 @@
         "generate_certs": false,
         "insecure": false,
         "cert_required": false,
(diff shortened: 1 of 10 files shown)


chef/data_bags/crowbar/template-nova.json
```diff
@@ -181,7 +183,7 @@
     "nova": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 305,
+      "schema-revision": 306,
```
jsuchome: Please rebase and change this, there already is 306 revision

bbobrov: done, thank you

PR #2347, opened 2020-01-24
Terminate nova ssl on haproxy (bsc#1149535)

bbobrov:

(cherry picked from commit 94fc788d401633f766474137e9493cb310e53c5d)


chef/data_bags/crowbar/template-cinder.json
```diff
@@ -182,7 +184,7 @@
     "cinder": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 207,
+      "schema-revision": 208,
```
jsuchome: cinder already has 209 in 5.0 branch

chef/data_bags/crowbar/template-nova.json
```diff
@@ -185,7 +187,7 @@
     "nova": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 211,
+      "schema-revision": 212,
```
jsuchome: nova already has 212 in 5.0 branch

chef/data_bags/crowbar/template-cinder.json
bbobrov: done, thank you

chef/data_bags/crowbar/template-nova.json
bbobrov: done, thank you

chef/data_bags/crowbar/migrate/cinder/210_add_haproxy_mode_http.rb
```diff
@@ -0,0 +1,12 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  key = "loadbalancer_terminate_ssl"
```
skazi0: what about the pemfile?

bbobrov: you are right, and the same problem exists in other pull requests. Thanks, will do

JanZerebecki:
Needs to be refreshed with the changes from master.

PR #2348, opened 2020-01-24
designate: don't use pacemaker for designate-producer (SOC-XXXX)

stefannica:
The designate-producer service has active-active capabilities
now [1], which means it doesn't need to be managed by pacemaker.

[1] https://docs.openstack.org/designate/rocky/admin/ha.html


stefannica:
I pushed the submit button too fast on this one. A closer look at the docs revealed my mistake:

```You will need to set a coordination backend_url. This needs to be a DLM that is supported by tooz, that supports group membership. See tooz driver list for available drivers. 
Warning: Failure to set a backend_url can cause unexpected consequences, and may result in some periodic tasks being ran more than once.```

PR #2349, opened 2020-01-28
rabbitmq: sync startup definitions.json with recipe (SOC-11077)

stefannica:
Keep the policy configured in the generated definitions.json
file used during startup in sync with the policy configured
at runtime by the rabbitmq chef recipe, to avoid inconsistencies.

Also see #1448 for more information on how we use definitions.json.

### DIFF
--- a/chef/cookbooks/rabbitmq/libraries/crowbar.rb
+++ b/chef/cookbooks/rabbitmq/libraries/crowbar.rb
@@ -28,4 +28,22 @@ module CrowbarRabbitmqHelper
   def self.get_management_address(node)
     get_listen_address(node)
   end
+
+  def self.ha_policy_regex
+    # don't mirror queues that are 'amq.*' or '*_fanout_*' or `reply_*` in their names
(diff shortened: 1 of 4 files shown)


stefannica:
I ran an all-round session of standby/online transitions for all 3 controller nodes in my environment and with this patch the rabbitmq policy configuration remained unchanged throughout all that.

chef/cookbooks/rabbitmq/libraries/crowbar.rb
```diff
@@ -28,4 +28,22 @@ def self.get_public_listen_address(node)
   def self.get_management_address(node)
     get_listen_address(node)
   end
+
+  def self.get_ha_policy_regex(node)
```
djoreilly: node is unused here?

stefannica: good catch, thanks. I fixed it to have no parameters.

PR #2350, opened 2020-01-29
[5.0] rabbitmq: sync startup definitions.json with recipe (SOC-11077)

stefannica:
(backports #2349)

Keep the policy configured in the generated definitions.json
file used during startup in sync with the policy configured
at runtime by the rabbitmq chef recipe, to avoid inconsistencies.

(cherry picked from commit bc3d7de745ef2863b9b485bce6634951685d5293)


stefannica:
With this fix applied, we have problems rebooting controller nodes (see https://jira.suse.com/browse/SOC-11082). Not to be merged until a solution is found.

stefannica:
Closed in favor of #2410

PR #2355, opened 2020-02-03
Revert "rabbitmq: sync startup definitions.json with recipe (SOC-11077)"

stefannica:
Reverts crowbar/crowbar-openstack#2349

With this fix applied, we have problems rebooting controller nodes (see https://jira.suse.com/browse/SOC-11082). Reverting until a solution is found.


stefannica:
Abandoned in favour of #2358

PR #2357, opened 2020-02-04
Designate: make sure dns-server is active on a non-admin node (SOC-10…

guangyee:
…636)

Designate uses Bind9 backend by default. Bind9 is managed by the
dns-server role. Since Designate does not rely on the DNS server
on the Crowbar admin node, we need to make sure DNS server is running
on at least one non-admin node. If HA is enabled, DNS server should be
running on all the Designate worker nodes. If not, we need to warn users
the need to do so.

### DIFF
--- a/crowbar_framework/app/models/designate_service.rb
+++ b/crowbar_framework/app/models/designate_service.rb
@@ -103,6 +103,26 @@ class DesignateService < OpenstackServiceObject
       )
     end
 
+    # Make sure dns-server role is applied to at least one non-admin node
+    dns_server_nodes = NodeObject.find("roles:dns-server")
+    dns_server_nodes.delete_if { |n| n.nil? || n.admin? }
+    if dns_server_nodes.empty?
+      validation_error I18n.t(
+        "barclamp.#{@bc_name}.validation.non_admin_dns_nodes_not_found"
+      )
+    end
+
+    # All designate-worker nodes should have dns-server role too. This is
+    # assuming that designate-working is deployed on all the controller nodes.
+    designate_worker_nodes = NodeObject.find("roles:designate-worker")
+    designate_worker_nodes.each do |node|
+      next if node.roles.include?("dns-server")
+      @logger.warn(
+        "Node #{node.name} does not have dns-server role. All "\
+        "designate-worker node(s) should have dns-server role too."
+      )
+    end
+
     super
   end
 
--- a/crowbar_framework/config/locales/designate/en.yml
+++ b/crowbar_framework/config/locales/designate/en.yml
@@ -36,3 +36,4 @@ en:
           ca_certs: 'SSL CA Certificates File'
       validation:
         invalid_email_address: 'E-mail address %{email} is invalid.'
+        non_admin_dns_nodes_not_found: 'No DNS server found on non-admin nodes. Dns-server role must be applied to at least one non-admin node prior to deploying Designate.'


crowbar_framework/app/models/designate_service.rb
```diff
@@ -103,6 +103,22 @@ def validate_proposal_after_save(proposal)
...
+    if dns_server_nodes.empty?
+      validation_error("No DNS server found on non-admin nodes. Dns-server role must be applied to at least one non-admin nodes prior to deploying Designate.")
+    end
+
+    # All designate-worker nodes should have dns-server role too. This is
+    # assuming that designate-working is deployed on all the controller nodes.
+    designate_worker_nodes = NodeObject.find("roles:designate-worker")
+    designate_worker_nodes.each do |node|
+      unless node.roles.include?("dns-server")
+        @logger.warn("Node #{node.name} does not have dns-server role. All designate-worker node(s) should have dns-server role too.")
```
jsuchome: Why is this just a warning in the log?

crowbar_framework/app/models/designate_service.rb
```diff
@@ -103,6 +103,22 @@ def validate_proposal_after_save(proposal)
       )
     end
 
+    # Make sure dns-server role is applied to at least one non-admin node
+    dns_server_nodes = NodeObject.find("roles:dns-server")
+    dns_server_nodes.delete_if { |n| n.nil? || n.admin? }
+    if dns_server_nodes.empty?
+      validation_error("No DNS server found on non-admin nodes. Dns-server role must be applied to at least one non-admin nodes prior to deploying Designate.")
```
nicolasbock: In other places we internationalize the actual message. Not that we have the messages in any other language than english, but for consistency it would be better to follow what's done elsewhere.

crowbar_framework/app/models/designate_service.rb
guangyee: Per my understanding, it is not a hard requirement as

1) for the existing deployments there's no guarantee that DNS is available on all the controller nodes.
2) Designate will still function even with one DNS node. Though in an HA situation, it will fail of that node goes down.
3) mkcloud currently does not provision DNS on all the controller nodes even if HA is enabled.

If I understand the comments in SOC-10636 correctly, we don't want to start making it a hard requirement as we don't want to disrupt the existing deployments. Instead, we will warn users that they *should* consider deploying DNS on all the Designate worker nodes. We will document the recommendation as well.

crowbar_framework/app/models/designate_service.rb
guangyee: Sounds good.

crowbar_framework/config/locales/designate/en.yml
```diff
@@ -36,3 +36,5 @@ en:
           ca_certs: 'SSL CA Certificates File'
       validation:
         invalid_email_address: 'E-mail address %{email} is invalid.'
+        non_admin_dns_nodes_not_found: 'No DNS server found on non-admin nodes. Dns-server role must be applied to at least one non-admin node prior to deploying Designate.'
+        designate_worker_node_is_not_dns_server: 'Node %{node} does not have dns-server role. All designate-worker node(s) should have dns-server role too.'
```
nicolasbock: Why aren't you using this in the call to `warn()`?

guangyee: I wasn't sure if we need to translate log message. Maybe I need to take it out of this file for now? What do you think?

nicolasbock: Hmm, ok, that's a good point. Frankly I don't see the use of this localization anyway. It's really just for consistency reasons :smile:

nicolasbock: A few lines up in the service file there is a log debug message that's not localized. I should have checked that before :smile: I think you are right, let's just leave the warning un-localized.

nicolasbock: Do we then still need this variable here?

guangyee: I've taken it out for now. Thanks for catching that Nick.

crowbar_framework/app/models/designate_service.rb
jsuchome: OK, I was just worried because in general we can't expect people to catch that warning message. Anyway if there's a plan to put it into documentation than it's good.

PR #2359, opened 2020-02-07
Designate: make sure dns-server is active on a non-admin node (SOC-10…

guangyee:
…636)

Designate uses Bind9 backend by default. Bind9 is managed by the
dns-server role. Since Designate does not rely on the DNS server
on the Crowbar admin node, we need to make sure DNS server is running
on at least one non-admin node. If HA is enabled, DNS server should be
running on all the Designate worker nodes. If not, we need to warn users
the need to do so.

### DIFF
--- a/crowbar_framework/app/models/designate_service.rb
+++ b/crowbar_framework/app/models/designate_service.rb
@@ -103,6 +103,26 @@ class DesignateService < OpenstackServiceObject
       )
     end
 
+    # Make sure dns-server role is applied to at least one non-admin node
+    dns_server_nodes = NodeObject.find("roles:dns-server")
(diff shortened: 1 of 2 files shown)


guangyee:
> Isn't that a backport from #2357 ? Could you cherry pick it?

This is the cherry-pick I think. I used 'git cherry-pick <commit ID of 2357>' to produce this patch.

nicolasbock:
Ah :smile: Could you run `git cherry-pick -x SHA` so that we get a reference to the cherry picked commit in the commit message? Thanks!

guangyee:
> Ah smile Could you run `git cherry-pick -x SHA` so that we get a reference to the cherry picked commit in the commit message? Thanks!

Ah, I forgot the -x. :-)

PR #2362, opened 2020-02-24
nova: Enable ephemeral volumes on SES (SOC-5269)

skazi0:
Add required configuration to enable usage of SES/Ceph storage backend
for ephemeral storage in nova.
If multiple SES/Ceph backends are defined in cinder, first one is used
by nova for ephemeral strorage.

### DIFF
--- a/chef/cookbooks/nova/recipes/config.rb
+++ b/chef/cookbooks/nova/recipes/config.rb
@@ -108,6 +108,7 @@ directory "/etc/nova" do
 end
 
 rbd_enabled = false
+ephemeral_rbd_settings = nil
 
 cinder_servers = node_search_with_cache("roles:cinder-controller")
 if cinder_servers.length > 0
@@ -117,7 +118,14 @@ if cinder_servers.length > 0
 
   if node.roles.include? "nova-compute-kvm"
     cinder_server[:cinder][:volumes].each do |volume|
-      rbd_enabled = true if volume["backend_driver"] == "rbd"
+      next unless volume["backend_driver"] == "rbd"
+      rbd_enabled = true
+      # use first rbd cinder backend for ephemeral storage settings
+      ephemeral_rbd_settings = volume[:rbd].clone
+      # override pool with nova section from settings
+      ses_config = SesHelper.ses_settings
+      ephemeral_rbd_settings[:pool] = ses_config[:nova][:rbd_store_pool]
+      break
     end
   end
 else
@@ -411,6 +419,7 @@ template node[:nova][:config_file] do
     use_baremetal_filters: use_baremetal_filters,
     track_instance_changes: track_instance_changes,
     ironic_settings: ironic_settings,
+    ephemeral_rbd_settings: ephemeral_rbd_settings,
     default_log_levels: node[:nova][:default_log_levels]
   )
 end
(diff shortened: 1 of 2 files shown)


chef/cookbooks/nova/recipes/config.rb
```diff
@@ -117,7 +118,10 @@
 
   if node.roles.include? "nova-compute-kvm"
     cinder_server[:cinder][:volumes].each do |volume|
-      rbd_enabled = true if volume["backend_driver"] == "rbd"
+      if volume["backend_driver"] == "rbd"
+        rbd_enabled = true
+        first_rbd_cinder_volume = volume if first_rbd_cinder_volume.nil?
```
jsuchome: Why are you checking for `if first_rbd_cinder_volume.nil?` Can it be something else then `nil` at this place?

chef/cookbooks/nova/recipes/config.rb
```diff
@@ -411,6 +415,7 @@
     use_baremetal_filters: use_baremetal_filters,
     track_instance_changes: track_instance_changes,
     ironic_settings: ironic_settings,
+    rbd_settings: first_rbd_cinder_volume,
```
jsuchome: Why are you calling the key `rbd_settings` if it's a volume? (I find it just a bit confusing)

chef/cookbooks/nova/recipes/config.rb
skazi0: If there are more than one cinder backend, we want `first_rbd_cinder_volume` to point to the first one (i.e. not be overwritten on subsequent loop runs).

skazi0: The whole loop could be changed to something like:
```
first_rbd_cinder_volume = cinder_server[:cinder][:volumes].first { |volume| volume["backend_driver"] == "rbd" }
rbd_enabled |= !first_rbd_cinder.volume.nil?

File: chef/cookbooks/nova/recipes/config.rb (hunk 2 above)
Comment (skazi0) [reply chef/cookbooks/nova/recipes/config.rb]: Because "If multiple SES/Ceph backends are defined in cinder, first one is used by nova for ephemeral strorage." as described in the commit message.

Comment (skazi0) [reply chef/cookbooks/nova/recipes/config.rb]: ...and to make it even more confusing, "volume" in Crowbar terms is a "backend" not an actual volume :oD

File: chef/cookbooks/nova/recipes/config.rb (hunk 1 above)
Comment (jsuchome) [reply chef/cookbooks/nova/recipes/config.rb]: Oh, right, I missed there's a loop ;-)

File: chef/cookbooks/nova/recipes/config.rb (hunk 2 above)
Comment (jsuchome) [reply chef/cookbooks/nova/recipes/config.rb]: That was not my point here, I would just name the key same way as the hash. But that's just a matter of taste I guess

File: chef/cookbooks/nova/recipes/config.rb (hunk 1 above)
Comment (skazi0) [reply chef/cookbooks/nova/recipes/config.rb]: I removed one nesting level to make Hound happy :o)

Comment (skazi0) [reply chef/cookbooks/nova/recipes/config.rb]: and got rid of the "if nil?" by breaking the loop on first hit.

File: chef/cookbooks/nova/recipes/config.rb (hunk 2 above)
Comment (skazi0) [reply chef/cookbooks/nova/recipes/config.rb]: then the confusion would move to the template...

**Comment by skazi0** [comment]
@jsuchome I updated the patch to use pool from nova section and reworked it a bit. hopefully it will be less confusing. please re-check.

**Comment by jsuchome** [comment]
confirming that it works after testing

**Comment by JanZerebecki** [comment]
This caused a regression caught by CI fixed in #2362 . The error looked like:
```
Instance failed to spawn: libvirtError: Secret not found: no secret with matching uuid 'f6313034-390c-43a3-9661-db9cf6680195'
```

PR #2364, opened 2020-02-27
[5.0] nova: Enable ephemeral volumes on SES (SOC-11119)

skazi0:
Add required configuration to enable usage of SES/Ceph storage backend
for ephemeral storage in nova.
If multiple SES/Ceph backends are defined in cinder, first one is used
by nova for ephemeral strorage.

port of #2362 and #2373


JanZerebecki:
I think this still needs a toggle to not switch existing installations.

PR #2366, opened 2020-02-28
keystone: Register SES RadosGW endpoints (SOC-5270)

skazi0:
If uploaded SES configuration contains non-empty radosgw_urls list first
of the URLs is registered in keystone catalog as object-store endpoint.
If Swift is deployed at the same time, SES configuration has higher
priority.


chef/cookbooks/swift/recipes/proxy.rb
```diff
@@ -234,23 +236,26 @@
...
-       port keystone_settings["admin_port"]
-       endpoint_service "swift"
-       endpoint_region keystone_settings["endpoint_region"]
-       endpoint_publicURL "#{swift_protocol}://#{public_host}:"\
-                          "#{node[:swift][:ports][:proxy]}/v1/"\
-                          "#{node[:swift][:reseller_prefix]}$(project_id)s"
-       endpoint_adminURL "#{swift_protocol}://#{admin_host}:"\
-                         "#{node[:swift][:ports][:proxy]}/v1/"
-       endpoint_internalURL "#{swift_protocol}://#{admin_host}:"\
+     # register RadosGW endpoints only if there's no SES based RadosGW present
```
JanZerebecki:
```suggestion
     # register swift-proxy endpoints only if no SES based RadosGW is being registered in ses.rb
```
Instead of this comment an alternative would be to call what is done in ses.rb here in an else block.

skazi0: Note that `ses.rb` is in keystone cookbook. Code above will not be called if swift is not deployed.

JanZerebecki:
```suggestion
     # register swift-proxy endpoints only if no SES based RadosGW is being registered in keystone/recipes/ses.rb
```
Right, so my alternative won't work. The above is a better fix for the comment.

skazi0: updated. thx!

skazi0:
@JanZerebecki it is documented that if you'd rather use Swift for object storage, you need to remove the `radosgw_urls` from SES yaml file.

PR #2361, opened 2020-02-21
[5.0] SES integration and (a bit of) Ceph cleanup

skazi0:
port of #2035 + #1962 (for clean backport)
and #2368 

prereq https://github.com/crowbar/crowbar-core/pull/2006

PR #2368, opened 2020-03-04
nova: Setup CEPH secrets later (SOC-11141)

skazi0:
Moved libvirt secret definition out of the config recipe which is
executed too early when libvirtd is not running yet. This should make
the CEPH integration complete on first chef-client run not in two runs
as it was before.


skazi0:
backport added to https://github.com/crowbar/crowbar-openstack/pull/2361

PR #2369, opened 2020-03-04
keystone: fix keystone node lookup (SOC-11333)

jgrassler:
In a cloud where a controller was added after initial
deployment and MAC addresses are out of sequence, the
search_for_keystone() method may return the wrong controller
in the first upgraded controller's crowbar_join phase. This
commit ensures this does not happen by removing controllers in
still in the 'prepare-os-upgrade' stage from the list of
controllers.

Note: I'm not sure if we actually need this for Cloud 9 (currently testing whether it's affected). If we turn out not to need it, I'll abandon this pull request and create one for Cloud 8 only.

### DIFF
--- a/chef/cookbooks/keystone/libraries/helpers.rb
+++ b/chef/cookbooks/keystone/libraries/helpers.rb
@@ -213,7 +213,13 @@ module KeystoneHelper
       return @keystone_node[instance]
     end
 
-    nodes, _, _ = Chef::Search::Query.new.search(:node, "roles:keystone-server AND keystone_config_environment:keystone-config-#{instance}")
+    nodes, = Chef::Search::Query.new.search(
+      :node,
+      "roles:keystone-server" \
+      " AND keystone_config_environment:keystone-config-#{instance}" \
+      " AND NOT state:crowbar_upgrade"
+    )
+
     if nodes.first
       keystone_node = nodes.first
       keystone_node = node if keystone_node.name == node.name


chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -214,6 +214,16 @@ def self.profiler_settings(current_node, cookbook_name)
     end
 
     nodes, _, _ = Chef::Search::Query.new.search(:node, "roles:keystone-server AND keystone_config_environment:keystone-config-#{instance}")
+
+    # Omit nodes with prepare-os-upgrade state (these may have pre-upgrade data bag content)
+    nodes.reject! do |n|
```
skazi0: can't we just add something like `AND NOT state:crowbar_upgrade` to the `search` above?

jsuchome: See e.g. https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/helpers/nodes_helper.rb#L23

chef/cookbooks/keystone/libraries/helpers.rb
```diff
@@ -213,7 +213,10 @@ def self.profiler_settings(current_node, cookbook_name)
       return @keystone_node[instance]
     end
 
-    nodes, _, _ = Chef::Search::Query.new.search(:node, "roles:keystone-server AND keystone_config_environment:keystone-config-#{instance}")
+    nodes, _, _ = Chef::Search::Query.new.search(:node,
+                                                 "roles:keystone-server " +
+                                                 "AND keystone_config_environment:keystone-config-#{instance}" +
+                                                 "AND NOT state:crowbar_upgrade")
```
jsuchome: I think you're missing the space before AND

jgrassler:
All Hound and review comments addressed, test cloud almost done redeploying...

jgrassler:
Upgrade worked fine with the code in its current state. We can merge as far as I'm concerned.

PR #2370, opened 2020-03-06
Revert 'tempest: Set port_admin_state_change to false when using linuxbridge' (SOC-10029)

stefannica:
The underlying problem has been fixed. The #2189 workaround can be removed.


stefannica:
Already covered in #2341.

PR #2371, opened 2020-03-11
heat: Simplify domain registration code + timeout increase (SOC-11103)

skazi0:
The code contained some workarounds for openstack client bugs which
were not needed anymore.

### DIFF
--- a/chef/cookbooks/heat/recipes/server.rb
+++ b/chef/cookbooks/heat/recipes/server.rb
@@ -113,7 +113,10 @@ my_public_host = CrowbarHelper.get_host_for_public_url(node, node[:heat][:api][:
 
 db_connection = fetch_database_connection_string(node[:heat][:db])
 
-crowbar_pacemaker_sync_mark "wait-heat_register" if ha_enabled
+crowbar_pacemaker_sync_mark "wait-heat_register" do
+  timeout 90
+  only_if { ha_enabled }
+end
 
 register_auth_hash = { user: keystone_settings["admin_user"],
                        password: keystone_settings["admin_password"],
--- a/chef/cookbooks/heat/recipes/server.rb
+++ b/chef/cookbooks/heat/recipes/server.rb
@@ -204,82 +204,54 @@ bash "register heat domain" do
   code <<-EOF
 
     # Find domain ID
-    id=
-    eval $(openstack #{insecure} \
+    HEAT_DOMAIN_ID=$(openstack #{insecure} \
         domain show \
-        -f shell --variable id \
+        -f value -c id \
         #{stack_user_domain_name})
 
-    HEAT_DOMAIN_ID=$id
-
     if [ -z "$HEAT_DOMAIN_ID" ]; then
-        id=
-        eval $(openstack #{insecure} \
+        HEAT_DOMAIN_ID=$(openstack #{insecure} \
             domain create \
-            -f shell --variable id \
+            -f value -c id \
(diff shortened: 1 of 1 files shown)


chef/cookbooks/heat/recipes/server.rb
```diff
@@ -204,65 +207,44 @@
...
+        HEAT_DOMAIN_ID=$(openstack #{insecure} \
             domain create \
-            -f shell --variable id \
+            -f value -c id \
             --description "Owns users and projects created by heat" \
             #{stack_user_domain_name})
-        HEAT_DOMAIN_ID=$id
     fi
 
     [ -n "$HEAT_DOMAIN_ID" ] || exit 1
```
jsuchome: Can you add some error message here, while you're at it?

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -204,65 +207,44 @@
   code <<-EOF
 
     # Find domain ID
-    id=
-    eval $(openstack #{insecure} \
+    HEAT_DOMAIN_ID=$(openstack #{insecure} \
```
jsuchome: Do we actually need to find out `HEAT_DOMAIN_ID` if we can use domain name (`heat`) in later calls?

chef/cookbooks/heat/recipes/server.rb
```diff
@@ -204,65 +207,44 @@
...
             --domain $HEAT_DOMAIN_ID \
-            --password #{node[:heat]["stack_domain_admin_password"]} \
+            --password #{node[:heat][:stack_domain_admin_password]} \
             --description "Manages users and projects created by heat" \
-            #{node[:heat]["stack_domain_admin"]})
-        STACK_DOMAIN_ADMIN_ID=$id
+            #{node[:heat][:stack_domain_admin]})
     fi
 
     [ -n "$STACK_DOMAIN_ADMIN_ID" ] || exit 1
```
jsuchome: + error message

jsuchome:
I think you can also drop the part that checks if `admin` role is present (`# Make user an admin` block) and always call `openstack role add`. This way we have one less CLI call in the initial run.

chef/cookbooks/heat/recipes/server.rb
skazi0: I don't think it's needed. `exit 1` will fail the bash block and chef will display STDOUT and STDERR of the whole script. This will contains outputs from all openstack commands and should be way more informative than simple "Domain creation failed" I could add here.

chef/cookbooks/heat/recipes/server.rb
skazi0: It would make sense only if we replace the 'check + create' for domain with 'create || true'. We would save one `openstack` call then. Otherwise it would not be much of a difference imo.

chef/cookbooks/heat/recipes/server.rb
skazi0: ditto

skazi0:
@jsuchome good point with the redundant check for role assignment.

chef/cookbooks/heat/recipes/server.rb
jsuchome: OK. You're right, we need to create it.

chef/cookbooks/heat/recipes/server.rb
jsuchome: ok

PR #2373, opened 2020-03-17
nova: Add option to disable ephemeral on ceph (SOC-5269)

skazi0:
New attribute was added to enable/disable ephemeral storage on CEPH/SES if
needed.

Added UI part:
Zrzut ekranu z 2020-03-17 23-08-25

bonus: remove redundant config line

### DIFF
--- a/chef/data_bags/crowbar/migrate/nova/306_add_use_rbd_ephemeral.rb
+++ b/chef/data_bags/crowbar/migrate/nova/306_add_use_rbd_ephemeral.rb
@@ -1,0 +1,11 @@
+def upgrade(template_attributes, template_deployment, attributes, deployment)
+  key = "use_rbd_ephemeral"
+  attributes[key] = template_attributes[key] unless attributes.key? key
+  return attributes, deployment
+end
+
+def downgrade(template_attributes, template_deployment, attributes, deployment)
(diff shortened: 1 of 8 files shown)


chef/data_bags/crowbar/migrate/nova/306_add_use_rbd_ephemeral.rb
```diff
@@ -0,0 +1,11 @@
+def upgrade(template_attributes, template_deployment, attributes, deployment)
+  key = "use_rbd_ephemeral"
+  attributes[key] = template_attributes[key] unless attributes.key? key
```
JanZerebecki:
```suggestion
  attributes[key] = false unless attributes.key? key
```
Then set the template to true, which would have the same effect as setting it in the model to true only on new installs. Not sure right now which one is better.

skazi0: Common pattern is to have migrations use default from template. I added enabling this for new deployments in `create_proposal()`

skazi0:
I've added a custom view control for the new flag. Please re-review.

skazi0:
backport added to https://github.com/crowbar/crowbar-openstack/pull/2364

PR #1608, opened 2018-03-29
Add ssl to magnum barclamp

sayalilunkad:

PR #2374, opened 2020-03-18
magnum: Add SSL support (SOC-9849)

skazi0:
Added options to configure SSL settings in Magnum barclamp.

Taken over from https://github.com/crowbar/crowbar-openstack/pull/1608


skazi0:
Tested again on a fresh deployment.

PR #2378, opened 2020-03-23
[5.0] magnum: Add SSL support (SOC-9849)

skazi0:
port of #2374


skazi0:
tested with SOC8 and default service image from mkcloud

PR #2380, opened 2020-03-24
fix travis CI to handle reverted commits properly (SOC-11180)

guangyee:
For reverted commits, Travis CI uses
'git clone --depth=50 --branch=<reverted branch> <repo url> <dir>'
instead of 'git clone --depth=50 <repo url> <dir>' when cloning the repo
for testing.

With 'git clone --depth=50 --branch=<reverted branch> <repo url> <dir>',
there's no master branch. Therefore, we should always use the branch name
instead of master branch when invoking gitlint.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
       script:
       - pip install gitlint
       - wget https://raw.githubusercontent.com/SUSE-Cloud/automation/master/scripts/jenkins/gitlint.ini
-      - gitlint --commits master..HEAD -C gitlint.ini
+      - gitlint --commits `git rev-parse --abbrev-ref HEAD`..HEAD -C gitlint.ini
     - name: "Syntaxcheck"
       gemfile: Gemfile
       script:


jeremy-moffitt:
The change looks fine to me, and seems highly unlikely to have triggered the failure in the mkcloud run:
```Failed to apply the proposal to:
data3 (d52-54-77-77-01-03.vc1.cloud.suse.de):

<pre>/var/chef/cache/cookbooks/keystone/libraries/helpers.rb:343:in `block in retry_request'
/var/chef/cache/cookbooks/keystone/libraries/helpers.rb:342:in `times'
/var/chef/cache/cookbooks/keystone/libraries/helpers.rb:342:in `retry_request'
/var/chef/cache/cookbooks/keystone/libraries/helpers.rb:267:in `post'
/var/chef/cache/cookbooks/keystone/libraries/helpers.rb:295:in `get_token'
/var/chef/cache/cookbooks/keystone/libraries/helpers.rb:259:in `initialize'
/var/chef/cache/cookbooks/keystone/providers/register.rb:141:in `new'
/var/chef/cache/cookbooks/keystone/providers/register.rb:141:in `block in class_from_file'
```

we may need to re-run that unless we want to force merge it, which is possible but probably not a great precedent

PR #2385, opened 2020-03-31
octavia: rework management network implementation (SOC-10904)

stefannica:
TL;DR: Crowbar has a more powerful mechanism of setting up additional networks and connecting interfaces to them. We leverage that here to set up a network for Octavia that doesn't require external routing, meaning that customers don't need to configure any external services or additional routes. This is better than the CLM octavia management network implementation, where customers must provide an external router that is able to route between the octavia management network and the OpenStack management network. Furthermore, this implementation achieves a complete isolation of the octavia management network from all other networks, providing better security. Having it entirely managed by Crowbar reduces the risk of a misconfiguration on the customer's part.

Rework the implementation of the Octavia management network as
a neutron provider network based on attributes taken from the
network.json octavia network, if present.
In this respect, the logic used to set up the Octavia management
network closely follows the logic already supported by Crowbar
allowing users to set up additional external networks [1], with
a few notable differences:
 * the `octavia` network is actually an internal neutron provider
 network (i.e. not usable as a source of floating IPs by other
 tenants)
 * the `octavia` network doesn't need a gateway, because controller
 nodes will communicate with amphora VMs directly, without requiring
 NAT
 * the `octavia` network needs to be set up on compute nodes even
 if DVR is not enabled, because amphorae will use it directly
 * we need to model two non-overlapping IP ranges for the octavia
 neutron provider network:
   * host range: used to allocate IP addresses for Octavia controller nodes
   * dhcp range: used to allocate IP addresses for amphora VMs
 Using a single range will eventually lead to situations in which an IP
 address allocated by neutron to an amphora VM is already in use by an
 Octavia controller node.

 * the octavia barclamp automatically creates the Neutron provider network
 and the subnet associated with the octavia network.json network
 * the controller nodes where Octavia is deployed have allocated
 IP addresses from the octavia network's `host` IP range

This rework also removes the `mgmt_cidr` attribute from the Octavia
barclamp, as it is rendered useless by the fact that all required
attributes to create the provider network are provided by the network.json
configuration. Given that controller nodes now have direct access to the
octavia network, the health manager configuration is also updated such that
if the octavia network is available, the service listens on the octavia
network IP locally configured directly instead of the admin network.

The actual creation of the octavia management network is moved
away from the neutron barclamp and into the octavia barclamp, to avoid
situations where the network isn't created because the neutron
`create_default_networks` attribute is explicitly disabled, as
well as to centralize all the logic responsible for creating
octavia prerequisites to the octavia barclamp. This breaks the
circular dependency between Neutron and Octavia (e.g. the name
of the management network is coded as an Octavia barclamp
attribute, which means Neutron would require Octavia to be
deployed to have access to this attribute, while Octavia relies
on this network to be already created by Neutron before its
services can be configured).

[1] https://documentation.suse.com/soc/9/single-html/suse-openstack-cloud-crowbar-deployment/#sec-setup-multi-ext-networks

### DIFF
--- a/chef/cookbooks/neutron/recipes/post_install_conf.rb
+++ b/chef/cookbooks/neutron/recipes/post_install_conf.rb
@@ -61,18 +61,22 @@ if ironic_net
   ironic_router = ironic_net["router"]
 end
 
-have_octavia = !(node[:network][:networks][:octavia].nil? || node[:octavia].nil?)
+octavia_net = Barclamp::Inventory.get_network_definition(node, "octavia")
+
+have_octavia = !octavia_net.nil?
 if have_octavia
-  octavia_net = node[:network][:networks][:octavia]
+  octavia_net_ranges = octavia_net["ranges"]
+  octavia_range = "#{octavia_net["subnet"]}/#{mask_to_bits(octavia_net["netmask"])}"
+  octavia_pool_start = octavia_net_ranges[:dhcp][:start]
+  octavia_pool_end = octavia_net_ranges[:dhcp][:end]
+  octavia_first_ip = IPAddr.new(octavia_range).to_range.to_a[2]
+  octavia_last_ip = IPAddr.new(octavia_range).to_range.to_a[-2]
+
+  octavia_pool_start = octavia_first_ip if octavia_first_ip > octavia_pool_start
+  octavia_pool_end = octavia_last_ip if octavia_last_ip < octavia_pool_end
+
   octavia_netname = node[:octavia][:amphora][:manage_net]
-  octavia_cidr = node[:octavia][:amphora][:manage_cidr]
   octavia_project = node[:octavia][:amphora][:project]
-  octavia_router = octavia_net[:router]
-  if octavia_net[:use_vlan]
-    octavia_vlan_id = octavia_net[:vlan]
-  else
-    Chef::Log.error("[FIXME] Add flat network")
-  end
 end
 
 vni_start = [node[:neutron][:vxlan][:vni_start], 0].max
@@ -94,6 +98,7 @@ openstack_cmd = "#{env} openstack #{ssl_insecure ? "--insecure" : ""}"
 
 fixed_network_type = ""
 floating_network_type = ""
+octavia_network_type = ""
 
 networking_plugin = node[:neutron][:networking_plugin]
 ml2_type_drivers_default_provider_network = node[:neutron][:ml2_type_drivers_default_provider_network]
@@ -102,11 +107,22 @@ when "ml2"
   # For ml2 always create the floating network as a flat provider network
   # find the network node, to figure out the right "physnet" parameter
   network_node = NeutronHelper.get_network_node_from_neutron_attributes(node)
-  physnet_map = NeutronHelper.get_neutron_physnets(network_node, ["nova_floating", "ironic"])
+
+  external_networks = ["nova_floating"]
+
+  # Add ironic to external_networks if ironic network is configured
+  external_networks << "ironic" if ironic_net
+
+  # Add octavia to external_networks if octavia network is configured
+  external_networks << "octavia" if octavia_net
+
+  physnet_map = NeutronHelper.get_neutron_physnets(network_node, external_networks)
   floating_network_type = "--provider-network-type flat " \
       "--provider-physical-network #{physnet_map["nova_floating"]}"
   ironic_network_type = "--provider-network-type flat " \
       "--provider-physical-network #{physnet_map["ironic"]}"
+  octavia_network_type = "--provider-network-type flat " \
+      "--provider-physical-network #{physnet_map["octavia"]}"
   case ml2_type_drivers_default_provider_network
   when "vlan"
     fixed_network_type = "--provider-network-type vlan " \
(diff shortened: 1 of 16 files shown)


chef/cookbooks/neutron/recipes/post_install_conf.rb
```diff
@@ -158,13 +174,9 @@ def mask_to_bits(mask)
 end
 
 execute "create_octavia_management_network" do
-  command "#{openstack_cmd} network create " \
+  command "#{openstack_cmd} network create --external " \
           "--project #{octavia_project} " \
-          "--provider-network-type vlan " \
-          "--provider-physical-network physnet1 " \
-          "--provider-segment #{octavia_vlan_id} " \
-          "--internal " \
```
djoreilly, line 166: Why change it from an internal to an external network? It won't be used for floating ips.

stefannica: @djoreilly that's a very good point. I used an external network because that's what a neutron provider network generally is. An external network doesn't include a managed gateway (i.e. that's what the `router:external=True` part means), which effectively disables NAT thus denying any type of external communication for Octavia amphora VMs.
But your question did bring up another issue: even though I'm not explicitly making the network shared, an external network is nevertheless visible to all other tenants and does allow them to use it as a source of floating IPs. Looks like I need to switch it back to internal, unless I find another way to hide it from other tenants.

stefannica: @djoreilly I found a way to hide the octavia network from other tenants: configure an RBAC policy that keeps the network hidden (more in [this upstream openstack bug](https://bugs.launchpad.net/horizon/+bug/1760832))

djoreilly: hmm - provider networks (type vlan or flat) can be used by themselves. They are not just for external networks for neutron routers.

The `--external` means neutron routers can be uplinked to the network (neutron router-gateway-set) and used for floating ips and/or snat.

Also by removing `--external` other tenants won't see it as it belongs to octavia_project.

stefannica: @djoreilly nevermind, I'll just switch back to using an internal network, to be closer to what CLM is using. Hopefully, that will still work with having host IPs assigned to the controller nodes themselves (will run some tests).

djoreilly: Okay. I assume crowbar knows not use IPs from the subnet allocation-pool for the controller hosts. And it knows to put the IPs on the right bridge - `br-octavia` ?

stefannica: @djoreilly yes, crowbar knows :) because I took care of that. There are two IP ranges, `dhcp` is used for VMs, `host` is used for controller hosts.

guangyee:
>       * host range: used to allocate IP addresses for Octavia controller nodes

Why are we treating Octavia services different than other services running on the controller? In CLM, OCTAVIA-MGMT-NET is used by the Amphoras only. Octavia services (API, Worker, Health, etc) using the internal management network. Route is setup between OCTAVIA-MGMT-NET and internal management network.

>       * dhcp range: used to allocate IP addresses for amphora VMs
>         Using a single range will eventually lead to situations in which an IP
>         address allocated by neutron to an amphora VM is already in use by an
>         Octavia controller node.

That should not be a problem if OCTAVIA-MGMT-NET and internal network are separated.

stefannica:
> > ```
> >   * host range: used to allocate IP addresses for Octavia controller nodes
> > ```
> 
> Why are we treating Octavia services different than other services running on the controller? In CLM, OCTAVIA-MGMT-NET is used by the Amphoras only. Octavia services (API, Worker, Health, etc) using the internal management network. Route is setup between OCTAVIA-MGMT-NET and internal management network.
> 
> > ```
> >   * dhcp range: used to allocate IP addresses for amphora VMs
> >     Using a single range will eventually lead to situations in which an IP
> >     address allocated by neutron to an amphora VM is already in use by an
> >     Octavia controller node.
> > ```
> 
> That should not be a problem if OCTAVIA-MGMT-NET and internal network are separated.

Crowbar has a more powerful mechanism of setting up additional networks and connecting interfaces to them. We leverage that here to set up a network for Octavia that _doesn't require external routing_, meaning that customers don't need to configure any external services or additional routes. This is better than the CLM octavia management network implementation, where customers must provide an external router that is able to route between the octavia management network and the OpenStack management network. Furthermore, this implementation achieves a _complete isolation_ of the octavia management network from all other networks, providing better security. Having it entirely managed by Crowbar reduces the risk of a misconfiguration on the customer's part.

stefannica:
Successfully tested the following scenarios (all Crowbar deployments have 3 controllers, 2 computes and passed tempest smoke, lbaas and octavia filters):

- full deployment, including Octavia without the use of an `octavia` network.json network (i.e. verifies that the previous way of configuring a `lb-mgmt-net` network manually still works: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/892/pipeline/
- cloud deployment excluding Octavia, then manually deploying Octavia with the use of an `octavia` network (i.e. validates that a customer can deploy Octavia with an `octavia` supplied neutron provider network on top of an existing cloud installation): https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/881/pipeline/
- complete cloud deployment, including Octavia, starting with the `octavia` network added to network.json:
  - using the OVS backend: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/887/pipeline/
  - using the linuxbridge backend: https://ci.suse.de/blue/organizations/jenkins/openstack-crowbar/detail/openstack-crowbar/888/pipeline/

stefannica:
@guangyee @djoreilly thanks, guys, glad to see this one finally merged :)

PR #2386, opened 2020-04-02
galera: Make sure checks are executed without password (bsc#1136928)

jsuchome:
Monitoring user that is used for clusterchecks is set up without
the password. Galera pacemaker setup needs to be adjusted so it is
using the empty password instead falling back to default ~/.my.cnf.

We have to wait and publish this only after the fix in resource agents is released:
https://github.com/ClusterLabs/resource-agents/pull/1473


jsuchome:
Note: the failure of HA gating jobs is exactly because the missing fix in resource agents package

PR #2387, opened 2020-04-06
neutron: fix neutron cli to use internal endpoint (bsc#1168512)

djoreilly:
The neutron cli does not understand OS_INTERFACE and is defaulting to
the public endpoint. Add OS_ENDPOINT_TYPE=internalURL to fix this.


jsuchome:
That's actually a port of https://github.com/crowbar/crowbar-openstack/pull/2291 ... I was under the impression it's not needed for [master] ...

djoreilly:
Master still uses neutron cli to update the dns-name. I see this comment:
https://github.com/crowbar/crowbar-openstack/pull/2290#issuecomment-543717458

PR #2388, opened 2020-04-06
Fix magnum tempest tests (SOC-9298)

gosipyan:
- magnum barclamp: set `verify_ca = false`. The magnum barclamp does not
support setting `openstack_ca_file` which leaves magnum unusable when
deploying crowbar using self-signed certificates. Setting `verify_ca =
false` ensures that it will work for any case.

- tempest barclamp: set the DNS nameserver for magnum to the crowbar DNS
server; and add `member` to the list of roles used for tempest.

(cherry picked from commit abae955fdf8f3775e0b4d695fd9e9c803e667d4f)

### DIFF
--- a/chef/cookbooks/magnum/templates/default/magnum.conf.erb
+++ b/chef/cookbooks/magnum/templates/default/magnum.conf.erb
@@ -27,6 +27,9 @@ region_name = <%= @keystone_settings['endpoint_region'] %>
 [database]
 connection = <%= @sql_connection %>
 
+[drivers]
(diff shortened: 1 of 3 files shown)


gosipyan:
fix for https://ci.suse.de/job/cloud-crowbar7-job-gate-ha-deploy-x86_64/286/testReport/junit/magnum.tests.functional.api.v1.test_cluster/ClusterTest/test_create_list_sign_delete_clusters_positive_/

PR #2392, opened 2020-04-16
Disable  magnum.tests.functional.api.v1.test_cluster (SOC-11224)

gosipyan:
Magnum test failing when ssl enabled
To unblock SOC7 gating jobs disabling tests till fix provided
Problem appear due to SOC7 missing https://review.opendev.org/#/c/529166
change.Heat failed to validate magnum certificates and cancel stack
creation.

### DIFF
--- a/chef/cookbooks/tempest/templates/default/run_filters/magnum.txt.erb
+++ b/chef/cookbooks/tempest/templates/default/run_filters/magnum.txt.erb
@@ -2,3 +2,6 @@
 
 # Fails when using 2+ controllers
 -magnum.tests.functional.api.v1.test_magnum_service.MagnumServiceTest.test_magnum_service_list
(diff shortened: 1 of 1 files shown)


chef/cookbooks/tempest/templates/default/run_filters/magnum.txt.erb
```diff
@@ -2,3 +2,5 @@
 
 # Fails when using 2+ controllers
 -magnum.tests.functional.api.v1.test_magnum_service.MagnumServiceTest.test_magnum_service_list
+# see SOC-11224
```
stefannica, line 5: Could you add a little more info to this ? something saying that magnum doesn't work when SSL is enabled for other services (e.g. heat) ?

PR #2395, opened 2020-04-20
nova: Hide setup_shared_instance_storage (SOC-11225)

skazi0:
This option is meant to be used only with Xen live migration.
Since Xen support has been removed, this option should not be presented
to the users anymore.

docs snippet
Zrzut ekranu z 2020-04-21 11-02-08

### DIFF
--- a/crowbar_framework/app/assets/javascripts/barclamps/nova/application.js
+++ b/crowbar_framework/app/assets/javascripts/barclamps/nova/application.js
@@ -14,28 +14,3 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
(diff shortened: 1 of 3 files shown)


jsuchome:
Why is the option intended for XEN setup only?

And if it indeed is, why not to remove it completely instead of just hiding it?

skazi0:
@jsuchome I added a piece of deployment guide to the PR description. It says that this feature is meant to be used only for Xen. Note that there's another "shared storage" option which requires manual setup and it is used in other places as well (e.g. in our CI :) )

jsuchome:
Thanks, I must have confused it with the other shared storage option.

What do you think about complete removal?

PR #2400, opened 2020-04-29
run keystone_register on cluster founder only when HA (SOC-11248)

flaviodsr:
Although it was fixed for some services (e.g. nova ) there are still
other services where keystone_register is called on all controllers with
a subsequent apache restart. From this, there is a risk of the
keystone_register call be executed while apache (and keystone) is
restarting and consequently failing the deployment.

This change make all calls for keystone_register to be made only on
one of the controllers for all services that supports HA.

### DIFF
--- a/chef/cookbooks/barbican/recipes/api.rb
+++ b/chef/cookbooks/barbican/recipes/api.rb
@@ -67,6 +67,7 @@ keystone_register "barbican api wakeup keystone" do
   port keystone_settings["admin_port"]
   auth register_auth_hash
   action :wakeup
+  only_if { !ha_enabled || CrowbarPacemakerHelper.is_cluster_founder?(node) }
 end
 
 # Create barbican service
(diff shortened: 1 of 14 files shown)


chef/cookbooks/swift/recipes/dispersion.rb
```diff
@@ -35,6 +35,7 @@
 
 keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
 
+ha_enabled = node[:swift][:ha][:enabled]
```
stefannica, line 38: Careful that not all roles can be clustered (this is one of them, you can check [the automation scripts](https://github.com/SUSE-Cloud/automation/blob/master/scripts/jenkins/cloud/ansible/roles/cloud_generator/defaults/main.yml#L282) for a summary). I don't know what could happen in this case (I don't know what `is_cluster_founder` returns for a non clustered role), but you might end up permanently skipping some code. You also have the guarantee that at maximum a single node will be allowed to have the dispersion assigned.

Another good hint that this case is different is the fact that you don't already have a pacemaker sync mark here to guard the keystone calls (i.e. you had to introduce the `ha_enabled` concept to a recipe that didn't know what HA was in the first place).

PR #2405, opened 2020-05-04
Fix monasca libvirt ping checks (bsc#1107190)

KeithMnemonic:
This patch is one of the pieces needs to fix the monasca
libvirt ping checks. The other patch is an rpm-packaging patch
to correct the sudoers file used by the monasca-agent.


JanZerebecki:
The mentioned patch is https://review.opendev.org/724689

KeithMnemonic:
> The mentioned patch is https://review.opendev.org/724689

do you want me to add that to the commit?

JanZerebecki:
Not necessary for me.

PR #2406, opened 2020-05-06
[5.0] ironic: Install deploy image with ironic

skazi0:
The official SUSE deploy image package will be installed by default
together with ironic-server role. Kernel and ramdisk images will be
uploaded to glance.

(cherry picked from commit a9507cef994bb1471e1c29287d0537d28d5a8fd7)

port of #2100 

**NOTE** This is needed to enable functional ironic testing in SOC8 CI

### DIFF
--- a/chef/cookbooks/ironic/attributes/default.rb
+++ b/chef/cookbooks/ironic/attributes/default.rb
@@ -20,6 +20,7 @@ when "rhel", "suse"
       "openstack-ironic",
       "openstack-ironic-api",
       "openstack-ironic-conductor",
+      "openstack-ironic-image-x86_64",
       "python-ironicclient",
       "python-openstackclient",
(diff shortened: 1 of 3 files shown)


JanZerebecki (approved):

JanZerebecki:
mkcloud ironic CI failed, that'll need to be fixed first.

skazi0:
@JanZerebecki this is a chicken/egg problem. This PR is part of work which was supposed to lead to a working ironic CI. IIRC there were some other problems which were causing CI failures in SOC8 ironic but I'm not sure if it's worth the effort to get back to fixing this.

JanZerebecki:
So that means this doesn't break anything? Then we can merge it.

skazi0:
It should not break anything.

PR #2407, opened 2020-05-07
nova: Remove deprecated options (SOC-7852)

skazi0:
Removed some deprecated options related to ironic/baremetal deployments.


JanZerebecki (approved):
Docs don't mention them any more: https://docs.openstack.org/nova/rocky/configuration/config.html https://docs.openstack.org/ironic/rocky/install/configure-compute.html

PR #2410, opened 2020-05-08
[5.0] rabbitmq: sync startup definitions.json with recipe (SOC-11077,SOC-11274)

stefannica:
(backports #2413)

In an HA setup, every time a rabbitmq broker is restarted, whatever the
reason may be (an RA monitor failure, a controller node reboot etc.), the
rabbitmq queue policy is changed from this:

```
openstack       ha-queues       queues  ^(?!(amq.)|(.*_fanout_)|(reply_)).*     {"ha-mode":"exactly","ha-params":2,"ha-sync-mode":"automatic"}  0
```

to this:

```
openstack       ha-queues       queues  ^(?!amq.).*     {"ha-mode":"exactly","ha-params":2}     0
```

This behavior isn't intentional, it's just a byproduct of neglect (the
startup rabbitmq configuration has over time fallen behind the runtime
configuration enforced by the chef recipes). It has a negative impact
on performance and reliability, because it abruptly makes fanout and
reply queues fully mirrored, making matters worse in the middle of a
restart operation that already puts a lot of strain on the rabbitmq
cluster.

The policy in the generated `/etc/rabbitmq/definitions.json`
file used during startup needs to be kept in sync with the policy
configured at runtime by the rabbitmq chef recipe, to avoid
inconsistencies. This is what this change is all about.

This change should improve the rabbitmq cluster reliability during
events such as broker restarts or controller node reboots.

### DIFF
--- a/chef/cookbooks/rabbitmq/libraries/crowbar.rb
+++ b/chef/cookbooks/rabbitmq/libraries/crowbar.rb
@@ -28,4 +28,22 @@ module CrowbarRabbitmqHelper
   def self.get_management_address(node)
     get_listen_address(node)
   end
+
+  def self.ha_policy_regex
+    # don't mirror queues that are 'amq.*' or '*_fanout_*' or `reply_*` in their names
+    "^(?!(amq.)|(.*_fanout_)|(reply_)).*"
+  end
+
+  def self.get_ha_policy_definition(node)
+    quorum = 1
+    if node[:rabbitmq][:enable_queue_mirroring] && node[:rabbitmq][:ha][:enabled]
+      quorum = CrowbarPacemakerHelper.num_corosync_nodes(node) / 2 + 1
+    end
+
+    {
+      "ha-mode"      => "exactly",
+      "ha-params"    => quorum,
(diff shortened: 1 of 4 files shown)


stefannica:
Testing with 3 similar jobs:
 * https://ci.suse.de/blue/organizations/jenkins/cloud-crowbar8-job-gate-ha-deploy-x86_64/detail/cloud-crowbar8-job-gate-ha-deploy-x86_64/341/pipeline
 * https://ci.suse.de/blue/organizations/jenkins/cloud-crowbar8-job-gate-ha-deploy-x86_64/detail/cloud-crowbar8-job-gate-ha-deploy-x86_64/339/pipeline
 * https://ci.suse.de/blue/organizations/jenkins/cloud-crowbar8-job-gate-ha-deploy-x86_64/detail/cloud-crowbar8-job-gate-ha-deploy-x86_64/340/pipeline

jsuchome:
It's the same as https://github.com/crowbar/crowbar-openstack/pull/2350, right? So why different PR?

PR #2411, opened 2020-05-14
[5.0] Tempest: enable test_volume_boot_pattern test (SOC-10874)

flaviodsr:
This reverts commit db7d6b3f4158874eff1e47e1146374d9c11c65e9.

As CI is now using Ceph for SOC 8, enable the `test_volume_boot_pattern`
tempest test.

### DIFF
--- a/chef/cookbooks/tempest/templates/default/run_filters/cinder.txt.erb
+++ b/chef/cookbooks/tempest/templates/default/run_filters/cinder.txt.erb
@@ -2,5 +2,7 @@
 +tempest.scenario.test_volume.*
 +tempest.scenario.test_encrypted.*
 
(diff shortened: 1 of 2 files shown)


chef/cookbooks/tempest/templates/default/run_filters/cinder.txt.erb
```diff
@@ -1,6 +1,3 @@
 +tempest.api.volume.*
 +tempest.scenario.test_volume.*
 +tempest.scenario.test_encrypted.*
-
```
stefannica, line 4: If it's not too much trouble, could you enable it conditionally ? By that, I mean blacklist it only if ceph is not enabled.

PR #2417, opened 2020-05-15
[4.0] magnum: fix check for image/flavor (SOC-11251)

flaviodsr:
Getting current images/flavors needs to be executed at the end of the
barclamp execution (delayed) as there are cases (e.g. changing keystone
api protocol to/from http/https) where the services (glance, nova) needs
to be restarted before running the checks to update its configuration or
it will fail.


stefannica:
@flaviodsr you should re-check that the image doesn't get uploaded if it's already present. I'm not 100% sure that the `if_not` expressions are indeed evaluated right at the time of execution (i.e. as a delayed action).

PR #2420, opened 2020-05-22
[5.0] Fix availability zone script (bsc#1171661)

djoreilly:
This script was failing with an auth error because
the project_name was not being passed correctly.


jsuchome:
Is it relevant to SOC8 only?

djoreilly:
> Is it relevant to SOC8 only?

There are more problems on SOC9. The keystone v2.0 endpoint is not available, and the novaclient does not have the "hosts" attribute. So it will require a bigger patch.

PR #2422, opened 2020-05-25
tempest: Enable shelve tests when using RBD ephemeral (SOC-11176)

flaviodsr:
The change [1] fixes the shelve feature when using ceph as backend. This
change enable those tests again.

1. https://review.opendev.org/#/c/708616/

This reverts commit 522b3598fd1e897df8dd83b48bac4c627d731cf3.


skazi0:
@flaviodsr what about soc8? the upstream change was not ported back there, was it?

flaviodsr:
@skazi0 no, it wasn't backported as this is not reproducible in pike.
Looking at the nova code I see that change [1] that might have introduced that bug.

1. https://github.com/openstack/nova/commit/e93bc57a73d8642012f759a4ffbe5289112ba490#diff-f4019782d93a196a0d026479e6aa61b1

PR #2425, opened 2020-06-05
Fix availability zone script for SOC9

djoreilly:
This patch changes the script to use the Keystone v3 endpoint as the 2.0
endpoint is not available in SOC9. Also the python-novaclient in SOC9
does not have the os-hosts bindings [1], so use the availability_zones
bindings to find the host's current availability zone instead.

[1] https://github.com/openstack/python-novaclient/commit/1d88717e751e743ca7edd978346b95b6e7b3f235


JanZerebecki:
https://travis-ci.org/github/crowbar/crowbar-openstack/builds/695052401 was successful, just failure to report.

JanZerebecki (approved):

PR #2432, opened 2020-12-02
Disable keepalive in Keystone Apache config (bsc#1179084)

djoreilly:
Horizon's cache is preventing the garbage collector from removing
objects [1] with connections to Keystone. Keystone is running under
Apache which closes these connections after 15 seconds and they move
to CLOSE_WAIT on the client side. This can lead to the Horizon process
running out of file descriptors.

This patch disables KeepAlive in the Keystone Apache config, so the
server sends "Connection: close" in the HTTP response header and the
client removes the connection immediately.

[1] https://bugs.launchpad.net/horizon/+bug/1793411

### DIFF
--- a/chef/cookbooks/crowbar-openstack/providers/wsgi.rb
+++ b/chef/cookbooks/crowbar-openstack/providers/wsgi.rb
@@ -67,6 +67,7 @@ action :create do
         ssl_keyfile: current_resource.ssl_keyfile,
         ssl_cacert: current_resource.ssl_cacert,
         timeout: current_resource.timeout,
+        disable_keepalive: current_resource.disable_keepalive,
         openidc_enabled: current_resource.openidc_enabled,
         openidc_provider: current_resource.openidc_provider,
         openidc_response_type: current_resource.openidc_response_type,
@@ -124,6 +125,7 @@ def load_current_resource
   @current_resource.ssl_cacert(@new_resource.ssl_cacert)
 
   @current_resource.timeout(@new_resource.timeout)
+  @current_resource.disable_keepalive(@new_resource.disable_keepalive)
 
   @current_resource.openidc_enabled(@new_resource.openidc_enabled)
   @current_resource.openidc_provider(@new_resource.openidc_provider)
--- a/chef/cookbooks/crowbar-openstack/resources/wsgi.rb
+++ b/chef/cookbooks/crowbar-openstack/resources/wsgi.rb
@@ -21,6 +21,7 @@ attribute :ssl_keyfile, kind_of: String, default: nil
 attribute :ssl_cacert, kind_of: String, default: nil
 
 attribute :timeout, kind_of: Integer, default: nil
+attribute :disable_keepalive, kind_of: [TrueClass, FalseClass], default: false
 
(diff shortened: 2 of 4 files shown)


JanZerebecki:
I don't have any good feeling how long individual keystone requests take and if there are some that are usually called in bulk or back-to-back. Can you tell something about the performance impact?

For public access there is not another reverse proxy in front of this, so this also affects clients that are not local?

A proper solution would probably be using a connection pool to reuse a small amount of existing connections. Or make the client close the connection after the response is received. Did you investigate any solution besides this patch?

Any idea what the upstream solution is? (The linked bug entry is probably a different situation? It seems to show few connections kept alive in a pool? Maybe that was changed in a version newer than Rocky?)

djoreilly:
> I don't have any good feeling how long individual keystone requests take and if there are some that are usually called in bulk or back-to-back. Can you tell something about the performance impact?
>
Did some quick tests on my system with Apache Benchmark. Mean time to validate a token is 18ms, of which 12ms is connection setup. So yes this patch could have an impact on services if they are frequently reusing connections.
> 
> For public access there is not another reverse proxy in front of this, so this also affects clients that are not local?
>
 Correct. But I guess these just use keystone to get a token every so often.

> A proper solution would probably be using a connection pool to reuse a small amount of existing connections. Or make the client close the connection after the response is received. Did you investigate any solution besides this patch?
> 
Horizon is using keystoneauth sessions already, which provide that. The problem is that it is not reusing these objects, but instead keeps creating new ones. Also it must be keeping a reference to the old ones because we would not see this problem if the garbage collector was allowed to delete them. I have tested the keystoneauth and neutronclient libs outside of Horizon, and they are okay with reusing connections, and the connection gets removed when the object is deleted or garbage collected.

> Any idea what the upstream solution is? (The linked bug entry is probably a different situation? It seems to show few connections kept alive in a pool? Maybe that was changed in a version newer than Rocky?)

I'm not sure there is an upstream solution at this time. See https://bugs.launchpad.net/python-keystoneclient/+bug/1780164/ . I have tried backporting the commits from stein that are supposed to fix the memory leak, but they don't seem to work. Maybe it's a python2/3 thing - need to look into that more. 

I'm going to see if I can reproduce this with the latest devstack. Interestingly devstack has the server closing connections:
https://github.com/openstack/devstack/blob/302ce78aefd5727a7cfd1f2ccad5ae0754a3c7c7/lib/apache#L297

JanZerebecki:
Do you think this might help?: https://opendev.org/openstack/keystoneauth/commit/b2b5ad3cb1ff05e08c22973bb079125214ba7bcf

djoreilly:
> Do you think this might help?: https://opendev.org/openstack/keystoneauth/commit/b2b5ad3cb1ff05e08c22973bb079125214ba7bcf

Thanks. I'll give it a try, but I don't see those warnings.

djoreilly:
> > Do you think this might help?: https://opendev.org/openstack/keystoneauth/commit/b2b5ad3cb1ff05e08c22973bb079125214ba7bcf
> 
> Thanks. I'll give it a try, but I don't see those warnings.

It did not make a difference.

djoreilly:
Closing. We decided to fix this in Horizon instead.

PR #2433, opened 2020-12-08
neutron: Fix handling of networks with non-ascii names (SOC-11429)

skazi0:
Chef uses default LC_ALL=C locale for all internal ShellOut calls.
These fails if command output contains non-ASCII characters.
The problem is bigger but it was really fixed only in Chef 12.
We could patch ShellOut to use different default be it could cause
some unwanted side effects in many places.
In this case the problem is with network names so it's easier to
limit the output to include only expected entries and not print all
names.

### DIFF
--- a/chef/cookbooks/neutron/recipes/post_install_conf.rb
+++ b/chef/cookbooks/neutron/recipes/post_install_conf.rb
@@ -121,7 +121,7 @@ package "python-openstackclient"
 
 execute "create_fixed_network" do
   command "#{openstack_cmd} network create --share #{fixed_network_type} fixed"
-  not_if "out=$(#{openstack_cmd} network list); [ $? != 0 ] || echo ${out} | grep -q ' fixed '"
+  not_if "out=$(#{openstack_cmd} network list --name fixed); [ $? != 0 ] || echo ${out} | grep -q ' fixed '"
   retries 5
   retry_delay 10
   action :nothing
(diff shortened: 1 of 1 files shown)


djoreilly (commented):
There are other places in the same file that do things like "openstack network list | grep fixed". Should these be changed to something like "openstack network list --name fixed" too?

skazi0:
@djoreilly I don't think it's needed. At least it didn't cause problems in any of my tests. Probably because the problematic strings never land on stdout.

djoreilly:
This fails for me with "codec can't encode character ..."
```
require 'mixlib/shellout'
cmd = Mixlib::ShellOut.new("openstack network list | grep fixed")
cmd.run_command
puts cmd.stdout
puts "error messages" + cmd.stderr
cmd.error!
```

skazi0:
@djoreilly I think I know what is going on there. If you use the exact command that we have in the cookbooks you'll see that it doesn't fail but it has some stderr output. We explicitly check for exit status of the openstack command and consider it a reason for not running the create commands (e.g. API not up yet). So if we have the networks created before we create the non-ascii one it should all be fine. BUT if we remove some of the default networks they will not be re-created on next chef-client run because of the non-ascii network (normally they would).

skazi0:
@djoreilly Even though the removal of pre-defined networks is probably not supported (it can fail in many different ways), I added the filtered outputs to `network list` commands. I don't think it can cause any damage and makes things a little bit more robust.

djoreilly (approved):
okay - thanks for the explanation.

JanZerebecki (approved):

PR #2435, opened 2021-02-25
monasca: fix monasca-server reinstall state check (SOC-11471)

jgrassler:
In the course of fixing bsc#1148158, an overzealous check for
the monasca-server node's state was introduced. This check
causes monasca-installer to not run when the monasca-server
node isn't in state `ready`. This commit changes the check to
only trip when the node is in state `reinstall`.

Note: I'm only targeting stable/pike with this because it does not apply to Cloud 9 where we did away with monasca-installer entirely.


skazi0 (approved):
lgtm

gosipyan (approved):

PR #2437, opened 2021-03-16
horizon: Add SSL protocol option.

OttoHollmann:
Current setting in cookbook templates exclude only SSLv2 and SSLv3.
`SSLProtocol all -SSLv2 -SSLv3`
but in these days TLSv1.0 protocol is already deprecated, so it will be nice to have any option how to disable this old protocol.
We can add protocol option (this PR) and keep it up to user to select which protocol he wants, disable TLSv1.0 in default settings or both of them.

### DIFF
--- a/chef/data_bags/crowbar/template-horizon.json
+++ b/chef/data_bags/crowbar/template-horizon.json
@@ -36,6 +36,7 @@
       "apache": {
         "ssl": false,
         "generate_certs": false,
+        "ssl_protocol": "all -SSLv2 -SSLv3",
         "ssl_crt_file": "/etc/apache2/ssl.crt/openstack-dashboard-server.crt",
         "ssl_key_file": "/etc/apache2/ssl.key/openstack-dashboard-server.key",
         "ssl_crt_chain_file": ""
@@ -52,7 +53,7 @@
     "horizon": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 302,
+      "schema-revision": 303,
       "element_states": {
         "horizon-server": [ "readying", "ready", "applying" ]
       },
--- a/crowbar_framework/app/views/barclamp/horizon/_edit_attributes.html.haml
+++ b/crowbar_framework/app/views/barclamp/horizon/_edit_attributes.html.haml
@@ -27,6 +27,7 @@
 
       #apache_container
         = boolean_field %w(apache generate_certs)
+        = string_field %w(apache ssl_protocol)
         = string_field %w(apache ssl_crt_file)
         = string_field %w(apache ssl_key_file)
         = string_field %w(apache ssl_crt_chain_file)
--- a/crowbar_framework/config/locales/horizon/en.yml
+++ b/crowbar_framework/config/locales/horizon/en.yml
@@ -31,6 +31,7 @@ en:
         apache:
           ssl: 'Protocol'
           generate_certs: 'Generate (self-signed) certificates'
+          ssl_protocol: 'SSL Protocol Version'
           ssl_crt_file: 'SSL Certificate File'
           ssl_key_file: 'SSL (Private) Key File'
           ssl_crt_chain_file: 'SSL Certificate Chain File'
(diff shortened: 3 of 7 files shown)


chef/data_bags/crowbar/template-horizon.json
```diff
@@ -36,6 +36,7 @@
       "apache": {
         "ssl": false,
         "generate_certs": false,
+        "ssl_protocol": "all -SSLv2 -SSLv3",
```
dirkmueller, line 39: note that -SSLv2 is redundant (there is no SSLv2 anymore)

chef/data_bags/crowbar/template-horizon.json
```diff
@@ -52,7 +53,7 @@
     "horizon": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 302,
+      "schema-revision": 303,
```
dirkmueller, line 56: in addition to bumping the schema version, you also need to add a migration file that adds the attribute.

crowbar_framework/app/views/barclamp/horizon/_edit_attributes.html.haml
```diff
@@ -27,6 +27,7 @@
 
       #apache_container
         = boolean_field %w(apache generate_certs)
+        = string_field %w(apache ssl_protocol)
```
dirkmueller, line 30: do we really want to expose that in ui?

OttoHollmann: I was not sure, but now it's removed.

chef/data_bags/crowbar/template-horizon.json
OttoHollmann: `SSLv2` removed

chef/data_bags/crowbar/template-horizon.json
OttoHollmann: See file 303_add_ssl_protocol.rb, hopefully it's correct.

crowbar_framework/config/locales/horizon/en.yml
```diff
@@ -31,6 +31,7 @@ en:
         apache:
           ssl: 'Protocol'
           generate_certs: 'Generate (self-signed) certificates'
+          ssl_protocol: 'SSL Protocol Version'
```
dirkmueller, line 34: as you didn't update edit_attributes.haml for this, I think this part of the change is unnecessary (I don't think we use localization for anything else.. I might be forgetting something..)

chef/data_bags/crowbar/migrate/horizon/303_add_ssl_protocol.rb
```diff
@@ -0,0 +1,9 @@
+def upgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["apache"]["ssl_protocol"] = template_attrs["apache"]["ssl_protocol"]
```
dirkmueller, line 2: please do this only if attrs["apache"]["ssl_protocol"] is not already defined (because of a backported earlier migration.

chef/data_bags/crowbar/migrate/horizon/303_add_ssl_protocol.rb
```diff
@@ -0,0 +1,9 @@
...
+  return attrs, deployment
+end
+
+def downgrade(template_attrs, template_deployment, attrs, deployment)
+  attrs["apache"].delete("ssl_protocol")
```
dirkmueller, line 7: please do it only if template_attrs doesn't still include ssl_protocol

chef/data_bags/crowbar/migrate/horizon/303_add_ssl_protocol.rb
OttoHollmann: I'm not sure which is preferred way, if `unless attrs.key?` or `if attr["..."].nil?`, but the former looks better.

chef/data_bags/crowbar/migrate/horizon/303_add_ssl_protocol.rb
OttoHollmann: done

crowbar_framework/config/locales/horizon/en.yml
OttoHollmann: Agree with you, removed.

PR #2440, opened 2021-08-11
keystone wakeup: get new session on any error (bsc#1189052)

djoreilly:
Before f0f3e481 the keystone wakeup action would always get a
new token, but now it tries to reuse the session object. The
sessions's token might be expired, and the retry loop only tries
to get a new session if the password changes. This patch chnages
that so it gets a new session on any error.

### DIFF
--- a/chef/cookbooks/keystone/recipes/server.rb
+++ b/chef/cookbooks/keystone/recipes/server.rb
@@ -509,7 +509,7 @@ keystone_register "wakeup keystone" do
   auth register_auth_hash
   retries 5
   retry_delay 10
-  reissue_token_on_error update_admin_password
+  reissue_token_on_error true
   action :wakeup
 end
 
--- a/chef/cookbooks/keystone/providers/register.rb
+++ b/chef/cookbooks/keystone/providers/register.rb
@@ -26,6 +26,7 @@ action :wakeup do
     break unless error && count < 50
     sleep 1
     next unless new_resource.reissue_token_on_error
+    Chef::Log.info "Problem finding /v3/services. Retrying with new session."
     KeystoneHelper.reset_session
     session
   end


chef/cookbooks/keystone/recipes/server.rb
```diff
@@ -509,7 +509,6 @@
   auth register_auth_hash
   retries 5
   retry_delay 10
-  reissue_token_on_error update_admin_password
```
skazi0, line 512: why not just set it to `true` here instead of removing the whole parameter?

djoreilly: > why not just set it to `true` here instead of removing the whole attribute?

I believe this is the only place that calls keystone wakeup - I don't see`reissue_token_on_error` anywhere else. So if the parameter is always `true` then it's redundant and just confusing?

skazi0: I'd call it "not used currently". Personally I'd rather limit the size of changes to this stuff as much as possible. But that's just my personal opinion :o) Feel free to merge as is.

skazi0 (dismissed):

chef/cookbooks/keystone/recipes/server.rb
djoreilly: done

skazi0 (approved):

KeithMnemonic (approved):
LGTM

JanZerebecki:
CI failed with timeout from http://download.nue.suse.com/update/build.suse.de/SUSE/Updates/SLE-SDK/12-SP5/x86_64/update/repodata/repomd.xml , when I try that URL it gives me a 404 but I can see the next subdirectory in http://download.nue.suse.com/update/ .

JanZerebecki:
Asked autobuild to please fix via mail.

JanZerebecki:
autobuild fixed it, retrying.

JanZerebecki:
The CI itself works now, but the mkcloud ha job failed. I see the same failure without this patch in the last cloud-mkcloud9-job-ha-x86_64 job run https://ci.suse.de/job/openstack-mkcloud/68001/parsed_console/ .

